PR #486, opened 2018-05-02
Add meta charset tag to HTML output generation.

Mornix:
Hint to browsers that the charset is UTF-8 due to "Système International d'Unités".


JacquesCarette:
Changed the reviewer to be me, since I just reviewed it.

Issue #490, opened 2018-05-04
Instance Model Section in the SSD probably Doesn't Plug in Properly


halonazhao:
I found out that the intro of IM is missing. After tracing it, I found that the problem is probably at [DocumentLanguage](https://github.com/JacquesCarette/Drasil/blob/master/code/Example/Drasil/DocumentLanguage.hs#L432) Line432. Since "SRS.inModel"  (in [SRS](https://github.com/JacquesCarette/Drasil/blob/master/code/Example/Drasil/SRS.hs#L66)) is used instead of "SSD.inModelF" (in [SSD](https://github.com/JacquesCarette/Drasil/blob/master/code/Example/Drasil/Sections/SpecificSystemDescription.hs#L177))which includes the intro section, the intro isn't plugged in properly.

@szymczdm @JacquesCarette

JacquesCarette:
Well done!  You are most likely correct.

Can you create a new branch (perhaps based on master, not your branch) where you try to fix this?

halonazhao:
I just solved it. They are posted in the new branch called "im_modify". I also tested it in my brach and it works. The intro shows up properly.

![screen shot 2018-05-05 at 12 52 57](https://user-images.githubusercontent.com/28347043/39668780-43b72eb4-50a6-11e8-9e70-94f7dfdeb992.png)

JacquesCarette:
Good!  Can you submit this as a Pull Request?  I want to test the process.

PR #494, opened 2018-05-06
Fix the intro of IM section for SSD.

halonazhao:
#490


halonazhao:
Yes, all the tests passed and the log files for all cases are empty. I did the manual check and I think this change doesn't effect the old way to generate the SSD section. 

But I don't know what is PR? @JacquesCarette

JacquesCarette:
Perfect.  Now merged.

PR #509, opened 2018-05-09
Replaced pi with π on line 272

elwazana:


JacquesCarette:
@szymczdm The CI builds failed on this. It seems to me that this is because of Haddock rather than the code itself.  But I may be misreading the symptoms. Could you check please?

szymczdm:
@JacquesCarette I just double-checked the CI log, something has changed in Chipmunk's log and that's causing the problem. I need to take a closer look.

szymczdm:
@JacquesCarette I took a look and Chipmunk's stable just needs to be updated. However, this changes the table of symbols so that the table of symbols defines π as Π. I don't think that's a good change, thoughts?

JacquesCarette:
That's indeed not a good change.  I'm thinking that this is due to the "noun phrase" which capitalizes things. Need to change which combinator is used. @elwazana can you investigate further?

elwazana:
@JacquesCarette  should I investigate "noun phrase" or which combinator needs to be replaced?

JacquesCarette:
You should look at the combinators in `NounPhrase.hs` and find the one(s) which do not actually do anything when you ask to capitalize a symbol, and use that one instead of the current one (`nounPhraseSP`).

elwazana:
@JacquesCarette I can't find a combinator that ignores symbols in <code>NounPhrase.hs</code> but I did find a list of strings that don't get capitalized.

Should I add <code>π</code> to the ignore list?

https://github.com/JacquesCarette/Drasil/blob/4a1c4f8f82eb50957d19494f45fd6b3c1dbfda7a/code/Language/Drasil/NounPhrase.hs#L223-L225

I think that the problem is that when I replaced <code>"pi"</code> with <code>"π"</code>, its considering it as a string not a symbol.

JacquesCarette:
Your analysis of the problem is correct.

Rather than adding to `doNotCaps`, I think that changing the function that does capitalization to ignore both `doNotCaps` and non-latin-alphabet letters would be the correct fix.

elwazana:
By function that does capitalization, are you referring to <code>CapFirst</code> and <code>CapWords</code>?

elwazana:
Alright, should I create a similar set of functions as:
https://github.com/JacquesCarette/Drasil/blob/4a1c4f8f82eb50957d19494f45fd6b3c1dbfda7a/code/Language/Drasil/NounPhrase.hs#L214-L225
but for non-latin-alphabet?

Or is their something else specific you had in mind?

elwazana:
I found this character classifier function, but I'm pretty sure it won't cover all non-latin-alphabet. Should I use it?
image

Or should I use this function?
image

Or a combination of the two?

JacquesCarette:
I think you want to use `isLatin1` and `isLetter`.

elwazana:
@szymczdm 
Hello Dan,
I've tried a few methods for fixing the issue, this is my latest attempt and closest attempt:
image
The files do get generated with this. However, the error itself it still not fixed and I'm unsure as to why. I was hoping if you can give some insgiht into it?

szymczdm:
@elwazana I'll come take a look. I can't see the change in the repo, so I'm assuming you haven't pushed it.

niazim3:
Found that changing the `at_start` from the `mkRefSec` definition 
https://github.com/JacquesCarette/Drasil/blob/4a1c4f8f82eb50957d19494f45fd6b3c1dbfda7a/code/Example/Drasil/DocumentLanguage.hs#L261
to `phrase` updated the table of symbols to have the symbol pi show up as lowercase. Not sure how relevant/helpful this may be (if at all), but thought I'd make a note here.

JacquesCarette:
Rather than use `head` and `tail`, please use pattern-matching.  i.e.
```haskell
isNotCap :: String -> String
isNotCap (c:cs) | not (isLetter c && isLatin1 c) = toLower c : cs
                | ((toLower c : cs) `elem` doNotCaps = toLower c : cs
isNotCap s = s
```

JacquesCarette:
Yes, this is definitely related to that.  But I've lost track of the exact relationship, use cases, etc.

szymczdm:
@elwazana try using the `pn` constructor instead of `nounPhraseSP`.

elwazana:
Yes, only one log was flagged as being different from stable and that was the Chipmunk one. The only thing in there is the change from `Pi` to `π`
image

JacquesCarette:
Excellent. That is a good change, so it should be accepted. Then I'll be able to merge this in.

elwazana:
Yep, updating stable is that last thing I can think of.

niazim3:
Created issue #529. Should the stable .tex file remain as is (i.e. containing the wrong correction)?

szymczdm:
It's not technically wrong, as the generated version has the appropriate symbol within it (ie. We want to be able to use π instead of \pi{}), we just need to add the appropriate package support so the LaTeX pdf renderer will display it properly.

JacquesCarette:
Right. As I said on another commit, I think it works with lualatex but not pdflatex. [Please test it!]. We can make pdflatex an option later.

szymczdm:
@JacquesCarette I tested the make and it resulted in nothing being printed. Lualatex was able to typeset the output, but not the unicode pi symbol. The unicode symbols in the SI description work fine though.

Example:

image

JacquesCarette:
Ok, that's definitely a problem then!  I was hoping that because the accented characters worked, that would mean that all would.  Can you poke at it a bit?

PR #518, opened 2018-05-10
Replaced and removed acroTest in Example and removed most occurences of acroNumGen

elwazana:
I've finished removing acroTest, I've kept the function definition in References.hs in Language.Drasil for now but every else acroTest was has been modified.

 I have also removed most occurrences of acroNumGen in Example and would like to know if I should finish removing the rest of acroNumGen?


JacquesCarette:
This requirement was in the original text version - so @smiths is really the one to 'justify' its existence. But the short of it is: it is not an error, it is wanted.

@szymczdm CI is failing for this too?

elwazana:
@JacquesCarette would you consider this an improvement:

image

This results in a link that takes you to the referred to assumption.

image

JacquesCarette:
I would move all the assumptions to its own separate file.

elwazana:
Should I take screenshots of the generated compared to the stable as I have done for acroNumGen?

elwazana:
@JacquesCarette Dr. Carette, I've removed acroA from SWHS, but I've noticed that in SWHS there are references to the assumptions made with two different methods. One involved the assumptions as `Contents` and the other after replacing acroA involved the assumptions as `Sentence`

The link is defined with assumptions as `AssumpChunk` type
image


While these links are defined with assumptions as `Contents` type
image

I'm not sure if this is what you would like? I believe this causes a consistency issue within the code, but I'm not sure how to go about resolving it if I need to at all?

JacquesCarette:
Good catch. I think we should be uniform. And I prefer the more meaningful approach, which means using the `AssumpChunk` type.

However, I think we should do this incrementally.  Already this PR is large, it has diffs and it has conflicts. So let's try to settle things 'as is', get it integrated into master, and then make those changes.

szymczdm:
This branch has test failures and conflicts with master. I think we should prioritize merging these changes in before continuing to remove the other acro-functions, since this PR was made to incorporate the removal of acroTest and most of acroNumGen.

I think you should fix the conflicts first, then if you're unsure about which updates to stable are good, I can jump in there so we can have the build passing.

Once that's done we'll merge it into master and move on to the other acro-functinos.

elwazana:
@szymczdm Dan, I've fixed the conflicts in the branch. I am unsure which changes you and Dr. Carette would prefer to have updated into stable.

szymczdm:
Some warnings need to be cleaned up, but otherwise these changes are good to go. Merging as soon as Travis CI gives the all-clear.

elwazana:
@szymczdm Should I create another branch and continue removing the rest of the `acro-` functions?

Issue #524, opened 2018-05-14
Removal of ConVar


niazim3:
As pointed out in issue #514, "it appears `ConVar` and `DefinedQuantityDict` are identical in terms of the class instances they implement. The only difference is in the way they're constructed. If anything it might be worthwhile to remove one of them entirely since they're not really saying anything different."

Since `ConVar` only has `ConceptChunk`, `Symbol`, and `Space` and `DQD` has `Quantity` (which has `unit`, `attribs`, `idea`, `space`, and `symbol`) and `ConceptChunk`, `ConVar` is the type that should be removed.

Other notes for convenience: 
"...we may want to change the way DQD is constructed to avoid the 'double uid' issue mentioned in the comments inside DefinedQuantity.hs.
Also, just to be clear, `ConVar` and `DQD` have exactly the same class instances. While `ConVar` always returns Nothing for the Maybe Unit, whereas `DQD` will sometimes have a non-Nothing value, they are databases of the same types of information.

Working on this issue in [this branch](https://github.com/JacquesCarette/Drasil/tree/remove-convar).

JacquesCarette:
[I've read all the conversation at #514]. I believe I agree.

**However** we should first figure out where `ConVar` is used/defined, and where a defined quantity is instead. There might be differences in intension that are not reflected in the implementations. If we prematurely unify them, we might regret it.

So I'd like to see say 5 different places each (with details!) of use "in context".

niazim3:
It looks like `DefinedQuantityDict` instances are not that common in the current examples: 
![image](https://user-images.githubusercontent.com/28247301/40010679-72c9bec8-5773-11e8-9a7c-e6af20217297.png)

Example of use #1:

Looking in [SSP/Unitals.hs](https://github.com/JacquesCarette/Drasil/blob/adding-attributes/code/Example/Drasil/SSP/Unitals.hs): 
https://github.com/JacquesCarette/Drasil/blob/dbd4b7f149c87c87794d7d9db21a1e0cd05d5972/code/Example/Drasil/SSP/Unitals.hs#L14-L16

, where `sspUnitless` is a list of `convar`s.

https://github.com/JacquesCarette/Drasil/blob/dbd4b7f149c87c87794d7d9db21a1e0cd05d5972/code/Example/Drasil/SSP/Unitals.hs#L390-L396

[...]

PR #526, opened 2018-05-14
Empty List Constructors

samm82:
In addition to our current work on adding attributes, I've also created the empty list smart constructors in this branch. We haven't removed `ConVar` yet, so could this branch be re-opened if you think that merging it is a good idea? @JacquesCarette


samm82:
@JacquesCarette Yeah - the problem is with `ConVar`, since it doesn't have `Attributes`. Should I go ahead and unify `ConVar` and `DQD`?

JacquesCarette:
Not quite. Can you instead tell me exactly what the consequences of that unification would have?

samm82:
See my latest comment on #524

niazim3:
@JacquesCarette Would it be considered to be better if the base branch for the `remove-convar` (purple) branch was changed to be `adding-attributes`(blue) (instead of `master`) since that was where this branch branched from?
image
 
Would changing the base branch be a bad idea this far ahead into it?
image

JacquesCarette:
I honestly don't know. It certainly isn't a bad idea. But I just don't know the full consequences this might have. And I don't know if we could easily back out such a change.

At this point, it might be safer to go via a merge rather than a rebase.

PR #528, opened 2018-05-15
Plurals.hs is no longer used

niazim3:


JacquesCarette:
It would have been better if this PR only contained the one change - it also has 2 other commits that are not related. Perhaps look into `git cherry-pick` ?  However, as this is all fine, I'll merge it as is.

PR #530, opened 2018-05-16
Devi

deviprasad135:
manual merge


JacquesCarette:
@szymczdm @deviprasad135 why has there been no movement on this?

deviprasad135:
I thought it was under review, so I din't touch it.

JacquesCarette:
It's still failing the Travis build, you should be able to investigate that yourself and fix it.

deviprasad135:
Okay.

deviprasad135:
@JacquesCarette, I did make and the log files of Glass-BR and Chipmunk are not empty.  So can I update the stable according to the changes in log files?

JacquesCarette:
You can figure out why they are not empty, understand the reason for the differences, and judge whether the difference is an improvement or a regression. Post the analysis as a comment. Then we can figure out what to do.

deviprasad135:
**GlassBR differences**

- The format of Theoretical Models, Data Definitions and Instance Models are different. I think the format in generated version is improved.

For example the Theoretical Model in stable version looks like
capture

The Theoretical Model in generated version looks like
capture g

Data Definition in stable version looks like 
dd s

Data Definition in generated version looks like
dd g

Instance Model in stable version looks like
im s

Instance Model in generated version looks like
im g

- The generation of equations are changed. "dmath" is used to generate equations. For example
equation

- Data Constraints section is missing in generated version.

**Game Physics differences**

- Assumptions Section is changed.
Assumptions in stable version looks like
assumptions

Assumptions in generated version looks like
assumptions g

- Theoretical Models and Instance Models are changed which is same as discussed in GlassBR example.

- Data Definitions and Data Constraints sections are missing in generated version.

smiths:
I agree that the newly generated documentation looks better than what is currently in stable.  However, I think we should wait for @szymczdm to confirm that the changes match what he is looking for.

deviprasad135:
Okay

szymczdm:
I'll be taking an in-depth look at this tomorrow. Cursory look seems like everything is good to go, so once I've double-checked I think we'll be able to merge.

szymczdm:
There are some non-trivial issues with the diffs between the build and stable on this branch. I'm going to continue removing the trivial diffs tomorrow morning, but then we'll need some issues up to figure out what exactly is going wrong and how to fix it (some are strictly printing issues for the IMs, some are a bit more complicated issues, in one case there's an entire subsection missing).

JacquesCarette:
Travis says GlassBR and Chipmunk logs are not empty, but I didn't see anything in the log itself. I thought the diffs used to show?

code/Example/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -158,7 +158,7 @@ mkIMField i _ l@(InConstraints) fs  = (show l,
   foldr ((:) . eqUnR) [] (map tConToExpr (i ^. inCons))) : fs
 mkIMField i _ l@(OutConstraints) fs = (show l,
   foldr ((:) . eqUnR) [] (map tConToExpr (i ^. outCons))) : fs
-mkIMField i _ l@(Notes) fs = (show l, maybe [Paragraph EmptyS] (map Paragraph) (i ^. getNotes)) : fs
+mkIMField i _ l@(Notes) fs = (show l, maybe [Paragraph $ S "Not applicable."] (map Paragraph) (i ^. getNotes)) : fs
```
smiths: This is an improvement, but @JacquesCarette was correct that not displaying empty notes is the best option.

JacquesCarette: So I think the 'right' thing to do is to change `[Paragraph $ S "Not applicable."] ` to `[]` and have the printer detect this as an empty `Notes` field and print nothing (including no header).

niazim3: Done in commit 38a401f.

szymczdm:
Looks like all the checks on this branch are now passing. I'm working on a merge with master atm, but it's going to be fairly slow and dirty.

szymczdm:
I'm going to finally merge this branch into master. @niazim3 can you please double-check that all unfinished issues discussed here have been created? I took a look and it seems like it, but a second set of eyes would be great.

niazim3:
@szymczdm After going through the comment and commit history, I agree that necessary issues have been made. My only concern is that the spacing between the elements in the lists that can be found in the descriptions weren't discussed or worked on for the PDF side of things...
image

szymczdm:
@niazim3 good catch! We definitely need to fix that, since it's already been done on the PDFs. Can you create an issue so it doesn't get lost?

PR #532, opened 2018-05-17
Moving contents of ExprRelat.hs to a more general file

niazim3:
Is there a reason why [this file](https://github.com/JacquesCarette/Drasil/blob/master/code/Language/Drasil/Chunk/ExprRelat.hs) is on its own?

Does it make sense for its contents to be in https://github.com/JacquesCarette/Drasil/blob/master/code/Language/Drasil/Classes.hs, so the data-type-files do not need to import Language.Drasil.Chunk.ExprRelat separately when they are instances of this class?


JacquesCarette:
Looks good -- @szymczdm can you merge?

PR #540, opened 2018-05-22
Started to remove and replace acroDD in SSP

elwazana:
@szymczdm @JacquesCarette I've highlighted how I started to replace acroDD, I just wanted to get it checked out before replacing more of it. I've also pushed the changes so you can take a closer look.

image


elwazana:
`make` works perfectly with this commit, I was wondering if the Travis ci checking fails if the logs aren't empty? or is the ci separate? 

Also the design review I asked for above is also related to acroGD, acroIM, and acroT. Just becuase the reference helper functions for these are extremely similar. So I've pumped the brakes on starting those for now until an apprpriate design is found.

szymczdm:
@elwazana the Travis CI build fails if any of a number of conditions are met:

1. The build fails
2. Generating Haddock documentation fails
3. The build succeeds, but there are differences between the generated artifacts and the stable artifacts (our test cases).

You can look at the Travis log to see which case occurred (although, it might involve combing through the log since 1 and 3 above show up before the haddock build is attempted).

As for your start to replacing acroDD, it seems like it's a lot of code duplication (namely `refDD (ddRefDB sspRefMDB)`. You might want to condense that into a helper function for this example. Otherwise I don't see any major cause for concern other than the CI tests failing -- is it just a minor difference?

elwazana:
It replaced the text "DD1" in a paragraph with a link named "DD1" that's the only difference. Although the links that are generated don't seem to work, I can't find the reason for them not working though.

JacquesCarette:
So the change from text "DD1" to link "DD1" is good; a dead link is bad. To stop the build from failing, you should also verify that the changes to 'stable' are good, and incorporate the changes from build into stable.

elwazana:
The `Referable` class seems to have errors from non-chunk instances such as `QDefinition`, `TheroyModel`, `GenDefn`, and `InstanceModel`, this error appears as the non-addressing of links (i.e. the `RefHelper` functions `refDD`, `refTM`, `refGD`, and `refIM` didn't actually create working links).

These links would be clickable but non-functioning, I've traced back the error to the `Referable` instances of the types of DDs, TMs, GDs, and IMs. 

To fix the issue:

Original:
```
instance Referable GenDefn where
  refName _ = error "No explicit name given for theory model -- build a custom Ref"
  refAdd  g = "GD:" ++ g^.uid
  rType   _ = Def
```

Fixed:
<pre>
instance Referable GenDefn where
  refName _ = error "No explicit name given for theory model -- build a custom Ref"
  refAdd  g = "GD:" ++  <b>concatMap repUnd</b>  (g ^ .uid)
  rType   _ = Def
</pre>

I just want to check-in with @JacquesCarette and get the go-ahead that the design of the fix is alright?

szymczdm:
That looks like a good fix for now, I think the use of uid was a hack to get things working, so we'll probably want to circle back to that at some point and create better referencing addresses for GDs, etc.

Can you create an issue for that as well?

JacquesCarette:
Can you give explicit occurrences where this is a problem?  The fix will likely depend on the details.

JacquesCarette:
So it seems like the definition of `displMtx` is in the wrong place?

elwazana:
That's what I initially thought but `displMtx` is also used within GenDefs.hs so I became unsure if I should move it or not.

From GenDefs.hs:
image

szymczdm:
If you move the definition of displMtx, then you can still import it into GenDefs, and that should still break the cycle (assuming it's the only thing being used by the DataDefs)

elwazana:
It isn't, however, the other imports seem to only used in the derivation of each Data Definition. Everything imported by DataDefs from GenDefs is `eqlExpr`, `displMtx,` and `rotMtx`. 

These are all occurences of them in DataDefs:
image
image
image
image

szymczdm:
So it seems those (`eqlExpr`, `displMtx`, and `rotMtx`) definitions should be pulled out of GenDefs and then imported where required. Where to put them really depends on what they represent.

elwazana:
They all seem to be related to the type `Expr`, should I create an auxiliary Expr file just for them in SSP? The only other place that might be viable is the Defs.hs file in SSP?

JacquesCarette:
What is the actual definition of `rotMtx` and `eqlExpr` ?  This will tell us the **meaning** of these, which is what we ought to use to figure out where to define them.

elwazana:
image
image

JacquesCarette:
You can create a new 'BasicExprs.hs' file for defining these re-usable `Expr`s.

elwazana:
@JacquesCarette Hello Dr. Carette, I don't know why those merges are appearing as commits I didn't commit any of them. Does git automatically commit after a merge?

JacquesCarette:
Those are triggered by the merge you did, I believe. And potentially by my merging of the PR. This part of git is still a tad mysterious for me as well.

elwazana:
@JacquesCarette Dr. Carette, are you satisfied with the replacement of acroDD? If so should I go through and update stable?

JacquesCarette:
Yes.

JacquesCarette:
Even though we're changing lots of stuff having to do with referencing, these changes are still good.

Issue #537, opened 2018-05-21
Remove Attributes


szymczdm:
As per the comments on #527 we should work on removing attributes. Here's the general idea on how to deal with each:

1. **Derivation**: create a `HasDerivation` lens with `getDerivation` accessor. Remove from Attribute. Add derivation field to `QDefinition`, and I think `DQD`. See what else.
2. **SourceRef**: create `HasReference` lens with `getReferences` accessor. Remove from Attributes. Add field to `QDefinition`
3. **ShortName**: create `HasShortName` lens. Merge this with `refAdd` lens (as a second step). Remove from Attributes. Add to same places that are currently instances of `Referable`
4. **Rationale**: remove. We will re-create this later.
5. **Uses**: remove. We will deal with this via 'inversion'.
6. **Attributes**: remove!

**Recommended order of implementation**: ~~Start with 5 and 4 as they're the simplest.~~ ~~Then 2's lens, 2 add field, 2 remove from attributes~~, ~~1 lens, 1 add field, 1, remove from attributes,~~ 3 lens, 3 add instances, 3 remove refAdd and replace with shortname, 3 remove from attributes, 6.

samm82:
@szymczdm @JacquesCarette If the end goal is to remove `Attributes`, and `QDefinition` is getting them from `QuantityDict`, shouldn't we add the Derivation and SourceRef fields to `QuantityDict` instead? Or are we passing the `Attributes` from `QuantityDict` to `QDefinition` and then lensing the `Derivation` and `References`?

samm82:
@JacquesCarette So then how should we go about implementing `SourceRef` - just wait for now?

szymczdm:
`SourceRef` should show up on certain chunks by design (TMs, GDs, DDs, and IMs). I think we should look at the manually produced documents and see where **Source** comes up and what it contains. We might also want to see where external sources are referenced, to see if they're attached to a particular piece of knowledge or not.

I think that'll give us a better idea of what we're dealing with.

samm82:
Also, @szymczdm, are the FIXMEs an error in the stable?
![ah](https://user-images.githubusercontent.com/35857611/40377799-cfe2c67a-5dbf-11e8-829f-81ab2ffc1afc.JPG)

niazim3:
@JacquesCarette 

For `ShortName`, where the instructions state to merge the `hasShortName` lens with `refAdd`, is the reasoning for this because one will be removed eventually?
[...]

PR #542, opened 2018-05-22
Adding SourceRef

samm82:
@szymczdm @JacquesCarette Is this the correct way to go about this adding `SourceRef` from #537? I left `Attributes` in the data definitions just as a placeholder until we're ready to fully remove it, and I have the constructors pass in empty lists for now, again, as placeholders until we remove `Attributes`. Then `References` will be passed in instead.

### DIFF
--- a/code/Example/Drasil/GlassBR/DataDefs.hs
+++ b/code/Example/Drasil/GlassBR/DataDefs.hs
@@ -42,7 +42,7 @@ risk_eq = ((sy sflawParamK) /
 
 -- FIXME [4] !!!
 risk :: QDefinition
-risk = mkDataDef' risk_fun risk_eq (aGrtrThanB +:+ hRef +:+ ldfRef +:+ jRef) [sourceref $ S "[4]"]
+risk = mkDataDef' risk_fun risk_eq (aGrtrThanB +:+ hRef +:+ ldfRef +:+ jRef) [] [sourceref $ S "[4]"]
 
 --DD2--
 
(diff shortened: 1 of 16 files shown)


samm82:
Should I be removing `Attributes` where I've added `References`? @szymczdm @JacquesCarette

JacquesCarette:
See the comments that I have added on each of the commits themselves.  If the attributes were only used for `SourceRef`, then removing them makes sense.  Otherwise, I worry that weird things could start happening.  But I guess since all the checks pass, it must be fine.

samm82:
@JacquesCarette But isn't the end goal to remove `Attibutes` completely?

JacquesCarette:
We are removing `Attributes` as an implementation methodology, yes. But right now, we're removing things incrementally, while trying to keep everything working.  So there are still uses of `Attributes` which have yet to be removed. These should continue to work, until they are removed.

JacquesCarette:
Sorry about the extra work.  No, don't add them there.  Especially as they *will* be removed soon!

samm82:
So the only thing we're waiting on is `sourceRef`, right?

JacquesCarette:
I haven't kept up with all the details of #537 so @szymczdm will have to chyme in here.

JacquesCarette:
You meant to ask "add them back", not "add", which threw me off.

Here is a different way to look at it: look at all uses of `mkDataDef`, and if none of them pass in any attributes, then there is no need to add them back.

code/Example/Drasil/GlassBR/DataDefs.hs
```diff
@@ -42,7 +42,7 @@ risk_eq = ((sy sflawParamK) /
 
 -- FIXME [4] !!!
 risk :: QDefinition
-risk = mkDataDef' risk_fun risk_eq (aGrtrThanB +:+ hRef +:+ ldfRef +:+ jRef) [sourceref $ S "[4]"]
+risk = mkDataDef' risk_fun risk_eq (aGrtrThanB +:+ hRef +:+ ldfRef +:+ jRef) [] [sourceref $ S "[4]"]
```
JacquesCarette: This `sourceref $ S "[4]"` is the kind of thing we'll also want to change, at least eventually. I'll let it go on this PR, it can be caught in a later pass.

PR #553, opened 2018-05-24
Changed type of RefName from Sentence to String

niazim3:


code/Language/Drasil/Chunk/Attribute/ShortName.hs
```diff
@@ -0,0 +1,6 @@
+{-# Language TemplateHaskell #-}
+module Language.Drasil.Chunk.Attribute.ShortName where
+
+import Language.Drasil.Spec (Sentence(..))
+
+data ShortNm = ShortName Sentence
```
JacquesCarette: Shouldn't this also be changed for consistency?

code/Language/Drasil/Printing/Import.hs
```diff
@@ -261,7 +261,7 @@ createLayout sm = map (sec sm 0)
 
 -- | Helper function for creating sections at the appropriate depth
 sec :: HasSymbolTable ctx => ctx -> Int -> Section -> T.LayoutObj
-sec sm depth x@(Section title contents _) =
+sec sm depth x@(Section title contents _ _) = --FIXME: should ShortName be used somewhere?
```
JacquesCarette: Yes, something here needs changed. But it can be done later. @Mornix is building expertize on this, so he's probably in the best place to do so.  Please file an issue though.

code/Language/Drasil/Chunk/Attribute/ShortName.hs
niazim3: Yes, updated in commit 02bbeb148.

PR #570, opened 2018-05-31
Replace RefName completely

niazim3:
- Replaced `RefName` alias with `ShortName` one
- moved instance definitions for `HasShortName` for the chunks into their own files to avoid import cycles


JacquesCarette:
I was also hoping to merge this, but it will have to wait until I have a bit more time.

niazim3:
This can be closed since this was merged into #575's branch, which has been merged into master.

Issue #562, opened 2018-05-28
[Design] De-embedding ALUR and Friends


Mornix:
## Rationale
Drasil knows too much about `AssumpChunk`, `ReqChunk`, and `Changes`. This propagates through most of the Language. This should be abstracted away.

### Introduction
This proposal is considering three different chunks: `AssumpChunk`, `ReqChunk`, and `Change`(Chunk), with primary design being driven by `AssumpChunk`.

### `AssumpChunk` Definition
```haskell
data AssumpChunk = AC 
    { _aid :: String
    , assuming :: Sentence
    , _refName :: RefName -- HACK for refs?. No spaces/special chars allowed
    , _attribs :: Attributes
    }
```
Where `aid` becomes the `uid` for `HasUID`.
`assuming` is the assumption.
`refName` is used to allow "links" to the assumption.
`attribs` is used to store a list of `Attributes`. This field is going to be largely ignored in its current state because of its removal (#537), and instead `shortname` will be considered as that is all `attribs` is used for in these chunks. 
`shortname` is the displayed text that is hyperlinked in the resulting document.

### Similarities
Both `ReqChunk` and `Change` contain similar fields as `AssumpChunk`, however both add a "switch" to toggle functional/nonfunctional and likely/unlikely respectively.

### Field Uses
To make this cleaner, the "switch" fields are going to be ignored from `ReqChunk` and `Changes`, and common names will be introduced for the remaining fields for the rest of this document.

#### Fields
`id` = `aid` (`AssumpChunk`), `id` (`ReqChunk`), `id` (`Changes`)
`body` = `assuming` (`AssumpChunk`), `requires` (`ReqChunk`), `chng` (`Changes`)
`refName` is named the same across all three.
`shortname` is an adaptation of `Attributes` across all three.

#### High Level Use
`id` is used both to implement `HasUID` and equality of chunks.
`body` stores a `Sentence` which becomes the "body" of the generated output.
`refName` becomes the hyperlink the backends use to reference a given chunk. There are restrictions on this field as indicated by the comment in the `AssumpChunk` definition above.
`shortname` becomes the text which `refName` will hyperlink.

#### `AssumpChunk`
`id` and `refName` seem to always be instantiated to the same value.
[...]

PR #571, opened 2018-05-31
Change instances of String to UID.

Mornix:
This pull request makes progress towards Issue #562, specifically the first bullet point outlined by @JacquesCarette in [comment 6](https://github.com/JacquesCarette/Drasil/issues/562#issuecomment-393544664).


Mornix:
Oops, missed reverting those. Good catch!

PR #573, opened 2018-06-04
Removed uses

elwazana:
This was already done, however, due to a merge the changes were reset so this PR is just to revert the changes.


elwazana:
@niazim3 and I decided to close this pull request and will remove uses from Attribute when we remove Attribute in general.

PR #606, opened 2018-06-08
General System Description Intro Update

niazim3:
As mentioned in smiths/caseStudies#91, the first sentence under the general system description section needed to be updated to be more accurate.
This change was quite easy to make.

Working on updating stable to reflect the appropriate changes so that the logs are not non-empty.


smiths:
@niazim3 the code looks good to me.  My understanding that when you make this update all of the examples will be fixed.  Assuming I am right, this is a nice example for @szymczdm to include when writing/promoting Drasil.

niazim3:
GamePhysics actually does have the same introduction and that section. It was not updated because of some duplication that went undetected at the time of the previous commits.

Issue #569, opened 2018-05-31
Making Import Statements more Specific


samm82:
As a time-filler task, we can make the import statements of files more specific - ie. don't import a whole module if the file only uses one or two functions from it. We should coordinate this so we don't miss files/go over files twice.

Also, make a separate branch/PR for **each** import statement in **each** file.

**Note:** Certain files in `Language.Drasil` need to be ignored until #558 gets finished and merged.

---
# TODO:
## Language.Drasil
- [x] NounPhrase.Core
- [x] URI.AST
- [x] ChunkDB
- [x] Classes
- [x] Config
- [x] Document
- [x] Expr
- [x] Format
- [x] Misc
- [x] NounPhrase
- [x] People
- [x] Reference
- [x] RefTypes
- [x] Setup
- [x] Space
- [x] Spec
- [x] Symbol
- [x] SymbolAlphabet
- [x] SystemInformation
- [x] Unicode

### Chunk @elwazana
- [x] Concept.Core
- [x] Constrained.Core
- [x] AssumpChunk
- [x] Attribute
- [x] Change
- [x] Citation
- [x] CommonIdea
- [x] Concept
- [x] Constrained
- [x] DefinedQuantity
- [x] Derivation
- [x] Eq
- [x] GenDefn
- [x] Goal
- [x] InstanceModel
- [x] NamedIdea
- [x] PhysSystDesc
- [x] Quantity
- [x] References
- [x] Relation
- [x] ReqChunk
- [x] ShortName
- [x] SymbolForm
- [x] Theory
- [x] UncertainQuantity
- [x] Unital
- [x] Unitary
- [x] UnitaryConcept
- [x] VarChunk

### ~~Expr~~ @niazim3
- [x] Extract
- [x] Math
- [x] Precedence

## Data.Drasil
- [x] Software.Products
- [x] Citations
- [x] Constraints
- [x] People
- [x] Phrase
- [x] SentenceStructures
- [x] Utils

### ~~Concepts~~ @samm82
- [x] Computation
- [x] Documentation
- [x] Education
- [x] Math
- [x] PhysicalProperties
- [x] Physics
- [x] Software
- [x] SolidMechanics
- [x] Thermodynamics

### ~~Quantities~~ @samm82
- [x] Math
- [x] PhysicalProperties
- [x] Physics
- [x] SolidMechanics
- [x] Thermodynamics

### ~~Units~~ @niazim3
- [x] PhysicalProperties
- [x] Physics
[...]

PR #625, opened 2018-06-11
Improved imports in Data.Units files

niazim3:
#569.

### DIFF
--- a/code/Data/Drasil/Units/Chemistry.hs
+++ b/code/Data/Drasil/Units/Chemistry.hs
@@ -1,7 +1,6 @@
 module Data.Drasil.Units.Chemistry where
 
-import Language.Drasil
-import Data.Drasil.SI_Units
+import Data.Drasil.SI_Units (mole, kilogram)
 
 
 densityU :: DerUChunk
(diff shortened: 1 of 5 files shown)


code/Data/Drasil/Units/Chemistry.hs
```diff
@@ -1,7 +1,6 @@
 module Data.Drasil.Units.Chemistry where
 
-import Language.Drasil
-import Data.Drasil.SI_Units
+import Data.Drasil.SI_Units (mole, kilogram)
```
niazim3: This file's contents are not used anywhere at the moment, which is why removing import of "Language.Drasil" didn't break anything. (That, and multiple definitions of `molarityU` on lines 8 and 9).

Can this file (i.e. Data.Drasil.Units.Chemistry) be removed? Or should it be treated as though it is being used somewhere (i.e. reinsert import, rename `molarityU` variables)  @JacquesCarette @szymczdm

szymczdm: This file should not be removed as it's relevant information (for a different domain -- we had an example that was chemistry related, but adding it has taken a backseat for the time being).

The multiple declarations of molarity are a problem. I'm not sure why they're both here with the same name and uid.

niazim3: I can update that and this PR will be ready for merge.

Issue #629, opened 2018-06-12
WorkFlow for issue #567


elwazana:
This is just the workflow @niazim3, @samm82 and I discussed (for issue #567) 

## Label
- [x] Create Label type

## isLabel
- [x] Create isLabel

## LabelledContent
- [x] Create LabelledContent type (MUST have its own uid and shortname separate from the label within it)
- [ ] Create smart constructors for LabelledContent

## Label all the things
- [ ] Add Label to all applicable Chunk types (MUST keep uid and shortname of the chunk type)
- [ ] Create ShortName lens

niazim3:
@szymczdm @JacquesCarette 
Clarification request: It says here that a labelled content "MUST have its own uid and shortname separate from the label within it". Is this true for the shortname?

Since a shortname is something a human will be seeing, must a label have a shortname as well? Or does it make sense to do one of the following instead:

1. label only has a UID and a label type (no shortname); labelled content has a label, its own UID, the actual contents, and maybe a shortname
2. label has a UID, a label type, and a shortname; labelled content has a label, its own UID, and the actual contents

Does it make a difference if either is implemented?

JacquesCarette:
I believe that the shortname for LabelledContent can be the same as for the content. So that option 2 above seems to make most sense to me.

szymczdm:
Option 2 makes the most sense to me as well.

niazim3:
Thanks for the input. Made changes in relation to that in the following commit range: d63663e52a31e0c6807088cc3eda83d6797fe6f6...99366f8156a15ddab3c114f1707f64fdea0b2b4b.

niazim3:
(I will be asking the following in today's upcoming meeting, but it may be useful to have it noted down here as well.)

In [PR #677's review](https://github.com/JacquesCarette/Drasil/pull/677#pullrequestreview-129645822), it was mentioned that the `RefAdd` from the Table (i.e. `Contents`) can most likely be pulled out into the `LabelledContent` constructor.

Such a change will influence anything that depends on the use of the `Referable` class and `HasShortName` class; which means that there needs to be a systematic approach to the remaining bits of this issue to hopefully avoid large PRs.

@szymczdm @JacquesCarette 
Would it be possible to get some feedback on the following workflow for labelled content?

1. create `LabelledContent` to have a label, UID, contents, where label has a UID of its own, a label type, and a maybe shortname (done)
2. leave reference addresses in `Contents` as they are (there will be some duplication in the next few stages)
[...]

Issue #567, opened 2018-05-30
Create and assign issues for Label, isLabel, and ContentChunk implementation


niazim3:
A lot of design decisions were discussed in yesterday's all hands meeting.

### `Label` Module
- create abstract data type `Label` which has a `uid`, and an enumerated type (`reference address`, `meta-link`, `uri`) and is not "attached" to anything
- `shortname` (external) is the display for a `reference address` (internal)
- multiple `mkLabel` constructors should be created for label creation
- make several `refer` functions that `:: <ctx> -> Label -> Reference` where `ctx` is some contextual information that may be useful in displaying the reference. So `refer` is to be used 'close to rendering'.
- take a label to build  a shortname in the Label module
- for anything that's referable (since anything that's referable should have a label), update smart chunk constructors to 
```
uid of chunk -> _ -> _ -> ... -> Label (uid) -> Chunk 
```
i.e. what should be passed in to the smart constructor should be a `Label` but what should be stored in the Chunk is its `uid` (only).

### `isLabel`
- create an `isLabel` class which will be associated with rendering strings
- no methods for now
- should be a sub-class of `HasUID`.
- can create an `HasLabel` class for those things which "have a label".  `isLabel` would then just be `HasUID` and `HasLabel` (i.e. no need for the sub-classing).

### `ContentChunk`
- `ContentChunk` creation as a way of getting around the problems that were popping up in #537 (i.e. `shortname` for `Contents` was resulting in errors)
- build off of the current `Contents` type by appending to it a (`RefAdd`) id and a `Maybe ShortName`?

Please feel free to update this post with notes that elaborate or fix any misunderstood comments made.

JacquesCarette:
Edited for clarity.  Should this remain assigned to @szymczdm and myself as a 'design'?  If so, it can likely be closed... but if it is for implementation, then I think the assignees should change.

niazim3:
The contents of this issue are mainly documenting discussed design, however, can this issue be closed once the above tasks are assigned and issues are created for `Label`, `isLabel`, and `ContentChunk`? Or should we just be assigning work to ourselves from here?

JacquesCarette:
I think we should somehow capture the fact that this issue contains valuable design information. On the wiki? Or a `DESIGN.md` file?  Then issues created & assigned for each, and then this can be closed.

niazim3:
[...]

PR #641, opened 2018-06-13
Create Label ADT and some constructors

niazim3:
This is for issue #629.

In this PR, the data type `Label` is created and some `mkLabel` constructors as well (which corresponds to the 1st and 3rd points in #567 under Label):
"
- create abstract data type `Label` which has a `uid`, and an enumerated type (`reference address`, `meta-link`, `uri`) and is not "attached" to anything
...
- multiple `mkLabel` constructors should be created for label creation
".

@szymczdm @JacquesCarette Would the next step for these issues be to create the `isLabel` class or make several of the `refer` functions in this module?

Edit: After thinking about it, I think both tasks can be done in parallel... since nothing is implemented yet in the examples, it should be alright starting with either or...

### DIFF
--- a/code/Language/Drasil/Label.hs
+++ b/code/Language/Drasil/Label.hs
@@ -1,0 +1,17 @@
+{-# Language TemplateHaskell #-}
+module Language.Drasil.Label (Label) where
+
+import Control.Lens (makeLenses)
+import Language.Drasil.UID (UID)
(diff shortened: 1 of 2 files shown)


JacquesCarette:
At this level, this seems to make sense.  We'll only really know once we start using it though.

PR #643, opened 2018-06-13
Created isLabel and HasLabel classes

niazim3:
This is for issue #629.

In this PR, the classes `IsLabel` and `HasLabel` are created, addressing the `isLabel` section under in #567.

@JacquesCarette @szymczdm Is it not desirable to have classes defined anywhere other than the Classes.hs file? I had to move the definitions into the Label.hs file due to cyclic imports (i.e. data type `Label` needs a `UID`, which means `Label` is an instance of `HasUID`, which needs to be imported from Classes.hs; this meant that defining the new classes in Classes.hs wouldn't work because in order to lens to a `Label`, the data type `Label` needs to be recognized in that file (i.e. Label.hs needs to be imported into Classes.hs)).


JacquesCarette:
The convention that I have used is to create `Language.Drasil.Label.Core` that defines the types and requires minimal imports.  Then `Language.Drasil.Label` can re-export that, as well as various utilities.

i.e. your `Label` would become `Label.Core` and your `LabelHelper`, `Label`.

niazim3:
I think this PR is ready for merge.

PR #649, opened 2018-06-14
Improved imports in Chunk/Constrained.hs as per issue #569

elwazana:


elwazana:
@JacquesCarette I've run this commit a few times now and works fine every time, I've asked @niazim3 to run it on her system and it runs fine there as well. So I'm not sure why Travis is failing.

JacquesCarette:
You must have been unlucky.  Builds fail like that because of weird network issues once in a while. I've restarted the build, and it worked.

PR #637, opened 2018-06-12
Updating Display of Uncertainties in the Examples

niazim3:
As mentioned in smiths/caseStudies#46, it would be preferred that the uncertainties of UncertainQuantities show up as a percent instead of a decimal between 0 and 1.

PR #667, opened 2018-06-14
Bw0And1 Creation

niazim3:
As @JacquesCarette mentioned in comments of PR #637, it would be beneficial to have the constructors for uncertainty to ensure that uncertainty is between 0 and 1, and throw an error otherwise.

Tested with values 0 and 5 threw the error (desired result).


niazim3:
Not sure why the build isn't working...

JacquesCarette:
About 5% of the builds seem to do weird things - looks like networking on the travis-ci servers just drop.  But once you kick it to restart, it usually passes.

PR #671, opened 2018-06-14
Update infrastructure of References to support ConceptChunks.

Mornix:
This is the second of (probably) a few pull requests to enable/support referencing of `ConceptChunk` s as per #562 [comment 6](https://github.com/JacquesCarette/Drasil/issues/562#issuecomment-393544664), third bullet point. 

This commit adds the necessary infrastructure to `RefDB` to facilitate the use of `ConceptChunk` s. 
This commit does not include being able to insert `ConceptChunk` s into the database. Nor does it add `Referable` support.

### DIFF
--- a/code/Language/Drasil/Reference.hs
+++ b/code/Language/Drasil/Reference.hs
@@ -3,12 +3,13 @@ module Language.Drasil.Reference where
 
 import Control.Lens ((^.), Simple, Lens, makeLenses)
 import Data.Function (on)
-import Data.List (partition, sortBy)
+import Data.List (concatMap, groupBy, partition, sortBy)
 import qualified Data.Map as Map
 
 import Language.Drasil.Chunk.AssumpChunk as A (AssumpChunk)
 import Language.Drasil.Chunk.Change as Ch (Change(..), ChngType(..))
 import Language.Drasil.Chunk.Citation as Ci (BibRef, Citation(citeID))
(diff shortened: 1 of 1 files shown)


code/Language/Drasil/Reference.hs
```diff
@@ -88,11 +93,25 @@ changeMap cs = Map.fromList $ zip (map (^. uid) (lcs ++ ulcs))
...
-        scs = sortBy citeSort cs
+        scs = sortBy uidSort cs
         -- Sorting is necessary if using elems to pull all the citations
         -- (as it sorts them and would change the order).
         -- We can always change the sorting to whatever makes most sense
 
+conGrp :: ConceptChunk -> ConceptChunk -> Bool
+conGrp a b = (cdl a) == (cdl b) where
+  cdl :: ConceptChunk -> UID
+  cdl x = if length (x ^. cdom) == 1 then head $ x ^. cdom
```
JacquesCarette:
```haskell
cdl x = the_dom $ x ^. cdom
  where
    the_dom [d] = d
    the_dom ds = error $ "Expected ConceptDomain for: " ++ (x ^. uid) ++
           " to have a single domain, found " ++ (show $ length ds) ++
           " instead."
```
Basically `head` should be avoided at all cost (same with other partial operations), even when guarded as you did. It is better to check for length 1 structurally.

PR #693, opened 2018-06-20
Created instances for LabelledContent as per issue #629 task 3

elwazana:


code/Language/Drasil/Chunk/ShortName.hs
```diff
@@ -9,6 +11,7 @@ class HasShortName  s where
                             -- A short name used for referencing within a document that can 
                             -- include symbols and whatnot if required.
                             -- Visible in the typeset documents (pdf)
+  shortNm :: Lens' s ShortName
```
JacquesCarette: We really don't need both `shortname` and `shortNm`. Especially since `shortname` should always be equal to `view shortNm`.

elwazana: @JacquesCarette Hello Dr. Carette, yeah I realized that having both is unnecessary but creating an instance of `HasShortName` is much easier with a Lens.
In that case, should I remove `shortname` or would you prefer to keep the getter?

JacquesCarette: I would then make 'shortname' be that Lens.

elwazana: Since I'll be removing the getter function and only have a Lens, should I go through in a separate branch first and add a `Label` to all the Chunk types?

JacquesCarette: You should not be adding a `Label` to *all* the chunk types.  Only the ones that need it.  Certainly all the ones that currently have a short name do. But that's probably it.

Having said that, yes, it is probably best to first switch from explicit short name to using label. And then only labels should have short names.

elwazana:
@JacquesCarette I created #695 to deal with switching from a getter to a Lens functions for `HasShortName`, so once that's done `HasShortName` will only have a Lens which will fulfil the requested change you asked for here.

szymczdm:
I see #695 is still open. Does that mean this PR is still waiting for the pending changes to be done?

elwazana:
@szymczdm Hello Dan, no the requested changes haven't been completed yet. Switching to a Lens only led to requiring the addition of `Label` to all `Chunk` types that previously had an independent `ShortName`. Now that the yuzhi branch is merged the addLabels branch can continue doing the stated above, however, because `Label` was added to a lot of types the majority of `drasil-example` is broken and needs updating.

@niazim3 is currently working on the addLabels branch and slowly working through the changes.

elwazana:
@JacquesCarette I believe this PR can be closed, with the planned implementation of `LabelledContent`, `UnlabelledContent`, `Contents`, and `RawContent` the changes here have become obsolete and unneeded.  The new design for `LabelledContent` can be followed here: https://github.com/JacquesCarette/Drasil/tree/updateContentsDesign

JacquesCarette:
Thanks for coming back to this, and indicating that this is now obsolete.

PR #694, opened 2018-06-21
Added Unlikely Changes to GlassBR

samm82:
As per smiths/caseStudies#100, an unlikely changes section was added to the SRS of GlassBR.
A Changes.hs was added to handle both likely and unlikely changes.


code/Example/Drasil/GlassBR/Body.hs
```diff
@@ -575,8 +576,13 @@ s7_1_req6 = [(Enumeration $ Simple $ [(acroR 6, Nested (titleize output_ +:+
 
 {--LIKELY CHANGES--}
 
-s8_list :: [Contents]
-s8_list = likelyChanges_SRS 
+s8a_list :: [Contents]
```
smiths: What does the 8 stand for?  Section 8?  If so, that isn't a great name.  All of the sections can move over time and the numbering is likely to change.  What about sLC_list and sUC_list?  I see LC and UC used elsewhere in the naming; I assume to mean Likely Changes and Unlikely Changes.

samm82: Yes and yes - should we make an issue for more descriptive section names?

smiths:
With the exception of the naming, which I noted in a comment above, the proposed changes look okay to me.

code/Example/Drasil/GlassBR/Body.hs
smiths: An issue sounds like a good idea to me.

samm82:
This can be merged now, as I made the section name fix in #708 to this branch

PR #697, opened 2018-06-21
Add UC to Tables of Acronyms and Abbreviations

niazim3:
Fixed issue in the comments of smiths/caseStudies#100, and updated stable in all examples.

### DIFF
--- a/code/Example/Drasil/GamePhysics/Concepts.hs
+++ b/code/Example/Drasil/GamePhysics/Concepts.hs
@@ -2,7 +2,7 @@ module Drasil.GamePhysics.Concepts (centreMass, twoD, chipmunk, cpAcronyms) wher
 
 import Language.Drasil
 import Data.Drasil.Concepts.Documentation (assumption, dataDefn, genDefn, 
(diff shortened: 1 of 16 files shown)


niazim3:
Should the table of abbreviations be automated like the table of symbols currently are? So that we won't have extra abbreviations or missing ones in the table and wouldn't need to make manual changes such as these... @szymczdm

szymczdm:
Ideally, yes. I believe @halonazhao's work with parsing the document for units, symbols, and whatnot could also be applied to the acronyms (though it might be a bit tricky), so that's definitely something to keep in mind going forward.

niazim3:
Should an issue be made for that then? I don't currently see any in relation to acronym/abbreviation extraction from the doucment...

szymczdm:
@niazim3 yes, an issue would be a great idea. Could you please create one? I think using the `enhancement` tag would also be appropriate.

niazim3:
Issue made; assigned @halonazhao for now; if anyone else should be assigned please let me know.

Issue #703, opened 2018-06-25
Class instance cheat-sheet now out of date


szymczdm:
The file [DataTypes.xlsx](https://github.com/JacquesCarette/Drasil/blob/master/People/Dan/DataTypes.xlsx) in the Dan folder is now (almost certainly) out of date. Can you please update/correct it?

We'll also likely want to move that information onto a Wiki page at some point (just having the table should be enough), though auto-generating it would be ideal if it's easy enough.

Whichever of you'd like to take this on, please feel free.

elwazana:
@szymczdm Hello Dan, I've started working on this and I'm just wondering if I should add `Label` and `LabelledContents`? The reason I'm asking is that neither of those data types are in the `Chunk` folder and everything in there before editing is.

szymczdm:
It should be there once the implementation is complete.

elwazana:
So should I put off doing it for now?

szymczdm:
Yeah, just leave this issue open and come back to it once #629 / #567 are done.

samm82:
@szymczdm What should be done about data types that aren't in the Chunk folder? Should I change the column to contain the name of the folder, instead of just a yes/no if it's in the Chunk folder?

szymczdm:
@samm82 that sounds like a great idea!

samm82:
PR open in #1847. I decided not to close this issue, since moving it to the Wiki/auto generating it could be done.

smiths:
Yes, autogenerating DataTypes.xlsx makes sense, as does posting it on our web-page as part of the Drasil documentation.

PR #706, opened 2018-06-25
WIP Updating instance class spread sheet

elwazana:
This is still being worked on and is waiting for `Label` and `LabelledContents` to be fully implemented. As per issue #703.


JacquesCarette:
So this shouldn't be merged until the latter is done, or can it be added now and updated later?

elwazana:
@JacquesCarette This can be merged as is, I just put W.I.P. to notify that it will need updating once `Label` and `LabelledContents` are implemented.

Issue #701, opened 2018-06-22
Overview of building .tex files for the examples


niazim3:
When building the latex documents for the examples, the following problems come up:

- images not found (which makes sense as only GlassBR has a filepath that allows access to images)
- undefined control sequence for 
    - `\tableofcontents`  (all examples)
    - `^{\circ}$C & temperature (centigrade)`
- `Missing } inserted.` messages
- `Missing \endgroup inserted.`
- overall inability to generate a PDF (SSP, SWHS)

Issues should be made to work on avoiding these messages, errors and warnings.

Details below:
***
## Chipmunk, GlassBR:
```
! Undefined control sequence.
<recently read> \

l.25 \tableofcontents
```
```
Underfull \vbox (badness 10000) detected at line 50
! Undefined control sequence.
<recently read> \

l.195 \end{longtabu}
```
[Chipmunk_SRS.pdf](https://github.com/JacquesCarette/Drasil/files/2128818/Chipmunk_SRS.pdf)
is the resulting file, which has an empty Contents section, an empty References section, and ?? links to the referenced sections and contents.

***
## NoPCM
(above warnings) and
```
! Undefined control sequence.
<recently read> \

l.45 ${
     }^{\circ}$C & temperature (centigrade)
```
```
LaTeX Warning: Citation 'parnasClements1986' on page 4 undefined on input line
182.

[4]

LaTeX Warning: Reference `Sec:UserChars' on page 5 undefined on input line 188.



LaTeX Warning: Reference `Sec:References' on page 5 undefined on input line 191
.


LaTeX Warning: Reference `Sec:IMs' on page 5 undefined on input line 191.


LaTeX Warning: Reference `Sec:IMs' on page 5 undefined on input line 192.


LaTeX Warning: Reference `Figure:SysCon' on page 5 undefined on input line 198.



LaTeX Warning: File `SystemContextFigure.png' not found on input line 201.


! Package luatex.def Error: File `SystemContextFigure.png' not found.

See the luatex.def package documentation for explanation.
Type  H <return>  for immediate help.
 ...

l.201 ...[width=\textwidth]{SystemContextFigure.png}

```
```
(load luc: C:/Users/Owner/AppData/Local/MiKTeX/2.9/luatex-cache/generic/fonts/o
tl/lmmono12-regular.luc)

LaTeX Warning: Reference `Figure:SysCon' on page 5 undefined on input line 202.



[...]

PR #717, opened 2018-06-27
commented out/removed Breqns

niazim3:
As discussed in #701, removing `Breqns` for now may be a good idea to get the pdfs to generate for the SWHS and SSP examples (along with removing some of the warnings coming up in the other examples).


JacquesCarette:
Apparently this change has also caused 'tiny' to break, so this can't be merged quite yet.

PR #731, opened 2018-06-28
Added Unlikely Changes into SWHS

elwazana:
@JacquesCarette I just wanted to get these changes reviewed before I update stable to match, just in case there is something you would like to change.


JacquesCarette:
And now the build fails?

elwazana:
@JacquesCarette it seems like `drasil.cabal` wasn't being tracked on my system so changes I've made there weren't committed with the rest, I just added `drasil.cabal` to my tracking list.

JacquesCarette:
What are all those cabal files in the `drasil.Cabal` directory?  I know that @samm82 is working on something 'like' that, but this seems weird.

elwazana:
@JacquesCarette I don't know what happened. I reset to the earliest commit I could find that didn't have the `drasil.Cabal`, which I didn't add, and that gets rid of all my additions because I wasn't tracking the renamed file for the Changes subsections.

elwazana:
@JacquesCarette Hello Dr. Carette, I've been trying to fix the conflicts that @niazim3 just fixed but every time I tried to merge in master it told me I was already up-to-date. 
image

I was wondering if you know why it was breaking on my system?

JacquesCarette:
What does `git status` give you?  Also did you first do
```
git checkout master
git pull
```
before switching back to your branch?  [Likely yes, but sometimes it's easy to forget to do a step, and then everything gets weird].

elwazana:
@JacquesCarette This is what I get when I run `git status`:
image

As for `git checkout master` then `git pull` if I remember correctly I did do this before switching onto `addLabels` but I may have forgotten and never realized it as you suggested.

JacquesCarette:
So if you now try to `git merge master`, what do you get?

Also, you should figure out what to do with those untracked files.

elwazana:
Ok I think its fixed because I got a bunch of merge messages

PR #754, opened 2018-07-05
Remove Import files no longer in use

niazim3:


szymczdm:
That makes a lot more sense as to why some functions (like `layout`) looked like they were in use, but really weren't.

PR #757, opened 2018-07-05
References now sorted by Author

samm82:
- The references aren't sorted by author anymore, but by last name, then by first name, and finally by year. 
  - A class `HasFields` was created with `getFields` to extract the fields from a `Citation`.
  - `getAuthor` and `getYear` functions were created to assist in sorting.
  - `Citation` and `People` were both made instances of `Ord` so that the `sort` function could be used instead of `sortBy`.
- Hardcoded references have had their number changes to align with the new sort, and comment added indicating which reference is being hardcoded (to make it easier to remove the hardcoding in future).
- Stables have been updated to match the new sort and hardcoding.

Can you make sure that the sorting is as desired, and should I add a sort by title as a last resort?


samm82:
Wait on this - merging SubPackages broke this

samm82:
Closed for now

PR #783, opened 2018-07-06
Improved imports in LanguageRenderer.hs

elwazana:
As per issue #569.

@JacquesCarette Hello Dr. Carette, I tried to explicitly import things from `AST.hs` in this file but there were a lot of things imported over from there so I've left it as is and will keep it un-checked in issue #569 so that I can come back to it.


JacquesCarette:
That's fine - `AST` is huge, and almost all of it is used.  A comment to that effect would be good.

Issue #729, opened 2018-06-28
Sort References by Author


samm82:
In GlassBR, as noted in the comments of #721, references are sorted by UID instead of by author. As the convention that UIDs should be of the form "authorYYYY" is not enforced, we need to make sure that correct sorting takes place.

Also, if two references have the same author, should they then be sorted by year or by title? 
@smiths @JacquesCarette

samm82:
So `Citation` is defined here:

https://github.com/JacquesCarette/Drasil/blob/fc388fc5cb9e30166a0feca9e55dd80e86d75ac3/code/Language/Drasil/Chunk/Citation.hs#L93-L98

and the sorting happens here:
https://github.com/JacquesCarette/Drasil/blob/fc388fc5cb9e30166a0feca9e55dd80e86d75ac3/code/Language/Drasil/Reference.hs#L95-L101
https://github.com/JacquesCarette/Drasil/blob/fc388fc5cb9e30166a0feca9e55dd80e86d75ac3/code/Language/Drasil/Reference.hs#L266-L267

I'm going to make an `authorSort` similar to `uidSort`. I'm going to make a class `HasFields` with a lens to pull out the `fields` from `Citation`; how should I go about extracting the author from the list of `CiteField`?

Can I proceed with this? @JacquesCarette @szymczdm @smiths

smiths:
Yes, you should proceed with this task.  You will have to wait for @JacquesCarette and @szymczdm for specific advice on your question about `CiteField`.

JacquesCarette:
There are combinators somewhere to grab the author from the `fields`. It should be an error if there isn't one.  I guess you can start by looking in the printing code. The combinator might not be there, but it should be used there (at least).  Secondary sort key should be year.  [And we can later make that a parameter too!]

samm82:
I added HasFields in 2621a9a, but I can't seem to find a combinator that grabs the author... What exactly am I looking for @JacquesCarette?

JacquesCarette:
Still looking for the right combinator.  But I did find `compCiteField` in L.D.HTML.Print which seems like a useful function for sorting.

JacquesCarette:
Looks like the kind of filter on `[CiteField]` that I was thinking of is no longer in use. Or it might have been on another related type.

samm82:
Now I have it to sort by `People`, but it cannot deduce (Ord Person). I need to sort by surname, but I can't seem to figure out a way to do that. Should I make a class `HasSurname`?

[...]

PR #805, opened 2018-07-10
Implemented better sorting of references

samm82:
As per #729, instead of being sorted by UID, references are now sorted with the `compareAuthYearTitle` function, that sorts first by author, then by year, then by title. This is done in a way that isn't hardcoded (via using `Ord` with `Person`, etc.) so that we can easily change the criteria for sorting the references.


code/drasil-lang/Language/Drasil/Reference.hs
```diff
@@ -271,8 +273,37 @@ instance Referable Contents where
...
+        getStr :: Sentence -> String
+        getStr (S s) = s
+        getStr ((:+:) s1 s2) = getStr s1 ++ getStr s2
+        getStr _ = error "Term is not a string" 
+
 citationsFromBibMap :: BibMap -> [Citation]
-citationsFromBibMap bm = sortBy uidSort citations
+citationsFromBibMap bm = sortBy compareAuthYearTitle citations
   where citations :: [Citation]
         citations = map (\(x,_) -> x) (Map.elems bm)
```
JacquesCarette:
```haskell
citations = map fst (Map.elems bm)
```

code/drasil-lang/Language/Drasil/Reference.hs
```diff
@@ -271,8 +273,37 @@ instance Referable Contents where
...
+        isAuthor _          = False 
+
+getYear :: (HasFields c) => c -> Int
+getYear c = maybe (error "No year found") (\(Year x) -> x) (find (isYear) (c ^. getFields))
+  where isYear :: CiteField -> Bool
+        isYear (Year _) = True
+        isYear _        = False
+
+getTitle :: (HasFields c) => c -> String
+getTitle c = getStr $ maybe (error "No title found") (\(Title x) -> x) (find (isTitle) (c ^. getFields))
```
JacquesCarette: you don't need () around `isTitle`.

samm82: I'm assuming this applies to `isAuthor` and `isYear` as well

JacquesCarette: Right, function names do not need to be wrapped in ().

samm82:
@JacquesCarette I made the proposed changes

PR #889, opened 2018-07-13
Minor GlassBR Changes

samm82:
Some various improvements to GlassBR in Drasil, such as:
- Fixed stakeholder capitalization
- Used AR for calculating J
- Added refrence to P<sub>b</sub>
- Added ln to Drasil and in DD9
- Changed TNT typical value to 1.0
- Removed last sentence of Data Constraints section


code/drasil-example/Drasil/GlassBR/Unitals.hs
```diff
@@ -337,7 +337,7 @@ notSafe       = dcc "notSafe"     (nounPhraseSP "not safe")
 probBreak     = cc prob_br
   ("The fraction of glass lites or plies that would break at the first " ++
     "occurrence of a specified load and duration, typically expressed " ++
-    "in lites per 1000.")
+    "in lites per 1000 [3]." {-astm2016-})
```
szymczdm: We really don't want hard-coded references like this `[3]`.

samm82: We can't really fix this until `Label` is fully implemented, so this is a temporary fix to (a) have the reference there and (b) track what it is with the comment so that it can be actually referenced once we have the appropriate tools.

samm82:
Should I bump the version number of `drasil-lang` because of the addition of `ln`/the version number of `drasil-example` because of these small changes? @szymczdm

szymczdm:
@samm82 yes, bump it by 0.0.1

code/drasil-example/Drasil/GlassBR/Unitals.hs
szymczdm: Hmm, alright I guess as long as the hard-coded references are clearly tracked (and not forgotten) it should be okay. I think the citation comment should be a bit more explicit, ie. `FIXME: cite astm2016`

samm82:
@szymczdm I made this pull request just to get some features merged in and because small PRs are better - should I bump the version number of `drasil-example` with every set of small changes or just once?

szymczdm:
If this is part of a bigger change I'd bump it just once for the whole thing. `drasil-example` is a bit weird in the sense it's actually a number of examples, so bumping it each time one changes might be a bit excessive (not necessarily bad). Really depends on what new features have been added in the change

samm82:
@szymczdm OK - I bumped it here, and won't again. I'm working on the Updating GlassBR SRS project - essentially just updating the SRSs of Drasil and caseStudies to match. 

If I have to change something in `drasil-lang` again though, should I bump its version?

PR #896, opened 2018-07-17
Symbol Sorting

halonazhao:
Fixed the sorting function for symbols. But there is still a potential leak of sorting function. 

So far the sorting string is all based on lower case. If giving list `q, Q` or `Q, q`, they would have different result. It's kind of depending on the input order. I searched on line, it seems that if we want it to be case sensitive, we have to add one more package. I am not quite sure whether it's a good way to be case sensitive by adding packages.

Do we put more effort to make it case sensitive? @JacquesCarette


szymczdm:
How much effort would fixing the case sensitivity problem be? If it's not much, then it's worth doing immediately. Otherwise, it'd be worth opening an issue (tagged as an enhancement) to remind us to do this later.

Also @halonazhao, can you fix the conflicts on this branch?

halonazhao:
@szymczdm Done.

halonazhao:
@JacquesCarette I have improved it to be case sensitive.

JacquesCarette:
"case sensitive" = "depends on case", "case insensitive = does not depend on case".  We want `q` and `Q` to be treated differently, so we want case **sensitivity**.

halonazhao:
The default `compare` function in Haskell is sorting all the lower case after the upper case, like `a` is after `Z`. That why every time I use `comparation` twice, once for making sure they are the same letter, once for `case sensitive`. @JacquesCarette

JacquesCarette:
That is an extremely important comment/remark, and that was not obvious from the code.  Please insert that as a comment in the code. [You can make that documentation change directly on master - I will merge this in now]. Thank you.

PR #897, opened 2018-07-17
[Workaround] Add --force-dirty to Force Examples to Reinstall

Mornix:


szymczdm:
I'm fine with this workaround until we have time to figure out what the real problem is. Can you throw up an issue (if you haven't already) that makes reference to this?

Issue #771, opened 2018-07-06
Remove Warnings in Make


samm82:
This will be a semi-ongoing issue, but the warnings while running `make` (ie. redundant imports, shadowed bindings, orphaned class instances, etc,) should be fixed. See PR #557 for some warnings that can be ignored, and note that warnings regarding `ufixme1` and `ufixme2` serve as reminders to revisit them and should be ignored.

samm82:
All warnings have been purged from `make`, and further code optimization will occur using hlint (including in issue #1249.

PR #900, opened 2018-07-17
W.I.P. removed multiple warnings

elwazana:
This PR is linked to issue: #771, there is still work to be done but this PR is small so merging shouldn't be too difficult.


niazim3:
This branch was merged into branch of #913.

Issue #321, opened 2017-06-27
Continue Refactoring of solChSpecF


hmadej:
The following has changed:
```
solChSpecF :: (NamedIdea a) => a -> (Section, Section) -> Sentence -> 
  (Sentence, **Sentence**, Sentence) -> 
  ([Contents], [Contents], [Contents], [Contents], [Contents], [Contents]) -> 
  [Section] -> Section
```

Bool has been changed to sentence, if your example has uncertainty replace the True with dataConstraintUncertainty, otherwise replace with EmptyS.

A new data type has been made to refactor solChSpecF, the new function is:

`solutionCharactersticCon :: (NamedIdea a) => a -> SOLsec -> [Section] -> Section`

a is the name of the program, [Section] is any additional subsection that we are not generating right now with this function.

it is a nested structure representing the solution characteristic subsection

SOLsec has 6 fields of type SOLsub (1 per subsection) 

to make both data types you call these functions:

```
mkSOLsec :: SOLsub -> SOLsub -> SOLsub -> SOLsub -> SOLsub -> SOLsub -> SOLsec
mkSOLsec as tm gd dd im dc = Sect as tm gd dd im dc

mkSOLsub :: Sentence -> Sentence -> Sentence -> Section -> [Contents] -> SOLsub
mkSOLsub start middle end refSection ys = Subs start middle end refSection ys
```

SOLsub are the subsections of the solution characteristic specification, each SOLsub takes a 3 sentence a Section and a list of contents. Please look at gamephys commit 9d5da47a  for more details on how to implement.

I am also looking for feedback or critique on the design and better names for the types.

hmadej:
this is related to issue #244

JacquesCarette:
6-tuples are usually not a good idea -- too easy to swap them by accident.

Better would be some kind of data-structure that contains the 'knowledge', and another which describes how to 'render' that knowledge into English (`Sentence`, `Contents`, `Section`, etc), and lastly an "assembler" which puts them together.  `solChSpecF` is kind of that assembler.

njericha:
Our current plan is to pass `[SecItem]` as the input to each section, and pass each section as the input to SolSub with "tags" shown below. SolSub would replace solCharSpecF and take care of how to display it as `Sentence`, `Contents`, and `Tables`.
```
solutionChar = SolSub (Assump [ProgName ssp, Ref imods, Cont sspAssump])
                                     (TMods [Cont sspTmods]) 
                                      ...
```
Where
[...]

PR #909, opened 2018-07-18
Added more section description languages

### LINKED ISSUE #368: Need more section description languages
We have languages for sections 1 (Reference Material), 2 (Introduction) and now references.  We still need to do the same for

- [x] Stakeholders
- [x] General System Description
- [x] Scope of the Project
- [x] Specific System Description
- [x] Requirements
- [x] Likely Changes
- [x] Traceability Matrices and Graphs
- [x] Values of Auxiliary Constants
- [x] Appendix

Once this is done `Verbatim` should be removed.

niazim3:
Closes #368.

### DIFF
--- a/code/drasil-example/Drasil/GamePhysics/Body.hs
+++ b/code/drasil-example/Drasil/GamePhysics/Body.hs
@@ -9,9 +9,9 @@ import Drasil.DocLang (DerivationDisplay(..), DocDesc, DocSection(..),
   Emphasis(..), Field(..), Fields, InclUnits(IncludeUnits), IntroSec(..), 
   IntroSub(..), RefSec(..), RefTab(..), SCSSub(..), SSDSec(SSDProg), 
   SSDSub(SSDSubVerb, SSDSolChSpec), SolChSpec(SCSProg), SubSec, TConvention(..), 
-  TSIntro(..), Verbosity(Verbose), ExistingSolnSec(..), assembler, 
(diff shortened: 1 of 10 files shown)


code/drasil-example/Drasil/GamePhysics/Body.hs
```diff
@@ -77,7 +77,8 @@ mkSRS = RefSec (RefProg intro [TUnits, tsymb tableOfSymbols, TAandA]) :
    IScope scope_of_requirements_intro_p1 scope_of_requirements_intro_p2, 
    IChar (S "rigid body dynamics") (phrase highSchoolCalculus) (EmptyS), 
    IOrgSec organization_of_documents_intro inModel solution_characteristics_specification EmptyS]) :
-   Verbatim general_system_description :
```
niazim3: GamePhysics is the only case study that relies on `assembler` to produce its sections; this `general_system_description` section was also generated using `assembler`. 
@JacquesCarette @szymczdm Is it alright to make this shift to using `GSDSec`? That is, generally, should `assembler` be used to create sections (mentioned in #321) or the section description languages that were added to issue #368?

szymczdm: `assembler` was, I believe, an ambitious attempt to do things in a generalized way, but I believe using `GSDSec` is better for consistency.

szymczdm: Create an issue to look into it. I don't remember the details on the implementation presently, since that would have an effect on what we do with it.

PR #914, opened 2018-07-19
[Small PR] Made Labels more meaningful in GlassBR

samm82:
As per smiths/caseStudies#106, labels for T1 and T2 were changed to contain useful information.


szymczdm:
@samm82 I believe some of the other PRs are causing conflicts here, can you please sort them out?

smiths:
I mentioned in comments that the Drasil names that have t1 and t2 in them are still not great names.  The order of t1 and t2 is arbitrary.

samm82:
Addressed in the addGlassDD6Notes branch (it was a small fix so I just tacked it on to the branch)

smiths:
Looks good.  I still like a prefix (like T) that tells me that this is a theoretical model, but it is best that the examples all be consistent.

samm82:
a) So do you want me to add the prefix?
b) Should I change the other examples with a similar fix?

smiths:
Hold off for now.  It is more of a stylistic preference, and we haven't really converged on a style for Drasil code.  Eventually we may want to put something in the contributor's guide related to this, but our conventions aren't yet clear.  I think we can close this issue for now.

Issue #926, opened 2018-07-20
Migrate from using `QDefinition`s for Data Definitions to using `DataDefinition` data type


niazim3:
Data definitions should be built using actual `DataDefinition`s instead of (what will eventually be) the simpler datatype `QDefinition`. In https://github.com/JacquesCarette/Drasil/commit/3b2993cfe177f88124d979fd0e6304ef27298bef#diff-d6f0b4483a86f614fa7fbae4e27fff46, `DataDefinition`s for GlassBR were created (this example specifically because the yuzhi branch had not been merged in by that time and it was meant for testing the data type).

#869 can be used as a guide on actually using the new lists of `DataDefinition`s (i.e. passing them into the `SystemInformation`). 

As mentioned in #747, the remaining examples should shift from using `QDefintiion`s to `DataDefinition`s. Please check off your task as it is completed.
- [x] SSP @niazim3
- [x] SWHS @elwazana 
- [x] NoPCM @elwazana 
- [x] GamePhysics @Mornix 
- [x] Tiny @elwazana 
---
- [x] Rename `mkDataDef` to `mkQD` or something similar, as it builds a `QDefinition`

@Mornix and @deviprasad135, if there are other issues that are higher priority, please feel free to re-assign  the examples to myself, @elwazana, or @samm82.

niazim3:
Hi @Mornix, what is the status of this issue for GamePhysics?

Mornix:
@niazim3 Totally slipped through the cracks. Should be up shortly.

PR #930, opened 2018-07-20
Propagated the use of DataDefinition into HGHC (tiny)

elwazana:
As per issue #926

### DIFF
--- a/code/drasil-example/Drasil/HGHC/HeatTransfer.hs
+++ b/code/drasil-example/Drasil/HGHC/HeatTransfer.hs
@@ -9,6 +9,9 @@ import Data.Drasil.Units.Thermodynamics (heat_transfer_coef)
 symbols :: [QuantityDict]
 symbols = map qw hghcVars ++ map qw htVars
 
+hghcVarsDD :: [DataDefinition]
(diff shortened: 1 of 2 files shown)


code/drasil-example/Drasil/HGHC/HeatTransfer.hs
```diff
@@ -31,6 +34,10 @@ cladCond     = vc "cladCond"     (cnIES "clad conductivity") (lK `sub` lC) Real
 htTransCladCool_eq, htTransCladFuel_eq :: Expr
 htTransCladCool, htTransCladFuel :: QDefinition
 
+htTransCladCoolDD :: DataDefinition
+htTransCladCoolDD = mkDD htTransCladCool [] [] ""
```
szymczdm: Why do these constructors have `[] [] ""`? Is that just because tiny is lacking a lot of information, or is it common for that to happen (and we should create a new smart constructor if that's the case)?

elwazana: The `[] [] "" `  is for the references in the source field and derivation, while the empty string at the end seems to be temporary. These fields are left empty because tiny is lacking information, the Data Definitions for tiny didn't have references or derivations so tiny is lacking information in that regard.

PR #933, opened 2018-07-23
[Small PR] Added Notes to Glass DD6

samm82:
And changed function names from using "t1" and "t2" to the more informative "pb" and "lr"


smiths:
We discussed this recently in #914.  I'm fine with the new names.  However, we might want to revisit our variable naming in the future to carry a bit more information through stylistic conventions.  For instance, we might later decide to add prefixes to variable names to give a quick mnemonic about their purpose, like T, D, I, etc.  This would be like a modified Hungarian notation:

https://en.wikipedia.org/wiki/Hungarian_notation

JacquesCarette:
Changes are fine, but there is now a conflict - can you fix please @samm82 ?

samm82:
Done 👍

PR #940, opened 2018-07-24
Added source fields to IMs according to issue #939

### LINKED ISSUE #939: Adding source field to Instance Models
As discussed in #887, the `fixme` placeholder in the generation of the InstanceModel tables is being removed and for some instance models replaced with references to the original material.

elwazana:
This is for adding a missing reference in the source field of an instance model (Probability of Glass Breakage) and other changes mentioned in issue #939 

closes #939


halonazhao:
@elwazana I have the same failure as yours. I guess the errors were caused by merging of the `sorting symbol`. I will merge the change of stable. If your log is only about the order of the symbols, you can wait until I merged. Yours error will go away.

JacquesCarette:
Thank you @halonazhao - #908 is now merged.  @elwazana if you merge master into here, that will likely fix things, and then I can merge this in too.

PR #942, opened 2018-07-24
562 Data

Mornix:
This PR is part of #562. It adds:
* A tree (chain at the moment) of domains to support functional requirements. `srsDom` -> `reqDom` -> `funcReqDom`
* Generic convenience functions to convert ConceptInstances to Contents. `mkConCC`, `mkConCC'`, and `mkConCC''`. The latter two are more "convenient" uses of the first.

### DIFF
--- a/code/drasil-docLang/Drasil/DocLang.hs
+++ b/code/drasil-docLang/Drasil/DocLang.hs
@@ -23,7 +23,7 @@ module Drasil.DocLang (
     -- Sections.ReferenceMaterial
     intro,
     -- Sections.Requirements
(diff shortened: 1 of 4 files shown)


JacquesCarette:
The master failures should be fixed -- please merge, that should allow this one to then be merged in as well.

Mornix:
You are absolutely correct, `mkConCC*` were in the wrong place for how specific they are. I moved them to DocumentLanguage. I also renamed `mkConCC''` to `mkEnumCC` since it is a targeted specialization to make `Contents` from the `Enumeration` constructor.

PR #943, opened 2018-07-24
[Small] Change argument order for cic

Mornix:
This PR is part of #562. It moves the domain argument to be the last in the `cic` constructor. 

This change is aimed to make using `ConceptInstance` more ergonomic by being able to `zipWith` over `[Concept -> ConceptInstance]`. The rational is a list of `ConceptInstance`s probably all belong to the same domain, so by being able to (easily) associate a domain to an entire list reduces inconsistencies, and helps avoid tedium while writing something like requirements or assumptions.


JacquesCarette:
The master failures should be fixed -- please merge, that should allow this one to then be merged in as well.

Issue #941, opened 2018-07-24
Printing rules for Expr multiplication and scientific notation.


szymczdm:
When we are printing out expressions involving multiplication we should follow a standard set of rules for explicitly showing the multiplication sign. We should decide on whether we want to use the dot, cross, or something else as the standard. I recommend showing the sign when:

- Multiplying anything by a number (ie. X\*2, 2\*2, etc.)
- Expressing scientific notation when not using E (essentially the same as above, since it's two numeric representations being multiplied together).

We also need to come up with (or find) a rule for how to determine when to automatically switch to scientific notation instead of showing all digits of a very large/small magnitude number. This also needs a standard presentation so we can have: n*10<sup>x</sup> or nE<sup>x</sup>, or some other variant.

@smiths and @JacquesCarette did I miss anything?

JacquesCarette:
These probably should have been 2 issues - one for Expr multiplication, and one for scientific notation.  Who was going to do this?  I now forget that part...

halonazhao:
**Rules for Multiplying:**
1. num `mul` symbol -> no `mul` symbol  eg. (2a) 
2. num `mul` num -> `mul` symbol (dot or star)  eg. (2*2 or 2 `dot`2) 
3. symbol `mul` num -> `mul` symbol (dot or star)   eg. (a*2 or a`dot`2)
   This rule can be improved to exchange the number and symbol to become `2a`
4. symbol `mul` symbol -> no `mul` symbol eg. (ab)
   This rule can be improved if the two symbols are the same, then (a*a) could be (a^2).

**Rules for large/small magnitude number:**
1. 1 < |number| < 1000 could be display directly eg. 369, 1200 should be "1.2`*`10^3"
   "exponent n is restricted to multiples of 3" 
   So for the decimal number apply the same rule. eg. 0.02, 0.002 should be "2`*`10^-3"
  Reference [Engineering notion](https://en.wikipedia.org/wiki/Scientific_notation)
2. "2`*`10^-3" can be represented as "2E^-3", but this is mostly used in Programming Languages.

Some comments on these rules?

JacquesCarette:
I agree with the four rules of multiplying -- but not the 'improvement'. If someone enters a*a, they should get (a a). GIGO. [Garbage in, Garbage Out]

I also agree with the rules for numbers. You should include that link in the comments to the Haskell code as well.

halonazhao:
[...]

PR #953, opened 2018-07-25
Add MINFRACT to swhs and correct the constraints for v_p.

halonazhao:


szymczdm:
Hi @halonazhao, can you please not leave the description for your PR's blank? It's a lot easier to tell what you're doing if you include a bit more detail than just the title.

code/drasil-example/Drasil/SWHS/Unitals.hs
```diff
@@ -446,6 +447,8 @@ tank_length_max = mkDataDef (unitary "tank_length_max"
   (nounPhraseSP "maximum length of tank")
   (sub (eqSymb tank_length) (Atomic "max")) metre Rational) 50
 
+frac_min_aux    = mkDataDef frac_min $ (dbl 1.0) * (10 $^ (negate 6))
```
szymczdm: We should probably keep this in mind for #941 as we don't want scientific notation hacks.

halonazhao: Ok. After scientific notation is done. I will check and correct all hacks.

halonazhao:
Alright. Next time, I'll try to be more specific.

JacquesCarette:
Unfortunately, there's another problem with this. As far as I can tell, the LaTeX ends up doing $MINFRAC$ (or the equivalent thereof), which ends up looking awful.  One should never display words in math notation.

Roughly speaking, the 'bug' is that `Atomic` should really have just a `Char` and not a `String`. [So the problem is much bigger than just this one example, but this one really makes it visible].  One potential hack would be that all `Atomic` of length >1 should be displayed (in LaTeX) using \mathsf. That would fix all problems at once.

@halonazhao can you file an issue about this please?

PR #954, opened 2018-07-25
InputDataConstraints Table update GlassBR

elwazana:
As per issue: https://github.com/smiths/caseStudies/issues/129


elwazana:
@JacquesCarette Hello Dr. Carette, After creating a PR for this branch I started working on updating Stable. This is when I noticed something I missed from before:

By adding `AR` to the data constraints table, it also adds `AR` to the list of inputs `gbInputs`. This list along with `glassBRUnitless` are used to generate the Table of Symbol. This leads to two `AR` s appearing in the table of symbols:
image

I've traced back the lists and how they end up generating the table, but there doesn't seem to be a way to add `AR` to the Data Constraints without having it appear again in the table of symbols.

elwazana:
@JacquesCarette The change from (> to >=) is done becuase manual was updated to allow `a` to equal `b`, this PR is just to propagate changes done to manual by this https://github.com/smiths/caseStudies/issues/129 issue. The reason for the change from (> to >=) is there.

As for the dropped ineqaulities, do you mean these:
image

Those are not needed now that `AR` has its own row in the data constraints table.

JacquesCarette:
Makes sense. I think what happened is that some of the trivial changes took a lot of lines of code to do, and some of these important changes were just 1-line, and so I missed their importance. Quite interesting that.

PR #962, opened 2018-07-27
`Contents`, `(Un)LabelledContent`, `RawContent` Implementation

niazim3:
Stable was updated to match the generated versions of the implementation in this branch since a lot of the discrepancies from the logs are stemming from the fact that data types do not have `Label` fields yet.

Once this branch is merged into master, the stable versions of the *.tex and *.html files that should be used for comparison in the logs are from this commit (i.e. when the examples first compiled: 79ef47ee9) and the [addLabels](https://github.com/JacquesCarette/Drasil/tree/addLabels) branch should be worked on to finish up the referencing infrastructure for Drasil.


JacquesCarette:
I will merge this in now, as the minor items I noticed are just that, minor. They should be fixed, but not so badly so as to hold up this big step forward wrt labels.

niazim3:
For the next step, master should be merged into the [addLabels branch](https://github.com/JacquesCarette/Drasil/tree/addLabels) where labels were added to other data types and continued to be worked on there, correct? @JacquesCarette

JacquesCarette:
I believe so.

PR #964, opened 2018-07-29
Use ConceptInstance for NoPCM Reqs

Mornix:
This PR is part of #562. It introduces the first use of `ConceptInstance` into the examples. This PR only affects the functional requirements of NoPCM. The old requirements still exist for compatibility with SWHS. As noted, there is a small formatting change with the new requirements as highlighted in the extended commit message of f123c6b. Below is a comparison of the requirements formatting:

# Old
screen shot 2018-07-27 at 1 46 04 am

# New
screen shot 2018-07-29 at 1 53 41 am

I have included @smiths as a reviewer due to this formatting change.

### DIFF
--- a/code/drasil-example/drasil-example.cabal
+++ b/code/drasil-example/drasil-example.cabal
@@ -57,7 +57,7 @@ executable glassbr
     drasil-code >= 0.1.1,
     drasil-printers >= 0.1.0,
     drasil-gen >= 0.1.0,
-    drasil-docLang >= 0.1.0   
+    drasil-docLang >= 0.1.0
   default-language: Haskell2010
   ghc-options:      -Wall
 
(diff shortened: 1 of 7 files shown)


JacquesCarette:
Can you re-merge master into here, so as to avoid the conflicts?

code/drasil-example/drasil-example.cabal
```diff
@@ -96,7 +96,7 @@ executable nopcm
     drasil-code >= 0.1.1,
     drasil-printers >= 0.1.0,
     drasil-gen >= 0.1.0,
-    drasil-docLang >= 0.1.1
+    drasil-docLang >= 0.1.2
```
Mornix: I bumped NoPCM's docLang requirement since it now uses mkEnumCC, but I'm not sure protocol for updating the version number for drasil-examples itself. Should I have bumped it in this commit?

JacquesCarette: If `drasil-examples` uses `mkEnumCC`, then yes. Until it does, then not necessary, and the bump can occur in that commit.

PR #983, opened 2018-08-08
Engineering Notation Implementation

halonazhao:
This PR is to represent data using Engineering Notation.


code/drasil-example/Drasil/GamePhysics/Body.hs
```diff
@@ -58,8 +58,10 @@ import Data.Drasil.SentenceStructures (foldlSent, foldlSent_, foldlList,
   SepType(Comma), FoldType(List), sOf,sAnd, sOr, foldlSentCol, foldlSP, 
   foldlSPCol, showingCxnBw)
 import Data.Drasil.Software.Products (videoGame, openSource, sciCompS)
-import Data.Drasil.Utils (makeTMatrix, itemRefToSent, makeListRef, bulletFlat,
-  bulletNested, enumSimple, enumBullet)
+import Data.Drasil.Utils (makeTMatrix, itemRefToSent,
+  makeListRef, bulletFlat, bulletNested, enumSimple, enumBullet)
+import Language.Drasil.Printers (PrintingInformation(..),
+ Notation(Scientific, Engineering), defaultConfiguration)
```
JacquesCarette: I think lines 63--64 should just be
```haskell
import Language.Drasil.Printers (PrintingInformation, defaultConfiguration)
```
[You should make sure that your PR is warning-free!]

code/drasil-example/Drasil/GlassBR/Unitals.hs
```diff
@@ -364,9 +364,9 @@ gbConstants = [constant_M, constant_K, constant_ModElas, constant_LoadDur, const
                 ++ gBRSpecParamVals 
 
 constant_M, constant_K, constant_ModElas, constant_LoadDur, constant_LoadDF, constant_LoadSF :: QDefinition
-constant_K       = mkQuantDef sflawParamK  $ (dbl 2.86) * (10 $^ (negate 53))
+constant_K       = mkQuantDef sflawParamK  $ dbl 2.86e-53
```
JacquesCarette: Excellent. Have you searched for all such notation hacks?

code/drasil-printers/Language/Drasil/HTML/Print.hs
```diff
@@ -24,9 +24,9 @@ import Language.Drasil.Printing.AST (Spec, ItemType(Flat, Nested),
   ListType(Ordered, Unordered, Definitions, Desc, Simple), Expr, Fence(Curly, Paren, Abs, Norm),
   Ops(Prod, Inte, Mul, Summ, Or, Add, And, Subt, Iff, Impl, GEq, LEq, Lt, Gt, NEq, Eq,
   Dot, Cross, Neg, Exp, Not, Dim, Cot, Csc, Sec, Tan, Cos, Sin, Log, Ln, Prime, Comma, Boolean, 
-  Real, Rational, Natural, Integer, IsIn), 
+  Real, Rational, Natural, Integer, IsIn, Point), 
   Expr(Sub, Sup, Over, Sqrt, Spc, Font, MO, Fenced, Spec, Ident, Row, Mtx, Case, Div, Str, 
-  Int, Dbl), Spec(Quote, EmptyS, Ref, HARDNL, Sp, Sy, S, E, (:+:)),
+  Int, Dbl, DblSc), Spec(Quote, EmptyS, Ref, HARDNL, Sp, Sy, S, E, (:+:)),
```
JacquesCarette: Do we need `DblSc` ? I understand that `Dbl` is printed using `showFFloat` and `DlbSc` using `showEFloat`, but the question is, can we generate `DblSc` ?

code/drasil-printers/Language/Drasil/Printing/Import.hs
```diff
@@ -40,19 +42,60 @@ p_space (DiscreteS a)  = "{" ++ (concat $ intersperse ", " a) ++ "}"
...
 -- | expr translation function from Drasil to layout AST
-expr :: HasSymbolTable s => Expr -> s -> P.Expr
-expr (Dbl d)            _ = P.Dbl   d
-expr (Int i)            _ = P.Int   i
+expr :: (HasSymbolTable s, HasPrintingOptions s) => Expr -> s -> P.Expr
+expr (Dbl d)           sm = case sm ^. getSetting of
+  Engineering -> P.Row $ digitsProcess (map toInteger $ fst $ floatToDigits 10 d)
+     (fst $ processExpo $ snd $ floatToDigits 10 d) 0
+     (toInteger $ snd $ processExpo $ snd $ floatToDigits 10 d)
+  _           ->  P.DblSc d
```
JacquesCarette: So if I read this code right, then in fact `DblSc` is the one which is always used, and `Dbl` is not anymore? In which case it would make sense to flip that.  So you should
1. explicitly label this case `Scientific`
2. have it output `P.Dbl`
3. change `P.Dbl` to print using scientific notation
4. remove `P.DlbSc`.

code/drasil-printers/drasil-printers.cabal
```diff
@@ -1,5 +1,5 @@
 Name:		drasil-printers
-Version:	0.1.0
+Version:	0.1.2
```
JacquesCarette: Good. You should make sure that all `drasil-examples` depends on version 0.1.2 of `drasil-printers` as well.

code/drasil-printers/Language/Drasil/HTML/Print.hs
halonazhao: If we don't use the build in function, we can write a similar function ourselves so that we don't need 'Dbl'. If this is what we want.

code/drasil-example/Drasil/GamePhysics/Body.hs
halonazhao: Warnings have been removed in d10ed89e52b0987b98192b916716fe095b1f85d1.

code/drasil-example/Drasil/GlassBR/Unitals.hs
halonazhao: I checked again. All hacks have been removed.

code/drasil-printers/drasil-printers.cabal
halonazhao: This has been fixed in 85d4e66d2a01eb9c79c0853e362a7056d581b09e.

code/drasil-printers/Language/Drasil/Printing/Import.hs
halonazhao: This has been fixed in 85d4e66d2a01eb9c79c0853e362a7056d581b09e.

code/drasil-example/Drasil/GlassBR/Unitals.hs
JacquesCarette: Good, thanks.

PR #986, opened 2018-08-09
Removed acroA

samm82:


samm82:
@JacquesCarette Now that I'm thinking about it, there might be less conflicts if all acro functions are removed in the same branch, but I can do them individually if you prefer.

JacquesCarette:
What kinds of conflicts do you see might arise?  More PRs is good. But I can merge things in without deleting the branch.

samm82:
Import lists, type declarations, etc. that have, say, (`acroA`, `acroGD`) on the same line will conflict if one is merged in but not the other. It isn't a big deal, but I feel like it would be a hassle to have to resolve them every time.

JacquesCarette:
That should be easier if you do all the work in the same branch, right? Then the only conflicts would be from merging in master, which will rarely touch those lines?

Main goal: I prefer to do this incrementally, but I don't want to cause you extra work.

samm82:
@JacquesCarette Yes - this would be fine if it is done in the same branch. However, if I make more changes, that gets added to the PR, no?

JacquesCarette:
I didn't think it did, I thought you had to explicitly add them to the PR. I could be wrong on that though!

JacquesCarette:
Let's see what happens now that I've merged this branch, but not deleted it.

samm82:
@JacquesCarette I'm almost ready to commit the removal of `acroGD`

PR #987, opened 2018-08-09
Removed all acro- functions

### LINKED ISSUE #276: Change acro- to use makeRef
This way, rather than generate just "T3", it will generate a link to the theoretical model.

edit: after doing some digging, this would require `RelationConcept` to be an instance of `LayoutObj`

### LINKED ISSUE #360: Assessment of acroTest Implementation
I have been working on a new version of referencing that would create a better hyperlink for html. However, I wanted some input/suggestions on what I was implementing and how I was implementing it.  In particular, I am concerned by two functions:

I) acroNumGen: This function takes an ordered list of contents and applies the correct enumerated abbreviation (e.g. LC1, LC2, LC3) to each item.  At the moment, the major concern for this function is the need to unwrap a sentence to create an abbreviation.
image

II) acroTest: This function creates a more accurate hyperlink for references e.g.:
image
However, I dislike the fact that two variables need to be passed (the content to be referenced and the 'enumerated' list of the contents) in order to make one reference. For instance `acroTest req1 s5_1_list_words_num`. Unfortunately, I am having difficulty thinking of a solution that does not involve two variables.

image

I have already altered the involved instances and types so that the code operates with the regular `makeRef` and the experimental `acroTest` so everything should run fine.  Most of these changes can be seen in commit dbfa8b990d32f1867899016587331b3bda6d7650.

samm82:
Closes #276 and closes #360

It looks like the previous commits are being "merged in again" - PRs just merge the entire branch.

EDIT: Yeah, I pushed my removal of `acroGS` to the branch, and it showed up here as well.

EDIT <span>#</span>2: Also, the Goal Statements are hardcoded, as it doesn't look like `Goal`s are implemented yet.


code/drasil-example/Drasil/SSP/GenDefs.hs
```diff
@@ -17,33 +34,6 @@ import Drasil.SSP.Unitals (baseAngle, baseHydroForce, baseLngth, baseWthX,
...
-l5  = mkLabelSame "normShrR" (Def General)
-l6  = mkLabelSame "momentEql" (Def General)
-l7  = mkLabelSame "netForcex" (Def General)
-l8  = mkLabelSame "netForcey" (Def General)
-l9  = mkLabelSame "hookesLaw2d" (Def General)
-l10 = mkLabelSame "displVect" (Def General)
-
 ---------------------------
 --  General Definitions  --
 ---------------------------
```
samm82: In this file, `GenDefn`s aren't explicitly built and the file exports `RelationConcept`s instead. Should this be changed? @JacquesCarette

JacquesCarette: No, as `RelationConcept` will get revisited, so that can be taken care of at that time.

PR #1021, opened 2018-10-11
Apply the specific types in DocumentLanguage.

halonazhao:
Since the cases are using specific types, the old way of using constraints was changed.


halonazhao:
Yes, that is `CI` not `IdeaDict`.

PR #1027, opened 2018-10-26
Update to use new TraceSec.

halonazhao:
Update NoPCM, SWHS and GamePhysics to use new TraceSection but with the old trace tables.


JacquesCarette:
For future reference - it was master that was failing the tests when Travis ran, it was not the fault of this PR. So it's ok to merge.

PR #1028, opened 2018-10-26
Extract Sentences and Exprs from DocDesc

halonazhao:
This PR constructed a structure where every `sentence` and  `expr` is collected. Then by collecting UIDs from sentences and exprs, all symbols are collected and this design is implemented to each case except GlassBR.

The issue left is that the new collecting structure is using some collecting functions from the old structure which was designed for `Document`. One solution is that I move those functions into the new structure in `DocLang`. The another is that we keep the imports. The choice depends on whether it is necessary to keep the collection structure for `Document`. @JacquesCarette


halonazhao:
Done. @JacquesCarette

PR #1031, opened 2018-10-29
SSP RFEM Removal from SRS

bmaclach:
This pull request removes the Rigid Finite Element Method and its associated outputs (displacement, local factor of safety) from the SRS for the SSP example.

@smiths and @JacquesCarette, I wasn't sure which one of you to assign for review, so I assigned both of you.


smiths:
This looks good to me, but I'll leave it to @JacquesCarette to approve because he can check the code.  It would be nice if we could keep things like Hooke's law in Data.Drasil, because Hooke's law is something that could easily come up in another context.  It looks like we haven't organized SSP in this way though, so now is probably not the time to reorganize the information.  We can always add it back in at some point in the future.

JacquesCarette:
Yes, there is information in here which should probably have gone into Data.Drasil.  However, the information is also coded with some horrible hacks (Drasil's support for vectors and indexing is still pretty bad), so that rewriting them from scratch may well be easier.

smiths:
That makes sense.  I don't like removing useful information, but I agree that cleaning things up and removing hacks outweighs any benefits of keeping "dead" code around.

PR #1032, opened 2018-10-31
Rearrange References between Assumptions and Other Chunks

halonazhao:
In this PR, things were done as below:
1. Add assumption references explicitly to chunks in `Note` field.
2. Remove all the chunk references in assumptions because this information can be traced back from the reference relations shown in Chunks.
3. Create Assumption file for NoPCM case because there are some new assumptions defined in NoPCM but different from SWHS. All the new assumptions were moved from Body file to Assumption file. After doing this, these assumptions should be able to be imported into other chunks' files.
4. After double-check every references, stable version was modified to match Drasil.
5. LikelyChange and Requirement references are currently kept since their structure is not fully designed yet. And they don't have any conflictions with other chunks.
6. No non-resolved instances occur in this PR. Every change looks good.


smiths:
@halonazhao are we going to have a situation where the manual case study versions no longer match the Drasil version?  I don't think we are quite ready to give up the security of the manual versions.  If there are changes that you make that should be propagated to the Case Studies repo, maybe you can create issues there so that we don't lose this information?

halonazhao:
I'm sorry that they do cause mismatching. But I just propagated these to `Nopcm` in case study [issue 201 in case study repo](https://github.com/smiths/caseStudies/issues/201). You can check the work there. If they are fine, I will finish the rest cases and the rework on Drasil. @smiths

smiths:
This change is fine with me.  I'm optimistic that information hasn't been lost here, but I cannot tell for certain.  We'll have to review the case study versions versus the Drasil versions when all the changes are made.  We have to do that anyway, since the Drasil versions aren't quite the same, so this shouldn't be a problem.

It looks like the merge has not conflicts, so I'll push the merge button.  😄

Issue #449, opened 2017-12-15
Stable to match manual for GD2 for ssp


smiths:
Issue #419 highlighted that the stable and manual versions of GD2 for ssp do not match.  I have verified that the stable version should be changed to match the manual version.  (The manual version is correct.)  Please update stable to match the manual version.

PR #1035, opened 2018-11-02
Improve Consistency of Manual version and Drasil

halonazhao:
When tried to move references in assumptions, found that there are too many inconsistencies, especially many references were not correct in SSP.

After this PR, even though there are still some inconsistencies, but all the references related to Assumptions are correct and complete. Those references should be matching manual version. 

Also, fixed issue #449.


JacquesCarette:
I am going to wait until I get confirmation on my comment on e007fdb to merge this. But that's the only thing I see in the way of doing that.

PR #1038, opened 2018-11-03
Change printers in HTML for Fig

halonazhao:
In this PR, these things were done:
1. Add the right path for tank.png in SWHS.
2. Change printers for `Figure` to use `<Figure>` and `<Figcaption>` tags.
3. Modify `figure` in CSS to make captions to be centered and bold.
4. Take `makeCSS` out to be an independent CSS printer in `drasil-printers`.


Mornix:
I notice as of 13463e0 that
* `<figure>`s have a nested `<p>` tag which (I think) is not necessary.
* The nested `<p>` tag is not closed.
* `` tags do not require a closing tag.

halonazhao:
Sorry, just double-check this. In https://www.w3schools.com/tags/tag_figcaption.asp, it did say '<figcaption>' and `<figure>` is adopted since HTML5. I think the link I referenced before is not reliable. 

And I was wrong again, I remember some closing tags were not apdopted since the beginning. But not <p> and yes I will add it back.
@Mornix Good memory!

JacquesCarette:
Note that I'm quite happy to have us switch to using 'figure' and 'figcaption'.

JacquesCarette:
Of course, you'll need to also update stable.  But this is otherwise looking ready to merge.

PR #1044, opened 2018-11-11
Populate ConceptDomain and Attach CIs to some Referable Chunks

halonazhao:
I think it's more convenient to merge the branch to master first since it has some preparation for us proceeding to the new references. Although there are some structures for collecting abbreviations in this branch, they are useful for the final implementation.

So if we merge this branch and the `newRef` branch, we can be more ready for the new design of References. What do you think? @JacquesCarette @Mornix


JacquesCarette:
Good idea.  Merging this one in now.  I won't delete the branch yet, just in case we want to tweak things.

PR #1045, opened 2018-11-11
implement a new Ref2 in Sentence (and Printing.AST)

JacquesCarette:
that will eventually replace Ref. Right now, nothing uses it.


JacquesCarette:
Because of how I have set things up, all PRs need reviewed, and cannot be reviewed by the creator, even though I am the owner of the repo.  So I've asked @halonazhao to review.  I could over-ride this, but it doesn't seem worth it.

PR #1049, opened 2018-11-20
Implement the New Sentence Constructor 'Ch'

halonazhao:
In this PR, things as following were done:
1. `Ch` is turn into take a `SentenceStyle` which can indicate the printer to render a chunk differently. `SentenceStyle` includes `short` which is the abbreviation of the term, `symbol` which is the symbol of the term, `plural` which is the plural of the term and `term` is the normal term.
2. `Ch` will take chunks' `UID`, which make every chunk got rendered until the printing level, which means the printer will query the DB(means `ChunkDB` here) to get the desired inforamtion.
3. A sufficient amount of terms were added into everything-ChunkDB for each case.
4. `the` constructor was redesigned to generate a NP instead of `namedChunk`.

The problem left for this PR is that because of the changes of the structure of the  `sentence`, the `capitalization` technique should be moved down to the printer level to be solved.


JacquesCarette:
Note that I will **not** delete this branch now, as it might still be useful for the other changes, as discussed in other check-ins. At least, I'm not sure, so that's the safest alternative.

PR #1051, opened 2018-11-22
ConceptInstance uses Ref2

Mornix:
Rebase and cleanup from the earlier version I posted comments on.

Changes of note:
* `RefProg` is now in `RefProg.hs` and not `Sentence.hs`
* "Hiding" `RefProg` constructors by providing nicer functions. (As a note, I want to use `defer`
for `RefProg` as it is used by `ShortName` now. So in `Reference.hs` (only place either is used) I've added a qualified import as part of the transition to support both until `Ref` is removed.)  (This point is mainly aimed for @JacquesCarette)
* Making the change so `P.Ref2` does not carry `UID` and lift `UID` from removed `PrependDomain` constructor to `Reference2`.
* Updating all examples to use `Ref2` for `ConceptInstance`.

PS. It should be noted that this pull request is to merge this branch onto abbAndAcr and _not_ master.


JacquesCarette:
Should I delete the branch?

Mornix:
I'm okay with the branch being deleted.

PR #1075, opened 2018-12-07
Remove the  "Referable Label"

halonazhao:
There was only one section reference still using the label.  It was turned to use a fake section now. And the Referable Label has been removed.


JacquesCarette:
Strictly speaking, this 'fix' is a hack, as it generates a dummy section instead of referring to the correct section. However, because the dummy section and the correct section have the same reference information, it all works. I will file an issue about this for later fixing [and assign it to myself], but accept this change now. The changes allow all of drasil-lang to be clean, and push the problem out to the outer layers, which is preferable to the current situation.

PR #1077, opened 2018-12-08
Fix unit collecting issue

halonazhao:
Somehow I didn't put the collecting unit structure in place and that's why the old way was taking effect instead of my new structure. So I went through the logic and push some design forward to finally get the function works. And now the collected unit are put into the `used ChunkDB` and I made the `DocumentLanguage` to use `unitMap` in `usedChunkDB` to generate the table of unit. And they passed all the tests for all cases including the Tiny one's stable version actually should be changed. So  please check this work and give me some feedback. @JacquesCarette Thanks.


halonazhao:
Actually, they technically cannot be both empty. But the first field was designed to be `maybe` is for some units like `m^3/N^16`. They are derived from SI base units or SI derived units like J, but they are not SI Units. I was keeping the first field for SI unit's symbol. And the second for the detailed form of the symbol.  So I actually not sure whether is that ok to keep two symbol field as `Maybe`. 

But all the constructors to generate a unit would make sure at least one of it is not empty. Even some units are constructed direct using the initial structure. If this unit will be used and printed out, the `usymb` function would cause an error because `usymb` function must get a symbol. 

Not very sure whether this way is ok? @JacquesCarette

Issue #359, opened 2017-07-20
SSP Data Defs and Table of Symbols cannot index


njericha:
I'm creating a new issue that stemmed from #265 being mostly fixed for clarity on what is still an issue.

The way `Index` works is that it is built into `Expr`. This works well for most cases except for two.
Data Defs used an index-ed variable as the output such as the example given below. But the `QDefinition` takes a `Quantity s, SymbolForm s' and `Expr` to equate them. This means I cannot index the quantity becuase `Index` returns an `Expr`.
![2017-07-20](https://user-images.githubusercontent.com/26230851/28434025-efba0030-6d5b-11e7-8ce6-e2a719d4e0ac.png)

The second issue is that I need to remove the subscript `i` from the symbols otherwise we would get duplicated `i`s when index-ing them. This means the table of symbols displays the symbol without the subscript `i`.

I'm not sure what the best solution would involve. We could;
- create a new data constructor for `Symbol` that can be a special index which would get ignored by `Expr`
- change the type of `Index`
- change the type of `QDefinition` and Table of Symbol constructors

njericha:
Making a note that @AKM11 has a similar issue in NoPCM.

JacquesCarette:
Can't we use some kind of pair as `Index`, to index by more than one thing "at once"?  [This is different than double-indexing].

njericha:
@JacquesCarette I'm not sure what you mean by "some kind of pair". Do you mean, "Can we index the whole equation, rather than just the individual quantities?"

JacquesCarette:
I mean something that would mean the same as `a[i,j]`, i.e. `Index(a, Pair(i,j))` as an `Expr`.  I don't know if we have the `Pair` constructor (or something like it) in `Expr` yet.

njericha:
We do not have something like that in `Expr` yet. That is not part this issue though. Right now, I'm treating "_U<sub>t,i</sub>_" as "_(U<sub>t</sub>)<sub>i</sub>_" where _U<sub>t</sub>_ is the symbol for the quantity and I am indexing it at _i_. The issue is that I cannot use `Index` on the left hand side of a `QDefinition` because you cannot index a symbol, only an `Expr`.

JacquesCarette:
I see, that is a rather different issue indeed.

[...]

PR #1081, opened 2018-12-21
SSP IM1 Revision

bmaclach:
IM1 and its derivation were updated to match the manual version. As part of this, a new assumption and two new data definitions were added, as they are used in the derivation.

### DIFF
--- a/code/drasil-data/Data/Drasil/People.hs
+++ b/code/drasil-data/Data/Drasil/People.hs
@@ -9,7 +9,7 @@ spencerSmith, henryFrankis, nKoothoor, dParnas, daAruliah, gWilson,
   fIncropera, dDewitt, tBergman, aLavine, jRobertson, sRobertson, 
   wlBeason, tlKohutek, jmBracci, qhQian, dyZhu, cfLee, grChen, dgFredlund,
   jKrahn, dStolle, yCLi, ymChen, tltZhan, ssLing, pjCleall, pGuo,
-  mCampidelli,   dmWiess, sPalmer, scottSmith :: Person
+  mCampidelli,   dmWiess, sPalmer, scottSmith, bKarchewski :: Person
   
 pjAgerfalk   = person    "PJ"                        "Agerfalk"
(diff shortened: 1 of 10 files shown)


smiths:
I've compared the generated version to the case study.  Looks great.  I do have a few minor comments
- x_wt is missing from the inputs in the Drasil code
- in the first equation for the derivation, should it be N'_i (should it be written as the effective force, with the prime symbol)
- I know it is planned for later, but we really should make FS be F_S, as it currently is written causes some confusion for reading the equations
- for DD: convertFunc1 and DD: convertFunc2, the labels in Drasil are Constant, they should be "First Function for ..." etc
- for DD: convertFunc1 and DD: convertFunc2, the subscripts do not seem to be correct.  The psi and the phi are missing subscripts, and for the second function there is a subscript n-1, instead of i-1

bmaclach:
Thanks @smiths. To address each item:

- This is an interesting case. I left x_wt out because it's actually not used in any of the equations (the x-coordinate is implicit when we write an equation for a value indexed by i). That being said, we probably should include it anyway since you don't know which y-coordinate to use without the corresponding x-coordinate. If you agree, I'll go ahead and add it.
- Yes, fixed in 24a194346be0358e282884bba9ea048266d195d7
- Okay, changed the symbol in 24a194346be0358e282884bba9ea048266d195d7
- Changed in 24a194346be0358e282884bba9ea048266d195d7
- The n-1 was a mistake, fixed in 24a194346be0358e282884bba9ea048266d195d7. As for the missing indices, there is a limitation where the left-hand side of a DD can't currently be an indexed symbol. This is discussed in #359.

smiths:
Sounds good @bmaclach.  Yes, I think we should put x_wt back in.  We could always remove it later, but for now I would like the Drasil version to match the manual case study version.  For now, I think we should still keep the two versions in sync.  However, we might be getting close to the time where we stop "maintaining" the manual version.

Thanks for making the updates.  I had forgotten about issue #359.  Thank you for the reminder.  😄

PR #1094, opened 2019-01-24
Gamephy srs

oluowoj:
Added new Data Definition "Torque"
I think I need a new symbol r?
r : is a position vector of the point where the force is applied, measured from the axis of rotation. I used another symbol "r_OB" for now.

### DIFF
--- a/code/drasil-example/Drasil/GamePhysics/Assumptions.hs
+++ b/code/drasil-example/Drasil/GamePhysics/Assumptions.hs
@@ -7,7 +7,7 @@ import Data.Drasil.SentenceStructures (FoldType(..), SepType(..),
   foldlList, foldlSent)
 import Drasil.GamePhysics.Concepts (twoD)
 import qualified Data.Drasil.Concepts.Physics as CP (rigidBody,  
-  cartesian, rightHand, collision, joint, damping)
+  cartesian, rightHand, collision, joint, damping, force, friction)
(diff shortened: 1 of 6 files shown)


smiths:
@oluowoj, if there are questions about the best way to present the physics, we should probably have those discussions with the manual case study.  In the interest of separation of concerns, we can sort out the best way to present the physics in the manual version and then just focus on the translation to Drasil for the Drasil version.

oluowoj:
sure, I will talk to you this afternoon in your office.

JacquesCarette:
I approved the changes 'visually', but now I see that the changes actually cause printing differences.  Please deal with these (and add a commit) to fix the CI build.

JacquesCarette:
Note that the builds are still failing.  Do they somehow work on your machine? It would be nice to merge this, but I can't until the builds are clean.

oluowoj:
The build works perfectly fine on my machine and I copied the .html and .tex file into stable folder but still no luck.

JacquesCarette:
Try merging `master` into your branch. That should also update the CI scripts, which should let you see what went wrong (Gabe improved some of the scripts).
Please do not do multiple pieces of work in the same branch / PR. I think that change f6023b0 is not related to the rest?  If so, it should be done as a separate PR, from a separate branch.

JacquesCarette:
Weirdly, it looks like the build completed fine, but didn't get picked up? I'll merge in.

oluowoj:
thanks,

PR #1111, opened 2019-02-12
SSP Effective Normal Force GD

bmaclach:
This pull request adds the effective normal force GD to the SSP example.

### DIFF
--- a/code/drasil-example/Drasil/SSP/GenDefs.hs
+++ b/code/drasil-example/Drasil/SSP/GenDefs.hs
@@ -8,26 +8,29 @@ import Language.Drasil
 
 import Drasil.DocLang.SRS as SRS (physSyst)
 
+import Data.Drasil.SI_Units (metre)
+
 import Data.Drasil.Concepts.Documentation (assumption, definition, 
   method_, property, value, variable)
 import Data.Drasil.Concepts.Math (normal, perp, surface)
 import Data.Drasil.Concepts.PhysicalProperties (mass)
 import Data.Drasil.Concepts.SolidMechanics (normForce, shearForce)
 
+import Data.Drasil.Concepts.Math (equation)
 import Data.Drasil.Quantities.Physics (force)
 import Data.Drasil.Quantities.SolidMechanics (nrmStrss)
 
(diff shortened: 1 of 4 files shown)


code/drasil-example/Drasil/SSP/GenDefs.hs
```diff
@@ -8,26 +8,29 @@ import Language.Drasil
...
 import Data.Drasil.Concepts.SolidMechanics (normForce, shearForce)
 
+import Data.Drasil.Concepts.Math (equation)
 import Data.Drasil.Quantities.Physics (force)
 import Data.Drasil.Quantities.SolidMechanics (nrmStrss)
 
 import Data.Drasil.SentenceStructures (foldlSent, foldlSent_, getTandS, ofThe, sAnd)
 
-import Drasil.SSP.Assumptions (newA2, newA3, newA4, newA5, newA6)
+import Drasil.SSP.Assumptions (newA2, newA3, newA4, newA5, newA6, newA7)
```
smiths: Is it possible for the assumptions to have more meaningful names?

bmaclach: I hope so! I'm just not sure why this naming convention was used in the first place. If there's no reason not to change them, I can change them, but should that be done as a separate PR?

smiths: Yes, I think the name change would be a separate pull request.  I commented on it here because it jumped out at me as I scrolled through the code.

Mornix: Just a heads-up, on the QDef-in-InstMod branch, as part of changing `AssumpChunk` to `ConceptInstance`, I changed the names there a bit:
https://github.com/JacquesCarette/Drasil/blob/8d22f45dc2ab2589159687b3900538d7303ede7b/code/drasil-example/Drasil/SSP/GenDefs.hs#L22-L23

I'm not sure how much better those names are overall, but I think they're an improvement over `newA*`.

smiths: Hello Mornix.  I definitely like your names better.  I'm glad you jumped in with your comment, since we don't want to be working at cross-purposes.

PR #1113, opened 2019-02-15
Move all Images From examples to datafiles.

Mornix:
This PR consolidates image files (some of which resided in drasil-examples/Drasil/*/*.png) to where they should be within datafiles. 

I updated the docs where appropriate to ensure they reference the datafiles variants. 

Additionally, I removed the image traceability matrices for NoPCM as they were old (@bmaclach did them in 2016), for SWHS, and @halonazhao's traceability matrices serve the same purpose.


Mornix:
Great question! Exact copies of all six of the files deleted from the GlassBR folder already existed in the datafiles/GlassBR folder prior to this commit.

PR #1122, opened 2019-02-20
Fix a few underfull {h,v}boxs

Mornix:
This is a tiny PR which fixes a few instances of TeX warning of "Underfull \hbox" and "Underfull \vbox".

One instance comes from `longtable` where it [expects a `\endfirsthead` or `\endhead` after the header](https://tex.stackexchange.com/questions/71096/underfull-vbox-in-each-longtable-can-it-be-fixed-or-must-it-be-ignored). I've included a screenshot below of a `longtable` in one of the examples and how it looks with a repeated header after a page break (a choice I have made which can be changed back –– I've included @smiths as a reviewer for this purpose).
screen shot 2019-02-20 at 2 21 34 pm

The other comes from [forcing `\\` in a paragraph leaving "unexpected empty space".](https://tex.stackexchange.com/questions/334246/what-does-the-phrase-underfull-hbox-badness-10000-in-paragraph-actually-mea)

### DIFF
--- a/code/drasil-printers/Language/Drasil/TeX/Helpers.hs
+++ b/code/drasil-printers/Language/Drasil/TeX/Helpers.hs
@@ -139,7 +139,7 @@ item' bull s = command1oD "item" (Just bull) s
 maketitle, maketoc, newline, newpage, centering :: D
 maketitle = command0 "maketitle"
 maketoc   = command0 "tableofcontents"
-newline   = command0 "newline"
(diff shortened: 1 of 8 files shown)


Mornix:
PS. It appears the status of builds is not being communicated between Travis and GitHub at the moment.

smiths:
Repeating the header after a page break seems like a good decision to me.

PR #1123, opened 2019-02-20
BibTex Fixes

Mornix:
This PR addresses two issues BibTex warns about during compilation.

1. A stray ", " at the end of a citation. This is caused by an author who only contains a last name rendering as "Last, <empty><empty>"

2. Months in BibTex should be rendered as `month=feb` instead of `month={feb}`.


Mornix:
PS. It appears the status of builds is not being communicated between Travis and GitHub at the moment.

PR #1127, opened 2019-02-21
SSP Scope of Requirements

bmaclach:
This updates the Scope of Requirements section of SSP to match the manual version. Also adds a "2D" CI to Data.Drasil since it is now used in multiple examples and is generic.

### DIFF
--- a/code/stable/ssp/SRS/SSP_SRS.tex
+++ b/code/stable/ssp/SRS/SSP_SRS.tex
@@ -191,6 +191,8 @@ ${ℓ_{s}}$ & Length of an Interslice Surface: from slip base to slope surface i
 Abbreviation & Full Form
 \\
 \midrule
+2D & Two-Dimensional
+\\
 A & Assumption
 \\
 DD & Data Definition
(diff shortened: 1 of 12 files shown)


code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -230,7 +232,7 @@ \subsection{Purpose of Document}
...
 \subsection{Scope of Requirements}
 \label{Sec:ReqsScope}
-The scope of the requirements includes stability analysis of a 2 dimensional slope, composed of homogeneous soil layers. Given the appropriate inputs, SSA identifies the most likely failure surface within the possible input range, and finds the factor of safety for the slope.
+The scope of the requirements includes stability analysis of a Two-Dimensional (2D) soil mass, composed of a single homogeneous layer with constant material properties. The soil mass is assumed to extend infinitely in the third dimension. The analysis will be at an instant in time; factors that may change the soil properties over time will not be considered.
```
JacquesCarette: Hmm, the capitalization here isn't good on "Two-Dimensional". Or at least, not what I expected.

bmaclach: Would "2-Dimensional" be better?

JacquesCarette: In inline text, I expect "two-dimensional".  It's not a title of something.

bmaclach: The reason for the capitalization is because this is where the "2D" abbreviation gets introduced. I thought (wrongly?) we should always capitalize for introduction of an acronym.

JacquesCarette: I don't think so? Yes, it happens that most acronyms seem to work that way, but not abbreviations.

JacquesCarette:
Also, this is going to need a merge of master.

bmaclach:
@JacquesCarette All right, master merged and "Two-Dimensional" changed to "two-dimensional". This capitalization is changed both in the paragraph text and also in the table of acronyms. Is that what we wanted, or did we want it to be capitalized in the table of acronyms?

PR #1128, opened 2019-02-21
SSP Characteristics of Intended Reader

bmaclach:
This updates how the Characteristics of Intended Reader section is built, allowing for the inclusion of topics that "would be an asset" if understood. This allowed the section in SSP to match the manual version. The change was done in such a way that the text in other examples did not change, though in the future we likely want to make this consistent between examples (issue about this on its way).


JacquesCarette:
Can you merge master into here? There's a conflict that you're in a better position to resolve than I.

bmaclach:
@JacquesCarette Master re-merged, should be good to go!

PR #1130, opened 2019-02-21
SSP Organization of Document

bmaclach:
This updates the Organization of Document section of SSP to match the manual version. Part of this was to add section references to a line that is common to all examples, so that change affected all examples (which I think is a good thing).


JacquesCarette:
I closed-and-reopened to get Travis to restart.

smiths:
The changes seem reasonable to me, although @JacquesCarette is a better judge.

bmaclach:
@JacquesCarette The way I did the links in the refine chain matches manual, but in manual we don't have meaningful labels for section references, so there is no duplication of information. So I see what you're saying... would the idea then be to remove the regular text, and have only the link? We would lose the redundancy at the expense of being grammatically correct.

JacquesCarette:
The idea would be to make the regular text **be** the link.  I guess the best example is to be found in most Wikipedia pages. The reason this is doable here is exactly because you pass in a pair of the word and the link information.

bmaclach:
@JacquesCarette Okay, I'll work on that change. Just to clarify, your comments apply to the changes in commit 74a9332e4a3fe547e3157d6b085c43e295e3c5ad, but do they also apply to the changes in commit 234421c7022c66de0d26e6800184b303dc346596? (The difference being that the latter commit is citations rather than internal referencing)

bmaclach:
@JacquesCarette I've looked into this some, and I think making the text itself the link would require a rework of how we print References in Drasil. Currently a `Reference` in Drasil is parameterized by a `ShortName` which corresponds to the text that will actually be rendered and user-visible. So if we want to define some text to be a link, we would either have to hack that text to be type `ShortName` or modify `Reference` to be parameterized by a `String` instead of a `ShortName`. That's the first issue. On top of that, we prepend a category like "Section:" or "Fig:" to the `ShortName` when we render it, so even if we used custom text for the link, we would still end up with this prepended bit. Our text would end up being:

"The _Section: goal statements_ are refined to the theoretical models, and the _Section: theoretical models_ are refined to the ..."

To optionally remove the prepended text would require an even deeper rework of Reference.

Is this what you were thinking, or am I missing an easier way to do this?

bmaclach:
@JacquesCarette Sorry for all the comments. Just to correct something from my last comment: Drasil does already have a way to tell the printer not to prepend anything to the reference text. So that's not much of a concern. The biggest issue is the proposed change to `Reference` to be parameterized by a `String` instead of a `ShortName`.

JacquesCarette:
Note that you can easily build a `ShortName` from a `String` (but not the other way around). And no, we don't want to do citations in quite the same way.  What we currently do is a bit of a hack, at least as compared to some of the citation combinators available in latex, but I don't quite want to go down that rabbit hole right now.

Referencing is so important that we need to take a close look at this.  We've already spent a huge amount of time implementing referencing mechanisms, but that's because it's tightly tied into traceability. And end-document readability is important to me, so I want to really understand the outstanding issues.

bmaclach:
If we keep `Reference` parameterized by `ShortName`, then in the case of this pull request, to make our `Sentence` the text for the link, we will need to convert from `Sentence` to `String` to `ShortName`, and then create a `Reference` with that. This also feels hacky to me, which is why I thought it might be better to change `Reference` to be parameterized by a `String` (though admittedly in that case, we still need to convert from `Sentence` to `String` and will need to convert from `ShortName` to `String` for all existing references, so it's also not a great solution).

JacquesCarette:
`Sentence` -> `String` is always a sign that we're doing something wrong. It might be that the original `Sentence` was 'premature'.

Can you pull the essence of this discussion out into a new issue, and merge master in here (there's a conflict)?  I'll then merge things in as is. The current state is not buggy, just not what I want long term. But fixing this is a bigger deal, so let's not hold things up that long.

bmaclach:
@JacquesCarette Master has been merged (as well as a fix for a double period I hadn't noticed before) and issue #1142 created.

Issue #1135, opened 2019-02-26
GlassBR self-referential DDs break generated code


bmaclach:
Some changes were made as part of commit f5ec138336b056dbf64223a8ee2a0e317ec5fa0c that broke the generated code for GlassBR (the changes to DataDefs.hs and IMods.hs are of particular note here). These changes seem to have been made intentionally to remove code-specific symbols from the GlassBR SRS, however they break the code because the DDs now say that, for example, `J` is calculated by the function for calculating `J`. The broken code looks like this:
```
public static double func_J(InputParameters inParams, double q_hat) throws Exception {
        return func_J(q_hat, inParams.AR);
}
```
And the correct code looks like this:
```
public static double func_J(InputParameters inParams, double q_hat) throws Exception {
    return Interpolation.func_interpZ("SDF.txt", inParams.AR, q_hat);
}
```
If I undo the changes from the above commit to DataDefs.hs and IMods.hs, the generated code will be fixed. However, the SRS will break. Here's what the SRS looks like now:
![image](https://user-images.githubusercontent.com/18634568/53449260-6bc7de00-39e7-11e9-9930-0a2e3a5aa9ca.png)
And here is what it will look like if I undo the change to fix the generated code:
![image](https://user-images.githubusercontent.com/18634568/53449364-a16cc700-39e7-11e9-89ae-6c5ae775dca3.png)
Note that the Equation makes no sense and there is a line in the Description that just says "is the interpZ".

I don't think even the current version of the SRS is correct, though it does match manual. My immediate question is whether I should put in the fix for the code at the expense of the SRS, or leave the code broken so that the SRS does not break even more.

smiths:
@bmaclach, in my opinion there is no way to fix this.  The requirements should be abstract, but the code needs to know what interpolation means.  What we need is the "design" that bridges between the two.  I have written a manual version of the Module Interface Specification that connects the SRS to the code:

https://github.com/smiths/caseStudies/blob/master/CaseStudies/glass/docs/Design/MIS/glassbr_mis.pdf

This document has no analogue in Drasil.  We haven't yet decided if this is the approach to use, or something else.  What is certain is that something has to happen between the requirements and the code.  The only reason so much of glassbr works is that the design is so trivial, since many equations are explicit.  The interpolation is more involved, so additional information is needed.

[...]

PR #1138, opened 2019-02-27
Modified self-referential GlassBR DDs to fix generated code

bmaclach:
This PR makes the changes discussed in #1135. Self-referential DDs in GlassBR no longer refer to themselves, instead referring to the correct interpolation function. This fixes the Calculations module of the generated code, though it results in some incorrect equations and descriptions in the SRS (issue about this coming soon).


JacquesCarette:
Considering the breakage, should this PR be closed until a better solution is found?

bmaclach:
@JacquesCarette I think the answer to that question depends on your input on #1139.

JacquesCarette:
Agreed - see my comments there.

bmaclach:
Implemented a hack as discussed in #1139 that allows symbols for implementation-stage functions to be used in the SRS. LaTeX and code should now both compile!

PR #1144, opened 2019-02-28
SSP System Constraints

bmaclach:
This updates the System Constraints section of the SSP SRS to match manual.

PR #1145, opened 2019-02-28
SSP System Context

bmaclach:
This updates the System Context section of the SSP SRS to match stable.


JacquesCarette:
Fine to me as well, but now there is a conflict, so master needs to be re-merged in, please.

bmaclach:
Master re-merged, though I fear we are going to get another conflict with import lists between this and #1144. These import lists are really messing with my supposedly independent pull requests, but I'm not sure if there's anything that we can do about it without combining PRs.

JacquesCarette:
Yes, you're right, this needs re-merged now that #1144 is done.

bmaclach:
Master re-merged.

PR #1149, opened 2019-03-04
General System Description intro text update

bmaclach:
This updates the small paragraph at the beginning of the General System Description section of the SRS to match manual for SSP. This paragraph is common between all examples, so the change affected all examples. I think this is good, however, because the new paragraph provides the same information but is more readable and grammatically correct.


smiths:
The new general system description intro looks fine to me.

PR #1161, opened 2019-03-15
Coefficient of Restitution

oluowoj:
Added new definition into Datadefs: Coefficient of Restitution.


oluowoj:
Description of CR was updated in datadefs.hs

oluowoj:
image
Pt 1: In my latest commit, I see that this is not an issue. Somehow it is showing as 'A'. I don't remember fixing this
Pt 2/Pt 3: Looking into this issue

> $v^{PB}$ has disappeared. I think this is on purpose, but it would have been nice to have indicated so in the log message.

This was updated in a previous branch, not sure how or why it disappeared as its being used in one of the DataDef modules. Could you point me to where you are looking pls?

oluowoj:
Yes for this branch, I need to re-merge. I will do that and push changes

oluowoj:
Looks like all checks have passed and should be ready to merge.
Your comment 3 was discussed at the ''All hands meeting today and I will send a ticket to @mornix"

PR #1162, opened 2019-03-15
WTNT DD to repair generated code

bmaclach:
This PR makes wTNT a DD in GlassBR, for the purpose of fixing the generated code. Now that it is a DD, the calculation of wTNT will be included in the DerivedValues module.

I also cleaned up some old QDefinitions and Relations that are not used anymore (because aspect ratio and stand-off distance were also made into DD's at some point, making their previous definitions obsolete).


JacquesCarette:
Good change, needs master merged.

bmaclach:
Master merged.

PR #1165, opened 2019-03-20
Code Generation - Fixing more errors in the generated GlassBR code

bmaclach:
This PR does the following:

- Code generation now considers more than just data definitions as values that need to be calculated (it also considers the "Definitions" field of the `SystemInformation`). This essentially undoes a change from e6fc5803a where these definitions were excluded from code generation for some reason.
- Since theoretical models and instance models are now considered in code generation, they needed some updates to avoid errors
- Thanks to the above two points, functions for calculating is_safeLR and is_safePB are now present in the generated code
- Also thanks to the above changes, the Control Module now correctly calls all of the functions in the Calculations module (before, it was not calling any of them)
- Also thanks to the above changes, I was able to remove the hack where the outputs of GlassBR were also added as inputs to GlassBR, which was the cause of at least one error when running the generated code


smiths:
I haven't looked at the new code, but the statement that tmods are included in the generation struck me as interesting.  The theoretical models are refined to instanced models in the SRS; therefore, I wouldn't think they would appear in the generated code.  I wonder if it was necessary in GlassBR because the theoretical models are wrong.  @vajihehm is working on changes suggested by me, and they include introducing theoretical models that are refined to instance models, where the instance models will look like the current theoretical models.

My comment isn't intended to generate any changes in this PR.  However, when we start thinking about the "proper" way to generate, I think we should consider whether the theoretical models should actually be included.  Once @vajihehm has completed her changes, we'll have a better idea of the consequence of those changes on code generation.

bmaclach:
@smiths I agree that it doesn't make much sense to include theoretical models in the generated code, and you are correct about why it was necessary here. Once the theoretical models are switched to instance models as part of @vajihehm's revisions, it should no longer be necessary to consider theoretical models in code generation.

bmaclach:
Also just a note that this PR introduces some duplicated code for the calculation of demand, because demand is both a data definition and an instance model. From what I can tell, as part of the revisions to the manual GlassBR SRS, it was changed from an IM to a DD, so I'm expecting that the IM will be removed from Drasil at some point as part of @vajihehm's GlassBR revisions.

smiths:
Sounds good @bmaclach.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -956,13 +956,14 @@ \subsubsection{Instance Models}
                                              \\ \midrule \\
                                              Output Constraints & \\ \midrule \\
                                                                   Equation & \begin{displaymath}
-                                                                             q=q\left({w_{TNT}},SD\right)
+                                                                             q=interpY\left(TSD.txt,SD,{w_{TNT}}\right)
```
JacquesCarette: Strictly speaking, this highlights a but as TSD.txt is a string, and should be in string quotes. The bug is likely in the renderer somewhere, as the model contains the correct information. [I won't let this hold up the merge]

PR #1167, opened 2019-03-20
Code generation - Fixing duplicate constraints

bmaclach:
This tiny PR fixes a bug where code for physical constraints was being generated twice and code for software constraints was not being generated at all.


JacquesCarette:
I've already talked with @Mornix about this, but this change should have caused test failures, in the sense that the generated code should have been different than 'stable'. Right now, we don't do that, but we should hurry up and do so!

PR #1175, opened 2019-04-01
Enable TeX building in Travis

Mornix:
This PR introduces TeX compilation in Travis. 

One of the changes introduced in this PR is to quiet the TeX building as there is a lot of noise in the output. So much so that Travis doesn't displays the entire log. With `SUMMARIZE_TEX=yes` set, both `lualatex` and `bibtex` are set to output the minimal amount (without redirecting `stdout` to `/dev/null` — in an attempt to be more Windows friendly). The TeX `.log` and BibTeX `.blg` files are `grep`ped to "summarize" the logs with any errors or warnings if the TeX build was successful. In the event of a build failure, the last `.log` is printed out as it is likely to contain the specific error and the `make tex` command will abort meaning the output is less likely to be lost among the other build commands. 

Related changes in this PR are:
* Changed SystemContextFigure.PNG to SystemContextFigure.png (Fix for building GlassBR TeX on case-sensitive filesystems)
* Changing `TEXFLAGS` to `BIBTEXFLAGS` for `bibtex` in generated Makefiles. (Because BibTeX has different arguments than `lualatex`.)


Mornix:
Closing and reopening to see if Travis Build will be marked as completed.

JacquesCarette:
I'm going to guess that the travis problems were intermittent, so I'll merge this now anyways. We'll see.

PR #1185, opened 2019-04-05
add new GlassBR TM symbols Pftol, pf

vajihehm:
@smiths Please confirm the changes before the merging. I have added the two symbols that I have considered "pfail : probability of fail" as a subset of "gbOutputs". However, in my opinion it should not be considered as an **output** for GlassBR, but I was not able to find a specified category for it that would include "**constraints**". So, I put this symbol into "gbOutputs" . If @JacquesCarette and you would have the better suggestion please let me know. For more information I have attached the related section from glassbr_srs.

TM


smiths:
@vajihehm, I might be missing something, but it seems to me that the theoretical model shouldn't have constraints.  The theoretical model certainly isn't part of the outputs.  The inputs and outputs, as used by the requirements, are part of the instanced model.  Maybe you are thinking that your pfail is constrained because of the equation where it is used, but the equation that defines it isn't a constraint, but the definition of the value of the Boolean that represents whether the situation is safe or not.

vajihehm:
> @vajihehm, I might be missing something, but it seems to me that the theoretical model shouldn't have constraints. The theoretical model certainly isn't part of the outputs. The inputs and outputs, as used by the requirements, are part of the instanced model. Maybe you are thinking that your pfail is constrained because of the equation where it is used, but the equation that defines it isn't a constraint, but the definition of the value of the Boolean that represents whether the situation is safe or not.

@smiths I do agree on that theoretical Model symbols should 'nt be part of the outputs. But, I think probability is a concept which has constraint between 0 and 1 and this constraint is not related to the Theoretical Model symbols. If you like to consider this symbol without any constraint I can put this symbol a part of "glassBRUnitless" without any constraint.

smiths:
Thank you for your response @vajihehm.  I believe I understand better now.  We certainly don't want to lose the information that probability is bounded between 0 and 1.  In the long run, it will help if we have types in Drasil, so that we can give probability a type that only has values from 0 to 1.  In the meantime though, we should record the constraint information for P_fail.  As we've observed in the past, we don't want to lose information.

I think our current approach should be to include P_fail as an output, if that is the only category that allows for constraints.  I've gone through several of our examples and I haven't seen another case where we did this with a variable that only shows up in the theoretical model, but I think it is worth the temporary confusion to not lose the constraint information.

Ideally what we want to do at some point is capture the refinement of the theoretical model to the instance model.  In some cases there is a significant refinement between the theoretical model and the instance model, but there is very little refinement in this case.  The only refinement is that we define a means to calculate P_fail, and call it P_b.  It would be nice if we knew that P_fail has a constraint (bounded between 0 and 1) and then we would know that P_b will inherit this constraint, and add a formula that defines the probability of failure.

Since we do not have a refinement mechanism like this, I think we will temporarily have to live with P_fail as an output.  @JacquesCarette may have a different idea though.

vajihehm:
@smiths Thank you Dr. Smith for your explanation. However, Brooks and I just had a chat with each other and he believed that if I consider "prob_fail" as output, we will get error in code generation such as after pushing  the changes last night, I have not had successful build on Travis CI. So, I have applied this approach that I considered this symbol independently as a separate group in Unitals.hs and I have added this group in the Symbols.hs and I have had successful build locally.

vajihehm:
> Did you mean to delete that completely? [I did not fully read @smiths 's comments, perhaps that is indeed the right thing to do]

Please review last commit (90b58ae), It is totally right solution and I have fixed your comments. However, I do not know why I do not have successful build on Travis although I have successful build locally and I have copied the build folder to stable folder as well.

smiths:
@vajihehm, since "Probability of Failure" cannot be an output, please add a note to it (using the Drasil field for notes) that the probability is bounded between 0 and 1.  We don't want to lose this information.  Eventually Drasil will catch up and we'll want the information on probability to propagate.  Really there is are bigger (more general) concepts of probability that should eventually find their way into the Drasil knowledge base.  For now, just add the Note field.

JacquesCarette:
@smiths Did you mean to make your comment into a 'must change' review?

@vajihehm Looks like one of the builds passed, and the other failed because of some problem with the latex in SSP (which isn't your issue). Once you deal with the remaining comments, I can merge this in.

smiths:
@JacquesCarette, yes, a must change review makes sense.  I'll do that.

vajihehm:
@smiths and @JacquesCarette  thank you for your review. @JacquesCarette as you know I has defined "probability of failure" as a value of type "**ConstrainedChunk**" as a separate category not a part of the output. So, @smiths regards this definition probability is bounded between 0 and 1 as the attached screenshot shows it and I think the requested change has been already applied and information will not be lost. If you agree please let me know, otherwise I will make the requested change.

smiths:
@vajihehm,  you told me earlier that you could not have a constraint for prob_fail.  That is what started this conversation.  You said that you cannot have it constrained unless it was an output.  When I said you can make it an output, you said, that isn't possible.  When I said, add the information on it being constrained, you said, you don't need to because you can make it a ConstrainedChunk.  As far as I can tell, we just went in a complete circle.  You could have just started with this solution, right?

vajihehm:
@smith I put this comment 5 days ago: https://github.com/JacquesCarette/Drasil/pull/1185#issuecomment-480421707 
According this comment I have applied the approach which I considered this symbol independently as a separate group in Unitals.hs not a **part of **output**** and I have added this group in the Symbols.hs and I have had successful build locally.

Issue #1179, opened 2019-04-03
NoPCM CodeGen; Bad Code


Mornix:
Generating code for NoPCM results in an invalid line being generated. This line is present regardless of the structure (ie. different from #1177) and is present only in the code generated from NoPCM. 
In `Control.py`, `func_get_input` —  defined in `InputFormat.py` — is called without the `InputFormat` prefix (or alternately having the import be `from InputFormat import func_get_input` in the Python output). 

One notable bit I've observed is the `otherwise` branch in `fApp`'s `doit` is being taken which would result in this behaviour of the omitted "prefix" (`InputFormat.`).

https://github.com/JacquesCarette/Drasil/blob/fff755edfbf71a8650fe6cbbda128ff3a37757d0/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs#L376-L385

This again comes back to `eMap` in `CodeSpec` generated from:
https://github.com/JacquesCarette/Drasil/blob/fff755edfbf71a8650fe6cbbda128ff3a37757d0/code/drasil-code/Language/Drasil/CodeSpec.hs#L231-L240

Why it is only affecting NoPCM and not GlassBR, I'm not entirely certain.

JacquesCarette:
I believe I understand this one, so there's that ;)

PR #1189, opened 2019-04-09
Fix for missing imports in GlassBR generated code

bmaclach:
This PR restores some imports that were missing in the generated GlassBR code. The problems were:

- The procedure being called in an `FProcCall` was not being recognized as a dependency for the module
- When functions are used in definitions as `Expr`, they weren't being recognized as dependencies because they were missing the "func_" prefix.


Mornix:
In the process of checking if this completely fixes #1179. It addresses at least part of it.

Issue #1186, opened 2019-04-08
Unit mismatch issue in GlassBR


bmaclach:
One of the issues with the generated GlassBR code is that interpolation fails when attempting to calculate `J`. The reason it fails is because all of the calculations assume units of Pa, but `q` actually has units of kPa, so it is off by 1000 times and out of bounds for the interpolation. The reason `q` has units of kPa is because it is interpolated from "TSD.txt", which I'm guessing has values in kPa units.

In the past, we had hardcoded unit conversions, but we intentionally removed such unit conversions (see issues #358, #948).

Would the desired solution here then be to fix our "TSD.txt" to have units in Pa?

smiths:
@bmaclach, I think the easiest solution is what you are proposing.  We could add a requirement that q be input in one unit and converted to the other.  We don't have a requirement for that right now, and adding it seems to add an unnecessary complication.  We can assume that the file is in the units of Pa.

JacquesCarette:
I agree with the short-term fix. But yes, our "data description language" ought to be able to define the units that the external data is in, and we should be able to generate conversion code to work in whatever internal units we want to use.

This might even be related to the 4-variable model, in the general case.

bmaclach:
The TSD.txt has been updated as of merge of PR #1190. Closing issue.

PR #1190, opened 2019-04-09
Updated TSD.txt to have values in Pa

bmaclach:
This PR updates the TSD.txt input file for GlassBR to have values in Pa, as discussed in #1186. This fixes one of the runtime errors that was happening in the generated GlassBR code.


smiths:
I agree with @JacquesCarette.

PR #1195, opened 2019-04-11
Fix for master

bmaclach:
The order in which my previous PR's got merged caused codefunc to be removed from an import list when it should not have been, which means master is currently building with errors. This fixes the issue.


smiths:
This sounds reasonable to me, but I will leave it to @JacquesCarette to approve the request.

Issue #1199, opened 2019-04-15
DD references IM in SSP


bmaclach:
There are some DDs in the SSP example that arise as part of the derivation of an IM. Thus, in the manual version, we reference the derivation of the IM in the DD, shown below. 
![image](https://user-images.githubusercontent.com/18634568/56147534-7b3ebe80-5f76-11e9-8ef6-c018778fd4c9.png)

We can't do this in Drasil because it would create a cyclic reference (IMs should reference DDs, not the other way around, see discussion in #1029).

Would the solution here be to write the derivation again for the DD? That would lead to some duplicated code. Or can we just not provide any information regarding derivation in the DD (i.e. just leave out the sentence shown in the above screenshot)? Since the IM will appear in the RefBy field for the DD, the reader may know to check there for more information even if we don't tell them to.

smiths:
@bmaclach, we should just remove the forward reference to the IM.  After all, the IM doesn't actually derive the DD; it just uses it.  The DD is simply the resistive shear without interslice normal and shear forces.  It is derived by summing the shear forces.  I believe that when the IM was originally derived, the deriver noticed that this collection of terms should have a name, and that the collection of terms had meaning.  We are just summarizing this name and meaning in a DD.

PR #1200, opened 2019-04-15
SSP General Definitions

bmaclach:
This PR updates the General Definitions section  of SSP to match manual, with the possible exception of the issue discussed in #1199. Depending on the resolution of that issue, I can update this PR or create a new one or do nothing.

### DIFF
--- a/code/stable/ssp/SRS/SSP_SRS.tex
+++ b/code/stable/ssp/SRS/SSP_SRS.tex
@@ -482,7 +482,7 @@ Label & Effective stress
 \end{minipage}
 \subsubsection{General Definitions}
 \label{Sec:GDs}
-This section collects the laws and equations that will be used in deriving the data definitions, which in turn are used to build the instance models.
+This section collects the laws and equations that will be used to build the instance models.
 \par~
 
 \noindent \begin{minipage}{\textwidth}
--- a/code/stable/ssp/SRS/SSP_SRS.tex
+++ b/code/stable/ssp/SRS/SSP_SRS.tex
@@ -115,7 +115,7 @@ $n$ & Number of Slices: the slip mass has been divided into & --
 \\
 $N'$ & Effective Normal Force: per meter in the z-direction for a soil surface, subtracting pore water reactive force from total reactive force & $\frac{\text{N}}{\text{m}}$
 \\
-$P$ & Resistive Shear Force: Mohr Coulomb frictional force per meter in the z-direction that describes the limit ofmobilized shear force a slice can withstand before failure & $\frac{\text{N}}{\text{m}}$
+$P$ & Resistive Shear Force: Mohr Coulomb frictional force per meter in the z-direction that describes the limit of mobilized shear force a slice can withstand before failure & $\frac{\text{N}}{\text{m}}$
 \\
 $Q$ & External Force: a force per meter in the z-direction acting into the surface from the midpoint of a slice & $\frac{\text{N}}{\text{m}}$
 \\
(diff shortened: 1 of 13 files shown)


code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -103,41 +103,41 @@ \subsection{Table of Symbols}
...
-$N$ & Normal Force: total reactive force for a soil surface subject to a body resting on it & N
+$N$ & Normal Force: total reactive force per meter in the z-direction for a soil surface subject to a body resting on it & $\frac{\text{N}}{\text{m}}$
 \\
 $n$ & Number of Slices: the slip mass has been divided into & --
 \\
-$N'$ & Effective Normal Force: for a soil surface, subtracting pore water reactive force from total reactive force & N
+$N'$ & Effective Normal Force: per meter in the z-direction for a soil surface, subtracting pore water reactive force from total reactive force & $\frac{\text{N}}{\text{m}}$
 \\
-$P$ & Resistive Shear Force Per Meter in the Z-Direction: Mohr Coulomb frictional force that describes the limit of mobilized shear force the slice i can withstand before failure & $\frac{\text{N}}{\text{m}}$
+$P$ & Resistive Shear Force: Mohr Coulomb frictional force per meter in the z-direction that describes the limit ofmobilized shear force a slice can withstand before failure & $\frac{\text{N}}{\text{m}}$
```
smiths: Please insert a space between "of" and "mobilized."

code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -389,8 +391,8 @@ \subsubsection{Theoretical Models}
                    \\ \midrule \\
                    Description & \begin{symbDescription}
                                  \item{${F_{S}}$ is the factor of safety (Unitless)}
-                                 \item{$P$ is the resistive shear force per meter in the z-direction ($\frac{\text{N}}{\text{m}}$)}
-                                 \item{$S$ is the mobilized shear force per meter in the z-direction ($\frac{\text{N}}{\text{m}}$)}
+                                 \item{$P$ is the resistive shear force ($\frac{\text{N}}{\text{m}}$)}
+                                 \item{$S$ is the mobilized shear force ($\frac{\text{N}}{\text{m}}$)}
```
smiths: Why is per meter in the z-direction removed?

bmaclach: This text comes from the short-name for the Unital. If I include the "per meter..." part in the short name, then it will show up everywhere (in these descriptions but also when the unital is referenced in assumptions, derivations, etc.). I originally included the "per meter..." part in the short name so that we would get that text in the descriptions, but when I noticed how it was changing everything else I changed it back. Having that text showing up everywhere made some passages unnecessarily bulky/awkward, in my opinion, though if we are okay with that I can change it back.

bmaclach: Note also that the "per meter..." information isn't lost because it will still show up in the longer definition in the Table of Symbols, and also the units here will show N/m.

smiths: @bmaclach, don't make any changes.  We'll review the whole thing again when it is complete and make sure everything "hangs" together.  The N/m part is something that I already think we might need to justify to a domain expert.  As I mentioned in the review, a domain expert will expect the force to be in N.

code/stable/ssp/SRS/SSP_SRS.tex
bmaclach: Good catch, fixed in 8b2bf80af01fe221d52bd87bbb671053812b1306.

Issue #1205, opened 2019-04-16
Case-study-specific paragraphs at start of Instance Models section


bmaclach:
In the manual SRS for SSP, this is the start of the Instance Models section:

![image](https://user-images.githubusercontent.com/18634568/56241463-ed410180-6063-11e9-86d3-2d72585bc008.png)

The first paragraph is generated automatically by Drasil for every case study. The following paragraphs are specific to SSP, but there is no way in Drasil to add these paragraphs to the IM section.

I checked the other manual case studies and the only other one with additional paragraphs like SSP has is SWHS.

Was it correct to include this information here in the manual version of SSP? If so, I assume we should add to Drasil the possibility of adding additional, case-study-specific paragraphs like these to the IM section?

JacquesCarette:
To include (or not) such paragraphs is an exposition choice -- which @smiths should make. We can then change the `DocLang` to allow it, and that is quite easy to do.

smiths:
@bmaclach, I definitely want to include the instance model introduction text for SSP.  For some cases, the IMs will need some overview comments to help orient the reader.  I think we should incorporate this into `DocLang`.  It is a shame that examples like this keep coming up, where we thought the Drasil version matched manual, but information was actually removed.  Thank you for noticing.

bmaclach:
Thanks for the replies, reassigning this to myself to change the `DocLang` and add the paragraphs for SSP. I can also add the paragraph for SWHS while I'm at it.

smiths:
Sounds good!

bmaclach:
This is done for SSP (#1240) and SWHS (#1248). As I mentioned before, the other case studies do not have any additional paragraphs for their Instance Models intro. However, I think that they probably should? At least, the sentence describing how the IMs map to the goal statements seems like the kind of sentence that we would want in all examples. Do you agree @smiths?

smiths:
Yes, @bmaclach, I agree.  Please proceed with propagating this to the other examples.

bmaclach:
With the merges of #1208, #1232, #1240, #1248, #1274, #1291, and #1315, this is complete. Closing issue.

PR #1208, opened 2019-04-17
Additional paragraphs in IM section

bmaclach:
This PR updates the `DocLang` to allow for additional paragraphs to be added to the beginning of the instance model section, as discussed in #1205. For now, none of the case studies use this. I will update SWHS and SSP in future PRs.


bmaclach:
Looking at all of the manual versions of the case studies, none of them have any additional paragraphs in the Theoretical Models, General Definitions, or Data Definitions sections. @smiths Can you foresee a scenario where we would want additional paragraphs for those sections?

smiths:
Yes, I can see scenarios where we need custom text to introduce the theoretical models, general definitions etc.  We might want to orient the reader to the reasoning for classifying some particular thing as a theoretical model, for instance, or maybe we might want to explain a relationship that exists between the general definitions.  However, the fact that this hasn't previously comes up suggests that it isn't a high priority.

In my opinion, if making DocLang more general to consider problem specific text for the different sections makes the design more uniform, we should do it.  If the change is tacked on and unnatural, we shouldn't do it.  I certainly wouldn't suggest that we spend a great deal of time on implementing DocLang support for problem specific text in theoretical models, etc.  We should only do it at this time if it is a relatively easy task.  If it is time consuming, we can put it off until the date in the future when the need for this kind of problem specific customization arises.

bmaclach:
Since I've already implemented it for the Instance Models section, doing the same for the other sections should be very easy. I expect the changes I will have to make for them will be directly analogous to the ones I've already made for Instance Models, so my vote would be to add the functionality for the rest of them now, to keep things consistent.

Then the only question is the one regarding the order of parameters. I think it makes sense to have the `[Sentence]` first since that introduction text will be rendered at the beginning of the section in question. @JacquesCarette if you agree, I'll go ahead and make the updates (adding the `[Sentence]` parameter to TMs, GDs, and DDs). Also, is it okay to do them all in this PR? Or a separate PR for each section?

JacquesCarette:
I agree that, in this case, it makes sense to have the new argument be first. Since they are all related, it makes sense to do them in this PR (but if you've already done this in separate PRs, that's fine too).

bmaclach:
Added the same functionality for TMs, GDs, and DDs.

Issue #1213, opened 2019-04-18
Representing a tuple in Expr


bmaclach:
IM4 in manual SSP appears like so:

![image](https://user-images.githubusercontent.com/18634568/56385596-a3842280-61ed-11e9-9612-1298a32ccc49.png)

Note that the left-hand side of the equation is a tuple, which we can't currently represent in Drasil as an `Expr`. 

The underlying problem here may be that a single IM has more than one output, which I know @oluowoj encountered during her updates for GamePhys as well. I'm not sure what the resolution was for her case, if there was one.

JacquesCarette:
Functions really ought to be able to produce a proper data-structure as output. That should be fairly easy to do.

What is a fair bit harder is to allow pattern-matching notation on the lhs of a definition to decompose that structure and assign each variable the right 'piece'. In a properly typed language, the technology to do this is understood. In Drasil, not so much.

For now, the best hack is to use 3 definitions: the function returns a 2-tuple, which is named, and then you use an explicit projection to extract the components. But this feature request really ought to be implemented too.

bmaclach:
I want to make sure I'm understanding your suggested short-term solution correctly. Would the "explicit projections" be in the form of two new data definitions that relate the 2-tuple to its components?

JacquesCarette:
Yes.  New DD (that should be quite generic, so maybe outside SSP) corresponding to `fst` and `snd` in Haskell.

bmaclach:
Thanks @JacquesCarette . I wasn't thinking of generic DDs... I'm confused about how that would work. First, I don't see any precedent for defining DDs outside of a specific example in Drasil. More importantly, if I were to define some generic DDs for `fst` and `snd`, how would I then tell Drasil that these DDs apply to the 2-tuple I'm using in SSP?

What makes more sense to me, and maybe this is what you meant in the first place, would be to define some new symbols, outside SSP, for `fst` and `snd`, and then define DDs, in SSP, that use these new symbols to relate the 2-tuple to its components. The DD in SSP would be built from something like this:
```
critCoordsDDEq :: Expr
critCoordsDDEq = apply second [returnedTuple]
```
where `second` is the generic symbol defined somewhere outside SSP, and `returnedTuple` is the name of the 2-tuple. Then my only question would be where best to define `first` and `second`. But I might also be completely on the wrong track.

JacquesCarette:
[...]

Issue #1083, opened 2018-12-24
Change in notation in SSP


JacquesCarette:
In change cbbeb75ba, I removed { } from being special. SSP used these inside the notation for one particular symbol.  That is rather odd notation...  It needs reconsidered.

smiths:
Yes, the  { } do lead to an odd notation.  The idea was that there are a set of coordinates.  We could probably re-work it to be the more traditional sequence of coordinates.  I'll keep this in mind when I'm doing a detailed review of the latest revision by @bmaclach.

bmaclach:
There are many places where we used this notation for a set of coordinates in the manual version of SSP (for the set of slope coordinates, the set of water table coordinates, the set of slip surface coordinates, and the set of critical slip surface). I would agree that representing them as a sequence would be better, especially since it is feasible and even likely that there would be repeats somewhere in the sequence, so sets are really not even an appropriate choice. Should I update all of the coordinate `Unital`s to use vector symbols, at least for the short term?

smiths:
@bmaclach, switching to vector symbols seems like the right move.  Yes, go ahead and do that.  :smile:

bmaclach:
With the merge of #1222, all unitals that were really sequences of values for each slice have been switched to use vector notation. @smiths I think this issue can be closed?

smiths:
Yes @bmaclach, I agree that this issue can be closed.

PR #1214, opened 2019-04-18
SSP Instance Models

bmaclach:
This PR updates the SSP IMs to match manual, with these differences still outstanding:

- Coordinates should probably be sequences of coordinates (see #1083)
- The last IM should return the critical slip surface coordinates in addition to the minimum factor of safety, but I don't know how we want to handle multiple outputs (see #1213)


bmaclach:
Is that a sign that the PR is too big?

JacquesCarette:
Yes, very much so.

bmaclach:
Okay, good to know.

Issue #1220, opened 2019-04-23
SSP Constraints cannot be expressed in Drasil


bmaclach:
Most of the constraints on input/output variables in SSP are more complicated then simply constraints on the range in which the value can lie, which seems to be the only type of constraint currently definable in Drasil.

The following screenshots show the constraints from the manual SSP SRS:
![image](https://user-images.githubusercontent.com/18634568/56612203-b6c63200-65e1-11e9-8b42-73343e556221.png)

![image](https://user-images.githubusercontent.com/18634568/56612223-c5ace480-65e1-11e9-971c-a324acee946a.png)

All of the constraints that are expressed in natural language in the manual version are not currently possible in Drasil. For the monotonicity constraint specifically, there has been some discussion in the past in issues #295, #453, but no solution has been decided upon yet.

JacquesCarette:
There used to be a hack for monotonicity... but it was such a hack that it prevented other progress, so was removed. We'll need to re-add something for this.

bmaclach:
Issues for the various "types" of constraints present here: #453, #1795, #1796, #1797, #1798, #1799, #1800.

PR #1221, opened 2019-04-23
SSP Inputs

bmaclach:
This PR updates the SSP inputs to match manual. The constraints for many of the inputs are left blank for now because we can't express them with Drasil's current Constraint possibilities (see #1220).


bmaclach:
@smiths I added all of the constraints to the Notes field of the critical slip surface identification IM. I also added some FIXMEs to remind us to replace that with actual constraints once that is possible.

smiths:
Sounds good.

PR #1232, opened 2019-04-25
SSP Referable Goals

bmaclach:
This PR makes `ConceptInstance`s out of the goal statements for SSP, so that the goal statements are `Referable`, which is needed for #1205.


JacquesCarette:
All good, but a merge is needed.

bmaclach:
Merged master.

Issue #918, opened 2018-07-19
Create drasil-theory Package


samm82:
The `drasil-theory` subpackage should be created with the code for general requirements knowledge.

## TODO
- [x] Move GenDefn (#1320)
- [x] Move DataDefinition (#1338)
- [x] Move InstanceModel (#1331)
- [x] Move Theory (#1340)
- [x] Add to wiki page

_Each move should be its own PR_

Mornix:
At least three of them (AssumpChunk, Change, ReqChunk) should be removed completely when #562 is done..

szymczdm:
Exactly what @Mornix said. Also, that will likely require Labels to be implemented as well.

samm82:
I've updated the README for the removed modules in #1242; should I proceed with moving the rest into docLang now that Label has been implemented? All prerequisite issues have been closed. @Mornix @szymczdm

Mornix:
As far as I see, this should be “good-to-go” now, but I’d wait on a comment from @JacquesCarette to be certain.

JacquesCarette:
Yes, please do go ahead.

samm82:
Two questions @JacquesCarette @Mornix:

1. Should each individual move be its own PR? (ie. a PR for GenDefn, a PR for Theory, etc.)
2. What other chunks should be moved to docLang?
   - Citation
   - CommonIdea
   - Concept
   - DataDefinition
   - DefinedQuantity
   - Eq
   - GenDefn
   - InstanceModel
   - NamedIdea
   - Quantity
   - Relation'
   - Theory
   - UncertainQuantity
   - Unital
   - Unitary
   - UnitaryConcept
   - UnitDefn

JacquesCarette:
Yes, each move should be its own PR.

`GenDefn`, `DataDefinition`, `InstanceModel` and `Theory` are all "requirements" knowledge. They are not, however, strictly "documentation" knowledge, they are more general than that. I think it would make sense to remove them from drasil-lang, but to put them into their own package. I am thinking drasil-theory.

samm82:
`ChunkDB` uses these modules that are being moved to `drasil-theory`. Should it be moved there as well, or should it be split into parts, so there is a `ChunkDB` in `drasil-lang` and in `drasil-theory`? @JacquesCarette 

Also, my current folder structure of `drasil-theory` is Drasil > Theory > [modules], with a Theory.hs that re-exports everything in Drasil/. Let me know if you think it should be something different.

[...]

Issue #700, opened 2018-06-22
Removing Sentence constructors


elwazana:
@szymczdm @JacquesCarette Hello Dr. Carette and Dan, @samm82 and I were looking through notes folder for stuff we could possibly work on and we were wondering what you meant by:
https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/notes/DrasilPotentialTasks.txt#L87

Do you mean these?
https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/code/Language/Drasil/Spec.hs#L39-L86

JacquesCarette:
I meant both those (which are almost always used as hacks, especially `sParenNum` and `sSqBr`). I also think that some of the cases of `Spec` perhaps shouldn't be there.  But that requires more analysis & thinking. I have already removed the ones that were easy to remove.

samm82:
## sParenNum

Used here for a list:
https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/code/Example/Drasil/GlassBR/Assumptions.hs#L44-L55

## sSqBrNum

Used here for a reference:
https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/code/Example/Drasil/GamePhysics/Body.hs#L192-L197

https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/code/Example/Drasil/GlassBR/Body.hs#L396-L398

https://github.com/JacquesCarette/Drasil/blob/628b54d0e17d0b818be0c36480416e5bf37a2139/code/Example/Drasil/SWHS/Body.hs#L833-L837

I'm not really sure how these usages are hacks.

samm82:
1. So should we define a new function for building a list and could we use `sParenNum` to achieve that?
2. @elwazana says that the change for `sSqBrNum` will be made once `Label` is fully implemented. Is that what you are referring to?

JacquesCarette:
1. Yes and yes. [`sParenNum` used internally in printing is fine. But it should not be exported]
2. Yes.

samm82:
Some observations (This probably isn't the best place for this - we might need to move it.) This is what the text reads now, in both gen and stable:

[...]

PR #1242, opened 2019-05-01
Documentation Update

samm82:
Minor PR with updates to documentation for #700 and #918.

Also, I like that `master` is a protected branch now 😁 - makes it a bit of a headache for small changes but a LOT less of a headache for big ones!


samm82:
What is the new workflow with the protected `master` - should I wait for @JacquesCarette or can I go ahead and merge this PR now that you've reviewed it @smiths?

smiths:
@samm82, we've never really discussed this, but I think one approval is enough to proceed.  If I am not confident in my opinion, I'll leave a comment, rather than an approval.  If @JacquesCarette feels that we need two approvals, or some other scheme, he'll let us know.

Issue #1246, opened 2019-05-01
Stable files have lots of tabs


danscime:
Going through some of the stable files, I've noticed that a lot of the tables have more and more tabs for every line in the table.

Are these tabs necessary?
If they are necessary, should they be changed to be only 2 spaces instead of 4?
When adding new Data Definitions for example (as per #1070), should a similar format be followed?

Mornix:
This is a bug in the TeX printer that @deviprasad135 is working on addressing.

deviprasad135:
I have already fixed this bug in the TeX printer.

smiths:
@deviprasad135, is this fix in a pull request?  It would help @danscime if you included a pointer to the pull request.

deviprasad135:
I have fixed this bug on indentTex branch in commit 0863b30. This bug is caused due to missing parentheses.

![OI](https://user-images.githubusercontent.com/30101689/57036786-4fccfc80-6c23-11e9-831d-20426745f40c.PNG)

The generated code for one of the data definition from SWHS after fixing the over indent bug and applying 4 spaces as indentation is as shown below.

![DD](https://user-images.githubusercontent.com/30101689/57037934-8c4e2780-6c26-11e9-9d59-36cfb6ea82a8.PNG)

JacquesCarette:
So can at least this one-liner fix be made into a PR asap?  Maybe @danscime can take care of it?

PR #1257, opened 2019-05-03
Minor latex tabs

danscime:
Addresses #1246.
Quick fix to printing files.
Then updates to stable to match tabs and other characters.


danscime:
I had done only 1 PR since I thought we wanted stable to match generated when creating a PR. Maybe I am misinterpreting things. Will be more careful with things.

smiths:
@danscime, using pull requests always takes a bit of time and experience.  If you have any advice to offer to future developers, please add it to the Contributor's Guide:
https://github.com/JacquesCarette/Drasil/wiki/Contributor's-Guide

PR #1270, opened 2019-05-06
hlint Camel Case

danscime:
So far I have only changed 1 item.
Am I correct in assuming each item with refactor should be its own commit?

Also, should I be working on this branch and pushing commits to this branch as I work on the Camel Case linting? As I push, the commits would be added to the request.

How often should I be making PRs and how should I be working on the tasks that are in a PR?

### DIFF
--- a/code/README.md
+++ b/code/README.md
@@ -25,7 +25,7 @@ top level
 7. Do not break operations across lines, unless they are sentence combinators (`:+:`, `+:+`, `(sC)`, etc.). Ex: 
 
 ```Haskell
-  S "The" +:+ (phrase $ heat_trans ^. term) +:+ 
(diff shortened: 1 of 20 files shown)


JacquesCarette:
1 commit per change is good.

For changes that are unlikely to break anything, it's ok to have a "long lived" branch (2-3 days) that accumulate commits. Once you're done, you then submit a PR.

You shouldn't submit a PR then keep adding commits to it. You should be doing that work locally instead, and submit a PR when it is ready to be merged.

For this particular case, it might be awkward to retract it. So it might be simplest if you continue to work on this branch (and PR) like you have for today. But when you're done for today, I'll pull it in, and you should start on a new branch tomorrow.

danscime:
Final commit for the day has been pushed.

Issue #1215, opened 2019-04-18
Goal Statements are not Referable


bmaclach:
The instance model intro text for SSP and SWHS, discussed in #1205, contains references to individual goal statements, but in Drasil each Goal Statement is simply a `Sentence` and is not `Referable`. The entire list of goal statements is `Referable`, but without being able to refer to individual goal statements we cannot say how each IM addresses a particular goal.

JacquesCarette:
Yep, that should be fixed.  This basically means that Goals are not 'important' enough as a unit of knowledge in Drasil, and that needs fixed.

Mornix:
Goal statements previously existed in drasil-lang but a bisect shows it was removed roughly six months ago. A comment at the top of the file mentioned `Goal` chunks should be moved to DocLang instead. The content captured by the old `Goal` chunk should be easily adapted to `ConceptInstance`.

bmaclach:
So what you're saying is I can create `ConceptInstance`s from my goal `Sentence`s, since `ConceptInstance` is `Referable`? If that's right, I can take ownership of this since I need it for #1205 and it shouldn't be too hard.

JacquesCarette:
Yes, that is indeed what he's saying.

bmaclach:
This is done for SSP (#1232) and SWHS (#1248). The other case studies do not reference goal statements anywhere in the SRS but I would think we should make their goal statements `ConceptInstance`s anyway for consistency and in case we need to reference the goal statements in the future (and if my latest comment on #1205 is correct, then we will definitely need `Referable` goal statements for all examples).

JacquesCarette:
Agreed.

smiths:
Also agreed.  😄

bmaclach:
Remaining to be done:
- [ ] gamephys
- [ ] glassbr
- [ ] nopcm

bmaclach:
With the merges of #1232, #1240, #1248, #1274, #1291, and #1315, this is complete. Closing issue.

PR #1274, opened 2019-05-06
NoPCM Instance Model Intro

bmaclach:
This PR contributes to #1205 and #1215 by adding the sentence describing how goal statements map to instance models to the instance model section intro for NoPCM.


code/stable/nopcm/SRS/NoPCM_SRS.tex
```diff
@@ -264,10 +264,10 @@ \subsubsection{Physical System Description}
 \end{figure}
 \subsubsection{Goal Statements}
 \label{Sec:GoalStmt}
-Given the temperature of the heating coil, initial temperature of the water, and material properties, the goal statement are:
+Given the temperature of the heating coil, the initial temperature of the water, and the material properties, the goal statements are:
 \begin{itemize}
-\item[GS1:]predict the temperature of the water over time
-\item[GS2:]predict the change in heat energy in the water over time
+\item[Predict-Water-Temperature:\phantomsection\label{waterTempGS}]Predict the temperature of the water over time.
```
smiths: These goals are a subset of the goals for SWHS.  I know it isn't part of this commit, but should we be thinking about ways to reuse these goals?  We have two members of  program family here.  We should find a way to only write their common parts once.  @bmaclach, if this is something that you agree makes sense for the future, can you please create an issue.

bmaclach: I actually did re-use the SWHS goals. If you look at the Goals.hs file I added for NoPCM, all it does is import these 2 goals from SWHS and re-export them. Did you have an idea for even further re-use, or is that what you had in mind?

smiths: Sounds great @bmaclach.  I didn't look too closely at the code.  What you describe is what I had in mind.  If the goal statements change, we only have to change them in one place.  This is another anecdote to keep in mind when writing about the advantages of Drasil.  (@szymczdm, this is an example you might want to highlight in your journal paper.)  :smile:

PR #1275, opened 2019-05-07
Correct Makefile diffs on Windows

### LINKED ISSUE #1272: Slash Discrepancy from Recent Merge
When running `make` on `master` on Windows 10, I get this in my NoPCM log

```
diff --strip-trailing-cr --ignore-all-space -r -X ../.gitignore -x '*.txt' ./stable/nopcm/src/java/Makefile ./build/NoPCM/src/java/Makefile
18c18
< SWHS/Control.class: SWHS/Control.java SWHS/InputParameters.java SWHS/OutputFormat.java SWHS/Calculations.java SWHS/InputFormat.java
---
> SWHS/Control.class: SWHS\Control.java SWHS\InputParameters.java SWHS\OutputFormat.java SWHS\Calculations.java SWHS\InputFormat.java
```

and a similar output in my GlassBR log. I'm guess that the issue has to do with me being on Windows. Any help would be appreciated. (Feel free to reassign to whoever can help - I'm not sure who made this change.)

Mornix:
This is a small PR which (should) close #1272. I have included @bmaclach and @samm82 as reviewers to ensure that it does indeed resolve the issue on Windows.

# What happened?
1. I tried to be "correct" and use the `System.FilePath` provided operator `</>` to join paths when generating code Makefiles. 
2. `System.FilePath` is smart enough to know it's being compiled/run on Windows and replaces `/` path separator used on Unix with the Windows-preferred `\`. 
3. `diff` doesn't care what operating system it is running on and is just comparing text in two files.

# Why this fix is "fine"
Even though `\` is the [preferred separator](http://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:pathSeparator) on Windows, it still [supports `/`](http://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:pathSeparators). To further corroborate this documentation is [a .NET CLR doc](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=netframework-4.8#remarks) states, "If you prefer to hard-code the directory separator character, you should use the forward slash (/) character. It is the only recognized directory separator character on Unix systems, as the output from the example shows, and is the AltDirectorySeparatorChar on Windows." And while we are (now) hard-coding `/`, this is more necessitated by `diff` and our testing infrastructure of comparing text output. So while we are not using Windows' preferred separator, this should cover all OS'.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
@@ -22,7 +22,6 @@ import Language.Drasil.Code.DataDesc (Ind(WithPattern, WithLine, Explicit),
 import Prelude hiding (log, exp, const)
 import Data.List (intersperse, (\\), stripPrefix)
 import System.Directory (setCurrentDirectory, createDirectoryIfMissing, getCurrentDirectory)
-import System.FilePath ((</>))
 import Data.Map (member)
 import qualified Data.Map as Map (lookup)
 import Data.Maybe (fromMaybe, maybe)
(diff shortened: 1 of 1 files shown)


bmaclach:
The fix works for me!

samm82:
So I left a approval (which happens to be the only review), and it says that there is 1 approving review by reviewers with write access, but I'm pretty sure I don't. (I tried to push to `master` my first day back and quickly discovered our new workflow. 😂) Does that mean I can merge this in? @JacquesCarette @Mornix

JacquesCarette:
I'm not sure... I didn't think so?  But I definitely do, so I'll do it now.

PR #1284, opened 2019-05-08
SSP Fix DD Mistake

bmaclach:
This PR fixes a mistake in the equations for the base hydrostatic force and surface hydrostatic force (they should be multiplied by the width of a slice rather than the length of the slice base/surface). After fixing this mistake, the `surfLngth` unital and `lengthLs` DD were no longer used, so I removed them.


JacquesCarette:
Someone's clock is weird!  Right now, it lists @bmaclach as having opened and commented on this '2 hours from now'!!!  Asked for my feedback from the future too. Neat.

bmaclach:
I noticed that too 😄. The pull requests I opened after this one have the correct time, so... yeah. Weird.

Issue #1277, opened 2019-05-07
Add derivation for surface hydrostatic force DD in SSP


bmaclach:
It is not obvious how the equation for surface hydrostatic force is derived. A derivation should be added to this DD. The following is an overview of the derivation, provided by @smiths:

"The derivation should start with the fact that we want to know the weight of the water on top of the slope of the soil surface.  We find this force as the volume of water multiplied by its unit weight (gamma_w).  (We should consider adding a DD for unit weight.)  We then make the geometric arguments for the volume.  We know the dimension into the plane is 1.  Therefore, we need the area of the trapezoid formed by the water on top of the soil slice."

bmaclach:
@smiths With regards to adding a DD for unit weight, unit weight is a direct input to SSP. So I don't think it makes sense to write a DD for that since we don't actually calculate it.

Also, the "Base hydrostatic force" DD equation is almost exactly the same as the one for surface hydrostatic force. The "Weight" DD follows a similar pattern as well. Should we write derivations for all of these DDs?

smiths:
Yes, you are right about the unit weight.  The data definition would be for the concept of weight.  The input is the volume and the specific weight, the output is a force (pointing in the negative y direction).  

Do we have Newton's second law anywhere?  It should be in game physics (I believe).  The theory is F = m*a.  Since we know that a = g, we have the weight (force) equal to the mass times g.  We know that the mass is the density times the volume.  So we have volume * density * g.  The term density*g is our specific weight.  If we want to go really crazy, g comes from Newton's Law of Universal Gravitation, with the mass and radius of the earth as inputs.  :smile:

Let us start with a derivation for just the DD we are discussing (hydrostatic force).  Once we like that, you are correct that we will likely modify the other similar DDs as well.

bmaclach:
This derivation has been added with #1286 so I'll close this issue. @smiths Do we want to now make similar changes for the DDs that are similar to surface hydrostatic force (base hydrostatic force, slice weight)? If so, I'll create new issues for those.

smiths:
@bmaclach, yes, let's see what it looks like if you propagate this approach to the other forces.  If it starts to get ugly, don't go too far down this path.  We might need to have another design review meeting.

Issue #1029, opened 2018-10-26
A Summary for References Who Cause Cyclic Importation Issues


halonazhao:
**The importing conflits are as below:**

Assumptions cannot import DataDefns;
Assumptions cannot import GenDefs;
Assumptions cannnot import IMods;
Assumptions cannnot import TMods;
DataDefns cannot import  GenDefs;
DataDefns cannot import IMods;
TMods cannot import DataDefns;
TMods cannot import GenDefns;
TMods cannot import IMods;

Since even some chunks in the code can be reached, those chunks may be made references by their labels instead of referenced directly, so the code for cases doesn't really present the problem. So I think the good way to show all occurances of issues is to show in document. [PDF shows all failing imports](https://github.com/halonazhao/Cyclic-Import) marked as Green.

Could you take a look see whether these information is sufficient? Could you get back to me once you see these? If I can get more instructions, I would still get time to make changes before the meeting. Thanks. @smiths @JacquesCarette

smiths:
Great summary @halonazhao.  I believe I have found a relationship between TM, GD, DD and IM that will remove these problems.  It will involve reclassifying some of the chunks we currently have in our case studies.  I believe there will be a significant benefit though - the distinction between the types of information will become much more clear.  I am going to run my idea by @JacquesCarette before we invest time into changing the examples.

JacquesCarette:
Sounds good.  Hopefully these thoughts are compatible with https://github.com/JacquesCarette/Drasil/wiki/General-Definitions which I wrote over the weekend. They are mainly about GD, but do also have some more general comments.

smiths:
@JacquesCarette, I don't see any inconsistencies between your wiki post and my thinking about the relation between GD, TM etc.

@halonazhao, I propose that we impose the following relationship between our chunks:

[RelationsBetweenTM_GD_IM_DD_A.pdf](https://github.com/JacquesCarette/Drasil/files/2527038/RelationsBetweenTM_GD_IM_DD_A.pdf)

The theoretical models are refined to general definitions and the general definitions are refined to instanced models.  The data definitions support the TMs, GDs and IMs.  Assumptions support everything.  The "may ref" arrows in the diagram show what is allowed to reference what.  Starting from the bottom, we would want to import in the following order: Assumptions, Data Definitions, Theoretical Models, General Definitions, Instanced Models.

[...]

PR #1286, opened 2019-05-08
SSP Surface Hydrostatic Force Derivation

bmaclach:
This PR adds the derivation for the surface hydrostatic force DD in SSP according to #1277.

Some notes for @smiths: I added a DD for the concept of weight as suggested, though I didn't go further back in the derivation than that (i.e. I didn't show how weight is derived from Newton's second law). As you thought, Newton's second law is indeed already defined in GamePhysics, but it is a Theoretical Model, and DD's cannot reference Theoretical Models according to the graph you made (https://github.com/JacquesCarette/Drasil/issues/1029#issuecomment-434077273).


bmaclach:
@smiths I left the Source field for the new DD for the concept of weight blank. What sources do we prefer to use for very general, common definitions like that one? I noticed in SWHS that the source for the definition of latent heat was the Wikipedia page for latent heat. Would it be all right to do a similar thing here, or do we want to minimize the use of Wikipedia as a source?

code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -144,8 +144,14 @@ \subsection{Table of Symbols}
 \\
 ${{\mathbf{U}^{R}}_{t}}$ & Right Surface Hydrostatic Forces on Slices: per meter in the z-direction from water pressure acting into each slice from standing water on the slope surface, assuming the entire slice has the height of the right side of the slice & $\frac{\text{N}}{\text{m}}$
 \\
+$V$ & Volume: the amount of space that a substance or object occupies. & $\text{m}^{3}$
+\\
 $v$ & Local Index: used as a bound variable index in calculations & --
 \\
+${\mathbf{V}_{wat}}$ & Volumes of Water: amount of space occupied by water for each slice & $\text{m}^{3}$
+\\
+$W$ & Weight: The gravitational force acting on an object & N
```
smiths: I thought for a second that having unbold W and bold W might be confusing, but the more I thought about it the more I'm okay with it.  unbold W on its own is not a component of bold W.  The components of bold W need an index to represent them, so there is no confusion between unbold W and bold W.

JacquesCarette:
Changes are good, but a merge still needs to happen.

smiths:
As far as the source field goes, I'm fine with reference to Wikipedia.  Technically, it is supposed to be better to cite a physical source, but I think we should relax this for common physics concepts.  We could cite a physics book I have on my shelf, or that is in the library, and nobody is ever going to follow up on the source.  If we cite Wikipedia, future users might actually go to the source to learn something.  This benefit outweighs, in my mind, the danger that Wikipedia may have false information.  When it comes to physical concepts, I've been impressed with the quality of the Wikipedia entries.

bmaclach:
@smiths We can discuss at tomorrow's meeting, but just to record my responses to your comments:

- The problem where weight is not indexed on the left-hand side of the equation is a known issue with Data Definitions where we can't modify the left-hand side (#359)
- I pushed a commit fixing the problem with γdry being referenced in the wrong DD, thanks for catching that.
- For your suggestion that "U_t = V_wat * gamma_w should be written for one slice, not for all slices", would it be enough to index the vector symbols in the equation (so they would still be bold but with an index)? Or do you want them unbolded and unindexed? Indexing them is easy, unbolding them not so. It would require defining 2 new unitals for the non-vector versions of these quantities.
- I agree with adding the new assumption.
- I agree that I'd like to consolidate the DDs if we can. The drawback is that it might look very messy with all of the different cases in a single equation.

JacquesCarette:
Re: Wikipedia or not - that seems like a false choice to me. I'm fine with using Wikipedia *and* many other sources.

smiths:
@JacquesCarette, great point about multiple sources, not just Wikipedia.

smiths:
Thank you for your feedback to my comments @bmaclach.  My response follows:

- The problem where weight is not indexed on the left-hand side of the equation is a known issue with Data Definitions where we can't modify the left-hand side (#359)

I forgot about this issue.  Thank you for the reminder.  I'm glad it is officially on our "to do" list.

- I pushed a commit fixing the problem with γdry being referenced in the wrong DD, thanks for catching that.

Great!

- For your suggestion that "U_t = V_wat * gamma_w should be written for one slice, not for all slices", would it be enough to index the vector symbols in the equation (so they would still be bold but with an index)? Or do you want them unbolded and unindexed? Indexing them is easy, unbolding them not so. It would require defining 2 new unitals for the non-vector versions of these quantities.

Indexes are fine.  The symbol can remain bold.

- I agree with adding the new assumption.

Great!

- I agree that I'd like to consolidate the DDs if we can. The drawback is that it might look very messy with all of the different cases in a single equation.

With the new assumption, I think we only have two cases: 1) one of the sides has the water table above the slope, or 2) neither side has the water above the slope.  We can discuss further tomorrow (Thursday).

bmaclach:
I addressed all the feedback with the last batch of commits, including changing the DDs to GDs, which we decided at the meeting. Since we wanted to reference the Newton's 2nd law TM from game physics, this is the first time that we needed a TM in drasil-data to reuse between examples, so I created a new Theories/ directory in drasil-data to store it. I also put the Weight GD in there since it is very generic. If the name of the directory should be changed, or if you would prefer to store TMs and GDs in separate files, let me know.

In making the requested changes, this PR got rather large. I can try to split it up into smaller PRs if you want, but they will all be dependent and need to be merged in a certain order @JacquesCarette

JacquesCarette:
This passes Travis, but there's one more conflict to resolve. You may as well hold off for ~15 minutes, as I'm about to merge something from @danscime that's likely to cause more conflicts.

JacquesCarette:
Hey, not too bad, just 2 files have conflicts.

JacquesCarette:
I guess I should wait for @smiths to re-review before merging it in, as he's the one who requested changes.

smiths:
Yes, please wait.  I'll get to my review this afternoon.

bmaclach:
For point 2., I do think that due to our new assumption the condition I have is correct. If one side is above zero, we are assuming that the second side will be too. But your suggestion is less confusing and is easy to implement so that seems like a good idea. The requested changes are on their way!

smiths:
@bmaclach, I believe I see what you are getting at with my point 2.  Since your test includes the case where the water table is exactly at the soil surface (ywt,i = yslope,i), you are correct that the case where the i-1 side is above zero is covered, because of our assumption.  I agree with you though that my suggestion is clearer.  Since it isn't hard to change, let's go with my suggestion.  Thank you!

bmaclach:
@smiths Both changes have been made in the last batch of commits.

JacquesCarette:
master needs re-merged, and @smiths should re-review.

JacquesCarette:
If @smiths agrees, this looks ready to merge.

smiths:
The revisions look good to me, although there appears to be conflicts now.

Issue #947, opened 2018-07-24
Empty Source Fields


elwazana:
Theoretical Models, General Definitions, Data Definitions, and Instance Models have a `Source` field that is sometimes empty when it should be filled with "--" to indicate the lack of information is intentional.

# TODO

- [x] Switch from `[Citation]` to `[Reference]` (#1295)
- Create constructors for intentionally-empty Source fields and modify old constructor to throw an error if Source Field is empty
    - [x] TM (#1310)
    - [x] DD (#1329)
    - [x] IM (#1327)
    - [x] GD (#1335)
- Chipmunk
   - [x] All TMs (#1310)
   - [x] All DDs (#1329)
   - [x] All IMs (#1327)
- NoPCM
   - [x] Simplified rate of change of temperature GD (#1335)
- SSP
   - [x] "Sums of the interslice normal forces" and "Sums of the interslice normal water forces" DDs (#1329)
   - [x] Heights of the left/right side of slices DDs (#1329)
- SWHS (same as NoPCM - probably one change will fix both)
   - [x] Simplified rate of change of temperature GD (#1335)
- Tiny
   - [x] DDs -> source code (#1329)
- [x] Display "--" if Source field empty (#1335)
- [x] Add design approach to Wiki

smiths:
Sounds good.  Hopefully labels are working soon.

JacquesCarette:
Hmm, I wonder what we should do when a source or refby is empty. Should that be an error?  If not, then maybe that part of the display should be omitted? We should make an explicit decision about which fields of these various things are *required*.

smiths:
Good question @JacquesCarette.  If source or refby are empty, the first question should be whether the lack of content is a mistake.  In some cases this information was left blank in the "manual" examples when it shouldn't have been.  However, I don't think I would go so far as to say a lack of content is an error. It feels more like a "warning" than an error.  

Not displaying the fields when they are empty is the cleanest looking solution.  I would be okay with that, although I marginally prefer the visual reminder of including the field, but explicitly showing the contents are empty, maybe with a dash.  The explicit dash shows the reader that this was considered and still left blank.  This is a clear indication that the content really is empty, and not just forgotten.

JacquesCarette:
As long as we do not have "easy combinators" for leaving those parts empty [because then people will use them without being forced to think about it], we could indeed just put -- as an indication of an explicit choice.

smiths:
Sounds reasonable.

[...]

PR #1290, opened 2019-05-08
Implemented Maybe [Reference] for Explicitly empty Source Fields

samm82:
As per #947, and awaiting your approval of/alternate approach to using `Maybe`, this PR allows for explicitly empty Source fields, as opposed to Source fields that are left empty by accident.


samm82:
Closed, since the new approach discussed in #947 is better and will most likely be done in a different branch.

PR #1293, opened 2019-05-09
add load,capacity and separate output symbols for Theoretical Model

vajihehm:


vajihehm:
@smiths I have considered separate input and output symbols and I have changed the definition for these symbols that be abstract to show a **concept** as the following snapshots. I am not sure that all these symbols would be required in particular in case of "capacity" and "load" as they are same with "LR" and "q" respectively. Please approve these symbols that I would able to use these symbols in the Theoretical model, then use TM in Instance models.

p1
p2
p3
p5

smiths:
The introduced symbols look fine to me.  We do need the new symbols, since these are more abstract concepts in the theory.  In GlassBR the load turns into q, but it will be something else in another problem.  Our goal here is to support long term reuse.

PR #1304, opened 2019-05-09
making changes for new TM based o the manual version'

vajihehm:
@JacquesCarette  I have already completed the changes on TM , this file would be Ok without any error but because the Instance Model is still the old version with the old outputs, we will get the error in the Travis CI. I am working on the another branch for Instance Model and I will create pull request for IM ASAP. Then it should work fine.


smiths:
@vajihehm, I'll hold off on my review until you have updated stable.

JacquesCarette:
The build is still failing - you should investigate that and fix it. Once it passes, I will review.

vajihehm:
@JacquesCarette as I have mentioned before I had already completed the changes on TM , this file would be Ok without any error but because the Instance Model is still the old version with the old outputs, we got the error in the Travis CI. As per of your last comment about camel-case and the errors that I got as per the changes of @danscime, I changed the name of symbols in TMods.hs and that is why I requested the review. 
 
Also in another branch I have completed the changes on IM, but as TM was old version I got the error in generated code, so I had to merge two branches  `glassSRS_TM` and `glassSRS_IM_SafePb`. Right now both changes IM and TM are available on `glassSRS_IM_SafePb` with the successful build and it is not required to merge `glassSRS_TM` no longer and this pull request can be closed without merging to master.

JacquesCarette:
So is this PR obsolete then? If so, please close it.  However, you should make sure that none of your changes are lost by doing that.  If the PR is not obsolete, it should be fixed.

vajihehm:
@JacquesCarette As I mentioned before I have merged to the `glassSRS_IM_SafePb`, So I no longer need these changes. However, I did not make sense why for this branch conflict have been arisen.

JacquesCarette:
The conflicts are because other people have made changes in parallel (which have been merged) in the same area as your changes.  In any case, I will close this PR and delete the branch, as you're saying it is no longer needed.

Issue #1294, opened 2019-05-09
Create drasil-database Package


samm82:
As mentioned in #918, the `drasil-database` subpackage should be created with the code that deals with "'database' aspects of information collection."

## TODO
- [x] Move Language.Drasil.ChunkDB (#1306)
- [x] Move Language.Drasil.ChunkDB.GetChunk (#1306)
- [x] Move Language.Drasil.SystemInformation (#1306)
- [x] Add to wiki page

samm82:
If `Language.Drasil.SystemInformation` uses `ChunkDB`, does it make sense to move it to `drasil-database` as well? @JacquesCarette

JacquesCarette:
Yes.

samm82:
@JacquesCarette Done in #1306.

PR #1306, opened 2019-05-10
Added drasil-database Subpackage

samm82:
As per #1294, the `drasil-subpackage` was created with the following modules moved to it from `drasil-lang`:

- `Language.Drasil.ChunkDB`
- `Language.Drasil.ChunkDB.GetChunk`
- `Language.Drasil.SystemInformation`

@JacquesCarette `make` runs fine with no differences, so everything "works" - I'm just not sure if I've matched the design you had in mind.

The main things that come to mind:
- Should `Database.Drasil` re-export everything like `Language.Drasil`?
- I had to add some imports to `Language.Drasil` for use in `Database.Drasil`. Are these hacks/should they be moved to the development package?


JacquesCarette:
`Database.Drasil` should export only those things that are needed externally. `Language.Drasil` was overly generous there. Now's a good time to clean up.

The few exports you added seem harmless enough. We can revisit later.

JacquesCarette:
So I'm going to wait just a little to merge this in, as it may cause other things to fail. But it is ready.

JacquesCarette:
And I was just about to, but now there's a conflict in the way...

samm82:
@JacquesCarette Fixed!

PR #1324, opened 2019-05-13
[Small PR] Updated +/- signs in Moment Equilibrium GD to match caseStudies

### LINKED ISSUE #919: Update "Moment Equilibrium" General Definition to match manual version (SSP)
This is the moment equilibrium general definition in Drasil: image and the following is the one from the case studies repo (differences have been hightlighted): 
image.
Since @smiths mentioned in #419 that the manual version has been corrected, the changes need to be propagated to Drasil's version of SSP.

samm82:
Closes #919


bmaclach:
I am pretty sure Drasil was correct in this case, and the manual version was wrong. Though the issue you linked does say otherwise, so this may need a close review.

I stopped updating the manual version of SSP so there were some things I changed while porting it over to Drasil that aren't reflected in the manual version. This is one of those cases. To see why I think the manual version was wrong, see the derivation of IM2. With the signs from the manual version, I think the whole result becomes negative? Double-checking now, but wanted to flag this.

If we decide the manual version is wrong, we should update it to avoid further confusion. It was an oversight on my part to not do that in the first place, so sorry for the confusion.

samm82:
I wondered about this, but decided to follow the original issue.

It doesn't matter if the whole result becomes negative, since it equals zero. I just figured the consistency would allow the issue to be closed @bmaclach. No worries! I was just trying to close some old issues.

bmaclach:
Sorry, I was vague again. By "whole result becomes negative" I was referring to the RHS of this equation from the IM2 derivation:
image
Then it would lead to a negative value of lambda. (The signs in the screenshot are what I believe to be correct).

samm82:
So should we just close this PR and issue? The current version (without this PR) matches this derivation.

bmaclach:
I worry now because @smiths said the manual version was correct in the linked issue. So I think we should get some confirmation from him, one way or the other.

samm82:
I'll close this PR so it doesn't get merged until @smiths takes a look at it. 👍

smiths:
I've had a look at this equation again.  This really highlights the need for a derivation section for our equations.  I know I've done this before, but I start at the beginning every time.  This is a bigger picture comment, but it highlights again the value of capturing the derivation knowledge once.

I'm afraid that I still like the case study version.  I'm taking moments about the base of the slice, at the midpoint in the x direction.  We can assume clockwise or counterclockwise moments are positive.  The choice won't matter because we can always multiply both sides of the equation by -1.  However, the length of the moment arm does change with the +(bi/2)*tan(alpha) or -(bi/2)*tan(alpha) depending on what side one is of the midpoint of the base.  Assuming positive moments are clockwise, the first time I get in the moment equilibrium is:

-Gi * (hzi + (bi/2)*tan(alpha_i))

This is different from the equation in Drasil.  This difference propagates throughout the equation.

Looking into the equation, I found some other causes for concern/change:

- the symbols in the Figure (Force Free Body Diagram) are different than the symbols in our documentation.  This isn't a big deal, but we should update the figure.  There is a XML file that describes how the figure is drawn under version control somewhere.  It would be really nice if it was updated automatically as symbols changed, but that is too much to ask for right now.  (Something cool for the future though!)
- of greater concern, there is an inconsistency in the figure and in the labelling by Brandon in his original papers.  Is h,z,w the height of the water table, or the height of the water force from the adjacent slice?  If it is the height of the force, then the equation is fine (with the modifications consistent with what I said above).  However, if it is the height of the watertable, then we need to multiply by 1/3.
- The definition, which starts with Brandon's papers, for hz is confusing.  It isn't really the height of the centre of a slice, as far as I can tell, unless the term slice is being overused.  The height of the midpoint of a slice is the height between the "slices" at i-1 and i.

I'm glad that this issue came up again.  We might have another example of the value of documenting derivations that we can share with @JacquesCarette and @szymczdm.  😄 

I think we need to settle this once and for all.  Given that @bmaclach has become the SSP expert, my recommendation is that @bmaclach document the derivation of the moment equilibrium equation.  The derivation doesn't have to be in Drasil to start.  It can be in the manual version, or even written on a piece of paper and scanned, but I think we need to see where we are all coming from.  Once we all agree on the derivation, I do think it should be added to Drasil, so that we can record this for once and for all.  If we find mistakes/omissions in Brandon's work, we can use this for future examples.

If you agree with this course of action @bmaclach, can you create an issue for documentation the derivation of the moment equilibrium equation?

bmaclach:
@smiths We definitely need some clarity, and documenting the derivation will hopefully provide that, so I agree with that course of action. I created #1326 for this.

smiths:
Thank you @bmaclach.  The new issue looks great (#1326) looks great.  I'm happy to meet with you to discuss this further, since the derivation might involve some solid mechanics you haven't seen before.  We don't want to have too much scope creep with this issue, but we might consider bringing in more material from the Game physics example, such as the definition of Moment (probably called torque in that document).

PR #1331, opened 2019-05-14
Moved InstanceModel to drasil-theory

samm82:
As per #918, `InstanceModel` was moved to the `drasil-theory` package.


JacquesCarette:
Needs a master merge.

PR #1333, opened 2019-05-14
Removed old NFR Code

### LINKED ISSUE #1229: Use new Non-Functional Requirement format in examples
All of the examples should be updated to use the new Non-Functional Requirements list-format discussed in #1223. 

# TODO
- [x] GamePhys (#1321)
- [x] GlassBR (#1300)
- [x] NoPCM (#1282)
- [x] SSP (#1273)
- [x] SWHS (#1282)
- [x] Remove generated NFR intro from non-priority reqs (#1333)
- [x] Remove old NFR code (#1333)

samm82:
Closes #1229 - removed the code for generating the NFR section the old way, as well as for generating the intro to the NFR section.

@JacquesCarette I commented out the code for generating the intros (since it wasn't correct and @smiths said it isn't worth it to fix right now since the section is essentially a placeholder) in case we wanted to use it in the future - if you think I should just remove it entirely, let me know.


smiths:
Thank you @JacquesCarette to be sensitive to the removal of text.  It is a good idea to scrutinize when something is removed.  In this case the change is what I was hoping for.  We should keep the NFRs simple, until we have a chance to think more deeply about how best to present them.  I'm fine with this PR (except for the conflicting file).

Issue #1307, opened 2019-05-10
Add Missing Quantities to FR1 Table in SWHS/NoPCM


samm82:
The table for FR1 in SWHS and NoPCM is missing some quantities which should be added:

![image](https://user-images.githubusercontent.com/35857611/57538630-d5813400-7316-11e9-9339-920848ccf521.png)

_Migrated from #1281_

# TODO
- [x] Add tolerance values (#1330)
- [x] Add time step (#1334)

samm82:
After adding the tolerances to SWHS, I realized that the word "tolerance" doesn't come up at all in the rest of the SRS - I think this is an issue, taking an input and not defining where its used.

@smiths Should something be added to the SSP SRS to account for these tolerance values? (ie. is this issue snowballing 😂)

samm82:
@smiths Also, there isn't a `time_step` unital defined in SWHS and I'm not sure what information it should contain. (Does it have uncertainty, etc.)

smiths:
@samm82, when you say tolerance values, you are referring to AbsTol, RelTol and ConsTol?  The problem with saying how they are used breaks the abstraction of the requirements document.  The requirements are to solve for the temperature, but we don't say how to solve for the temperature.  However, when we actually solve it, we do it numerically and the calculations require tolerance values.  The tolerances are inputs, so they are in the requirements document, but their meaning comes from the design, so the equations aren't given.

How best to deal with this isn't entirely clear.  This would be a place where our code generator will eventually help.  We can decide in the design stage what numerical parameters are needed, and they can be automatically added to the requirements and to the code, with a pointer to the design decisions.  As we change these design decisions, the code and documentation can be automatically generated.

At this time, I think it would be distracting to do a deep dive on the tolerance values.  At least for now, we can leave them as they are.

The same thing applies for the time_step.  We are solving with an ODE solver, so we introduce a time step.  However, there are other ways to solve an ODE.  For instance, we could solve it symbolically.  In that case, we wouldn't need a time_step.  The presence of this input reveals a design decision.

Again, I think we want to avoid getting to involved in this at the moment.  We need the inputs in the requirements for the code generation.  As the design approach takes shape, the separation between requirements and design should be more obvious.  @JacquesCarette and @bmaclach, this is something we should think about when we are thinking about the bridge between requirements and code.

samm82:
@smiths So just to clarify, for now, I should leave the tolerance values in the table, but not add the time step? Or should I not add any and close this issue?

[...]

PR #1334, opened 2019-05-14
Added Time Step to SWHS

samm82:
As per #1307, added time step to SWHS.

_Closes #1307_


samm82:
@JacquesCarette The issue with making them `ConstrConcept`s is that the list `swhsInputs` is of type `[UncertQ]`, since the Table of Input Data Constraints has an Uncertainty column. The instant fix that comes to my mind is changing `UncertQ` to have a `Maybe Uncertainty` that will display "--" if `Nothing`, but that just seems like the same hack as zero uncertainty.

JacquesCarette:
Not all examples have uncertainties. So take a look at the other examples too. I'm guessing that at least one other involves mixed inputs, and you'll find a better solution there.

samm82:
@JacquesCarette I've looked through the examples, and in every example with a Table of Input Data Constraints, each value has an uncertainty. Most of them are 10%, which might be placeholders.

JacquesCarette:
Indeed - that's a problem. This is a bigger issue, and should have its own issue. And this PR will be blocked on that, as we do want to design this properly rather than just hack it in.

samm82:
What if I set these values to the default 10% uncertainty, then I'll open up an issue with for determining what the uncertainties should actually be/how best to denote no uncertainty? This way, these values will be kept in the same way as the others (and time step won't be lost), and we can make a pass over all examples/quantities for this new issue.

JacquesCarette:
Sounds like a good plan.

samm82:
I'll make the issue, and this PR is now good-to-go.

PR #1336, opened 2019-05-14
[Small PR] Fixed Display of "and" for Constraints

### LINKED ISSUE #865: Input Data Constraints Table | a and b vs. AR
- [x] Switch "and" to "∧" in  Drasil (left) (#1336)
- [x] Switch the use of _a_ and _b_ to _AR_ in Drasil - fixed in smiths/caseStudies#129
- [x] Switch to "0 < b < a" in Drasil
- [x] Switch to "0 < b < a" in caseStudies

ab/ar

samm82:
Updated the display of constraints to use a helper function that "ands" the expressions together, instead of converting them to `Sentence`s and using `foldlList` on them.

The "and" symbol displays in italics in the HTML, which might be undesired; if it is, it is still different than what this PR fixes, so it should be tackled in another issue.

Closes #865 (FINALLY!! It took me soooooo long to find where the display of the constraints was controlled so I could make the fix!)


code/drasil-data/Data/Drasil/SentenceStructures.hs
```diff
@@ -233,20 +233,15 @@ constraintToExpr c (EnumeratedStr _ l) = isin (sy c) (DiscreteS l)
...
-foldIfNotEmpty _ [] = EmptyS
-foldIfNotEmpty c e  = E $ foldConstraints $ map (constraintToExpr c) e
-
-foldConstraints :: [Expr] -> Expr
-foldConstraints []     = error "foldIfNotEmpty didn't catch an empty list of constraints."
-foldConstraints [x]    = x
-foldConstraints (x:xs) = x $&& foldConstraints xs
+foldConstraints :: (Quantity c) => c -> [Constraint] -> Sentence
+foldConstraints _ [] = EmptyS
+foldConstraints c e  = E $ foldl1 ($&&) $ map (constraintToExpr c) e
```
samm82: `intersperse` isn't applicable anymore when using the `($&&)` operator on the expressions (since I'm no longer hardcoding `S "∧"` as a `Sentence`), but I think this is a nice compromise. @JacquesCarette

JacquesCarette: Agreed.

PR #1365, opened 2019-05-17
[Small PR] Sorted FR Input Table in GlassBR

samm82:
Sorted FR Input Table in GlassBR by symbol. Just a small change I'd make to improve consistency between the other tables.

### DIFF
--- a/code/drasil-example/Drasil/GlassBR/Requirements.hs
+++ b/code/drasil-example/Drasil/GlassBR/Requirements.hs
@@ -6,6 +6,7 @@ import Data.Function (on)
 import Data.List (sortBy)
 
 import Language.Drasil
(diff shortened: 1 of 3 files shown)


samm82:
I figured that this table doesn't correspond to a table that would be generated for another example, so sorting it here is part of generating it, since we might want to build a table that isn't sorted by symbol in another example (or in the same example). Should I make a separate issue for discussing this or just delete the branch? We could pass in a "sortBy" parameter when generating tables. @JacquesCarette

JacquesCarette:
Ah, indeed. Ok, if this table only exists here, then it's ok to sort it just here. Infrastructure changes can come later, we need to be careful to prioritize those properly.

PR #1369, opened 2019-05-18
Research grant

oluowoj:
@smiths, @JacquesCarette , would you have a template that I can use?


JacquesCarette:
Why is this PR titled "research grant" but seems to be about trying to merge old changes in? Did you want to do a PR or an issue here?

oluowoj:
I pushed the branch that I am going to be using for the Catalyst grant proposal, I actually don't even need a pull request yet but it was a pull request.

oluowoj:
> I pushed the branch that I am going to be using for the Catalyst grant proposal, I actually don't even need a pull request yet but it was a pull request.

And I asked a question about the template

JacquesCarette:
What I'm saying is that the researchGrant branch shouldn't have all sorts of other work on it. You should start a clean branch based off of current master for that.  And you shouldn't be asking questions based on a PR, but on issues instead.

I'll answer the question here (but then close this PR): no, we don't have a template.

PR #1370, opened 2019-05-18
making changes for instance model based on manual SRS version

vajihehm:
@JacquesCarette I had a successful build locally, however I had a trivial problem that I was not able to create space between two instance model tables. Could you please let me know how I can create space between two tables of instance models?


JacquesCarette:
Don't worry about the lack of space between the two tables, as if you've done things correctly, that is an infrastructure problem.

vajihehm:
@JacquesCarette I just ran `make clean` and then `make glassbr_diff`, I coped the build files into stable folder, however I have not already had changes to commit as the following snapshot.

JacquesCarette:
The output of `make glassbr_diff` should have contained some warnings in the log, and you need to fix those.

vajihehm:
@JacquesCarette I took a look at the log file, It seems I should copy `src` folder from `build` folder to `stable`, I was not coping this folder before! Should I do? It seems there are differences between source code folders.

JacquesCarette:
And you also need to fix the warnings, they won't go away on their own.

You should not just copy the new sources to `stable`. You should first figure out if each change is expected (i.e. can be explained by your changes) and, if so, then accepted. Otherwise you should file issues so that these can be investigated more deeply.

vajihehm:
@JacquesCarette I am wondering why some functions were being generated twice (e.g. in `Calculations.cpp` like `func_q` in `stable` and `func_is_safePb` in the `build ` folder), should I change  manually in the `stable` version? as neither build folder is correct nor stable folder.

JacquesCarette:
You fixed part of the problems with the build, there are more.

vajihehm:
@smiths Please take a look at the `Notes` for the theoretical model, I think it should 'nt mention to `glassbr` as it is the concept of safety which is something that should generalize from the GlassBR example. do you agree to take out the `glassbr` from the Notes in TM?

smiths:
@vajihehm, yes, I agree that the theoretical model should not mention glass.  It should be more abstract than this specific problem.  It should just be about the concept of safety.  Please go ahead and modify the wording to not mention glass.

JacquesCarette:
So what is the holdup on this branch?  I did not think that modifying the wording would be so much work - and there has been 2 full working days elapsed since that was asked for.  And now, of course, conflicts have arisen.

vajihehm:
@JacquesCarette No, I did not work on it and I know It will not need much work. I just wanted replace `glassbr` into `system`, but it seemed not to be suitable word and I prefer create another branch for other issues like this. At this time this branch is absolute.

vajihehm:
@JacquesCarette Should I did not make sense why these conflicts have been arisen, Should I resolve manually from command line as the following steps or manually:

JacquesCarette:
You need to do the merge as per those steps. Note that you will have to carefully edit things to actually 'resolve' the conflicts.

vajihehm:
@JacquesCarette I was wondering why these conflicts have been arisen.

JacquesCarette:
Because many other people are working in parallel, and their changes to the same files got merged in before yours did - and you both made changes to the same lines, so they couldn't be merged automatically.  https://imagej.net/Git_Conflicts

bmaclach:
Just a heads up to @Mornix - with this merged in, I confirmed that the last errors on the generated GlassBR code are gone. If you were planning on adding actually building/running the code to our Travis builds (I think you were?) that should be possible now.

Issue #1332, opened 2019-05-14
Add derivation for base hydrostatic force and slice weight in SSP


bmaclach:
Derivations should be added for base hydrostatic force and slice weight in SSP, similar to what was done for #1277. These derivations will require references to General Definitions, so the DDs for base hydrostatic force and slice weight will also need to be switched to GDs. If this gets too messy, a new design may need to be discussed.

bmaclach:
With #1345, #1380, and #1392 all merged, this is done.

smiths:
Great!

Issue #1326, opened 2019-05-13
Document Derivation of Moment Equilibrium Equation for SSP


bmaclach:
The moment equilibrium equation in SSP has inconsistencies between manual and Drasil and even within manual. See #1324 for in-depth discussion on these inconsistencies and other ambiguities. We should document the derivation of this equation to ensure we have the correct equation and to avoid future confusion. The derivation and the free body diagram in the SSP SRS should agree with each other.

bmaclach:
@smiths I pushed a first draft of the derivation to the case studies repo. Can you please review it?
https://github.com/smiths/caseStudies/blob/master/CaseStudies/ssp/docs/SRS/SRS.pdf

I brought in the Torque DD from GamePhysics. I also updated the IM2 derivation to be consistent with the moment equilibrium equation from the GD.

smiths:
Thank you @bmaclach.  This is a good start.  I did notice a few things we need to correct/clarify.  Rather than type it up, hopefully we can have a quick discussion during tomorrow's (May 16) meeting.  If we don't have time then, we'll make a quick appointment for another time.

smiths:
@bmaclach thank you for your efforts on this.  I think after you make the changes we discussed yesterday, we will have a good derivation.  @danscime, this is relevant for you because refining the derivation found mistakes in what we did previously, and in the original papers.  The papers mis-defined a variable.  In some cases it was the height of the water table, but in other cases it was considered to be the height where the hydrostatic force was applied.  In the Zhu paper there is a missing assumption about the calculation of the moment created by the soil mass under earthquake loading.  @bmaclach will be adding an assumption about this to our documentation.

bmaclach:
@smiths I've addressed the feedback on the case studies repo. Can you please let me know if you spot any other issues with the derivation before I port it over to Drasil?

smiths:
Looks great @bmaclach.  The only thing that occurs to me is that we don't consider the case where the water table is below the base of the slice (the height of the water table as shown in Figure 4 would be negative in this case.)  Is there an assumption that the water table is always above the slip surface?  If it is below, then the formula will be wrong because the hydrostatic force shouldn't be contributing anything.  Oh, I know what happens.  H_i would be zero (by DD4), so the term related to hydrostatic force won't enter into the equation.  We don't need any special cases in GD10.

Looks good to me.

[...]

PR #1380, opened 2019-05-21
SSP Slice Weight Derivation

bmaclach:
This PR adds a derivation for slice weights, in a similar style to what was done for surface/base hydrostatic forces. This includes consolidating the 3 slice weight DDs into a single GD. This contributes to #1332.


bmaclach:
@smiths I am waiting for your feedback on the revised moment equilbrium deriviation before bringing it to Drasil (see #1326)

The base hydrostatic force equation in Drasil does not match Brandon's code. We had thought we found a mistake with Brandon's code where he was multiplying by ell_i (the length of the base) instead of b_i (the width of the slice), but now it seems like that's not the case and Brandon's version was right all along.

After reading your explanation and drawing it out, I mostly understand the derivation but I do have a couple questions. A quick meeting would be helpful. You seem to be booked up for the next couple of days though. Should we wait until Thursday? I'm available any day except Friday this week.

smiths:
Yes, I have #1326 in my e-mail inbox to remind me to look over it again.  I'll get to that soon.

I just deleted what I was writing because I think you are right when you say that Brandon was right.  I 
was originally writing an explanation that the surface load is different from the base load and then I realized that they aren't different.  We can't just talk about the weight of the water from the surface slice, like I originally told you.  The water will be pushing normal to the slope surface along the length of the surface.  The weight of the water as we calculated would be parallel to the y-axis, which ignores part of the force.  Also, the weight of the water passes directly through the axis of rotation, so it doesn't actually generate any moment.  We need to consider the resultant from the load distributed along the surface.  I'm pretty sure the equation will be the same as what Brandon had before.

Although  Brandon was likely right, we are still highlighting the value of deriving the equations.  If Brandon had shown the derivation of the original equation, our task would be verifying that derivation, not deriving it ourselves.  @szymczdm, if you were planning on using the Brandon example that I previously highlighted for the calculation of the surface hydrostatic force, the point is a little more nuanced than I originally thought.  😄 

My schedule for tomorrow (Wednesday) looks to be opening up, especially in the afternoon.  If we can find a time to meet to discuss the documentation of design, you and I could meet immediately before or after that meeting.

JacquesCarette:
(and another master merge will be needed)

bmaclach:
Merged master.

Also discussed with @smiths and decided that, since this PR isn't introducing anything wrong, we should be able to merge this one in. I will fix the mistakes with the base/surface hydrostatic force derivations in a fresh branch.

smiths:
@bmaclach, I'll merge this pull request.  Please fix the mistakes with the base/surface hydrostatic force derivations in a fresh branch, as we discussed.   I'll also delete the branch.

PR #1397, opened 2019-05-23
Began Work on Projectile

samm82:
- Added blank SRS template to Projectile
- Added manual source code for Projectile (with manual test cases)
- Added notes on equations and requirements for Projectile
- Made a change to `drasil-printers` to only define `symbDescription` in the preamble if it's used

Feel free to not delete this branch - I'll just keep working in it.


JacquesCarette:
You really should have made the bug fix in a separate branch. But, be that as it may, I'll merge the whole thing in now.

PR #1405, opened 2019-05-24
[Review] Preliminary Projectile Documentation (SSD)

samm82:
I started on the SSD for Projectile, and I think I have enough that I can start working on the SRS, I just wanted a review before I did so. In particular, I'm not super sharp on the differences between TMs, GDs, DDs, and IMs (and can't find the chart for relations between them), and I'm not sure if I've assigned them correctly. Not sure how the best way to do a review would be, so I just made a PR.

_This branch can be deleted once merged. I forgot that it's easier to make a separate branch for every change, since all of the previous commits (that have since been merged into `master`) are still visible here._


smiths:
You are off to the right start @samm82.  I suggest your next step is to follow the feedback provided by @bmaclach.  That is a nice overview of TM versus GD etc.  @bmaclach is correct that your TM is the relationship between acceleration, velocity and distance.  Your simplified ballistics problem ignores air drag, so you can formulate it as a purely kinematics problem.  That is, you don't actually need forces to enter in to the equation.  You don't need any of Newton's laws, just the kinematic relationship between acceleration, velocity and distance.

In your hand written notes, it looks like you have an acceleration in the x-direction, which either means your vertical coordinate is x, or you have gravity operating in both the x and y directions.  Let us keep things simple.  Define the vertical coordinate as y, and only have an acceleration component in the y direction to start with. 

Your requirements document should explicitly introduce the coordinate system and the sign convention.

samm82:
Thanks @smiths and @bmaclach! Your feedback is super helpful and I'll get to implementing it right now!

@smiths There is an acceleration variable in one of my original equations, but later is set to zero, with a note "assume no wind resistance/friction".

Also, I feel like the referencing chart of DDs, TMs, GDs, and IMs, as well as a summary of each, should be added to the Contributor's Guide - I'll add it, I'm just not sure what to call the section.

samm82:
@smiths @bmaclach I've made all the updates - one thing I'm unsure about is how to document the notes of (what is now) GD3 ("v at max height = 0" and "t at max height = t_total/2"). It would make sense for them to be TMs, right?

bmaclach:
@samm82 The updates look great! For the notes, my instinct would be to just add them to the "Notes" field of the GD in Drasil, or the "Derivation" field. The notes are only used as part of the derivation for the equation, right?

samm82:
@bmaclach Correct 👍

bmaclach:
Good, so when it comes time to putting this into Drasil, you could include the derivation you wrote on paper, and the notes would be captured somewhere in that derivation.

JacquesCarette:
And make sure you keep careful notes of the 'process' you used (and are using, and will be using) to create this whole example. That kind of introspection exercise will be very valuable to us shortly.

samm82:
So should the summary of GDs, etc. go in the Projectile wiki page for now?

JacquesCarette:
I am not sure what should go in the wiki vs other places!

samm82:
If everyone's happy with this, once this PR is merged in, I'll start work on adding them to the SRS.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
```diff
@@ -0,0 +1,44 @@
...
+
+Elements:
+	1) The object
+	2) The ground
+
+Assumptions
+1) The projectile motion is in 2D.
+2) The object is a point mass.
+3) The object's path starts and ends at ground level.
+4) The effects of drag, wind, and the Earth's curvature are ignored.
```
smiths: Make your assumptions atomic - that is, you should split each of these into its own assumption

samm82: Should I do the same for A3 (ie. "The object's path starts at ground level." and "The object's path ends at ground level."), or are they related enough that they can be grouped?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
```diff
@@ -0,0 +1,44 @@
...
+5) The flight is free; there are no collisions during the trajectory of the object.
+6) The positive x-direction is the direction from start to the end of the object's path.
```
smiths: I don't consider the coordinate system an assumption.  My advice is to put it in the physical system description.  This work starting on a fresh project is a good chance to revisit the rational for separating our knowledge.  I'm sure we aren't 100% consistent in the examples, but the assumptions are intended to define the scope of the problem.  The coordinate system is a modelling decision.  You could have decided to use a polar coordinate system and you would get the same answers (once they are transformed between coordinate systems.)

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
```diff
@@ -0,0 +1,44 @@
...
+7) The positive y-direction is the direction away from the ground.
+
+Goal Statments
+1) Check if the final position from user input is reached by the object with initial conditions.
```
smiths: From the physics point of view, I think the goal is simply to calculate the landing position of the projectile.  You might have requirements related to how you convey this information to the user, but everything is based off of the goal of finding the final position.  If we want to reuse this ballistics trajectory, we are better off not revealing that we are using it as a "game" in the modelling portion.  In this way we can use the same goals for a different game, or simulation, simply by changing the requirements.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: @smiths I based this assumption off of these ones from Chipmunk:

"coordinateSystemTy: The library uses a Cartesian coordinate system."
"axesDefined: The axes are defined using right-handed coordinate system."

Would it be better to say (in the physical system description) "The physical system is modeled in such a way that the positive x-direction is the direction from start to the end of the object's path, and the positive y-direction is the direction away from the ground.", and if so, should I make an issue to update Chipmunk (and potentially others) in a similar fashion?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
```diff
@@ -0,0 +1,44 @@
...
+2) If the final position is not reached, provide the offset and if the object went long or short.
+
+Theoretical Models
+1) v = ds/dt
+2) a = dv/dt
+
+General Definitions
+1) v_f = v_i + a*t
+2) d = v_i_x*t + a_x*(t^2)/2
```
smiths: You will need to add an assumption for constant acceleration, which implies an assumption that the sum of the unbalanced forces and the mass are constant.  If we had air-drag, which would add a force opposing the direction of flight, this general definition would no longer apply.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
```diff
@@ -0,0 +1,44 @@
...
+3) d = 2*(v_i^2)*sin(θ)*cos(θ)/(|a|) (from A4, GD2)
+	- v at max height = 0
+	- t at max height = t_total/2
+
+Data Definitions
+1) v_x = v*cos(θ)
+2) v_y = v*sin(θ)
+
+Instance Models
```
smiths: I think I'd prefer these instance models as requirements.  I think the instance model is simply solving for the distance travelled by the projectile.  In the current set-up for Drasil though, it might be easier to have instance models.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: @smiths Combining your proposed fix for splitting assumptions, what about:

"4) Acceleration is constant in the x-direction.
5) The effects of the Earth's curvature are ignored."
?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: So should I leave it for now?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: Thinking about A3, I think you do have one assumption, but ground level isn't really part of it.  Your assumption is that the object and the target are at the same height.  The notion of ground isn't really necessary.  It could be one aircraft launching a cannon at another aircraft at the same height.  In this way you only have one assumption.

You might actually choose to make the height of the projectile launcher and the target inputs.  If the heights input are the same, then you have met this assumption, but allowing the heights to differ, you wouldn't need this assumption at all.  This would be an example of a requirements level variability.

For now, I suggest just changing the assumption as I mentioned above - the launcher and the target are at the same height.  (You will later introduce the modelling convenience that this height is zero.)

samm82: @smiths I was thinking of adding this as a variability, but right now I just want to get this instance completely working first. If I update this assumption to be that the launch and target heights are equal (for now), should I also remove the concept of ground from the system?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: I'm following the example for the requirements for 3dfim+.  We haven't discussed this example previously.  You can find it at:

https://gitlab.cas.mcmaster.ca/SEforSC/se4sc/blob/git-svn/GradStudents/Mojdeh/SRS_3dfim+/3dfim+.pdf

The locations for knowledge in the previous example were guided by the best judgement of myself and the students working on the projects, but we weren't consistent.  The judgement wasn't as guided by principles as it should be, because the principles are not obvious.  This exercise, and the rigorous approach in Drasil, is forcing me to look more explicitly at the assumptions.  However, I'm not 100% sure that the advice I'm giving you will be the same advice I would offer at the end of this exercise.  Let us keep our work to this example, and at the end of the process ask ourselves what principles we applied, and how they should be applied to the other examples.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: I would say that the acceleration is zero in the x-direction, because there are no forces acting in this direction.  The acceleration is constant in the y-direction, as dictated by the value of acceleration due to gravity.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
bmaclach: I'll jump in here just to say that part of my motivation for suggesting adding ground was because having the object as the only element of the physical system seemed incomplete. I agree with @smiths that saying "ground" is unnecessarily restrictive. So I would replace the "ground" element of the physical system with "target". But maybe @smiths disagrees.

smiths: Yes, I think we can remove the concept of ground.  It isn't like you are modelling the ground.  It is just the name that you give the height of zero.  You could introduce ground under terminology, if you think it is a helpful concept.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: That's what I thought but not what I typed 😂 final answer:

"4) Acceleration is zero in the x-direction.
5) Acceleration in the y-direction is only caused by gravity.
6) The effects of the Earth's curvature are ignored."

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: I'd like to see what @bmaclach thinks.  It depends on the code generator, and I don't know if the code generator starts from the instance models or the requirements.  I know that it should eventually start from the requirement, which reference the instance models, but I'm not sure if that is how things are implemented.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: Yes, the target should be part of the physical system.  Thank you @bmaclach.  I've been modelling the problem in my head as "launcher" and "target," but we need to make these concepts explicitly part of the documentation.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: I'll add a Coordinate System section that says "The coordinate system is set up in such a way that the positive x-direction is the direction from start to the end of the object's path and the positive y-direction is the direction away from the ground." Sound good?

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: That's a great start @samm82.  I'm sure we'll refine as we go along, but I like how the assumptions are shaping up.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
bmaclach: The code generator uses the instance models to calculate the outputs, and then just returns the values of those outputs. So if we don't include these instance models, the generated code would return the displacement and nothing else. I'm pretty sure the code generator ignores requirements altogether.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: Should we still include the launcher? For example, the object could be "launched" by something being thrown at it, not necessary being "launched" in the traditional sense. If so, does "launch site" make sense as a component of the system? @smiths @bmaclach

bmaclach: I like the idea of "launch site" as an element of the physical system.

smiths: The notion of launcher is abstract.  How it happens doesn't matter.  I think we want parallelism between our concepts of launcher and target.  If we say launch site, then we should say target site, and then it seems like we are adding redundant terminology with the word site.   I think I prefer launcher and target, unless there is a synonym for launcher that sounds better.

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
samm82: @smiths Since we've removed ground, would the better alternative be "the positive y-direction is the direction opposite the direction of gravity" for consistency, or "the negative y-direction is the direction of gravity" for brevity?

smiths: You can just say up and down for the definition of the coordinates and sign convention.  Up is the direction of increasing y coordinate and y is decreasing height.  Gravity is defined after the coordinate directions are assigned.  We could potentially define gravity as acting upward.  If it really is a game, then we can use some strange game physics.  :-)

code/drasil-example/Drasil/Projectile/ProblemDescription.txt
smiths: I think in the long term we want the requirements to be more meaningful, but for the current situation, we can use the instance models.  We should still write requirements though.

samm82:
I think all reviews have been resolved, so it should be good to go (unless someone wants to do a second pass).

smiths:
I think we are good with this PR.  I think it is a good point to start on the SRS.  Our next review can be of that document.

JacquesCarette:
should I delete the branch, or keep it going?

Issue #1390, opened 2019-05-23
Create drasil-utils SubPackage


samm82:
As per comments on #1388, there should be a `drasil-utils`. The file structure will be:
```
+ Utils.Drasil
| --- Fold.hs (contains folde, foldlList, foldlSent_, etc. from SentenceStructures)
| --- Misc.hs (everything else until we re-allocate them)
| --- Phrase.hs (phrase builders)
| --- Sentence.hs (sentence constructors like sAnd, ofThe, etc. from SentenceStructures)
```

# TODO
- [x] Move marked functions from SentenceStructures.hs to `Utils.hs` (#1411)
- [x] Make Sentence in `utils` (#1436)
- [x] Move Phrase to `utils` (#1470)
- [x] Make Fold in `utils` (SentenceStructures) (#1448)
- [x] Make Misc in `utils` (Utils) (#1510)
- [x] Move rest of SentenceStructures to Misc (#1544)
- [x] `printers` should depend on `utils` now, not `data` (#1448)
- [x] Update Wiki

JacquesCarette:
If it is going to be mostly filled with utilities to deal with `Sentence`s, it's tempting to call it `drasil-english`. If it's a mixture of generic Haskell code and other things, `-utils` is probably right.

And yes, many of the things in `SentenceStructures.hs` are not really in the right place.

samm82:
Most functions create a `Sentence`; the only exceptions are `figureLabel` that creates a `LabelledContent` (marked with a FIXME wondering if it's in the right file) and the term formatters that return an `ItemType` (`tAndDWAcc`, tAndDWSym` and  `tAndDOnly`).

There is also a section of functions at the bottom marked "TO BE MOVED TO EXAMPLE/DRASIL/SECTIONS", which is now `docLang`, so I'll move those to there.

JacquesCarette:
The stuff that generates pieces of documents should all go into `docLang` for now.

samm82:
There already exists a `Data.Drasil.Utils`. Should this, and `Data.Drasil.SentenceStructures` (once all functions unrelated to `Sentence`s are removed) both be moved to `drasil-utils`?

samm82:
Also, instead of moving these functions to `docLang`, I'm moving them to `Utils.hs`, since they have potential to easily be reused at some point in the future. (If that's OK with you @JacquesCarette)

samm82:
[...]

PR #1411, opened 2019-05-24
Moved Functions from SentenceStructures to Utils

samm82:
As per #1390


JacquesCarette:
Changes are fine, but a merge is needed.

PR #1428, opened 2019-05-28
[Review] Start on Projectile SRS

### LINKED ISSUE #1430: Add Physical System Description Section to Projectile
This section includes a figure of the problem, which is:

projDiagram

We might also want to add the positive directions (up and to the right) at some point.

### LINKED ISSUE #1431: Add Requirements Section to Projectile
The FRs should be added using `FReqsSub`, and the NFRs should use `NonFReqsSub`.

### LINKED ISSUE #1493: Add Values of Aux Constants to Projectile
It should include the relative tolerance constant, among others.

### LINKED ISSUE #1577: Coordinate Systems Concepts
The concepts `xAxis`, `xDir`, `yAxis`, and `yDir` should be added, so that "y-direction" displays as "*y*-direction". I think they should be added to `Concepts.Math`, and the concepts for cartesian and right-hand should probably be moved here as well from `Concepts.Physics`. Does this sound OK? @JacquesCarette @smiths

### LINKED ISSUE #1584: Traceability Matrix for Projectile
@samm82, I think we should add a traceability matrix to Projectile.  My understanding is that this will not take much additional code, and the matrix will help me with reviewing the SRS documentation (#1428).

### LINKED ISSUE #1617: Add Hibbler Reference to Projectile SRS
Here is the textbook that we can use for the citations to Source in the SRS for Projectile.  Please create a bibtex entry for this book and add it to the SRS documentation.

https://www.amazon.ca/Engineering-Mechanics-Dynamics-Russell-Hibbeler/dp/0133915387

### LINKED ISSUE #1620: Update Physical System Diagram
Add:
- the coordinate system
- the direction of gravity
- the offset

### LINKED ISSUE #1621: Improve Cartesian Coordinate System Definition
@smiths you mentioned that the definition of a Cartesian coordinate system should be improved:

image

I just used the definition already defined in Drasil, and it is the same definition as the one on Wikipedia. What information is missing?

### LINKED ISSUE #1622: Improve Links
@smiths You mentioned that it is hard to see if something is a link:

image

What should be done to improve the display of links?

### LINKED ISSUE #1623: Constraint on Offset
@smiths You mentioned that the shortest possible flight is one where *p<sub>land</sub>* = 0, but since the constraint on the angle is 0 < θ < π/2, this isn't the case. Should we change the constraint on the angle to allow for a shot straight up, or the rationale on the new offset constraint (ie. *p<sub>land</sub>* is always positive)?
image

samm82:
So far, I added the Goal Statements, Assumptions, TMs, DDs, GD, and IMs, and I think this is enough to review. In particular, I'm not 100% if the concepts/unitals are defined clearly enough/if I'm using the default physics concepts too heavily. Also, I think technically the velocity entered is speed since it is a scalar with units (m/s), but since an angle is specified as well, I think it might be OK?

Closes #1430, closes #1431, closes #1493, closes #1577, closes #1584, closes #1617, closes #1620, closes #1621, closes #1622, and closes #1623


samm82:
Since the input is the launch speed, with no direction in m/s, as well as a (separate) launch angle, I think it would make sense to transition them all to scalars for consistency (but I might be wrong). I think originally I just used `velocity` from `drasil-data` and went with it. There isn't currently a `speed` concept, and I think it should be added. However, the definition of `velocity` given is "The rate of change of a body's position". Should `speed` just be defined as "The magnitude of a body's velocity"?

This would mean that the definition of `acceleration` would be weird since there aren't separate terms for vector/scalar. Would we have to define one as "rate of change of velocity" and one as "rate of change of speed" and pick which one gets used?

bmaclach:
It might be okay to make them all scalar, but you will need to use components in some cases. For example, for final velocity, v<sub>f</sub> = v<sub>i</sub> + a*t is not quite right if they are all scalar quantities defined as the magnitude of the corresponding vector. It would have to be v<sub>f,y</sub> = v<sub>i,y</sub> + a<sub>y</sub>*t (or similar for x, but the version of the equation with y-components is the one we actually use).

Defining speed as the magnitude of the velocity seems fine to me. You could similarly define the scalar acceleration as the magnitude of the acceleration vector. You could also just have a unital for a<sub>y</sub>, in which case the definition could simply be "the y-component of acceleration".

samm82:
For the v<sub>f</sub> = v<sub>i</sub> + a*t equation, would it make sense to have the original and then a refined model separately? And if so, should the original be a TM? @bmaclach

bmaclach:
@samm82 I think both can be GDs. I think of TMs as models that can't be derived. The v_f equation is derived from the definitions of v and a (our current TMs), so I think it's a GD. 

In terms of whether it's good to define the vector and scalar versions in separate GDs... I'm not sure, but I think that makes sense. I worry that it's going overboard because not every intermediate equation in our derivation needs to have its own GD/DD, but I think both versions of this equation are meaningful enough to warrant having their own GD. You could imagine re-using either of them in different examples.

samm82:
Unless it's important for derivation purposes, I think I'll leave the vector equations out, since we don't use vectors at all in this problem. (Technically we do, but we keep the magnitude and direction separate throughout the whole process.) @bmaclach @smiths Thoughts?

bmaclach:
If we are planning on explicitly adding the derivations of our equations, my opinion is that it would be best to start from the vector equations. If not, the scalar equations might be enough. Maybe start with the scalar equations for now, and if @smiths thinks we should include the vector equations too you can always add them at that point.

samm82:
I think I got all of the changes mentioned, but feel free to do another pass everyone!

bmaclach:
Great, looks like all of my comments were addressed except for one: the final velocity equation. v<sub>f</sub> = v<sub>i</sub> + a*t is only true if the velocity and acceleration are in the exact same direction. That's why I think you should use the y-components instead, v<sub>f,y</sub> = v<sub>i,y</sub> + a<sub>y</sub>*t, that way the equation is always correct (and I don't think we use the x-component version of the equation, so it's okay to restrict it to y-components). 

One other thing is that I don't think it's right to say "x/y-component of speed/distance/scalar acceleration". I don't think it makes sense to have components of a scalar. I think the equations/symbols are correct, but their descriptions should be changed to the form "x/y-component of velocity/displacement/acceleration"

samm82:
Would "distance in the x-direction" be an appropriate substitution?

bmaclach:
Yes, I would say so.

smiths:
@bmaclach and @samm82, the conversation about vectors versus scalars is great.  I did notice some mismatches in the equations, but I think those have been sorted out.  I haven't checked the most recent version of the documentation, but what we want is to use vector notation in the theoretical models.  The theoretical models are supposed to be abstract.  We haven't yet made any decisions about coordinate system, or even the dimension of the problem.  The concept of vector is more general than just a component-wise representation of the vector.  The relationship between acceleration and velocity and velocity and position are not tied to any coordinate system.  (There needs to be a non-accelerating frame of reference, but we don't need to get into that.)  We want our theoretical model to be reusable in any project that has acceleration and velocity, even if that problem has a non-orthogonal curvilinear 3D coordinate system.  :smile:

We transform the theory into an instance model by making decisions.  When we introduce the dimension of the problem (2D), the coordinate system, and the origin of the coordinate system, the concept of vector can be turned into components in each coordinate direction.  These are the scalar equations.  In 2D each vector contains two scalars.

samm82:
@smiths Right now, the TMs are speed and scalar acceleration. Should they be changed to velocity and acceleration (as a vector)? If so, should the speed and scalar acceleration TMs be moved to another section, or replaced with a definition of _v_ = |**_v_**|?

smiths:
The TMs should be abstract.  They definitely should be vectors.  You need to be very careful in thinking about the speed as the magnitude of the velocity vector.  Your initial speed is the magnitude of the vector, but your calculations will mostly deal with the scalar components of the velocity vector.

@samm82, I think it might help you to have a look at the original source material for the SRS template.  I mentioned elsewhere that we'll have a look at Lei Lai's beam example.  The papers that use that example are in the se4sc repo under the names:

SmithLaiAndKhedri2007fulltext.pdf
SmithAndLai_2005.pdf

samm82:
I've changed the TMs to be vectors and am kind of unsure as to how to proceed. @smiths 

I think one of my biggest issues is conceptualizing how things "connect". The concept of "initial speed" isn't defined, since it is an input (and will be specified in the requirements section, correct?), so I'm not sure how to encapsulate the idea that it is "tied" to the velocity of the projectile, since we get the notion of its velocity from its initial speed and launch angle, inputted separately.

Also, should the IMs use _r<sub>x</sub>_ instead of _d_, because _r<sub>x</sub>_ is the symbol for the concept of distance in the x-direction (defined in a GD), and _d_ is a variable name I assigned (and probably shouldn't have)? If so, _d<sub>target</sub>_ should be renamed to _r<sub>target</sub>_ for consistency,

samm82:
How are things looking now @smiths?

smiths:
@samm82, I'll need to think about your example from the start.  I think the parts you have are correct, but they aren't fitting together.  I'm planning on doing this exercise so that I can explain to myself and to you what the parts are and what the relationship is between them.  This exercise has highlighted that it isn't as obvious as I implicitly thought.  

Yes, you should use use r_x instead of d.  That makes the connection between the vector and its components more explicit.  The fact that we have r as the vector is unfortunate, since r makes sense in the Game Physics example, but less sense here.  Normally the position vector would be \bf{u}.

Drasil is supposed to allow us to separate the concept from the symbol used to represent it.  How difficult would it be to benefit from the definition of a position vector in Game Physics, but change the symbol to be \bf{u}.  If this is easy to do, then it is a nice little example for @szymczdm.

smiths:
@samm82, I added a sketch of an SRS for projectile motion to the Notes folder in your branch:

https://github.com/JacquesCarette/Drasil/commit/742681df04e4ab9c880bed14a44fbf304f0862a0

I wrote this sketch without re-looking at your SRS.  I wanted to come at it with "fresh" eyes.  I've looked at your SRS now and there is much in common.  My version brings in more information, especially derivations, and it (I believe) refines the knowledge from abstract theory to the concrete instance model in a rigorous way.  I made an effort to write the knowledge generically when possible.  I believe that I've used the DDs so that they do not reference anything else.

My sketch is not written exactly like an SRS, especially in the beginning sections where I was collecting my thoughts.  The division into TMs, GDs etc follows the SRS, but then I didn't use the standard fields from our template.  Hopefully you can take what I have written and rearrange it to better fit the template.

As a first step, please review what I have written and make sure that it makes sense.  Please ask if you have any questions.  @bmaclach, can you please review my sketch of an SRS too?  (We can create an issue for the discussion, if that is helpful.)  I think the projectile motion example is a good chance for us to work out the relationship between knowledge in our problems.  We started the discussion for SSP about what is a GD and what is a DD.  I think DDs are what is on the bottom.  The knowledge we take as axiomatic.  If it depends on other knowledge in the problem domain, then it will be a theory or a GD.  What is a DD in one problem, might be a GD or a TM in another problem.  Once we have all looked over this example, I'd like to discuss it again.  We should also include @JacquesCarette in the discussion, but that can wait until we've converged on an actual SRS for projectile.

samm82:
How should A4 and A5 be articulated? "The positive y-direction is up" and "The positive x-direction is to the right"?

smiths:
@samm82, I don't really have a problem with how A4 and A5 are written, but if you find it confusing, we could add in reference to horizontal and vertical?  Maybe you want to add that these are relative to the ground.  Parallel to the ground is the x coordinate direction, with the positive direction being to the right.  Perpendicular to the ground is the y-direction, with up (moving away from the ground) as positive.

I think you will need a terminology section in the SRS to define ground, launcher (or whatever you decided to call it) and target.

We also need an assumption that the target is also on the ground (y = 0).  I forgot to include that in my sketch.

samm82:
@smiths We decided to forego the definition of ground in this problem. The positive x-direction was defined as the direction from the launcher to the target (with an assumption that they are the same height), and the positive y-direction was defined as opposite to gravity. Is this still valid?

smiths:
@samm82, I remember the discussion about the ground.  When I started writing the sketch of the SRS, it seemed the easiest way to keep things simple.  In retrospect, the assumption that the launcher and the target are at the same height, keeps things generic, but also simple.  Let us go with what we said in the first place and not introduce the concept of the ground.  That means we can't say they are at the same height, since height is relative to the ground.  They are both at a "special" height, y = 0.  We have an assumption that the origin of the coordinate system is located at the launcher.  Let us add an assumption that the vertical coordinate for the position of the launcher is at 0.

I like the definition that the positive x-direction is in the direction from the launcher to the target.  This considers the case of the target either being to the right or the left.  I thought for a second that this might confuse the definition of the angle theta, but I don't think there will be any problems there.

For the positive y-direction, I would rather not invoke the gravity direction.  I feel like the coordinate system comes first, and then we add the concept of gravity.  I'd like to add the possibility (in the future) of a x-component of the gravity.  If we had some kind of weird game physics, we might even have gravity acting "up."  

We can define the y-coordinate axis as perpendicular to the x-coordinate axis.  For the direction, we can define it so that we have a right handed coordinate system with the z-axis coming out of the page.  That is, the cross product of the unit basis vector for the x-axis with the y-axis should point out of the page.  (If you curl the fingers of your right and from the positive x axis to the positive y axis, your thumb will point up.)

samm82:
Is it enough to define a right-handed coodinate system - that would define the y-axis as both perpendicular to the x-axis and the direction of the position y-axis? Or would "the y-coordinate axis is perpendicular to the x-coordinate axis" need to be a separate assumption?

smiths:
I would like to keep the idea of perpendicular and right-handed.  It is technically possible to have a non-orthogonal basis and still have the notion of right-handed ness.

samm82:
I merged the new assumptions with the old ones - let me know what you think @smiths.

bmaclach:
@smiths I like how the SRS sketch explicitly refines from the TMs to IM. A couple mistakes I noticed:
- The DDs for v_x and v_y are backwards -- v_x should use cos and v_y should use sin
- Towards the end of the IM derivation, it says "To find the vertical distance travelled," where it should say "To find the horizontal distance travelled"

And some questions and comments I have:
- What's the reason for having a single IM with 2 equations/outputs? Why not have an IM for calculating time and another for calculating distance? 
- Defining DDs as "axiomatic" definitions that don't reference anything else is interesting... I wonder how that differs from TMs? And if a DD in one problem might be a GD or TM in another, how would that be decided? In the "Thoughts" section of your SRS sketch, you say "The data definitions don't reference anything. If the properties of vectors were used, then the DDs would become GDs.". I don't really understand this... are we not using the properties of vectors in this example?
- We've discussed briefly before how the offset calculation should be a requirement and not an instance model, and how the code generator will thus need to look at requirements and not just the models, but I want to understand why the offset calculation is a requirement in this case and not an IM? How do we decide which outputs warrant an IM and which are only a requirement?

bmaclach:
@smiths @samm82 Also, it just occurred to me that, given that we don't care about "ground", the assumption about neglecting the curvature of the earth seems unnecessary. We have other assumptions related to the launcher and target being at the same height.

samm82:
@bmaclach You're totally right!

bmaclach:
@smiths With regards to your assumptions, @smiths made a note in his "Requirements" section that the relative tolerance constant needs to be defined somewhere and that in the past that has been done with an assumption. So you may want to add an assumption about that (it sounds like a good idea to me, at least).

samm82:
@bmaclach @smiths Wouldn't it make more sense to be in a table of constants?

bmaclach:
@samm82 That also sounds like a good idea 😄 

Do we have a Table of Constants in any of our current examples?

samm82:
@bmaclach Yeah, the Values of Auxiliary Constants (GlassBR and many others have it).

bmaclach:
@samm82 Ah, right, thanks. I agree that it makes more sense in that table than in an assumption.

smiths:
I should have created a separate issue for the review of my SRS sketch.  The conversation is likely get a little confused now, especially since @samm82 and @bmaclach work so fast!

Thank you @bmaclach for your review at https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-498808556.  You are correct about my mistakes.  My responses to your questions follows:

- The IM that combines t' and d' should be split into 2 separate IMs.  I put them together because my first thought was that they are calculated together, but there isn't actually any coupling of the equations.  t' can be its own IM and then the d' IM can reference the other one.
- With respect to the difference between DD and TMs, I'm still thinking about it.  The TMs are more abstract than the DDs.  The TMs are refined into lower level GDs, while the DDs are simply used by the TMs and the GDs.  We have been thinking about the relationship between knowledge as simply "may reference."  We are likely going to have to refine this notion.  For my specific comment about the DDs using vectors, I meant that we find the x and y components of the vector from the vector's magnitude and initial angle.  When we are doing this we are using the fact that vectors can be decomposed using a special basis and the properties of vector addition.  In a more detailed version of Projectile, the data definitions would include this information and the component decomposition would then be promoted to a "higher" level.
- The offset calculation as a requirement feels like the cleanest way to organize the document to me, but I don't know if I have a principle that puts it there.  The IM to calculate the final position and time are reasonably reusable in other projectile, non-game, situations.  The offset calculation is starting to be game specific.  It is also much simpler than the other calculations.  I remember our conversation now and I now recall that there is no way to turn the requirements into code.  @samm82, let us take my offset calculation requirements and add an associated IM that actually does the calculation.  This captures the information and allows for code generation.  Down the road the difference between requirements and IMs is likely a documentation decision anyway.  The main thing is to capture the equations.

samm82:
I updated the GDs and have two questions for @smiths and/or @bmaclach:

1) Should there be any mention of vectors in the first two GDs? There are in my implementation because I just used the symbols that made sense to me, but there is an unholy mix of vectors and scalars that I don't think should be there. Also, since no values are underlined in @smiths's notes, I'm assuming they should all be scalars, but I'm just confirming.
2) There are some "empty" integrations (eg. ∫dx) in the derivations. Right now, there isn't a way to implement that, so I wrote them as ∫1dx. Is this acceptable, or should I make an issue for allowing "empty" integrations?

bmaclach:
1. I think the GDs should just use scalars because at the beginning of the derivation you refine the problem to one dimension: "The motion in TM: acceleration is now one-dimensional". This line is only there for the 1st GD but I think a similar line (along with a similar version of the rest of that paragraph) should also appear in the 2nd DD, justifying the use of scalars for both.
2. @oluowoj has run into the same issue and has said she plans on creating an issue for that (see #1447).

samm82:
Right now, position is defined as a vector (and used in GamePhysics as such). Does this make sense? Shouldn't it be a scalar, since a position doesn't really have a direction? @smiths @bmaclach

bmaclach:
@samm82 Since the problem is 2D, position is a vector (it has x and y components). The "direction" of position is the direction from the origin to the position. I still think we should refine it to a scalar in GD2, though.

smiths:
Position is a vector.  It does have a direction with respect to the coordinate system origin that is selected.  The direction is from the origin to the position at the end of the "position" vector.  By modelling position as a vector we can do vector arithmetic with positions.

smiths:
@bmaclach is correct, position becomes a scalar in GD2.  It just occurred to me that @samm82 might not know how to read my hand-written sketch of an SRS with respect to vectors.  If a symbol is underlined it is a vector quantity.  If it does not have an underline, it is a scalar.  (This is the typographic convention for hand-written notes, since you can't really show bold face with hand writing.)

smiths:
@samm82, in response to https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-499177435:

1. yes, vectors are underlined in my notes, scalars are not.  You should watch the dimension of your equations.  In previous versions of the SRS you had a vector on one side of the equation equal to a scalar on the other.  You can't mix vectors and scalars in this way.
2. You shouldn't need any "empty" integrations in your derivations.  Unless I'm really missing something, this doesn't come up in the derivations that I sent you.

code/drasil-example/Drasil/Projectile/Assumptions.hs
```diff
@@ -78,10 +77,6 @@ pointMassDesc :: Sentence
 pointMassDesc = (S "size" `sAnd` S "shape") `ofThe'` phrase projectile `sAre`
                 S "negligible" `sC` S "so that it can be modelled as a point" +:+. phrase mass
 
-neglectCurvDesc :: Sentence
-neglectCurvDesc = S "The" +:+ phrase distance `sIs` S "small enough that" +:+.
-                  (S "curvature" `ofThe` S "Earth can be neglected")
-
```
smiths: No, we shouldn't remove this assumption.  For one reason because @JacquesCarette likes this assumption.  :smile:  For another, better reason, removing this assumption is removing information.  It is true that we have the assumption that the height of the launcher and the target are the same.  We also have the assumption that we can use a Cartesian coordinate system.  These assumptions are possible, in part, because we have assumed that we can ignore the curvature of the earth.  These are the dependencies between assumptions that have been mentioned elsewhere.  Instead of removing assumptions, we should keep them all and explain the relationships between them.  In a different problem context some assumptions will drop out and others will stay.  Since we don't know what family members we might want in the future, we should keep all assumptions.

smiths:
From the previous comment https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-498815255, I agree that we should make use of the table of constants.  I had hoped that we had something like that, but I wasn't quite sure.  :smile:  Of course a value like the allowed tolerance could easily become something we set at run-time, instead of compile time.  If we set it at run-time, then it is no longer a constant, but an input.  We can worry about that when we do another pass to think about the family of projectile programs.

samm82:
> 2. You shouldn't need any "empty" integrations in your derivations.  Unless I'm really missing something, this doesn't come up in the derivations that I sent you.

This is what I meant by "empty" integrations:

Empty Integrations

And I'll make a `scalarPos` concept/quantity - let me know if I do it right @smiths @bmaclach

samm82:
Also, I guess now is as good a time as any to clarify the outputs of this program. My interpretation was that the program would output if the target was hit or not, and if it wasn't output if the projectile went long or short, and by how much. However, I noticed that time is an output in @smiths's sketch. What would be the desired outputs, and followup, the desired output format (at least for now until we get this example fully working and start thinking about variabilities). @smiths @bmaclach @JacquesCarette

smiths:
The instance model outputs do not all need to be required outputs.  You can still write requirements as you originally planned.  Write all of the IMs that we have discussed, but then in the requirements, select the ones that you need for your output.  (These IMs will be using the other ones, so there won't be any parts that aren't relevant.)

smiths:
You are correct @samm82.  As you showed in https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-499210987, there are integrations where the integrand is implicitly 1.  @oluowoj has met the same issue.  For now, I'm fine if you list explicitly list the integrand as 1.  We'll just need to make sure that we keep track of this issue for the future.

samm82:
Once again, I think I have completed everything. @smiths @JacquesCarette @bmaclach @oluowoj Feel free to review again. (Hopefully for the last time!)

code/drasil-example/Drasil/Projectile/Assumptions.hs
samm82: Added back in 2807130

bmaclach:
Looking good, @samm82! My comments are mostly minor:

- There are many unused assumptions, which we've talked about how some assumptions will only be used in other assumptions. Have you tried referencing an assumption in an assumption yet? I agree with what I think you said earlier, that it will probably work already.
- When TM2 is used in the derivation of GD2, the equation is slightly different. In TM2 it uses displacement but in GD2 it uses position. These should be consistent. I think position is the correct one to use here.
- In the derivation of IM2, it says "To find the vertical distance travelled" which I think should be "To find the horizontal distance travelled"
- There is inconsistency with regards to the output of IM2. The description of IM2 calls it "landing distance", the description of the symbol d' calls it "launch distance" and the goal statement calls it "landing position". We should choose one and make sure the equations align with that choice. I'm thinking position might be the one to go with, so that we don't need to introduce the concept of distance. Since the initial position is assumed 0, numerically the result will be the same whether we call it position or distance.

samm82:
I'm not sure where the following assumptions should be referenced: yAxisPerpend, rightHandAxes, posXDirection (should this be referenced in the IM for distance?), constMass, neglectCurv, and timeStartZero.

samm82:
@bmaclach Changing "launch distance" to "landing position" leads to this in IM3: "_isShort_ is the variable that is assigned true when the target distance is greater than the landing position", and it doesn't make sense to compare distance and position like that. Should "target distance" be changed to "target position" as well?

bmaclach:
@samm82 Yep!

samm82:
I'm also going to change the _d_ symbols to _p_ for consistency.

samm82:
Now I have a similar problem in _isHit_:
image

Do you think this is OK, since _p<sub>target</sub>_ is implicitly the distance from the origin to the target? @bmaclach

bmaclach:
@samm82 I think that's okay... the units work out and d_offset is equivalent to `p_target - p'`, and if you were to make that substitution then the equation looks just like any typical relative error calculation. We may want to define somewhere (maybe a DD) that `d = p_2 - p_1` or more specifically `d_offset = p_target - p'` (or both? Probably @smiths has an opinion).

samm82:
I'm guessing that `d = |p_2 - p_1|` would work as a DD and `d_offset = |p_target - p|'` is just the IM we have already (and could then be derived from the DD), so it could work to have both if @smiths agrees.

bmaclach:
@samm82 Good point, I'd forgotten we had that IM already.

bmaclach:
@samm82 My thoughts on the unused assumptions:
- yAxisPerpend seems very related to the use of a Cartesian coordinate system. Perhaps this assumption (yAxisPerpend) can be used in that one?
- rightHandAxes - I don't know either... this is another assumption that seems unnecessary to me (but last time I said that I was wrong, so let's wait for @smiths before deleting anything!)
- posXDirection - It actually seems to me like this assumption is our justification for the constraint that `p_target` must be >0. There's no way to provide rationale for a constraint in Drasil yet, so maybe you can reference this assumption in the "Notes" field for the IMs that use `p_target`
- constMass - again I'm unsure. Maybe this can be referenced in the pointMass assumption?
- neglectCurv - @smiths has said this assumption is part of the justification for the cartSyst and the targetXAxis assumptions, so it can be used in those.
- timeStartZero - In the derivations for GD1 and GD2 when you have a definite integral from 0 to `t`, this assumption is where the 0 comes from.

samm82:
@smiths I've added more references to assumptions, but @bmaclach and I are unsure about where yAxisPerpend, rightHandAxes, and constMass fit in.

smiths:
Thank you @samm82 and @bmaclach for your hard work on this issue.  I was trying to trace through the above comments, but it is too difficult to follow everything.  Moreover, it is difficult to reply to individual points.  I have read all of the comments and discussion, so I know that this document is now very close to done.  I am going to review it by reading the generated documentation and providing my feedback.  I'll try to get to that tomorrow (actually today) (June 12).

In the mean-time, can you please add requirements?  I know already that that will be part of my feedback.  Although requirements do not have a central role in our code generation, at least not at the moment, they are arguably the most important part of the SRS.  They are after all where the R comes from.  :smile:.

samm82:
@smiths I'm assuming you want the requirements added in this same PR?

smiths:
Same PR is fine.

samm82:
Requirements added

samm82:
@smiths Can you double check that I updated the assumptions correctly?

smiths:
@samm82, in the interest of managing my time, I'll check the assumptions when I check the revisions of all the parts.  I find it easier to look over the complete document, and to rethink it, all at once.  It is also a more efficient  use of time.  :smile:

samm82:
To summarize (and so I don't forget), I think the rest of what needs to be done is:
- Fill in tables (I had issues with terms not being in the `ConceptMap` when I tried, but I'll keep at it.)
- Format the derivations better (which I think should be a larger issue and not on this branch)
- Add parentheses around integrals (@smiths I looked it up and I was wrong; there should be parentheses around the different terms of the integrand. Not sure how to generate them in Drasil however.)
- Add the sources (@smiths Is Wikipedia OK or should I wait for the resource that you have?)
- Shorten up names (ie. IM1, IM2; should also be a larger issue not tackled here.)
- Extract typing information for vectors (#1558) | _related_ Fix units of vectors
- Improve labels of GDs (a smaller fix I haven't gotten to yet)
- Add inputs and outputs for GDs and DDs? (Bigger issue)
- Reference figure in DDs
- Revisit IM intro (Bigger issue?)

smiths:
@samm82, I have some comments to add to https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-502262119:

> To summarize (and so I don't forget), I think the rest of what needs to be done is:
> 
> * Fill in tables (I had issues with terms not being in the `ConceptMap` when I tried, but I'll keep at it.)
> * Format the derivations better (which I think should be a larger issue and not on this branch)
> * Add parentheses around integrals (@smiths I looked it up and I was wrong; there should be parentheses around the different terms of the integrand. Not sure how to generate them in Drasil however.)

I spoke to @JacquesCarette about this and he confirmed that mathematically the parentheses are NOT necessary.  Maybe down the road we could provide an option for stylistically adding the brackets, for those that prefer them, but for now, I think we can stick with what we have.

> * Add the sources (@smiths Is Wikipedia OK or should I wait for the resource that you have?)

I'll send you a resource @samm82 that we can cite.  I think you should also add Wikipedia link (or links).  Wikipedia isn't right as the only source, but as a convenient support for a more traditional format, it has its place.

> * Shorten up names (ie. IM1, IM2; should also be a larger issue not tackled here.)

Agreed.

> * Extract typing information for vectors (#1558) | _related_ Fix units of vectors
> * Improve labels of GDs (a smaller fix I haven't gotten to yet)
> * Add inputs and outputs for GDs and DDs? (Bigger issue)

I spoke with @JacquesCarette about this and we decided that we don't want to go down this road.  In a different context the inputs and outputs for a GD could change, even for the same equation.  Right now, are inputs and outputs are implicit.  The output is the symbol on the left hand side of the equation, the inputs are everything else.  The same formula could be rearranged to change the outputs, and we wouldn't want the new form to be new knowledge.  We aren't going down that road at the moment, so lets leave the GDs and DDs as they are with respect to inputs and outputs.

> * Reference figure in DDs
> * Revisit IM intro (Bigger issue?)

I'm not sure what this is, but you could start by creating a question issue describing the proposed changes.  After discussion, we could turn it into a full issue or issues, as appropriate.

smiths:
@samm82, I've finished reviewing your SRS for Projectile.  Rather than scan it and upload a large file, can you come by my office today (June 17) to pick up the hard-copy.  We can also chat for a minute, especially about the isHit and isShort IMs.

samm82:
@smiths How would you feel about merging "launch velocity" and "initial velocity" into the same idea, since they use the same symbol (v<sup>i</sup>) and capture the same (if not, a similar) idea. On a related note, should we replace "landing position" (p') with "final position" (p<sup>f</sup>) for consistency? Or should these concepts be kept separate, since we'd probably have to define what it means to be "final" (the first time it hits the ground/reaches the same elevation as the target).

samm82:
@smiths @JacquesCarette Should the definitions for the coordinate systems (cartesian and right-hand) be moved to `Data.Concepts.Math` from `Concepts.Physics`? I'm going to add `xDir`, `xAxis`, and the "y-" variants so that "y-direction" displays as "*y*-direction", as mentioned by @smiths, and I'm not sure where they should go.

samm82:
Other than my previous question about "launch velocity" and "initial velocity", and "landing position" and "final position", I think this PR is good for another pass of review/for merge. @bmaclach @oluowoj @smiths @JacquesCarette. The formatting issues were moved to #1575, and the coordinate question was moved to #1577.

bmaclach:
@samm82 I reviewed the SRS. Some minor comments:

- x and y components of initial position have units of m/s instead of m
- In terminology and definitions, I would change "Target: Where the projectile should be launched." to "Target: Where the projectile should be launched to."
- There was discussion at the last meeting about the purpose of Terminology and Definitions. I have always thought of it as a place to define any term that might not be known, and with that reasoning I would include "rectilinear" in that section. Maybe @smiths disagrees. But I had to look it up 😆 
- I think you should reference A:timeStartZero in the detailed derivations of GD3 and GD4, when you say that initial time is t=0.
- In the detailed derivation for GD4, where the components of the position vector are shown, why aren't p<sub>x</sub> and p<sub>y</sub> italicized? (Looks like that's only in HTML. Might be a bug unrelated to what you did)

Good work!

samm82:
@bmaclach I noticed the non-italicized vector; it also happens with the velocity vector in GD3. I think it happens to the first vector in each detailed derivation, and I'm not sure why. I noticed it earlier and forgot to document it, so thanks for bringing this up 😁

I'll make the other changes (I agree with adding rectilinear to the Terms and Defs).

smiths:
I agree with the comments from @bmaclach given at

https://github.com/JacquesCarette/Drasil/pull/1428#issuecomment-503333521

Specifically, rectilinear should be added as part of the terminology.

smiths:
@oluowoj, it might get lost in all the other e-mails, so this is a reminder for you to review the SRS for Projectile.  Any questions or comments from you will be valuable for improving the documentation, and your familiarity with the example will pay off going forward.

samm82:
I've changed everything but the first vector not being italicized, since it seems like a more general issue than Projectile. Should I make an issue for it?

smiths:
Yes, please create an issue for the absence of italics.  Include a screenshot in the issue.  Does it happen in other examples too?

samm82:
There is only one instance of a vector in the examples (SSP, in a GD) and it displays properly. I wonder if it's because of a helper function I use - I'll try building the derivation more "manually" and see if that fixes it.

samm82:
I tried making the derivations more "manually" and the bug persisted. I made an issue for it at #1585.

samm82:
@oluowoj @bmaclach @smiths I've added the DD for the magnitude of velocity, so this is ready for the (final?) set of reviews!

bmaclach:
@samm82 I reviewed it once more, just a couple comments:
- Right-handed coordinate system is in Terminology and Definitions but isn't used elsewhere in the document.
- Should v<sub>launch</sub> be used in the equations for the IMs instead of v<sup>i</sup>?

samm82:
@bmaclach Changes made - thanks!

code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -86,9 +86,9 @@ \subsection{Table of Symbols}
 \\
 ${{F_{S}}^{min}}$ & Minimum Factor of Safety: The minimum factor of safety associated with the critical slip surface & --
 \\
-${F_{x}}$ & X-Component of the Force:  & N
+${F_{x}}$ & $x$-component Of the Force:  & N
 \\
-${F_{y}}$ & Y-Component of the Force:  & N
+${F_{y}}$ & $y$-component Of the Force:  & N
```
samm82: The capitalization of "Of" in these rows is because the `NounPhrase` has to be built from a `Sentence` to have the *x* and *y*. I think in the future the SSP ToS should be revisited, since it is the only example that has the name of the quantity followed by its definition, so this could be left for now? @smiths

samm82: EDIT: There is a lot of in-depth instances where `xDir`, etc. should be used in SSP, so I won't do that in this PR.

smiths: When you ask whether this should be left for now, what do you mean?  The capitalization of "Of"?  It does look odd to have a capital like that.  We should create an issue related to this so that we don't forget it is something that we want to eventually fix.

samm82: Yes, that is what I meant, but at least for now (in the Projectile branch), I switched SSP back from using `yComp`, etc. to avoid this for now. I'll make an issue for swapping SSP over to using these concepts.

smiths: Sounds good.  Thank you @samm82.

samm82: Issue made at #1626

samm82:
Barring any unforeseen issues, Projectile is ready to be launched! (Pun very much intended) @smiths @JacquesCarette

samm82:
@smiths Other than the Hibbeler reference - I totally forgot to ask in the meeting! Would I be able to grab the book today?

smiths:
I just sent you an e-mail about the book.  :smile:  Please come by my office to pick it up.

smiths:
Since @JacquesCarette will not be able to review this for at least a week, and possibly longer, I've merged it into master.  We can create new branches as we continue to develop certain aspects of Projectile, such as the code generation.

PR #1435, opened 2019-05-28
Finally Tagless GOOL

bmaclach:
This PR adds the new, finally tagless version of GOOL to Drasil. Renderers are currently written for Java, Python, C#, and C++, which are the four languages Drasil currently generates. The code and makefiles generated for GlassBR work for all 4 languages (except for one error which already exists in Drasil and which @vajihehm has fixed in her open PR). 

@JacquesCarette I have a "Brooks" folder that has examples written in GOOL that I used to test new renderers when I wrote them. Do you want me to remove that folder before merging this in? Also, at what point should we remove the old GOOL AST and renderers?


JacquesCarette:
one more master merge needed.

bmaclach:
@JacquesCarette They can be useful but it's up to you if we want them - see my comment about the "Brooks" directory at the top of this PR.

JacquesCarette:
And, of course, because of the wait, there are now 3 conflicts.

Brooks/Example/FileTests.hs
```diff
@@ -0,0 +1,38 @@
+module Example.FileTests (fileTests) where
```
JacquesCarette: I think this would fit better under `code/CodeGen/Test/FileTests`. And the same with most of `Brooks/Example` (I'll assume so, and comment explicitly on exceptions).

Brooks/Example/FileTests.hs
```diff
@@ -0,0 +1,38 @@
...
+fileTests :: (PackageSym repr) => repr (Package repr)
+fileTests = packMods "FileTests" [fileDoc (buildModule "FileTests" [] [] [fileTestMethod] [])]
+
+fileTestMethod :: (RenderSym repr) => repr (Method repr)
+fileTestMethod = mainMethod "FileTests" (body [writeStory, block [readStory], 
+  goodBye])
+
+writeStory :: (RenderSym repr) => repr (Block repr)
+writeStory = block [
+  (varDecDef "e" int (litInt 5)),
```
JacquesCarette: So `hlint` doesn't complain about these extra parens?

In any case, I would want to try to make this whole block "look like" code as much as possible.  For example, by inserting a blank line between the declarations and the executable code.

Brooks/Example/HelloWorld.hs
```diff
@@ -0,0 +1,151 @@
...
+    printLn (float) (csc (litFloat 1.0)),
+    printLn (float) (sec (litFloat 1.0)),
+    printLn (int) (var "a"),
+    printLn (int) (inlineIf (litTrue) (litInt 5) (litInt 0)),
+    printLn (float) (cot (litFloat 1.0))]])
+
+helloElseBody :: (RenderSym repr) => repr (Body repr)
+helloElseBody = bodyStatements [
+  printStrLn "These will print but (by intention) not compile"]
+  -- printLn (int) (notNull (var "a")),
```
JacquesCarette: I don't like having this commented out. It should either be proper code or completely deleted.

Brooks/GOOLFinallyTagless.cabal
```diff
@@ -0,0 +1,37 @@
+Name:		GOOLFinallyTagless
```
JacquesCarette: Do we want to keep this as a separate package? [It needs a better name!]. It might indeed be a good idea to do so.

Maybe just GOOL.cabal?

Brooks/GOOLFinallyTagless.cabal
```diff
@@ -0,0 +1,37 @@
...
+build-type:     Simple
+Synopsis:	A reimplementation of GOOL using finally-tagless style
+
+library
+  hs-source-dirs: ./
+  exposed-modules:
+      New
+    , Helpers
+    , NewLanguageRenderer
+    , LanguageRenderer.NewJavaRenderer
```
JacquesCarette: All the `New`s should be removed.

Brooks/GOOLFinallyTagless.cabal
```diff
@@ -0,0 +1,37 @@
...
+executable helloworld
+  hs-source-dirs: Example
+  main-is: Main.hs
+  other-modules:
+  Build-Depends:
+    base >= 4.7,
+    containers >= 0.5.5.1,
+    pretty >= 1.1.1.1,
+    directory >= 1.2.6.2,
+    GOOLFinallyTagless >= 0.0.1
```
JacquesCarette: Should it self-depend?

Brooks/Helpers.hs
```diff
@@ -0,0 +1,159 @@
+{-# LANGUAGE TupleSections #-}
+
+module Helpers (
```
JacquesCarette: Why are these so far outside the hierarchy?

Brooks/Helpers.hs
```diff
@@ -0,0 +1,159 @@
...
+tripFst :: (a, b, c) -> a
+tripFst (c, _, _) = c
+
+tripSnd :: (a, b, c) -> b
+tripSnd (_, n, _) = n
+
+tripThird :: (a, b, c) -> c
+tripThird (_, _, b) = b
+
+liftA4 :: Applicative f => (a -> b -> c -> d -> e) -> f a -> f b -> f c -> f d -> f e
```
JacquesCarette: making sure we actually use all of these would be worthwhile too.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
+{-# LANGUAGE TypeFamilies #-}
```
JacquesCarette: Is the option actually needed?

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+
+module LanguageRenderer.NewCSharpRenderer (
```
JacquesCarette: Is this the right place in the hierarchy for this?

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+
+import New (Label,
+    PackageSym(..), RenderSym(..), KeywordSym(..), PermanenceSym(..),
+    BodySym(..), BlockSym(..), ControlBlockSym(..), StateTypeSym(..),
+    UnaryOpSym(..), BinaryOpSym(..), ValueSym(..), NumericExpression(..), 
+    BooleanExpression(..), ValueExpression(..), Selector(..), FunctionSym(..), 
+    SelectorFunction(..), StatementSym(..), ControlStatementSym(..), ScopeSym(..),
+    MethodTypeSym(..), ParameterSym(..), MethodSym(..), StateVarSym(..), 
+    ClassSym(..), ModuleSym(..))
+import NewLanguageRenderer (Terminator(..), fileDoc',
```
JacquesCarette: All those trailing `DocD` are for weird historical reasons, and should probably all be removed.

JacquesCarette: (I mean from the name)

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+import Helpers (oneTab, tripFst, tripSnd, tripThird, liftA4, liftA5, liftA6, 
+  liftA7, liftList, lift1List, lift3Pair, lift4Pair, liftPairFst, liftTripFst)
+
+import Prelude hiding (break,print,(<>),sin,cos,tan,floor)
+import qualified Data.Map as Map (fromList,lookup)
+import Control.Applicative (Applicative, liftA2, liftA3)
+import Text.PrettyPrint.HughesPJ (Doc, text, (<>), (<+>), parens, comma, empty,
+  equals, semi, vcat, lbrace, rbrace, render, colon, render)
+
+newtype CSharpCode a = CSC {unCSC :: a}
```
JacquesCarette: I wonder if you could `derive` `Functor` and friends?

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+    return = CSC
+    CSC x >>= f = f x
+
+instance PackageSym CSharpCode where
+    type Package CSharpCode = ([(Doc, Label, Bool)], Label)
+    packMods n ms = liftPairFst (sequence ms, n)
+
+instance RenderSym CSharpCode where
+    type RenderFile CSharpCode = (Doc, Label, Bool)
+    fileDoc code = liftTripFst (liftA3 fileDoc' (top code) (fmap tripFst code) bottom, tripSnd $ unCSC code, tripThird $ unCSC code)
```
JacquesCarette: I'm thinking that using a record for `(Doc, Label, Bool)` would enable nicer names than `tripFst` et al, and thus make the code nicer.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+    intListType p = listType p int
+    floatListType p = listType p float
+    boolListType = listType dynamic bool
+    obj t = return $ typeDocD t
+    enumType t = return $ typeDocD t
+    iterator _ = error "Iterator-type variables do not exist in C#"
+
+instance ControlBlockSym CSharpCode where
+    runStrategy l strats rv av = 
+        case Map.lookup l (Map.fromList strats) of Nothing -> error $ "Strategy '" ++ l ++ "': RunStrategy called on non-existent strategy."
```
JacquesCarette: I don't like super-long lines, I would prefer to use a new line and less indent.  Also, a local function using `where` could be good here.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+            v_temp = var l_temp
+            l_i = "i_temp"
+            v_i = var l_i
+        in
+        (bodyStatements [
+            (listDec l_temp 0 t),
+            for (varDecDef l_i int (getB b)) (v_i ?< getE e) (getS s v_i)
+                (oneLiner $ valState $ v_temp $. (listAppend (vold $. (listAccess v_i)))),
+            (vnew &= v_temp)])
+        where getB Nothing = litInt 0
```
JacquesCarette: you could use `maybe (litInt 0) id b` instead of defining `getB`.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
```diff
@@ -0,0 +1,561 @@
...
+    iterVar = var
+    
+    inputFunc = return $ (text "Console.ReadLine()", Nothing)
+    argsList = return $ (text "args", Nothing)
+
+    valName (CSC (v, s)) = case s of Nothing -> error $ "Attempt to print unprintable Value (" ++ render v ++ ")"
+                                     Just valstr -> valstr
+
+instance NumericExpression CSharpCode where
+    (#~) v = liftPairFst (liftA2 unOpDocD negateOp v, Nothing)
```
JacquesCarette: looks like there is a definite opportunity to make soem helper functions to make the code below nicer.

Brooks/New.hs
```diff
@@ -0,0 +1,484 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module New (
```
JacquesCarette: Not a good name...

Brooks/NewLanguageRenderer.hs
```diff
@@ -0,0 +1,675 @@
...
+asinOpDocD = text "asin"
+
+asinOpDocD' :: Doc
+asinOpDocD' = text "math.asin"
+
+acosOpDocD :: Doc
+acosOpDocD = text "acos"
+
+acosOpDocD' :: Doc
+acosOpDocD' = text "math.acos"
```
JacquesCarette: This seems like a Java-ism ? So it seems out of place.

Brooks/filename.txt
```diff
@@ -0,0 +1,5 @@
+Red
```
JacquesCarette: ????

GOOLCodeGen/GOOLCode.cabal
```diff
@@ -1,5 +1,5 @@
 Name:		GOOLCode
-Version:	0.0.1
+Version:	0.0.2
```
JacquesCarette: So `GOOLCodeGen` was outside of the `code/` hierarchy to allow independent experimentation. It doesn't make sense to have 3 copies (i.e. the one under `New`, this one, and the one under `code/`).

GOOLCodeGen/GlassBR/Main.hs
```diff
@@ -2,7 +2,14 @@ module Main where
 
 import System.Directory
 
-import Language.Drasil.Code
+import Language.Drasil.Code.Imperative.New (Label)
```
JacquesCarette: BTW. the `New` names can stay for the first PR, then be removed in a second pass.

code/drasil-code/Language/Drasil/Code/Imperative/Build/Import.hs
```diff
@@ -3,63 +3,61 @@ module Language.Drasil.Code.Imperative.Build.Import (
...
-      (Runnable nm no ty) = runnable c
+      (Runnable nm no ty) = r
       buildTarget = makeS "build"
 
-renderBuildName :: Config -> Package -> NameOpts -> BuildName -> MakeString
-renderBuildName _ (Pack _ m) _ BMain = makeS $ getMainModule m
-renderBuildName _ (Pack l _) _ BPackName = makeS l
-renderBuildName c p o (BPack a) = renderBuildName c p o BPackName <> makeS (packSep o) <> renderBuildName c p o a
-renderBuildName c p o (BWithExt a e) = renderBuildName c p o a <> if includeExt o then renderExt c e else makeS ""
+renderBuildName :: [String] -> ([(Doc, Label, Bool)], Label) -> NameOpts -> BuildName -> MakeString
```
JacquesCarette: I'm uncomfortable with hard-coding `(Doc, Label, Bool)` here. It's likely to change. Should be abstracted.

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
```diff
@@ -1,18 +1,29 @@
-module Language.Drasil.Code.Imperative.Helpers (
+{-# LANGUAGE TupleSections #-}
+
+module Language.Drasil.Code.Imperative.Helpers (Pair(..),
```
JacquesCarette: So this looks like a complete copy of the `Helpers` above?

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
```diff
@@ -1,18 +1,29 @@
...
     blank,spc,oneTabbed,oneTab,vertical,verticalComma,verticalNewLine,
     angles,doubleQuoted,doubleQuotedText,capitalize,containsAll,
     makeLiteralNameValid,makeVarNameValid,makeClassNameValid,powerSet,
-    hmap,himap,vicat,vibcat,vmap,vimap,vibmap, reduceLibs
+    hmap,himap,hicat,vicat,vibcat,vmap,vimap,vibmap, reduceLibs, mapPairFst,
```
JacquesCarette: and a number of these pretty-printing helpers are also in `drasil-printers`. We should (eventually) pull them out.

code/stable/glassbr/src/cpp/Calculations.cpp
```diff
@@ -18,52 +15,55 @@ using std::vector;
 using std::ifstream;
 using std::ofstream;
 
+#include "InputParameters.hpp"
+#include "Interpolation.hpp"
+
 bool func_is_safePb(InputParameters &inParams, double P_b) {
-    return P_b < inParams.P_btol;
+    return (P_b < inParams.P_btol);
```
JacquesCarette: Do you know why there are extra `()` here? [Not a big deal, but it would be nice to not have them. Probably future work.]

code/stable/glassbr/src/cpp/Makefile
```diff
@@ -12,7 +12,7 @@ endif
 
 build: GlassBR$(TARGET_EXTENSION)
 
-GlassBR$(TARGET_EXTENSION): InputParameters.hpp DerivedValues.hpp InputConstraints.hpp OutputFormat.hpp Calculations.hpp ReadTable.hpp InputFormat.hpp Interpolation.hpp Control.cpp InputParameters.cpp DerivedValues.cpp InputConstraints.cpp OutputFormat.cpp Calculations.cpp ReadTable.cpp InputFormat.cpp Interpolation.cpp
+GlassBR$(TARGET_EXTENSION): Control.cpp InputParameters.cpp DerivedValues.cpp InputConstraints.cpp OutputFormat.cpp Calculations.cpp ReadTable.cpp InputFormat.cpp Interpolation.cpp InputParameters.hpp DerivedValues.hpp InputConstraints.hpp OutputFormat.hpp Calculations.hpp ReadTable.hpp InputFormat.hpp Interpolation.hpp
```
JacquesCarette: Any chance you could flip the order, so that the `.hpp` files come first?

code/stable/glassbr/src/csharp/Calculations.cs
```diff
@@ -3,60 +3,58 @@
 using System.Collections;
 using System.Collections.Generic;
 
-namespace GlassBR {
```
JacquesCarette: Any reason the `namespace` disappeared?

bmaclach:
@JacquesCarette I didn't do a good job of explaining the Brooks/ directory. As you noticed, a lot of the stuff in that directory (Basically all of "New" GOOL) has a copy in the actual Drasil hierarchy infrastructure. In other words, the Brooks/ directory is completely optional. We could remove it entirely and nothing would break. The reason it exists was because when I originally wrote the finally tagless GOOL for CAS 761, I needed to do it "outside" of Drasil, so I created this new directory. When it came time to plug the new GOOL into Drasil, I did that by copying everything from Brooks and then updating what needed to be updated to make it work in Drasil.

The only "unique" things in Brooks/ are the tests under Example/, which I maintained because they helped when writing new renderers. However, they were written ad-hoc and are rather rough, which you noticed (the large commented-out parts, the strange "filename.txt"). They also aren't automated. I really don't know if they are even worth keeping.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: It pained me when I resorted to using triples and having to write `tripFst` et al. This was never intended to be a long-term solution. We've talked before about making our instances `Reader`s so that we don't have to explicitly pass around these `Label`s and `Bool`s like we do here. However, it will be easy to replace this with a record for now, so I'll do that.

Brooks/New.hs
bmaclach: For the record, this file exists in Drasil already (in the code/ directory). It's equivalent to the current AST.hs, but it doesn't really define an AST since we switched to finally tagless. Would Symantics.hs be a good name?

code/stable/glassbr/src/csharp/Calculations.cs
bmaclach: I saw that the code worked without it and didn't see why it was necessary/useful, so I dropped it. If there's a reason for it, I can look into adding it back, though I think this is something that will be much easier to implement once we have `Reader`s for our instances, so that we can read the `Package` name whenever we need it.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: Yes, for type declarations within typeclasses.

Brooks/NewLanguageRenderer.hs
bmaclach: It's used in Python and Lua (though we haven't written the Lua renderer yet). Since more than one language uses these (in the long run), I think it's okay to keep them where they are?

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: If I just drop `DocD`, we will end up with a lot of function names that are the same as function names in the symantics typeclasses. For example, `packageDocD` would become `package` which is the same as the `PackageSym` function for creating a package. In these cases, can I just drop the `D` and have the name as `packageDoc`? But then for consistency I would want to keep `Doc` in the names for all of these functions. I think it makes sense to have `Doc` in these names, since they are all functions for creating `Doc`s.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: `Functor` can be derived if we use the DeriveFunctor language extension. When I attempted to derive `Applicative` and `Monad`, GHC gave an error and recommended I use the GeneralizedNewtypeDeriving extension. I tried using that extension but it did not help, so I think we'll need to keep the `Applicative` and `Monad` instances explicit.

Is the cost of using the DeriveFunctor extension worth it to derive `Functor`?

Brooks/Helpers.hs
bmaclach: Checked all of the helpers and removed unused ones in 3c6f8c9cba941ab386ef6ba8fcb6ea0d88d0cdc4.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: Added records and removed functions for working with triples in 77139c28eaa4cc4d2b155e47de2fcfe1c7682fbb and 4755f9b996c33bd70c0bd9b98f9236ff367a50d5.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: Style fixes in 7e619d656288976863ec517a37896b3c6427d26a and 54f17c8450fc6193959a07ceeadf329eb7cf5028. Simplified this function in c1e48fb5503d21482c789a131932270e748dee33.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: Done in bd294b8deb14bcb7196ea30f0fc4dfd819c4be78.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: Various helper functions implemented in 237a480698f8dc661f69409273d05e7399d56bd8.

code/drasil-code/Language/Drasil/Code/Imperative/Build/Import.hs
bmaclach: Replaced (Doc, Label, Bool) with a new record type in 77139c28eaa4cc4d2b155e47de2fcfe1c7682fbb.

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
bmaclach: #1477 created so we don't forget.

code/stable/glassbr/src/cpp/Makefile
bmaclach: Done in d0a32aa0b91419250d364e2ec5dcd577410854b5.

bmaclach:
@JacquesCarette I believe I've responded to all of your comments. If any don't have an explicit response, it's because they are related to my comment above (https://github.com/JacquesCarette/Drasil/pull/1435#issuecomment-498282302).

code/stable/glassbr/src/cpp/Calculations.cpp
JacquesCarette: Please file an issue. It's a fun feature to implement, but it is 'work'.

bmaclach: #1513 created.

Brooks/LanguageRenderer/NewCSharpRenderer.hs
bmaclach: @JacquesCarette here was our previous discussion of deriving Functor, etc.

JacquesCarette: I'm pretty sure we could use the (even newer) `DerivingVia` to do it.  Maybe insert that as a comment, and we can get to it later.

PR #1446, opened 2019-05-30
Input format fixes

### LINKED ISSUE #1179: NoPCM CodeGen; Bad Code

bmaclach:
Fixes #1179. The way we are currently set up, the function in the input format module needs to be called "get_input" because we've hardcoded a call to that function in our generate code for the Control module (ideally, we would be able to read the function name from the module and call it, but that would require some design, since there is currently no way to differentiate the InputFormat module from the other modules). Some of the examples had accidentally called the function "get_inputs" instead, so this changes those to "get_input". 

This also removes a hack where the "get_input" function was hardcoded as an export of "InputFormat". That export is read automatically from the list of modules we pass in, so the hack is not needed.

### DIFF
--- a/code/drasil-code/Language/Drasil/CodeSpec.hs
+++ b/code/drasil-code/Language/Drasil/CodeSpec.hs
@@ -238,8 +238,7 @@ modExportMap :: [Mod] -> [Input] -> [Const] -> ModExportMap
 modExportMap ms ins _ = Map.fromList $ concatMap mpair ms
   where mpair (Mod n fs) = map fname fs `zip` repeat n
                         ++ map codeName ins `zip` repeat "InputParameters"
-                        ++ [ ("get_input", "InputFormat"),
-                             ("derived_values", "DerivedValues"),
+                        ++ [ ("derived_values", "DerivedValues"),
                              ("input_constraints", "InputConstraints"),
                              ("write_output", "OutputFormat") ]  -- hardcoded for now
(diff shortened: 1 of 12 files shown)


JacquesCarette:
We're definitely going to want to fix this "get_input" hack. But this fix is definitely worthwhile in the meantime, merging.

PR #1447, opened 2019-05-30
Impulse derivation in datadefs.hs

oluowoj:
Added a new definition in datadefs.hs (Impulse - Vector)
Files updated: 
image

Could not produce the equation exactly how it should be but very close. 

I used "defint" to construct the impulse derivative equation, and it takes 4 expressions as input, however, the second part of the expressions needs only 3 inputs and I don't find any functions for an empty expression like we have "Empty S" ( for empty sentences).

I will make a ticket to have a function for empty expression.

Here is what I currently have, and also the upper and lower bounds of the integral are staggered, looks like that's the best we can have:
Drasil output:
image

Manual version:
image

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Physics.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Physics.hs
@@ -5,7 +5,7 @@ module Data.Drasil.Concepts.Physics
   , gravitationalAccel, gravitationalConst, position, distance
   , time, torque, weight, fbd, angular, linear, tension, compression, stress
   , strain , angDisp, angVelo, angAccel, linDisp, linVelo, linAccel, joint
-  , damping , cohesion, isotropy, twoD, threeD, physicCon, physicCon', kEnergy
+  , damping , cohesion, isotropy, twoD, threeD, physicCon, physicCon', kEnergy, chgInVelocity
   ) where
 --This is obviously a bad name, but for now it will do until we come
 --  up with a better one.
@@ -23,7 +23,7 @@ physicCon = [rigidBody, velocity, friction, elasticity, energy, mechEnergy, coll
   gravitationalAccel, gravitationalConst, position, distance,
   time, torque, weight, fbd, linear, angular, tension, compression, stress, 
   strain, angDisp, angVelo, angAccel, linDisp, linVelo, linAccel, 
-  joint, damping, pressure, cohesion, isotropy, kEnergy]
+  joint, damping, pressure, cohesion, isotropy, kEnergy, chgInVelocity]
 
 
 physicCon' :: [CI]
@@ -35,9 +35,7 @@ rigidBody, velocity, friction, elasticity, energy, mechEnergy, collision, space,
   gravitationalAccel, gravitationalConst, position, distance,
   time, torque, weight, fbd, linear, angular, tension, compression, stress, 
   strain, angDisp, angVelo, angAccel, linDisp, linVelo, linAccel, 
-  joint, damping, pressure,cohesion, isotropy, kEnergy :: ConceptChunk
-
-  -- joint, damping, pressure, cohesion, isotropy :: ConceptChunk
+  joint, damping, pressure,cohesion, isotropy, kEnergy, chgInVelocity :: ConceptChunk
 
 twoD, threeD :: CI
 
@@ -160,6 +158,9 @@ cohesion = dccWDS "cohesion" (cn "cohesion") (S "An attractive" +:+
 isotropy = dccWDS "isotropy" (cn "isotropy") (S "A condition where the" +:+
   phrase value `sOf` S "a" +:+ phrase property +:+ S "is independent of" +:+
   S "the direction in which it is measured.")
+
+chgInVelocity = dccWDS "velocity" (cn "velocity") (S "Change in" +:+ phrase velocity +:+ S "of a"
+   +:+ S "body.")
 
 twoD = commonIdeaWithDict "twoD" (pn "two-dimensional") "2D" [physics]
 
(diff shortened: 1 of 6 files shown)


JacquesCarette:
There seems to be an `I` missing from the "rearranging" step in the manual version? Also, the definite integrals look really weird in the above.  I'll try to check the latex to see what the problem might be.

code/drasil-data/Data/Drasil/Concepts/Physics.hs
```diff
@@ -161,6 +159,9 @@ isotropy = dccWDS "isotropy" (cn "isotropy") (S "A condition where the" +:+
   phrase value `sOf` S "a" +:+ phrase property +:+ S "is independent of" +:+
   S "the direction in which it is measured.")
 
+chgInVelocity = dccWDS "velocity" (cn "velocity") (S "Change in" +:+ phrase velocity +:+ S "of a"
+   +:+ S "body.")
```
JacquesCarette: it would be better to not split "of a body" this way.

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
```diff
@@ -245,6 +246,56 @@ chaslesThmDesc = foldlSent [S "The linear", (phrase QP.velocity),
...
+impulseVDerivSentence3 :: [Sentence]
+impulseVDerivSentence3 = [S "Integrating the right hand side "] 
+
+impulseVDerivEqn1 :: Expr
+impulseVDerivEqn1 = (sy QP.force) $= (sy QPP.mass) * (sy QP.acceleration) 
+                    $= (sy QPP.mass) * (deriv(sy QP.velocity) QP.time) 
+
+impulseVDerivEqn2 :: Expr
+impulseVDerivEqn2 = (defint (eqSymb timeT)(sy time_1)(sy time_2)(sy QP.force))
+                     $= sy QPP.mass * defint (eqSymb QP.velocity)(sy velo_1)(sy velo_2)1
```
JacquesCarette: Do you really want to integrate `1` here?

code/stable/gamephys/SRS/Chipmunk_SRS.tex
```diff
@@ -138,6 +140,10 @@ \subsection{Table of Symbols}
 \\
 ${\mathbf{v}^{BP}}$ & Velocity Of the Point of Collision P in Body B & $\frac{\text{m}}{\text{s}}$
 \\
+${\mathbf{v}_{1}}$ & Velocity Of the Firstbody & $\frac{\text{m}}{\text{s}}$
```
JacquesCarette: Something is weird "Firstbody" - there is a space missing.

code/stable/gamephys/SRS/Chipmunk_SRS.tex
```diff
@@ -885,13 +891,57 @@ \subsubsection{Data Definitions}
...
 RefBy & 
 \\ \bottomrule \end{tabular}
 \end{minipage}
+ Derivation of impulse (vector) - Newton's second law of motion states:
+\begin{displaymath}
+\mathbf{F}=m \mathbf{a}=m \frac{d\,Δ\mathbf{v}}{d\,t}
+\end{displaymath}
+Rearranging :
+\begin{displaymath}
+\int_{{t_{1}}}^{{t_{2}}}{\mathbf{F}}\,dt=m \left(\int_{{\mathbf{v}_{1}}}^{{\mathbf{v}_{2}}}{1}\,d\mathbf{v}\right)
```
JacquesCarette: the latex is fine - maybe it was the html you were showing?

JacquesCarette:
And a merge, as usual.

code/stable/gamephys/SRS/Chipmunk_SRS.tex
oluowoj: The latex is fine, its just the html that looks kind of weird

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
oluowoj: No, I dont want to integrate but I had to put '1'  in there in order to provide 4 arguements required by the defint function. Is there another way to do this? like some 'empty' expression?

JacquesCarette: What are you actually trying to write down? (Give me the 'math').

In a way, integrating the constant 1 is sometimes actually the right thing - it's just that we tend to abuse notation and elide it. But I'm not sure if this is such a case.

oluowoj: Here is what the actual formula looks like, everything without the '1'
image

But I had to put '1' there as a placeholder in order to be able to use 'defint' function. I don't think it makes any difference in the calculation though.

JacquesCarette: It does not, and is the correct change to make. Please file an issue to record that we might want to pretty-fy the output of integrating over `1`.

JacquesCarette:
What about `Firstbody`? And, as usual, another merge of master will be needed.

code/drasil-data/Data/Drasil/Concepts/Physics.hs
oluowoj: fixed

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
oluowoj: > It does not, and is the correct change to make. Please file an issue to record that we might want to pretty-fy the output of integrating over 1.

#1539

smiths:
The equation shown at the beginning of this issue for F = m a = m dv/dt shows a Delta before the v, but I just had a look at the code generated by this branch, and the Delta has (correctly) been removed.  I don't know at what point it was removed, but I'm glad to see it go.  :smile:

code/stable/gamephys/SRS/Chipmunk_SRS.tex
oluowoj: fixed

oluowoj:
Yes, it was resolved with this branch. Not sure how that came about, I saw that as well 2 or so days ago. The concept map error fix resolved it.

PR #1466, opened 2019-06-01
Travis Code Building

Mornix:
This PR changes Travis to use the existing `make code` target to build generated code. I've added an extension to the code building rule which checks for NoPCM and allows failed code to "pass" for that example as there are still a few outstanding issues there for that code gen. 

I've also setup a time-bomb in the sense that when NoPCM does compile, `make code` will fail with the message:
```
NoPCM surprisingly did not fail to compile!
Failing this build because you should check the Drasil makefile and alter the code as indicated by the comment for the currently invoked target.
```
and "fails" the build. This is to ensure that when NoPCM can and does compile that to merge those changes in the `Makefile` NoPCM suppression must be removed. 

Currently to indicate that there is non-standard behaviour in place for NoPCM is the message 
`NoPCM failed to compile as expected. Letting it slide...for now.` when `make code` runs and building NoPCM fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,14 +36,19 @@ addons:
   apt:
     packages:
       - libgmp-dev
+      - python3
+      - graphviz
+# TeX Building
       - fonts-lmodern
       - texlive-bibtex-extra
       - texlive-latex-extra
       - texlive-math-extra
       - texlive-xetex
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Excellent on the time bomb!

PR #1467, opened 2019-06-03
PR - Small PR to rename wrapAux and wrap_genAux

danscime:
Just a small naming change to keep naming more consistent throughout project.


smiths:
I don't have a problem with this change, but @JacquesCarette is better suited to determine if the name changes fit with the (stated and unstated) style conventions for Drasil.

Issue #1249, opened 2019-05-01
Correct HLint Suppressions


Mornix:
#1117 evaluated HLint for use in Drasil and #1147 made use of HLint in Drasil. While #1147 took care of most of the issues, there were five outstanding classes of lints:
- (104) Move brackets to avoid $
- (140) Eta reduce
- (2194) Redundant bracket
- (133) Redundant $
- (488) Use camelCase
where the number indicates the number of instances the lint appears.

While #1117 mentions obtaining HLint on macOS, a "better" way to acquire HLint is using `cabal install hlint` which will install the latest version (important because newer versions address false-positive "Parse Error" lints). 

While there are a lot of individual lints to fix overall, the speed in which this issue addressed isn't important; there is no rush. There are really three separate "phases" to address this issue:
- [x] Eta reduce (#1262)
- [x] Use camelCase (#1468)
- [x] Move brackets to avoid $ (#1339, #1343, and #1357)
- [x] Redundant bracket (#1385, #1396, #1438, #1443)
- [x] Redundant $ (#1384)

To remove these suppressions, remove the associated line in the `.hlint.yaml` file in the `code` folder.

samm82:
Just an FYI @danscime I'm fixing `off_the_shelf_solutions_2dlist` and friends in GamePhysics Body, so I'll just fix their camel casing issues, too.

danscime:
Once #1468 and #1508 are both successfully merged, the `.hlint.yaml` file can be updated and this issue can be closed.

JacquesCarette:
Both need one more merge, and they're ready to go.

PR #1468, opened 2019-06-03
PR - Hlint camelCase fixes

danscime:
Removing underscores in favour of camelCase for naming as per #1249.


danscime:
Working on merging conflicts.

danscime:
Fixed merge. Builds on local machine.

JacquesCarette:
Sorry, merge of previous stuff caused a conflict.

danscime:
Fixed.

JacquesCarette:
and again...

danscime:
Fixed

JacquesCarette:
Tried to do a hand-merge with the web interface... let's see if I managed to do it properly!

JacquesCarette:
So the problem now is as follows (from the log file):
```
diff -r -X ../.gitignore -x '*.txt' stable/nopcm/src/cpp/InputParameters.cpp build/NoPCM/src/cpp/InputParameters.cpp
37c37
<     if (!(((rho_W_min < inParams.rho_W) && (inParams.rho_W <= rho_W_max)))) {
---
>     if (!(((ρ_W_min < inParams.rho_W) && (inParams.rho_W <= ρ_W_max)))) {
```
stable is indeed correct. And, as far as I can tell, the SWHS (which is where these are defined) does it right. So I have no idea at all what's causing the output to have unicode -- @bmaclach this is likely related to some of the changes you did (and that I merged in from master).

bmaclach:
@JacquesCarette Yes, it was related. We need to use `mkQuantDef'` to conserve the stage-dependent symbol. `mkQuantDef` uses only the `Equational` stage symbol. Thinking about this again, though, and in light of the trouble it's caused, I don't see why we would ever want to "overwrite" our stage dependent symbol like `mkQuantDef` does. I'm thinking we can probably just remove `mkQuantDef` and rename `mkQuantDef'` as `mkQuantDef`... I pushed a fix given what we have currently but if you agree with this proposed change let me know and I can do it (in a separate PR probably?)

JacquesCarette:
I agree - no idea what we were thinking with `mkQuantDef` !! And yes, separate PR please.

JacquesCarette:
Sigh. And the build is failing because of hlint parsing problems?!? Any idea? Maybe @Mornix knows?

bmaclach:
The errors seem related to my GOOL changes, which is weird because the build passed for that branch. And  I get no errors when running hlint locally, plus the code it's complaining about looks fine to me. So hopefully @Mornix knows!

Mornix:
Took a quick look. Looks like HLint pushed a new version ~4hours before this build was run. The way HLint is configured into the CI we use whatever the latest version is when running, One of the in-progress changes is switching Haskell parsing libraries from HSE to ghc-lib-parser. Looking into why exactly it's marking things as parse error's however I'm guessing it has something to do with `static` as that is present on every line marked.

Mornix:
Did a quick test, HLint is indeed failing on `static`. I'm assuming with the switch they're treating all extensions as "loaded" and thus [static pointers](http://downloads.haskell.org/~ghc/8.2.2/docs/html/users_guide/glasgow_exts.html#static-pointers) are in play and it's recognizing `static` (incorrectly) as part of the language extension.

JacquesCarette:
In this meantime, it might be a good idea to rename `static` to something else? @danscime can you take care of that?

PR #1495, opened 2019-06-04
Minor Fixes to SWHS

samm82:
Minor typo fixes in SWHS, as well as using more `Sentence` constructors.


JacquesCarette:
The issue with those combinators is that they're not really capturing semantic information, just English patterns. I used to think that was worth doing (it is indeed fun!), I'm less sure now.

samm82:
It sounds like they should maybe be used in the back end to capture "semantics" (like commas and "ands"/"ors" for lists) and not externally in the examples. But I'm assuming that this is not a high priority thing to worry about.

PR #1496, opened 2019-06-05
NoPCM Inputs

### LINKED ISSUE #1450: NoPCM Inputs
Some of the inputs to NoPCM aren't included in the inputs field of the `SystemInformation`. Specifically, `timeStep`, `rel_tol`, `abs_tol`, and `cons_tol` are inputs but are not included in the `SystemInformation` for NoPCM. Also, `tau` is mistakenly included as in input in DataDesc.hs instead of `timeStep`. This breaks the generated code.

bmaclach:
This PR closes #1450 by adding the missing inputs and correcting the input file format.


bmaclach:
I noticed that there were still some places in NoPCM where a different input list was being used, so I added a commit such that the same input list is used everywhere.

code/stable/nopcm/Website/NoPCM_SRS.html
```diff
@@ -1326,6 +1331,15 @@ <h3>Data Constraints</h3>
 <td>10<em>%</em></td>
 </tr>
 <tr>
+<td><em>t<sub>step</sub></em></td>
+<td>
+<em>0&thinsp;&lt;&thinsp;t<sub>step</sub>&thinsp;&lt;&thinsp;t<sub>final</sub></em>
+</td>
+<td>--</td>
+<td><em>100.0&sdot;10<sup>-3</sup></em> s</td>
```
smiths: This is odd.  Why not just write 0.1? or 10^-1?

bmaclach: I did some digging, and found that we based this off of the standards for engineering notation described here: https://www.khanacademy.org/science/electrical-engineering/introduction-to-ee/intro-to-ee/a/ee-numbers-in-electrical-engineering

It says that for engineering notation, the number to the left of the decimal should always be in the range from 1 to 999. However, it also says "The number format rules are not rigid. As long as the point you are trying to make is clear and unambiguous, you may make exceptions. The blink of an eye may be clearly written 0.350 seconds if you intend the reader to compare the value to one second."

Thus, if we wanted, we could simply print values between 1000 and 0.001 as they are, and save the engineering notation for values greater than 1000 or less than 0.001?

Regardless, you've also discovered a bug because the number here is supposed to be 0.01, not 0.1. There's a mistake in how we print numbers in the range [0.01, 0.1). This bug will be an easy fix, coming shortly.

bmaclach: Bug fix in c4c99d61e1d18193148dcb8990fdce97638dce44, and I'll wait for feedback before changing anything else with the printing.

JacquesCarette: I certainly prefer numbers between 1000 and 0.001 to be done 'as is'.

JacquesCarette:
Merge needed.

JacquesCarette:
Waiting for @smiths approval of the requested changes.

JacquesCarette:
But of course there's now a conflict to resolve first.

JacquesCarette:
Sigh, and I merged something else in, which caused another conflict.

PR #1507, opened 2019-06-05
NoPCM Code generation - remove duplicated function

### LINKED ISSUE #1455: Duplicated function in NoPCM generated code
In the Calculations module for NoPCM generated code, the function for calculating q_C is generated twice, which causes errors in Java, C#, and C++. I think this is happening because the definition for q_C is included in `SystemInformation` in both the `datadefs` and `definitions` fields. It should only be included in one of them.

bmaclach:
This tiny PR fixes #1455.


JacquesCarette:
Because of the GOOL merge, this now needs a master merge.

PR #1509, opened 2019-06-06
NoPCM and SWHS Greek symbol fix

### LINKED ISSUE #1454: Variable names with Greek characters in generated code do not work in Java, C++
In the generated code for NoPCM, the variable for the water density is called `ρ_W`. The use of a Greek character in a variable name causes errors in Java and C++.

Variable names for quantities with Greek letter symbols should spell out the greek letter to avoid these errors. For example, this variable should be called `rho_W`.

bmaclach:
This PR updates some quantities in NoPCM and SWHS with Greek symbols to have a different symbol in the implementation stage because Greek symbols can't be used in some languages. Fixes #1454. Also fixes a bug where sorting in the SRS was based on the Implementation-stage symbols instead of the Equational-stage symbols.


JacquesCarette:
And yet another master merge needed.

PR #1510, opened 2019-06-06
Created Misc in utils

samm82:
As per #1390, created Misc in `utils` from Utils.hs in `lang` and `data`. @JacquesCarette is there anything that should be moved out of Misc that you can see?


JacquesCarette:
[master merge needed again]
What I mean is, the display of lists changed, and I don't see why that's a good thing.

Issue #1502, opened 2019-06-05
All examples should have one source for inputs


bmaclach:
Some of the examples define their list of inputs in different places - usually one list of inputs to pass to the `SystemInformation` and another list to use in the functional requirement for accepting inputs. All examples should be updated such that the inputs list is only defined once, and is used everywhere it is needed.

- [x] GamePhys
- [x] GlassBR
- [x] NoPCM
- [x] SSP
- [x] SWHS

smiths:
Good issue!

bmaclach:
#1496, #1511, and #1540 close this issue.

PR #1511, opened 2019-06-06
SWHS Inputs

bmaclach:
This PR makes SWHS use a single input list everywhere, contributing to #1502.


JacquesCarette:
Woohoo, a PR that doesn't need a master merge!  In it goes...

PR #1525, opened 2019-06-10
Hold HLint at v2.1.22

Mornix:
Quick Travis fix for HLint. Hold HLint at version 2.1.22 until their transition from HSE to ghc-lib-parser is complete and they correctly enable/disable GHC extensions.


JacquesCarette:
Can you also please file a bug at the HLint repo, so that we can upgrade in the future?

Issue #1512, opened 2019-06-07
Change HTML indent to 2 spaces


danscime:
I believe that at one point, we said that indent should use 2 spaces for stable just to display everything a bit more nicely. But the end goal is that the user can choose their amount of indent.

https://github.com/JacquesCarette/Drasil/blob/a75925ca802f861e51edd5da480372d74b721f8f/code/drasil-printers/Language/Drasil/HTML/Helpers.hs#L150-L151

JacquesCarette:
Correct.

danscime:
This indent size is now implemented.

PR #1530, opened 2019-06-10
PR - Changing HTML indent size from 0 to 2 spaces

danscime:
Addresses #1512.

My process was changing the number used to create spaces from 0 to 2 in my first commit.
Then I changed the makefile to ignore whitespaces, so the only changes to be made were to fix line breaks.

Once make ran without error, I removed my previous addition to the makefile.
Now, the only diffs would be due to whitespace. So I replaced each file in stable with the build files (after doing brief scans for correctness.)

It would have been infeasible to do all of the indents manually, so I think this is probably the best way I could have done it.


JacquesCarette:
Don't know why Travis doesn't seem to be terminating. In any case, merge needs to happen.

Mornix:
Travis finished (successfully) but it appears to have not relayed that information to GitHub.

JacquesCarette:
and of course, since I merged something else in, this now need yet another master merge.

danscime:
Fixed

JacquesCarette:
And, of course, I merged some things that were submitted earlier, so another merge is needed... But if you hurry, yours is likely to sneak in next!

danscime:
Fixed.

Issue #1526, opened 2019-06-10
Rename static


danscime:
Rename `static` as discussed in #1468.

danscime:
Renaming via 8e38267be3d2be82775ed1cd9b4589d470122727.

@JacquesCarette Are these suitable names?

Also `dynamic` is a term that is used as well pretty much wherever `static` is used, so I think I should rename that as well in order to be consistent.

danscime:
Renaming dynamic to dynamicRepr and Dynamic to DynamicCon via 584621ec6bfb970f029c81fffcc6f24db905ab35.

JacquesCarette:
Do we need to rename `Static` ? We do need to rename `static` for sure.

adding `Repr` is ok (super ugly, but I can't think of anything better). I would prefer to not add `Con` if we can at all avoid it.

danscime:
@Mornix would `Static` need to be renamed?

Mornix:
Only lowercase `static` causes the false-positives with HLint.

JacquesCarette:
Right - so let's just change that. And I've changed my mind: `static_` rather than `staticRepr` please.

danscime:
I will delete my current branch and implement this via a new branch.

danscime:
Should `dynamic` be changed to `dynamic_` for consistency?

JacquesCarette:
yes.

PR #1532, opened 2019-06-10
PR - Rename static and dynamic

danscime:
Based on #1468.

Renaming `static` to help avoid new Hlint errors.
Renaming `dynamic` for consistency.


danscime:
Being redone as discussed in #1526.

PR #1536, opened 2019-06-10
Clearer printing of doubles

bmaclach:
This PR updates printing of doubles such that values between 0 and 0.001 (inclusive) get printed directly, without any scientific notation. This is as discussed in https://github.com/JacquesCarette/Drasil/pull/1496#discussion_r291226461 (printing of numbers between 0 and 1000 was already how we want it, so no changes there).


JacquesCarette:
Note: that last comment was meant for #1736 not this one.

PR #1545, opened 2019-06-11
New module Potential Energy in datadefs.hs

oluowoj:
Added new data definition to Datadef.hs

Files updated:
image


oluowoj:
dependent on PR #1546 I think

JacquesCarette:
Because #1546 has been merged, I'll be hopeful that this indeed works. Let's hope.

JacquesCarette:
The build failure does seem to be dependent on that, so let's hope.

PR #1548, opened 2019-06-12
Don't generate useless files

bmaclach:
This PR updates all of our language renderers such that, if a file's main contents are empty, the file doesn't get generated. (Previously we would still generate a file with the "top" and "bottom" portions but nothing in between, i.e. just a file with imports, which was useless).


JacquesCarette:
unlucky... @danscime just made a change to the same files (two renames) that apparently can't be automatically merged.

Issue #1453, opened 2019-05-30
Code Generation - Loose input structure option leads to broken code


bmaclach:
Choosing "Loose" as the input structure for code generation (as opposed to "AsClass") leads to very broken code. The differences between Loose and AsClass seem to be the following:

- Loose combines the declaration of input variables, checking input_constraints, and calculating derived values into a single module whereas AsClass has separate modules for each
- Loose does not have a class for storing all of the input variables. It simply declares/defines different variables for each input. 

A lot of our generator assumes that there is an "InputParameters" object and that the DerivedValues and InputConstraints modules exist, but these assumptions do not hold when "Loose" is chosen (see related issue #1177).

I don't know how best to fix this because I don't know what the intention behind the "Loose" option was. @JacquesCarette do you know what code generated with this option was supposed to look like? Is the idea that all the input variables would be global? Or local variables that get passed to all of the functions that use them?

smiths:
@bmaclach, this issue reminds me of our recent discussions with @samm82 and @JacquesCarette.  We want implementation variabilities in our generated code.  We could have code that follows the human's design, or code that ignores the imposed design structure.  I think the problem with AsClass and Loose is that they aren't really defined.  The exercise with Projectile should help us get a better idea of the design variabilities we want to control.

JacquesCarette:
I think the idea was that `AsClass` (really `AsRecord`) should pass a single record for all inputs, while `Loose` (really `Unbundled`) passes all variables explicitly, "unbundled" as it were.

The code should work equally well either way. That, of course, doesn't mean that the current generator is set up to deal with it in either configuration!

bmaclach:
For the `Loose` option, passing each input explicitly to the functions that use them is now working. But one thing I still don't understand about the `Loose` option is how it's supposed to work with setting the values of the input variables. The `func_get_input` function (i.e. the function that reads an input file and sets the values) would have to return multiple values if the inputs are all unbundled, which we can't currently do with GOOL (unless we use a class, but the point of `Loose` seems to be to not use a class). Should we just inline the functionality of `func_get_input` in our Control module? @JacquesCarette @smiths

smiths:
I agree with your proposal to inline the functionality into the control module.

[...]

PR #1549, opened 2019-06-12
Making derived_values optional

### LINKED ISSUE #1451: Code generation: derived_values always called with a parameter even if it takes none
The call to derived_values in generated Control modules is hardcoded to call derived_values with the input parameters object as a parameter. However, if there are no derived values, then the function takes no parameters (and does nothing), so in this case the generated call to derived_values throws an error.

Some pontential fixes:
1. Modify generation of derived_values so that derived_values always take a parameter
2. Modify generation of the Control module to call derived_values with the correct parameters
3. Modify the code generation such that, if there are no derived values, the DerivedValues module and function do not get generated at all, nor does the call to derived_values in the Control module.

I lean towards option 3 unless we have a reason to generate a DerivedValues module even when it does nothing? @JacquesCarette @smiths

bmaclach:
This PR closes #1451. To avoid making this PR too large, I focused only on the derived_values function for now. That means that there are some inconsistencies with how derived_values gets handled compared to other functions. I'll update the generation for the other functions to be consistent with derived_values next, but in a separate branch/PR (though it will be dependent on this one).

Note that I did things differently than my proposed plan in https://github.com/JacquesCarette/Drasil/issues/1451#issuecomment-499989429 with regards to de-hardcoding the module dependency map. Removing the map altogether proved to be a bigger task than I originally thought, as it would require me to update all of the functions that generate any code to return not only the code but also the dependencies for that chunk of code. We may need to do that in the future, but for now it seems unnecessary, because I also realized that the correct module export map and the `Choices` record should be enough information to generate the dependency map prior to generating code (like we currently do). So that's how I did it. Plus, @JacquesCarette mentioned in #1451 that we will probably want to generate the module dependency map in the future, and so that should be much easier if we are able to build it this way.


smiths:
@bmaclach, I've noticed in your generated code (in stable) that you have unpacked the function arguments as separate inputs, instead of bundling them all in a record.  The original PR doesn't suggest (at least to me) that this will be part of the change.  I can see an argument for unpacking the arguments, but I also can see an argument in favour of bundling them together.  One of the random functions that I looked at had 12 separate input arguments.   If a human being is ever going to use a library call, this isn't the easiest interface for them.  I feel like we should discuss this design choice and not just have this change go in as part of another change.  If we had design variabilities, then the structure of the input arguments could be a design variability.

bmaclach:
@smiths So one of the design variabilities we already have is the choice of input structure as either `Loose` (i.e. unbundled) or `AsClass` (i.e. bundled). The NoPCM example uses `Loose`, and thus has unbundled inputs. In the past, the code generator for `derived_values` was not recognizing this and thus was still trying to pass in an `inParams` object to the function, even though such an object didn't exist. So the change you see here is actually not a change, but a correction.

bmaclach:
@smiths That being said, the `Loose` option could use some design discussion. Some of that has started in #1453.

smiths:
Got it!  Thank you for the information @bmaclach.  Linking to the associated discussion of `Loose` was a good idea.  I'm glad what we are trying to do is what I was hoping we were trying to do.  :smile:.

PR #1552, opened 2019-06-12
Improving input_constraints generation

bmaclach:
This PR is similar to the previous one (#1549) but for the input_constraints function. It corrects the function to always have and get called with the correct parameters, and makes the function disappear if there are no inputs with constraints.


code/stable/nopcm/src/cpp/Control.cpp
```diff
@@ -23,7 +23,7 @@ int main(int argc, const char *argv[]) {
     string inputfile = argv[1];
     InputParameters inParams = InputParameters();
     func_get_input(inputfile, inParams);
-    input_constraints(inParams);
+    input_constraints(A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D, T_W, E_W);
```
JacquesCarette: Does this compile? The previous and next line uses `inParams` and here they are all exploded.

code/stable/nopcm/src/java/SWHS/Control.java
```diff
@@ -14,7 +14,7 @@ public static void main(String[] args) throws Exception {
         String inputfile = args[0];
         InputParameters inParams = new InputParameters();
         InputFormat.func_get_input(inputfile, inParams);
-        InputConstraints.input_constraints(inParams);
+        InputParameters.input_constraints(A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D, T_W, E_W);
```
JacquesCarette: and this I am quite sure can't compile.

bmaclach: @JacquesCarette Right, NoPCM has never been able to compile. It's getting better but still quite broken.

PR #1554, opened 2019-06-13
Don't generate empty files when a module is empty

bmaclach:
My last PR about this succeeded in making modules empty if they had no functions/classes, but I realized an empty file would still be generated. This PR makes it so that empty files do not get generated.


bmaclach:
@Mornix Any insight into the build error I got here?

Mornix:
Stack appears to have had an intermittent outage or disconnected the socket before the content was completed downloading.

bmaclach:
@Mornix Okay, thanks. Is closing/reopening the PR the only way to re-run the build?

Mornix:
If the PR build builds fine, then this should be okay to merge. /cc @JacquesCarette

Mornix:
To get a push-build to rerun, either @JacquesCarette has to do it from Travis itself or you have to push a blank or blank amended commit.

PR #1557, opened 2019-06-14
More code generation de-hardcoding and consistency improvements

bmaclach:
For the get_input function, I could not do all of the improvements that I've done for the other functions because, unlike the others, get_input is defined explicitly in Drasil and then passed to the code generator. Eventually we will need to generate our input format reading just from the inputs and `Choices`, because sometimes we don't even want it to be a function (for example if the `Loose` input structure is chosen), but that will be a bigger change that will come later. For now, I've made get_input consistent with the others as far as I could, which allowed me to remove the old `fApp` function and replace it with the new one.

This PR is dependent on #1556. That PR should be merged first.


JacquesCarette:
#1556 is merged - but there are still conflicts, which is surprising. In any case, great stuff!

JacquesCarette:
As this is all codegen, I'll merge now.

PR #1559, opened 2019-06-14
Improvements to input parameters code generation

bmaclach:
This PR de-hardcodes parts of the code generator related to the input parameters (the input parameters class for `AsClass` input structure and the input parameter variables for `Loose` input structure). 

This PR depends on #1557 and #1556. Those PRs should be merged first.

PR #1561, opened 2019-06-14
Finish de-hardcoding module dependency map

bmaclach:
This PR de-hardcodes the last couple entries in the module dependency map.

This PR depends on #1556, #1557, and #1559. They should be merged first.

This should wrap up my "de-hardcoding" series of PRs, at least for now :)


bmaclach:
@JacquesCarette I added an intermediate record between `SystemInformation` and `CodeSpec` to decrease the number of parameters for these functions. I then had a record within a record with `CodeSpec` and so I had to update the places where information was extracted from it accordingly. I'm not sure if there was a cleaner way to do this?

I don't know why there are still conflicts either... After googling it seems like maybe what I should have done was branch from my previous branch instead of from master.

JacquesCarette:
Yes, I think than branching from the branch is the way to go.

The code didn't get much cleaner because you did a pattern-match on the record on input (which is fine), though maybe using the field extractor functions (especially for the fields that are used only once) might make things shrink more.

bmaclach:
@JacquesCarette Good point. I did some more cleanup based on that suggestion.

Issue #512, opened 2018-05-09
Remove `defaultValue` and `defaultValue'` hacks


niazim3:
As explained in issue #356 by @JacquesCarette, these hacks may prevent extending things and were likely put in place to accomodate for how some languages must have a value for all variables when they get created. 

Further mentioned: "the reason I call it a hack is that it is based on the type (in this case StateType) to have a "default value". Rather, the backend code generator should insist that all values have a reasVal instead. That way the "default" value can be application-specific rather than type-generic."

`defaultValue` and `defaultValue'` are defined in [code/Language/Drasil/Code/Imperative/AST.hs](https://github.com/JacquesCarette/Drasil/blob/cad5e614b0cfa68ce05f69ef7944bdfa53df77a6/code/Language/Drasil/Code/Imperative/AST.hs#L253-L265)

bmaclach:
Note that `defaultValue` and `defaultValue'` have been replaced with type-specific `default` functions in finally tagless GOOL, but these are still hacks and should still be removed:

https://github.com/JacquesCarette/Drasil/blob/b72f7c1627f89acd04ba0ff1b3cdc06d9510270f/code/drasil-code/Language/Drasil/Code/Imperative/New.hs#L144-L148

From @JacquesCarette : "The best way to do this is to insist that all variables get initialized."

I'll note that while using a `reasVal` for a default value should work, there is one other place where the `default` functions are used: GOOL offers a `listExtend` function which extends a list by filling in the extended piece with the default value for that type.

JacquesCarette:
Definitely a solid improvement.  We should see if any of those `default*` routines are used currently, and remove the ones that aren't. Is `listExtend` used? If not, remove it!

bmaclach:
@JacquesCarette The `default*` routines are all used when initializing input variables. However, if we use your suggestion of using the typical value as the default value, then those uses of `default*` can be removed.

As for `listExtend`, it is unfortunately used when reading input files. And the typical value solution won't always work in this case. For example, when reading the TSD.txt and SDF.txt files in GlassBR, the values we are reading do not each have a typical value - these values don't even exist in Drasil (though perhaps they should, and then maybe the typical value solution could work)

JacquesCarette:
It would be best if we could remove all 'dummy' initialisations -- whether through default or typical.
[...]

PR #1589, opened 2019-06-19
Removed listPopulate

bmaclach:
Contributes to #512. `listPopulate*` were more functions that used the `default*` functions (I missed them before). They were only implemented in Python and used nowhere, so they seem very safe to remove.


JacquesCarette:
Because the two branches were done independently, there's now a conflict to resolve.

bmaclach:
Darn, I thought they were independent. Fixed now.

PR #1594, opened 2019-06-20
Implemented TypeData instead of pair for types in GOOL

bmaclach:
As suggested by @JacquesCarette in #1581, this PR replaces the pair for types in GOOL with a new data structure.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/LanguageRenderer.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/LanguageRenderer.hs
@@ -38,7 +38,7 @@ import Utils.Drasil (capitalize, indent, indentList)
 import Language.Drasil.Code.Code (CodeType(..))
 import Language.Drasil.Code.Imperative.Symantics (Label, Library)
 import Language.Drasil.Code.Imperative.Helpers (Terminator(..), ModData(..), md,
-  angles,blank, doubleQuotedText,hicat,vibcat,vmap)
+  TypeData(..), td, angles,blank, doubleQuotedText,hicat,vibcat,vmap)
 
 import Data.List (intersperse, last)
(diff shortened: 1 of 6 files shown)


code/drasil-code/Language/Drasil/Code/Imperative/LanguageRenderer.hs
```diff
@@ -184,8 +184,8 @@ constructDocD _ = empty
 
 -- Parameters --
 
-stateParamDocD :: Label -> (Doc, CodeType) -> Doc
-stateParamDocD n (t, _) = t <+> text n
+stateParamDocD :: Label -> TypeData -> Doc
+stateParamDocD n (TD _ t) = t <+> text n
```
JacquesCarette: might be even better to not pattern-match here and instead use the projector, so that if things get added to the data type, this code wouldn't have to change. Same with the other places where you pattern match on `TD`.

bmaclach: @JacquesCarette Done for all of the cases of pattern matching except where I was pattern-matching on a specific type... I suppose I could use a `case` statement for those scenarios, but is that still preferred over pattern matching?

Mornix:
Looks like the PR Travis build finished successfully, but Travis itself doesn’t recognize it’s done and will probably “terminate” within the hour with some “No progress” reason. However, it has completely successfully despite not being marked as such.

code/drasil-code/Language/Drasil/Code/Imperative/LanguageRenderer.hs
JacquesCarette: No, pattern matching is better than `case` in almost all situations.

PR #1595, opened 2019-06-20
Changes Take [ConceptInstance]

### LINKED ISSUE #1471: Changes to take [ConceptInstance]
I see that work has been started to make an `LCsSec'` constructor to take a `[ConceptInstance]`, and I couldn't find who was working on it. @smiths @JacquesCarette @Mornix Anyone know who was working on this, and can I continue the work on it?

Likely changes has been started, so I'll continue it, taking inspiration from the work I did on the FRs and NFRs, and I'll also do the unlikely changes.

samm82:
Both Likely and Unlikely Changes take a `[ConceptInstance]`. I also cleaned up the exports of `docLang`

Closes #1471


code/drasil-docLang/Drasil/ExtractDocDesc.hs
```diff
@@ -15,10 +15,9 @@ egetDocSec IntroSec{}           = []
 egetDocSec (StkhldrSec s)       = egetStk s
 egetDocSec (GSDSec g)           = egetGSD g
 egetDocSec (SSDSec s)           = egetSSD s
-egetDocSec (ReqrmntSec r)       = egetReq r
-egetDocSec (LCsSec l)           = egetLcs l
-egetDocSec LCsSec'{}            = [] -- likely changes can't lead to Expr?
-egetDocSec (UCsSec u)           = egetUcs u
+egetDocSec ReqrmntSec{}         = [] -- requirements can't lead to Expr?
+egetDocSec LCsSec{}             = [] -- likely changes can't lead to Expr?
```
JacquesCarette: Are you absolutely sure of this? I am not.

samm82: There is a requirement in GlassBR that has an `Expr` (which might be removed in the future) so that (at least for now) is wrong. I think I should assume that `Expr`s can show up in the changes as well, as there isn't any reason I can think of as to why they couldn't. Does this sound good @JacquesCarette?

JacquesCarette: Yes. So the code should be adjusted to traverse things properly.

Mornix: I'm going to say to hold off on this (the LCs and UCs traversals). I have this fixed in a local branch which I'm in the process of cleaning up. No point in doubling up on the work. 😄

samm82: @Mornix Should I still fix the requirements traversal?

Mornix: Already taken care of as well.

samm82: Works for me!

Issue #1606, opened 2019-06-22
Fix Symbol Corners


Mornix:
The title to this issue is fairly vague, and that's because I'm not certain I understand the intricacies of what needs to be done. I do know that this is an issue.

Symbols defines a constructor `Corners` which allows the user to add arbitrary super/subscripts to any of the corners of an inner symbol.

https://github.com/JacquesCarette/Drasil/blob/76913ecbd3ef11258b38294318fdfb6d81502585/code/drasil-lang/Language/Drasil/Symbol.hs#L27

The issue arises that the common smart constructors always wrap an existing symbol with a new instance of corners:

https://github.com/JacquesCarette/Drasil/blob/76913ecbd3ef11258b38294318fdfb6d81502585/code/drasil-lang/Language/Drasil/Symbol.hs#L130-L143

Moreover, symbol rendering can only handle top right and bottom right corner rendering *individually* in both the HTML and LaTeX printers. Want to use a combination or something else? You're SOL at the moment. 

LaTeX:
https://github.com/JacquesCarette/Drasil/blob/76913ecbd3ef11258b38294318fdfb6d81502585/code/drasil-printers/Language/Drasil/TeX/Print.hs#L82-L86

HTML:
https://github.com/JacquesCarette/Drasil/blob/76913ecbd3ef11258b38294318fdfb6d81502585/code/drasil-printers/Language/Drasil/HTML/Print.hs#L119-L123

Fun fact: Now that I see the HTML super and subscript renderer code, it can render nested `Corners` cases incorrectly and is inconsistent in behaviour with the LaTeX printer. 

So really, this comes down to the fact that I think `Corners` as an idea needs to be better understood. 

* What do items in different corners mean?
* Can multiple corners be populated at the same nesting "level"?
* Should the way `Corners` is represented have some sort of precedence where nested `Corners` "resets" precedence similar to brackets?
* Visually how do you represent (and understand) nested and multiple corners?

JacquesCarette:
Yeah, these definitely are bugs. This needs a proper design.

balacij:
To my knowledge, 'prescripts' are written in LaTeX (and similarly in KaTeX) using a null text placed _before_ the 'middle' symbol. For example, ${}^1_2a^3_4$ is written with `${}^1_2a^3_4$` (edit my comment to have a peek!).

Issue #1605, opened 2019-06-22
Improve Symbol Decoration Infrastructure


Mornix:
Drasil currently supports symbol decorations in the form of:
* Hat
* Prime
* Vector

This is a good start, but there are a few "hacks" which have propagated. One is `∆` which (in the current use cases) represents "difference of value". The symbol sorting algorithm considers this to be a decoration (albeit hardcoded) and strips it off when comparing. 

https://github.com/JacquesCarette/Drasil/blob/76913ecbd3ef11258b38294318fdfb6d81502585/code/drasil-lang/Language/Drasil/Symbol.hs#L56-L71

Similarly, `|| ||` is hardcoded in a few examples to indicate the "magnitude of some vector". (This one isn't hardcoded when symbol sorting and thus the magnitudes will always sort lower than the vectors and other symbols beginning with the same character).

It may be worth encoding these (and others I haven't noticed) as decoration within `Symbols` to make sorting more consistent and remove these hacks encoding semantic meaning through use of a literal symbol.

JacquesCarette:
I agree. And we should revisit the Δ hack too. Maybe we should simply remove it?

PR #1607, opened 2019-06-22
Symbol Sorting Fixes

Mornix:
This PR addresses some inconsistency when sorting (discovered making innocuous changes on a yet-unpushed branch) where sorting of symbols could change (in the bad way, not the stable sort way) when un-equal items in the input list were in a different order. 

This PR also slightly altered the behaviour of symbol sorting such that having symbols v<sub>i</sub>, v<sub>f</sub><sup>AB</sup>, and v<sub>f</sub> would be sorted as: 
v<sub>f</sub>, v<sub>f</sub><sup>AB</sup>, v<sub>i</sub> instead of the current
v<sub>f</sub>, v<sub>i</sub>, v<sub>f</sub><sup>AB</sup>

If this is an undesirable change, I can back it out.

This PR also spurred two issues involving Symbols: #1605 and #1606.


smiths:
@Mornix, I like your new sorting order better than the old one.  Keeping the symbols with the same "base" and subscript together makes sense to me.

Issue #1432, opened 2019-05-28
Make Terminology and Definitions Sections Consistent


samm82:
Some examples use bulleted lists, some use numbered lists, and some just list them (much like how assumptions are listed). Which is the preferred method for the Terminology and Definitions section? @smiths @JacquesCarette 

![chipmunk](https://user-images.githubusercontent.com/35857611/58499849-f2a76680-814e-11e9-81ff-aa63bf558014.png)

![glassbr](https://user-images.githubusercontent.com/35857611/58499888-0226af80-814f-11e9-9d36-963268f5e3e9.png)

![ssp](https://user-images.githubusercontent.com/35857611/58499919-14085280-814f-11e9-92a4-55d4620d42d9.png)

(I think I'll use a similar workflow to #1403 - implement the desired result in Projectile and go back for the other examples later.)

Smaller note - the period is missing from the end of the intro in Chipmunk.

JacquesCarette:
I prefer bullet (or to use the 'definition' environments in latex/html). There is no point to numbered lists since the definitions will not be referenced that way.

This is of course another area where the `docLang` should be used to gain uniformity.

smiths:
I agree with @JacquesCarette.

samm82:
Since the Terms and Defs section is generated like so:

`SSDProg [SSDSubVerb probDescription, ...]`

Should this issue be expanded to generate the Problem Description section from `docLang`? It's a small(ish) section and shouldn't too difficult. @smiths @JacquesCarette

samm82:
Part of #990

smiths:
That sounds good to me, but @JacquesCarette will see the whole picture better than me.  If this issue is expanded, please create another issue, rather than piggyback on this one.

samm82:
In GlassBR, there are subsections like so:
![glass](https://user-images.githubusercontent.com/35857611/58725881-3d72e980-83ae-11e9-8918-435a3a02988e.png)

If we wanted to automate this section, it would be a lot easier to collapse the list into one flat list. We would sort the whole list alphabetically (#1245), so it would still be easy to find a definition. The sub-definitions of glass type should be changed (ie. from "Annealed (AN)" to "Annealed (AN) glass").

What are your thoughts on this? @smiths @JacquesCarette

smiths:
The definitions are currently subsections?  I agree that having a list of definitions would be better.  Numbering them isn't necessary, but sorting them alphabetically would be great.

[...]

Issue #1244, opened 2019-05-01
Merging Combinators for making Sentences for Terminologies with Definitions 


samm82:
(Continuation of #835)

https://github.com/JacquesCarette/Drasil/blob/46967c44b13f1c82253ecd6ffa63e85cf69ee9cc/code/drasil-example/Drasil/GlassBR/Body.hs#L205-L216

As noted in the comment at the bottom, the last entry (probability of breakage) should be reworked, since it is similar to `tAndDWAcc`. It is required because `ConceptChunk` doesn't have a symbol, so it is supplied from the `ConstrainedChunk`.

Based on the comments here, I think it would make sense to treat acronyms and symbols as equivalent, since in all contexts I can think of, they both function as a short form of the term. This would allow for easy merging of these functions. Does this sound like an appropriate fix? Not sure about the design implications of this. @smiths @JacquesCarette

JacquesCarette:
Acronyms and symbols serve different purposes. An acronym is meant to be used in an English context, while generally the symbol is in an equation context. So wouldn't want to merge these without looking at all the examples very closely.

PR #1613, opened 2019-06-24
Terminology and Definitions section Automated (almost)

samm82:
Part of #1432. The Terminology and Definitions section is now automated in all examples except for GlassBR, due to #1244 and discussion in #1432. Sorting is also not done yet, because that will likely need some deeper design.


samm82:
@JacquesCarette Ready to be merged 😊

PR #1616, opened 2019-06-25
GOOL Input/Output functions

bmaclach:
This PR adds functions to GOOL for generating functions based on their inputs and outputs,  and for generating calls to these functions, which allows for a generated function to return multiple outputs, which is needed for fixing #1453.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Code.hs
+++ b/code/drasil-code/Language/Drasil/Code/Code.hs
@@ -18,4 +18,4 @@ data CodeType = Boolean
               | List CodeType
               | Iterator CodeType
               | Object String
-              | Void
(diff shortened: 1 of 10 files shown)


bmaclach:
I'm going to temporarily close this because I want to change the parameters of `inOutFunc` from `(Label, repr (StateType repr))` lists to simpler `repr (Value repr)` lists. At first I thought this would be a hack because it means the function will need to convert `Value`s to `Parameter`s, but thinking about it more I think it actually makes sense to build `Parameter`s from `Value`s, because a parameter is nothing more than a sort of `Value` declaration.

JacquesCarette:
Good stuff!

PR #1627, opened 2019-06-26
Generate input format

### LINKED ISSUE #1497: Disallow conflicts between inputs and input file format definition
As mentioned in #1450, we had made a mistake in our definition of NoPCM's input file where it included a variable `tau` that wasn't actually an input, and was missing the actual input variable `timeStep`. This mistake went undetected for some time. 

Drasil should catch such conflicts between inputs and input files. This could be done by collecting the unitals used in the input file, comparing them to the inputs, and throwing an error if there is a difference. 

Another possibility is to prevent such mistakes from ever happening by reworking how we define input files. I am thinking that an input file format could be automatically generated based on some high-level choices and the known inputs (in the `SystemInformation`). Thus, the "user" of Drasil would not have to be responsible for ensuring all inputs from the `SystemInformation` are included in the input file, and it would be impossible for a non-input to make its way into the input file.

bmaclach:
This PR closes #1497 and contributes to #1453 by updating the generation of the input format function/module to be more automated, similar to the way we automate other functions/modules, instead of having it hardcoded in each example. As mentioned in #1497, this forces a "one input per line" input format for now but the idea is that we will add more choices for input format in the future.


bmaclach:
I must have done something crazy with Git that seems to have caused my commits to have been repeated twice... The final result is correct though. Sorry about that.

Issue #1640, opened 2019-06-27
Bold Table Headings in LaTeX


samm82:
The headings of the reference tables should be bold in LaTeX:

![image](https://user-images.githubusercontent.com/35857611/60296264-b56ff900-98f3-11e9-99e2-50334fb2b88b.png)

smiths:
@samm82, this issue isn't descriptive enough for someone else to understand it.  Please add a screen shot of the table that doesn't have bold in the heading.  This will help us remember what we were talking about, especially if we don't get back to it right away.  It will also help others judge whether this is a real issue, or just a personal preference of @smiths.  😄

JacquesCarette:
So that "should be bold" could be backed up by reference (even Wikipedia!) on style that recommends this, for example.

smiths:
Yes, a reference that backs up our formatting would be great.  Eventually we want the formatting style to be a user selected variability.  My motive at the moment is just to make it easier to review these documents.  When the table of symbols spilled to a second page the header and the entry in the table were formatted in the exact same font.  This made it difficult to quickly determine whether the first entry in the table was the header or an entry.

@samm82, can you investigate the recommended style for tables in scientific writing.  I'm sure you will find resources that you can provide us with links to.

samm82:
A lot of sources don't have the headers bolded, but [this one](https://dl.sciencesocieties.org/files/publications/style/chapter-05.pdf) does, and the example on page 5-3 resembles our Table of Symbols, etc. This, paired with the reasoning presented by @smiths, is more than enough to justify our use of bold headers. Should this information be kept somewhere? @smiths @JacquesCarette

smiths:
@samm82, for now, I think the best place to have the rationale for this change is in this issue.  You have a link above to an "authoritative" source, and we have a rational argument in the comments above.  I can't think of another spot to record this at the moment.  Two things in the future that will give us a better solution:

1. Drasil in Drasil
2. Incorporation of rationale along with knowledge

On its own, I do not think bolding the headings of a table or not is an interesting enough variability to encode as a user's choice.   However, in the future, I could envision us having "plain" and "fancy" options.  Even better would be if we had options based on style guides, like the "Chicago" style, or "Harvard" style etc.  I think this would be too much of a tangent at the moment.  Getting code generation and design decisions working is more important, but if is nice to know that we could incorporate some documentation recipes.

PR #1644, opened 2019-06-27
Minor LaTeX Table Fixes

samm82:
- Bolded all non-chunk table headers in LaTeX
   - This also bolds the headers of the trace tables; is this OK/desired? @JacquesCarette @smiths
- Increased chunk column width so everything fits on one line

_Closes #1640 and closes #1638_


smiths:
The bold tables look good to me, as does the output constraints.  (I can see that we could use the white space more effectively with a narrower column and a new line between "Output" and "Constraints", but  I think we don't want to get too caught up in fine tuning the appearance of the generated documentation.  The changes in this PR already make the LaTeX docs look better.)

samm82:
@smiths Unfortunately, line splitting in LaTeX isn't as simple as I'd hoped it would be, or else I would have gone that route.

smiths:
There are ways to tell LaTeX how to format the contents of a cell.  It seems to me that I've used something called ragged right for this in the past.  Ragged right is discussed at:

https://tex.stackexchange.com/questions/12703/how-to-create-fixed-width-table-columns-with-text-raggedright-centered-raggedlef

Among other places.

It might let us reduce the column width, but still have it look nice.  If a quick check of this doesn't work @samm82, I'm fine with not spending any more cycles on this.  The LaTeX tables already look better than they did.

samm82:
`raggedright` is actually a lot simpler than I thought, so I went that route now instead. I might have spent a little too long making the tables the "perfect" size, but at least no one should have to fix it in the future.

code/drasil-printers/Language/Drasil/TeX/Print.hs
```diff
@@ -197,7 +197,7 @@ makeTable :: [[Spec]] -> D -> Bool -> D -> D
 makeTable lls r bool t =
   pure (text ("\\begin{" ++ ltab ++ "}" ++ (brace . unwords . anyBig) lls))
   %% pure (text "\\toprule")
-  %% makeRows [head lls]
+  %% makeRows [map (\x -> S "\\textbf{" :+: x :+: S "}") (head lls)]
```
JacquesCarette: You really shouldn't hand-build latex like this, but rather use the facilities in the pretty-printer built for that purpose.

code/drasil-printers/Language/Drasil/TeX/Print.hs
```diff
@@ -333,7 +333,7 @@ makeDefTable :: PrintingInformation -> [(String,[LayoutObj])] -> D -> D
 makeDefTable _ [] _ = error "Trying to make empty Data Defn"
 makeDefTable sm ps l = vcat [
   pure $ text 
-  $ "\\begin{tabular}{p{"++show colAwidth++"\\textwidth} p{"++show colBwidth++"\\textwidth}}",
+  $ "\\begin{tabular}{>{\\raggedright}p{"++show colAwidth++"\\textwidth}>{\\raggedright\\arraybackslash}p{"++show colBwidth++"\\textwidth}}",
```
JacquesCarette: hard-coded widths like this seem like a bad idea? This doesn't feel future-proof.

JacquesCarette: Oh, I see that these are percentages of \textwidth - that's not too bad then.

PR #1645, opened 2019-06-27
Fixed Spacing of Derivations

### LINKED ISSUE #1641: Spacing of Symbols in Derivatives
In both HTML and LaTeX, the spacing of the change in a variable (ie. *dt*) in derivatives should be fixed to match the display in integrals:

image

samm82:
Closes #1641


smiths:
Yes, the derivatives look better without the extra space between the d and the t in dt (for instance).  I'm not sure why that extra space was added in the first place?

samm82:
I think the order of items in the list got switched. (Not sure how else to word it 😂)

Issue #1639, opened 2019-06-27
Display Variable Names Better in LaTeX


samm82:
Variable names with "words" (like "offset") shouldn't be displayed in italics and with proper kerning in LaTeX (using \mathit{} or \text{}), possibly using a new constructor for `Symbol`.

![image](https://user-images.githubusercontent.com/35857611/60295433-ae47eb80-98f1-11e9-8194-e8c92f366952.png)

smiths:
A screenshot is really necessary here.  Nobody else, other than you and I, will understand what you are talking about based on the description.

samm82:
What other "phrases" should be displayed like this? "max" and "min" jump out at me. @smiths

samm82:
Also, is the HTML display OK?

![html](https://user-images.githubusercontent.com/35857611/60367515-604bea00-99bc-11e9-89b4-b6d8efbc7a72.png)

JacquesCarette:
To get this right, we will indeed need to change `Symbol` to have some kind of "word" constructor that says that something is a 'thing' instead of a variable.
We might need more than one -- to me *land* and *target* are quite different in kind than *max* and *min*.

samm82:
You mentioned on a commit that we should be capturing semantic information; what information should we be capturing and how?

JacquesCarette:
Shoot, it has now been too long, and I don't remember -- and the commit wasn't linked!

samm82:
https://github.com/JacquesCarette/Drasil/commit/b7582bb0be5b081e8266f0cf6a20d4facc5d52c1#commitcomment-34128032

JacquesCarette:
Thanks, that helped.

The idea is to find a common theme to the subscript. `land` and `target` are domain "objects" (for example).  `min` and `max` are (abbreviations for) specific concepts. At other times, the subscript is the name of a variable (quantity) of interest.

In the first two cases, I think we want to display those in Romain, while in the latter, in the usual Math font.

If I had to take a guess, I'd say that the separation is that the `land` and `max` cases are tied to pure concepts, while the latter are tied to quantities. But we'd need more examples to figure out a better classification scheme.

smiths:
I like the idea of @JacquesCarette to decide the formatting based on the semantic information, but I think we shouldn't invent any rules.  We should follow the standard typographic conventions, as summarized at:

https://en.wikipedia.org/wiki/Typographical_conventions_in_mathematical_formulae

[...]

PR #1651, opened 2019-06-28
Minor Fixes from Projectile

samm82:
- Added vector units to notes in Table of Symbols section
- Added Wikipedia reference to Cartesian concept
- Fixed display of strings
- Improved implementation of some constructors.

_Closes #1646_


samm82:
@smiths This change isn't for non-italicized font (that's what #1639 is for) - it's for the display of strings as a data type, which is why the quotations show up.

smiths:
Okay @samm82.  I see that what you are doing is different than what I thought.  You still need to satisfy @JacquesCarette before we can merge this.  :smile:

PR #1652, opened 2019-06-28
Reworked log message for function call

### LINKED ISSUE #1635: Logging of function calls is not always correct for functions generated with inOutFunc
When we turn the logging option on for our generated code, the log messages corresponding to function calls currently have this form:
```
function functionName(value of first parameter, ..., value of last parameter) called
```
Or as a concrete example: 
```
function func_lin_interp(1.058491334523, 21.16368512241, 1.067889548895, 21.16116642865, 1.0666666666666667) called
```
This worked fine before, when all functions were generated using the `function` function, which gets passed the function parameters explicitly, but now we generate some functions with `inOutFunc`, which determines the parameters based on the inputs and outputs. For Java and Python, the parameters directly correspond to the inputs. But for C# and C++, the parameters are a combination of the inputs and outputs (because the outputs are realized by pass-by-reference parameters). Thus, the parameters are different depending on the language, and our log messages don't account for that. Currently they print as if the parameters are the inputs, so they are correct for Python and Java, but incorrect for C# and C++, as they give the impression that the function takes fewer parameters than it actually might.

One solution would be to write different log messages for functions generated with `inOutFunc` than functions generated with `function`, but it would be better to continue having one function that consistently generates logging for all function calls. For this to work, the form of the log message would have to change so as to not list out parameters. I'm thinking something like:
```
function func_lin_interp called with inputs:
x_1 = 1.058491334523,
y_1 = 21.16368512241,
x_2 = 1.067889548895,
y_2 = 21.16116642865,
and x = 1.0666666666666667
```

bmaclach:
Small PR that closes #1635 by reformatting the log messages for a function call in generated code.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
@@ -425,15 +425,21 @@ loggedMethod n vals b =
     return $ block [
       varDec v_outfile,
       openFileA v_outfile (litString $ logName g),
-      printFileStr v_outfile ("function " ++ n ++ "("),
-      printParams vals v_outfile,
-      printFileStrLn v_outfile ") called",
+      printFileStrLn v_outfile ("function " ++ n ++ " called with inputs: {"),
+      multi $ printInputs vals v_outfile,
+      printFileStrLn v_outfile "  }",
       closeFile v_outfile ]
       : rest
   where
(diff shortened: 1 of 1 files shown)


JacquesCarette:
It's kind of disappointing that this isn't causing any test failures... means that this feature isn't being used+tested at all!

PR #1657, opened 2019-07-02
[Small PR] Added Refname to HTML

### LINKED ISSUE #1643: Add Short Label to Information Presented for Each Chunk in HTML SRS
I would like it if each "major" chunk types (TM, GD, DD, IM) incorporated its short label into the information presented for each chunk.  When we reference the chunk a label is created that has a nice name, but we don't actually see this short label when we review the original definition of the chunk.  Consider the example of "Position vector as a function of time for 2D motion under constant acceleration":

NoShortLabel

In IM: calOfLandingTime, I can find out that the short name for this is GD: posVec, but there is nothing in the definition of the GD that actually tells me what short name will be used to reference it.  I noticed this issue when reviewing #1628.  The short names were used, but I had to deduce the short names from the long names to read and understand the issue.  We used to keep the short names (in the form of a number) in the manual versions of the case study.  This made it easy to look up a particular chunk.  It is now actually harder, even though the names are better.

I would like to see a new field for each chunk that just reproduces its short name.  (I can't remember if I should say name or label.)  We already have the short names available, so hopefully this isn't too difficult.  @JacquesCarette and @Mornix will likely have advice to offer.

samm82:
Closes #1643

### DIFF
--- a/code/drasil-printers/Language/Drasil/HTML/Print.hs
+++ b/code/drasil-printers/Language/Drasil/HTML/Print.hs
@@ -259,7 +259,8 @@ makeColumns = vcat . map (td . pSpec)
 -- | Renders definition tables (Data, General, Theory, etc.)
 makeDefn :: L.DType -> [(String,[LayoutObj])] -> Doc -> Doc
 makeDefn _ [] _  = error "L.Empty definition"
-makeDefn dt ps l = refwrap l $ table [dtag dt] (makeDRows ps)
+makeDefn dt ps l = refwrap l $ table [dtag dt]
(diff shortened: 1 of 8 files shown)


smiths:
@danscime, you might want to take note of this PR and the comment of @JacquesCarette above.  Sounds like a good example for your paper.  😄

danscime:
@smiths did you mean to @ the other Dan?

smiths:
Thanks @danscime.  Yes, I did mean to @ the other Dan: @szymczdm.  However, if you want to write a paper too, that would be great.  :smile:

PR #1660, opened 2019-07-03
Improved Display of Derivations

samm82:
Fixed spacing and bold of derivations.

_Closes #1636_


JacquesCarette:
> These paragraphs give the desired spacing after a chunk that doesn't have a derivation (so that the end of a chunk is more clearly visible). I think this change was brought up in a meeting about Projectile.

Feels like a hack. The problem is real, the solution doesn't seem 'right'.

JacquesCarette:
A conflict to resolve, but this is definitely ready to go now!

PR #1662, opened 2019-07-03
Polymorphic GOOL functions

### LINKED ISSUE #1611: Take advantage of Values holding type information in GOOL to simplify functions
With the merge of #1610, `Value`s in GOOL are all `ValData`, so they hold type information accessible with the `valType` function:
https://github.com/JacquesCarette/Drasil/blob/54d6d290449ebca55966674255495cb2333fc25c/code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs#L59

Reading types from `Value`s should allow us to clean up many of the GOOL functions in Symantics.hs (and implemented in the different renderers in the LanguageRenderer directory). We should look over all of these functions and clean them up where we can.

If a function gets passed a `repr (StateType repr)` but also a `repr (Value repr)` which holds the needed type information, the `StateType` parameter can be removed and the type read from the `Value` instead. An example is `listSlice`, but there may be more:
https://github.com/JacquesCarette/Drasil/blob/47050c6081fc3a663b23b69fac0ee7c28679f839/code/drasil-code/Language/Drasil/Code/Imperative/Symantics.hs#L93-L95
Just ensure that any `StateType` parameter removed actually corresponds to the existing `Value` parameter passed to the function (this will have to be determined on a case-by-case basis by understanding how each parameter gets used)

If a function gets passed a `Label` and a `repr (StateType repr)` and both get passed to `var` in the function body to construct a `Value`, the two parameters can be replaced by a single `repr (Value repr)` parameter. An example is `get` but there may be more:
https://github.com/JacquesCarette/Drasil/blob/47050c6081fc3a663b23b69fac0ee7c28679f839/code/drasil-code/Language/Drasil/Code/Imperative/Symantics.hs#L291
https://github.com/JacquesCarette/Drasil/blob/47050c6081fc3a663b23b69fac0ee7c28679f839/code/drasil-code/Language/Drasil/Code/Imperative/LanguageRenderer/PythonRenderer.hs#L297

If there is a group of functions where each operates on a specific type, the group can be replaced by a single polymorphic function that pattern matches on the type and then performs the correct actions for that type. For example, `getIntInput`, `getFloatInput`, `getBoolInput`, `getCharInput`, and `getStringInput` can be replaced by a single `getInput` function.
https://github.com/JacquesCarette/Drasil/blob/47050c6081fc3a663b23b69fac0ee7c28679f839/code/drasil-code/Language/Drasil/Code/Imperative/Symantics.hs#L376-L380

bmaclach:
This PR closes #1611 by combining groups of functions where each function acts on a specific type (like `getIntInput`, `getFloatInput`, etc.) into a single polymorphic function (`getInput`) that works for all types.

This PR depends on #1659, so that should be merged in first.


JacquesCarette:
Not sure why there are conflicts as, as far as I can tell, you're doing the right thing with branches. So I don't quite understand - but it is worth investigating.

Mornix:
If I were to guess, I'd assume the "conflicts" arise because the way they're being merged is the commits are being squashed so it's not a simple fast-forward from master to the current branch.

ie. By squashing the commits the tree for `master` and `polyGOOL` no longer share the same root, `master` is 
original `master` -> squashed previous commits
and `polyGOOL` is:
original `master` -> unsquashed series of commits -> `polyGOOL` commits
and while the code is the same between the two, the hashes and so forth don't match.

JacquesCarette:
So I should try to 'rebase' and merge then? [Though it's too late for this one]

Mornix:
I believe a rebase merge should work.

PR #1663, opened 2019-07-04
Fixed SSP Table of Symbols

samm82:
Removed definitions and used sentence case and coordinate math concepts (`xCoord`, etc.)

_Closes #1626_

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Math.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Math.hs
@@ -11,8 +11,8 @@ mathcon :: [ConceptChunk]
 mathcon = [angle, area, calculation, cartesian, change, constraint, diameter,
   equation, euclidN, euclidSpace, gradient, graph, law, matrix, norm, normal,
   normalV, number, orient, parameter, perp, perpV, pi_, probability, rOfChng,
-  rate, rightHand, shape, surArea, surface, unitV, unit_, vector, xAxis, xComp,
-  xDir, yAxis, yComp, yDir, zAxis, zComp, zDir]
+  rate, rightHand, shape, surArea, surface, unitV, unit_, vector, xAxis, xCoord,
+  xComp, xDir, yAxis, yCoord, yComp, yDir, zAxis, zCoord, zComp, zDir]
 
 mathcon' :: [CI]
 mathcon' = [pde, ode, de]
@@ -20,8 +20,8 @@ mathcon' = [pde, ode, de]
 angle, area, calculation, cartesian, change, constraint, diameter, equation,
   euclidN, euclidSpace, gradient, graph, law, matrix, norm, normal, normalV,
   number, orient, parameter, perp, perpV, pi_, probability, rOfChng, rate,
-  rightHand, shape, surArea, surface, unitV, unit_, vector, xAxis, xComp, xDir,
-  yAxis, yComp, yDir, zAxis, zComp, zDir :: ConceptChunk
+  rightHand, shape, surArea, surface, unitV, unit_, vector, xAxis, xCoord,
+  xComp, xDir, yAxis, yCoord, yComp, yDir, zAxis, zCoord, zComp, zDir :: ConceptChunk
 
 pde, ode, de :: CI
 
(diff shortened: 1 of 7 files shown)


bmaclach:
@smiths When you say you "searched through the original documentation" do you mean the manual version in the case studies repo? If so, the reason you didn't find "surface angles" is because the manual version of the table of symbols shows the definition only, not the "name" of the symbol. So it is there with the definition "angle between the surface of a slice and the horizontal".

I had forgotten that the manual version showed the definitions in the table of symbols. I wonder if it would be possible to do the same in Drasil (i.e. show only the definition, not the name), @samm82? Or should our preference be to show both name and definition as was done previously, @smiths?

samm82:
Should I manually add the missing definitions to the Terms and Defs sections, or should this be automated, so that every term in the ToS shows up in the Terms and Defs? A current problem with automating (and with only displaying the definitions, as @bmaclach suggested) is that some terms don't have definitions:

noDefs

Also, some definitions don't fully make sense:

yCoord
others

"y-coordinate in the Cartesian coordinate system" makes sense as a definition (although it is self-referential), but just "in the Cartesian coordinate system" doesn't.

Also, should all definitions be included, including simpler ones like "acceleration"?

---

@bmaclach We totally would be able to only display the definition, but I'm not sure if that makes sense to do here when we have another section devoted to definitions. We would also have to add the aforementioned missing definitions and fix the incomplete ones.

bmaclach:
@samm82 I suppose I think of the Terminology and Definitions section as being there to define general concepts, which may or may not correspond to a symbol. For example in SSP I we have "effective forces and stresses" in the Terminology and Definitions section, but there are various symbols that define different effective forces and stresses more concretely. That's why I don't think it would be redundant to show only the definitions in the Table of Symbols.

Definitely we should update the definitions to make sense on their own. I think when I was working on SSP I was trying to make the text look like the manual version, which I did by defining the name and definition such that they make sense when read together, but looking back on it now, that wasn't the right way to do it.

bmaclach:
Thinking about it more, though, showing only the definition could be bad for another reason. For something like "Shear Strength", I think it's important to include the name "shear strength" in the table of symbols. But for other symbols the name is not enough (like with "Surface Angles", as @smiths originally noted). Which leads me to think that showing both name and definition, like was done prior to this PR, is the best option (but still I think the definitions should be updated such that they make sense on their own).

samm82:
I'm thinking I should close this PR. 😂 Who should be in charge of fixing the definitions? Is that something you should do @bmaclach?

bmaclach:
@samm82 If you're looking for something to do, you could do it. But if you have other things to work on I can also do it. I'm not sure how highly it should be prioritized -- @smiths may have thoughts about that (and about what should be included in the table of symbols -- names, definitions, or both. He may not agree with what I eventually landed on).

samm82:
I think I'll do it then, but I'll wait for confirmation from @smiths 👍

smiths:
The first goal is to not lose information.  When I mentioned the original documentation, I meant the before the pull-request version of the Drasil generated documentation.  The new documentation is the term I used for the pull request version.  Some information, like the clear definition of surface angles, is lost in going from the original (generated doc) to the new (pull request doc).  I thought we used the definitions elsewhere, but it appears that the definitions only appear in the table of symbols.

We definitely want to have names and definitions, and we want the definitions to stand on their own.  We also need to give more thought to how we use the definitions.  It appears that we don't really have a standard way of using the definitions.  As @bmaclach has highlighted the terminology and definitions section isn't about providing a definition for every symbol.  

Thinking about it now, I feel like the definition should be used the first time the symbol is introduced.  For instance, the definition of surface angle would be used when presenting the figure that describes a slice of soil.

I think in the short-term, we should leave the ssp example as it is, with the definitions in the table of symbols.  We don't want to loose information.

We should also try to understand how we intend to use names and definitions.  @samm82, how are definitions used in Projectile?  Are they used at all?  We can discuss this further during our meeting on Thursday.

smiths:
I agree with @JacquesCarette.  We don't want to lose information, but we also don't want to lose good work either.  :smile:

Issue #1592, opened 2019-06-20
Add constants to Drasil-Data


oluowoj:
We currently do not have constant values stored in Drasil. This needs to be added

bmaclach:
The way we handle constants in the examples currently is by constructing them with `mkQuantDef` which is a constructor for `QDefinition`s (I said in the meeting that it was a constructor for `DataDefinition`s -- I remembered that wrong).

`QDefinition`s are defined as:
https://github.com/JacquesCarette/Drasil/blob/5e3356ef40baaad3da39b593c51c95fc46cdaa57/code/drasil-lang/Language/Drasil/Chunk/Eq.hs#L20-L21
Which seems like basically what we want. The `Expr` for constants would simply be their value. (Thoughts @Mornix @JacquesCarette)?

samm82:
I personally like this, so that we can define constants for percentages (as `Perc` is a constructor of `Expr`).

JacquesCarette:
Yes, that is indeed what we want. Good!

samm82:
Assuming you're still working on this @oluowoj, the gravitational constant (defined in `Projectile.Unitals`, I'm pretty sure) should be pulled out to `drasil-data`, and its definition updated to "the approximate acceleration due to gravity on Earth at sea level".

JacquesCarette:
Any update on this? It seems this should have been done and closed.

JacquesCarette:
Is this done @samm82, @bmaclach ? It looks like it is.

bmaclach:
@samm82 and I added the constant value for pi to get projectile code generation working, but I think @oluowoj needed different constant values (gravitational constant, acceleration due to gravity) which are not yet done.

JacquesCarette:
Indeed - but now we know how to do it, so let's hurry up and get it done, so that the oldest outstanding PR can be merged!

oluowoj:
on it now.

JacquesCarette:
@oluowoj has everything about this issue been dealt with?

oluowoj:
Checking the SRS now, realized that g (gravitational acceleration) is missing from the table, that's the only item left for this to be completed.

JacquesCarette:
Ok, thanks for checking and documenting.

smiths:
The original issue was to add constants to Drasil-Data.  Hasn't that been done?  When I look at this file:

[...]

Issue #1395, opened 2019-05-23
Projectile Code Review


samm82:
As I mentioned on the [Projectile wiki page](https://github.com/JacquesCarette/Drasil/wiki/Projectile-Example), I used the generated GlassBR code as a guide to develop source code for Projectile (found in the [src/](https://github.com/JacquesCarette/Drasil/tree/projectile/code/drasil-example/Drasil/Projectile/src) folder in Projectile). Let me know if there are any issues you find.

This was just a first pass, so in the future, (at least) the following will be variable:
- Input Format
- Output Format
- Units
- Planet (Gravitational Constant)
- Criteria for Hit

I think this is mostly a question for @smiths, but would automating the test cases (ie. with pytest) be a good idea at this point?

bmaclach:
The code looks good to me. Note that what you have currently is great as one "version" of the code. When you say that, in the future, Input Format etc. will be variable, you should do that by writing an entirely new "version" of the program, rather than by updating your existing program. What we want is a family of similar programs. So you could write another "version" of the program where, for example, the output format is a graph, and yet another "version" where the user of the program decides at run time whether they want graph or textual output (thus changing the decision on output format from compile time to run time).

JacquesCarette:
Most definitely adding test cases and automating them is good!  And I agree with all of @bmaclach 's comments as well.

smiths:
I agree with @bmaclach and @JacquesCarette.  I definitely like the idea of automating test cases, but I don't think we currently have the infrastructure for that.  We should definitely aim for this though.  It would be great to have an example (even a small example) that does everything.  😄

samm82:
Would the end result of this example involve building each variant of the program when running `make`, with each one being in the stable for Projectile, or would we choose a default to build with `make`, and passing in params would change the built variant (and the location to look for the corresponding stable version)?

Mornix:
I would assume it would be similar to how SWHS and NoPCM are split at the moment. Where each variant would have its own “example” and respective stable and build folders so the entire SRS, ???, and code “chain” is specialized for the given variant. 

@smiths and @JacquesCarette may have a different idea as i wasn’t around for the SWHS, NoPCM “split” so I’m not sure if that was a conscious decision made to organize it that way.

samm82:
[...]

PR #1664, opened 2019-07-04
Projectile Code Generation

samm82:
Not sure if this is ready to be merged, but I think PRs are a good place to have discussions about changes. @bmaclach and I got the code generation for Projectile working, and a few things came up.

1) Related to #1592, we had to declare a `piConst`, that might have to be removed/changed in the future. Not sure what the progress on #1592 is @oluowoj, but just something to be aware of.
   1) I think we'd want to use the value of pi in the different languages instead of the hardcoded value, which could be a separate issue.
   2) If we decide to keep the hardcoded value, the value of pi shows up in the Values of Auxiliary Constants table. I'm assuming this would be desired if we don't go with (1i), but I'm also guessing we'd want to do (1i). @smiths 
2) As far as I can tell, `defSequence` isn't used in code or document generation. @bmaclach said that it might be deprecated. Does it do anything, and if not, can it be removed? @Mornix @JacquesCarette 
3) I don't have C# or Java on my computer, so @bmaclach said he would test those to make sure they work. Python and C++ are fine though. 👌
**EDIT:** The Travis CI build shows errors with both:

| Language | Error
|---|---
| **C#** | `Calculations.cs(20,26): error CS0161: 'Calculations.func_s(InputParameters, double)': not all code paths return a value`
**Java** | `./Projectile/Calculations.java:35: error: missing return statement`
4) I'm guessing #1395 can be closed with the merge of this PR? @smiths


bmaclach:
Java and C# are failing to compile the generated function `func_s`:
```
public static String func_s(InputParameters inParams, double d_offset) throws Exception {
        if ((Math.abs((d_offset / inParams.p_target)) < 2.0e-2)) {
            return "The target was hit.";
        }
        else if ((d_offset < 0)) {
            return "The projectile fell short.";
        }
        else if ((d_offset > 0)) {
            return "The projectile went long.";
        }
    }
```
They think that there's a path through the code that does not have a `return` statement. We will have to change the way we translate `Case` `Expr`s to GOOL. We could replace the last `else if` branch with `else`, or add a 4th `else` branch with some dummy return, or initialize a variable with some placeholder value, then assign to the variable in each branch of the `if`, then return the variable. I prefer the first option because it doesn't require any dummy or placeholder assignments, but it assumes that every `Case` `Expr` in Drasil is set up properly so that the last case is equivalent to an "else" statement. Thoughts @smiths @JacquesCarette?

samm82:
I like the first option. We could even enforce how cases are built in Drasil to take the cases, and then a default expression, to make it explicit.

bmaclach:
@samm82 I agree that we should enforce it in Drasil. In the SRS, would we be okay with showing the condition for the third case as "true", or would we still want to show "d_offset < 0"? (More of a question for @smiths). If we still want to show "d_offset < 0",  then the expression would still need to carry a condition for the default case, which seems a little weird but might be fine.

It would also be cool if we could check somehow if the cases are exhaustive, and throw an error if not. That seems like it would be very complicated, though.

Mornix:
It may think there is an additional path due to NaN.

bmaclach:
@Mornix Good point. I tested adding a condition to catch NaN but still got the error (in both languages), so I don't think that's it.

Mornix:
Another question, is this exclusive to FP or a general language rule (i.e does it enforce an `else` even if the cases are absolutely exhaustive).
If you try something along the lines of
```csharp
int a = 0;
if (a > 0) { return false;}
else if (a < 0) {return false;}
else if (a == 0) {return true;}
```
(Note the primitive `int`)
Do the compiler messages still appear?

bmaclach:
@Mornix I replaced the body of `func_s` with the block of code you posted and still got the same compiler errors.

Mornix:
@bmaclach Thanks, I was curious if either language’s compiler would be able to “prove” `else if` branches are exhaustive but that does not appear to be the case.

Mornix:
Follow up: Although it compiled, I’d assume g++ would emit a warning about a missing return statement if `-Wall` was passed as an argument.

bmaclach:
@Mornix Just checked, C++ does indeed throw a warning with `-Wall`

Mornix:
Re: Point 2, indeed `defSequence` is unused in the backend. I'm not sure if it should be used or was ever used, but it looks like the examples which provide a value for the field which is not empty are:

GlassBR: where it is `qDefns`
SSP: where it is `[Parallel (qdFromDD (head dataDefns)) (map qdFromDD (tail dataDefns))]`
GamePhysics: where it is `blockQDefs`
NoPCM: where it is `[Parallel dd1HtFluxCQD []]`

I'm not quite sure what it's purpose is, even going through the history, @JacquesCarette may know though.

smiths:
With respect to the value for pi, we have choices.  We could hard-code it in the generated code wherever it appears, or we could define it with a constant in each case.  If we define it with a constant, the location of that definition is another variability.  Although it isn't that interesting a variability, I don't think this is something we should decide for all examples.  It seems like something we could leave as a design time variability.  This wouldn't apply to just pi, but to all symbolic constants.  I'm sure that @JacquesCarette will have a thought on this too.

With respect to the case statements, the simplest solution is to add an else statement at the end that returns an error message.  Something completely unexpected happened.  I like the ideas of proving the coverage of our cases, but that probably is a difficult problem in general, or the compilers would already be able to reason about it.  We don't need the catch-all error case in our SRS, since it isn't part of the requirements.  It is there in case we get the requirements wrong.

With respect to the question by @bmaclach at:

https://github.com/JacquesCarette/Drasil/pull/1664#issuecomment-508564308

I would rather keep the test for the value of d_offset than replace it with True.  I know it is logically the same, but the d_offset test matches the requirements and is easier to read by an external person.  I could be convinced otherwise, especially in the generated code, but being able to quickly scan the cases is a good format for human verification activities.

bmaclach:
@smiths @samm82 I pushed a fix for the case calculations, adding an else that throws an error, as @smiths suggested. Hopefully the build passes now!

smiths:
@samm82, how long did it take to get the code generation working (not counting the problems with the case statement)?

samm82:
A little under an hour! Most of it was just setting up Projectile to be able to generate code (like filling in fields of `SystemInformation`),

bmaclach:
We scheduled a meeting for 2 PM and had working generated code by 2:34. Also I think I was a minute or 2 late to the meeting 😛 

Also, a lot of that time was sitting and waiting for Drasil to build.

JacquesCarette:
I don't think a PR is the best place to discuss a bunch of issues... a bunch of issues is the place to do that!
Regarding `if-then-else`: this is a more complicated issue. For code generation, some languages enforce syntactically (i.e. via having an `else` clause) that a conditional is complete. Completeness of piecewise-defined functions is a semantic property that can be arbitrarily hard to verify -- even though all the ones in our current examples are trivial. What this means is that we need to know a priori whether our "cases" are complete or not; when they are, we can generate code straightforwardly by dropping the last condition and using `else`. When they are not, or not marked complete, we should generate code with an extra `else` that throws an error.
But please go through the whole thread and create (complete) issues for every separate problem that arose.
Note that I have no idea whether I should be trying to merge this PR or not!

samm82:
@bmaclach Is your fix satisfactory, or should we create a new issue for the else clauses?

bmaclach:
@samm82 Based on @JacquesCarette's comment, I think we should create an issue. Though I think my fix might be satisfactory for now, so that we can get projectile code generation merged in.

samm82:
I'll make an issue for that, and for the unicode translating that we mentioned yesterday.

bmaclach:
Sounds good, thanks!

PR #1666, opened 2019-07-04
GOOL Usability Improvements

bmaclach:
GOOL defines `Function`s, which are commonly-needed functions on a given value, like `listSize` and `listAccess`. Since a `Function` exists independently from the `Value` it acts on, it can't read the type or doc of that `Value`. In cases where it needs this information from the `Value`, the type or the `Value` itself must be passed as a parameter to the function for creating the `Function`. This leads to code duplication. For example, if you have some list variable `lst`, and some integer variable `i`, then to access the `i`th element of `lst` in GOOL you would have to do: `lst $. listAccess (listInnerType $ valueType lst) i` -- you annoyingly need to specify the `lst` variable twice.

This PR adds functions to GOOL for applying a `Function` to a `Value` to yield a new `Value`, all in one function, so that the properties of the `Value` can be passed to the `Function` without any additional effort by a user of GOOL. Thus, the list access described above is simplified to: `listAccess lst i`.

This PR depends on #1662, so that one should be merged in first.


JacquesCarette:
And I'll wait to merge, since #1662 is not merged.

PR #1670, opened 2019-07-05
More improvements for the logging option for codegen

bmaclach:
I noticed a couple more issues with the logging, fixed by this PR:
- For logging of assignments, log statements were only generated for assignments defined explicitly with the `assign` function in GOOL. But many other GOOL functions perform implicit assignments (ex. `varDecDef`, `listAppend`, `getFileInput`), and there was no logging when those functions were used. I updated the function for logging assignments to not be tied to the `assign` function. This allows the generator to log a variable's value whenever it wants, not just when it uses `assign`. So, I updated the generator to generate logs whenever a variable's value changes.
- Some log statements for assignments were missing a module name (i.e. the message would be "var x assigned 1 in module "). This is now fixed.

PR #1666 should be merged in first.


JacquesCarette:
Unfortunately, there's still a conflict, even though I tried to merge things in ways to avoid that!

bmaclach:
For the commits from the branches that were merged, the commit hashes in this PR don't match the hashes for those same commits in master. I talked to @Mornix about this and he said that choosing the "Create a merge commit" option when merging might work (correct me if I got that wrong, @Mornix).

JacquesCarette:
Rather than a rebase merge then?

bmaclach:
Yes, seems like rebases rewrite the commit hashes so git doesn't realize they are the same commits.

PR #1673, opened 2019-07-05
Remove unused variables when reading input

bmaclach:
There were some variable declarations being generated for input-reading that were only used for certain formats of input. When they were unused, the C# compiler was throwing warnings. This PR updates the generator to only declare these variables if they are necessary and will be used.

#1670 should be merged first.


bmaclach:
@smiths If you are just looking at stable, I suggest looking for the latest "Updated stable" commit in each PR, and checking the diff for that commit only. That way it shouldn't highlight changes from other PRs.

smiths:
Yes, that works @bmaclach.  I tend to use the tab at the top of PR, but your commit messages always make it clear where stable is updated.  I had a look and everything looks fine in stable.

PR #1674, opened 2019-07-05
Hyperlinked URLs in Citations in HTML

samm82:
1) I had to add "https://" to some of the references to get them to properly link to the webpages (instead of trying to access a file on my computer), which might be a design flaw.
2) The following reference in GlassBR is broken: https://pdfs.semanticscholar.org/cf57/27a59801086cbd3d14e587e09880561dbe22.pdf

_Closes #1667_


smiths:
The changes to the urls look fine to me.

Issue #1355, opened 2019-05-16
Strings: Primitive or Derived? (Notation Table)


samm82:
In the manual MIS Notation section, strings are listed as both primitive and derived data types, with two separate definitions:
![image](https://user-images.githubusercontent.com/35857611/57885268-8f394280-77f8-11e9-8076-d023d3b1d5cb.png)

I have two questions:
1) Which definition is preferred?
2) Are strings primitive or derived? (This is tricky since they are primitive in some languages and derived in others. Not sure what approach we should take in the MIS.)

smiths:
As we discussed elsewhere, we don't want to get distracted by this right now.  The MIS that you are working on is a good starting point for discussion of our design documentation, but I'm sure it will change over time.  Devoting too much effort to the boiler plate introduction wording isn't a good use of our time.  Having an infrastructure in place where we can include the same text in every example is good, but the specific content of that text can wait until we have a firmer idea of the structure of the MIS.

Let us delete all of this, and replace it with references to Ghezzi et al, Hoffman and Strooper and Gries and Schneider.

PR #1676, opened 2019-07-08
[Review] Projectile Design Documentation

samm82:
I'm not sure who else should be reviewing this @smiths, but I think @bmaclach might want to at some point, especially to make sure it matched the generated code. I think I'm almost done the Design Documentation for Projectile. Some things that need

1) A link to the full documentation and implementation of Projectile. Since we don't have a stable version, should this be removed, or should we link to the new Drasil website?
image
2) A uses hierarchy for Section 7 of the MG (Use Hierarchy Between Modules), which I want to wait on until everything else is confirmed to be correct.
image
3) Some comments left over from SWHS by @smiths - I'm not sure what should be done with these.
(MG)
image

_Closes #1648_


smiths:
@samm82, for your questions above, like links to other resources, let us hold off.  I don't want a perfect document; I just want an example that we can discuss.  The goal here is to help us determine if the MG/MIS documentation is what we want to pursue.  The GlassBR example and the SWHS examples have too much going on to easily judge the feasibility of this approach.  Projectile is nice because it is much smaller.

Don't worry about links between documents at this point.  You also don't need to worry about drawing the uses hierarchy, or uses relation.  If we decide to commit fully to the MG approach, these diagrams will be auto-generated.  The same goes for the traceability matrix.  My comment about tracing to the code will be relevant if we document an MG.  Which code file corresponds to which module is relevant for making changes and for traceability.  For now though, we won't worry about it.

I have only one change to request for the MG.  I don't believe that lists or arrays are used at all in Projectile, so the software decision hiding module for sequences should be removed.

For the MIS, I suggest the following changes:

- do the input constraint checking inside of `get_input`
- `func_s` is not a very descriptive name, please change it to something more meaningful
- Is there a reason to have `func` as a prefix for every function?  It seems redundant to me.
- The Calculations module uses Specification Parameters for g and epsilon

I already see a potential problem with our MIS.  We have assumed `Bundled`, but the code generator by @bmaclach also allows for `Unbundled`.  We might be too close to the code implementation.

If you can make the changes mentioned above @samm82, we can close this issue.  The goal here is to get a sense of what the MG/MIS documentation would look like, so that we can discuss it as a group.

I have another idea for the design documentation.  I'll create a separate issue so that we can discuss.

samm82:
Since strings aren't a primitive type (in some languages), they are defined as lists of characters. Wouldn't the list infrastructure be necessary to create strings? This came up partially in the old MIS issue #1355. Should the whole table be removed and replaced with "references to Ghezzi et al, Hoffman and Strooper and Gries and Schneider?"

The name of `func_s` and the prefix `func_` for the calculation functions came from the code, as I remember reading that the function names of the MIS should match those of the code. I'll make the other two changes for the MIS.

smiths:
I believe that strings are available in all of the languages that we target, but that is fine if you want to leave the sequence module in the MG.  However, to be technically complete, you should then add it to the MIS.  :-)

Sure, you can replace the whole table with the references you mentioned.  That will hold off @JacquesCarette getting distracted by the introduction to the MIS document.  😄  Don't worry much about the introduction though.  That isn't the point of this exercise.  We want to see what the documentation of the modules looks like.

JacquesCarette:
I'm kind of holding off on reviewing this one until @smiths has had his full turn.

smiths:
@JacquesCarette, I am okay with the design documentation for Projectile.  I think we can merge it into master, as a helpful example and discussion point.  There is work that could be done to improve the MG and MIS, but I do not currently think this is a good use of our time.  @samm82 developed the example to the point where I'm fairly sure that we don't want to generate the MIS or MG, at least not now.  The problem with the MG and MIS for Projectile is that they don't say much interesting.  They are just an abstract view of the code, but code itself is fairly easy to understand.  

The reason that Projectile is not very interesting is that we have "baked in" many of the design decisions for all of our examples, and for those that we haven't baked in, the MIS isn't where we specify them, although it would show them.

The biggest baked in design decision is that all of our code follows the "input-calculate-output" design pattern.  All of our examples follow this.  Many applications in scientific computing follow this.  We do have some variabilities on where we put the verification of the input between examples, but we have pretty much converged on the decision that this really belongs in the input module itself.  Having a separate module for verification isn't useful, since it only happens once, right after the inputs are read.

The design decisions that we make include Bundled versus Unbundled, Warnings versus Exceptions etc.  This is probably where we should put our efforts.  What are design decisions that we can make for the code.  We should also think about a way to document these design decisions, but I don't currently think that an MG/MIS is the right place.  When it comes to documenting the variation points, we should probably be looking at program family (software product line) documentation for ways to do this.  Maybe we want a feature tree?

At some point our design decisions might be so involved that we want something like an MIS to provide an abstract view of the code.

While I'm typing about design, I think we will need further documentation between the SRS and the code for examples where algorithm and data structure choices are being made.  My current thinking is that when we have computing knowledge (like the algorithms for the ODE solver), we should document this knowledge pretty much the same as we document the physics knowledge.  That is, I think we will have a version of the SRS that documents the computing  knowledge.  In the case of algorithms though, we might want to think more deeply about how we show the variation points.

JacquesCarette:
Ok, in it goes!

PR #1680, opened 2019-07-08
Doxygen-style comments in GOOL

bmaclach:
This PR adds functions to GOOL for generating files/classes/methods with doxygen-style comments.

#1666 should be merged in first.


smiths:
@bmaclach, why not just generate actual doxygen comments?  That would allow running doxygen on the files to produce actual doxygen generated web-pages, pdf files, etc.  I do like the additional print messages; they definitely improve the code.  

Is it possible to turn this printing documentation information to files off and on?  That is, is this another example of a design time variability?

bmaclach:
@smiths I have a sort of chain of dependent PRs open right now, which I think is causing some confusion. This PR depends on some others that have not been merged in yet, so GitHub is showing the commits from those other PRs as part of this one. The print statements you see added to stable are actually from previous PRs (and yes, they are a variability that can be turned on and off).

This PR does focus on generating actual Doxygen comments, but you don't see those changes in stable because this PR just adds the functions for generating these comments to GOOL -- actually using them in Drasil will be the focus of my next PR.

smiths:
Okay.  I understand @bmaclach.  I also mis-interpreted Doxygen-style comments.  I thought that meant that your comments were inspired by Doxygen, but did not actually follow it.  I will leave reviewing the changes to the Drasil code to @JacquesCarette, but I'm looking forward to the upcoming PR with doxygen comments added to the output.  I really like this idea.  It seems to be in the category of providing a significant return on investment.  If we can get doxygen comments working, then @Mornix and @danscime can add that output to the web-page too.  😄

PR #1690, opened 2019-07-10
Moved Data Constraints to Properties of Correct Solution

samm82:
- moved Data Constraints to Properties of Correct Solution
- made both sections take a list of quantities, with Data Constraints taking an extra `Sentence`, and Properties of Correct Solution taking a `[Contents]`

_Closes #1624_


JacquesCarette:
Code changes are good. Another master merge needed though. An explicit review by @smiths is probably best here, though it definitely seems that this was done with his knowledge of the details.

JacquesCarette:
Oh look, when I did a merge of something else, that resolve the previously identified conflict.

smiths:
Yes, @JacquesCarette, the movement of the output constraints to the "properties of a correct solution" section was my idea.  That is really where I should have put them from the start.  😄 I've looked over the changes to stable and it looks like what I was hoping to see.

Issue #444, opened 2017-12-06
Implementation choice for generating code comments


palmerst:
We want a choice for generating code comments (and sub-choices for specifics on what should be commented).  The commenting features of GOOL will need an overhaul to do this right.  Currently comments in GOOL are implemented as part of the Statement syntax, so they can only be put with/between statements.  There is no way to make comment blocks for modules/classes/functions.

bmaclach:
On the GOOL side of things, there are currently two ways to do comments. `comment` for single-line comments, represented as Statements, and `addComments` for adding a single-line comment above and below a Body of code, as a sort of delimiter with a short description. We can easily add a `blockComment` function. I am thinking this function should reside in a new typeclass (thinking from the perspective of finally tagless GOOL). Then we can add to the `MethodSym`, `ClassSym`, and `ModuleSym` typeclasses functions like `methodWithComments`, `classWithComments`, `moduleWithComments` which are parameterized by a `BlockComment` in addition to all of their regular parameters. Our renderers would then display this comment directly above the corresponding method/class/module. The reason for not adding `blockComment` to the `BlockSym` typeclass is that we would then need to parameterize functions for creating classes and modules with a Block, which could easily be abused (by putting actual code in the Block instead of a block comment).

On the Drasil side of things, our `Choices` data structure currently holds a `Comments`, where `Comments` is defined by:
```
data Comments = CommentNone
              | CommentFunc
```
Currently this option is just ignored when actually generating the code. My thinking is that a user should be able to decide whether they want functions, classes, or modules to be commented, or any combination thereof. So I would update `Choices` to take a list of `Comments`. Then define `Comments` as:
```
data Comments = CommentFunc | CommentClass | CommentModule
```
Where the `CommentNone` option is removed because it would simply be represented as the empty list. Then, any time we are generating a function, class, or module, we check whether the corresponding `Comments` data constructor is present in the list in the `Choices` structure, and generate a block comment if so. 

Another option is to remove `Comments` from the `Choices` structure altogether. With it in `Choices`, it is a "global" choice, which means the user can't decide to include comments only on one module that they feel warrants an explanation, and not others. To give them this flexibility, we could redefine `Comments` as:
```
data Comments = NoComments | WithComments
```
[...]

PR #1696, opened 2019-07-11
Implementing the choice for comments

bmaclach:
This PR implements the choice for turning comments on or off in the generated code, closing #444. The comments can be turned on for functions, classes, or modules (or any combination thereof). The comments are in doxygen-style, so they can be compiled into doxygen documentation.

You'll notice that many of the doxygen comments for parameters in stable say "No description given", which happens when the parameter comes from something that does not have a description in Drasil. I plan on adding descriptions for everything in a future PR, but wanted to focus on just getting the comments generated and working in this PR.

Currently I've turned on comments for GlassBR and kept them off for NoPCM. @smiths, given your idea to add links to generated doxygen docs to our website, should I turn the comments on for both examples?

#1693 should be merged in first.


JacquesCarette:
Needs a merge, but is otherwise good to go.

Issue #1688, opened 2019-07-10
Allow for "internal" GOOL functions that are not exposed


bmaclach:
Some low-level functions in GOOL should not be exposed, and instead should just be used internally in higher-level GOOL functions. Some examples of such low-level functions are:
- `cast`
- Any function in the `FunctionSym` and `SelectorFunction` typeclasses ending in `Func`
- `printSt`
- potentially others

To do this, functions that should not be exposed should not be in the same typeclasses as functions that should be exposed. Then, only the typeclasses with "external" functions should be exported. However, @JacquesCarette notes:

"We might also need to play around with separate 'smart constructors', as Haskell makes it hard to not export classes."

(This issue based on discussion in #1662)

PR #1707, opened 2019-07-12
Remove uses of cast outside of GOOL

bmaclach:
This PR is a first step towards #1688 by removing external uses of GOOL's `cast` function (it is now used only internally in GOOL renderers).

I removed the need for using `cast` outside GOOL by adding some higher-level functions in GOOL to do the tasks that needed casts. In doing this, I made some simplifications to our `DataDesc` type, which is our language for defining formats for input data. With these simplifications, there are some input formats that we no longer support:

- If a line contains inputs, none of the line can be "junk" (though we can still have junk lines)
- If a line contains a repeated pattern, the values on that line must all be entries for a list variable. (This definitely makes sense because if they were values for a non-list variable, they would just be overwritten during the read of the next "repeat" of the pattern)
- If a line does not contain a repeated pattern, the values must be for non-list variables (again, this seems fine because if you are filling in a list, a repeated pattern seems like the logical way of doing that)

If we want to support any of these possible formats, let me know and I'll create an issue for adding that support.


JacquesCarette:
So now I see on this PR your explanation of what was removed. Comments on each (same order):
- but some of that input line can be a "comment" ?
- I'd like a bit more documentation of the pattern language; but I think this does make sense
- ditto
I'm going to re-review the code with that in mind.

bmaclach:
In response to your first point, no, comments wouldn't be allowed unless the entire line is a comment.

PR #1709, opened 2019-07-12
Unicode Symbol Conversion

samm82:
Dependent on #1664 

`Implementation` `Symbol`s are now automatically generated by replacing unicode with the unicode symbol name (ie. "θ" -> "theta")

_Closes #1705_


bmaclach:
@samm82 Do you mean `Implementation` `Symbol`s? `Equational` is the SRS stage and `Implementation` is the code stage, so it should be the `Implementation` symbols that get replaced.

samm82:
Yup 😬 - fixed it 👍

smiths:
@JacquesCarette raises some good points.  It is great to automate, but the user still needs to be involved, so that we don't have name clashes.

JacquesCarette:
So I still see calls to `autoStage`. Is this complete?

samm82:
`autoStage` is just a helper function for translating the symbol to the unicode variant for the code generation. It isn't called by default; only by the user's choice. Is this OK?

JacquesCarette:
Yes - as long as it is very well document as to what it actually does.

code/drasil-lang/Language/Drasil/Chunk/DefinedQuantity.hs
```diff
@@ -35,10 +35,13 @@ instance HasSymbol     DefinedQuantityDict where symbol = view symb
 instance Quantity      DefinedQuantityDict where 
 instance MayHaveUnit   DefinedQuantityDict where getUnit = view unit'
 
--- For when the symbol is constant through stages
+-- For when the symbol is constant through stages (maps unicode to strings for code gen)
```
JacquesCarette: Doesn't this mean that the comment here is obsolete?

code/drasil-lang/Language/Drasil/Chunk/Unitary.hs
```diff
@@ -41,12 +40,6 @@ unitary :: (IsUnit u) =>
 unitary i t s u space = UC (mkQuant i t s space (Just uu) Nothing) uu -- Unit doesn't have a unitDefn, so [] is passed in
   where uu = unitWrapper u
 
--- | For when the symbol changes depending on the stage
-unitary' :: (IsUnit u) =>
```
JacquesCarette: Doesn't the removal of `unitary'` now make it impossible for a user to hand-stage a unitary?

samm82:
It used to be "angle" because @bmaclach and I manually defined the code symbol to be "angle" to get the code generated. In a previous meeting, @smiths said that "θ" should map to "theta" in the automatic way of Unicode conversion.

JacquesCarette:
Right, I guessed that that was the path-through-history. I'm asking whether we should revisit that one more time. [Or maybe another one like it].

samm82:
That would be a question for @smiths - he seemed to like the idea of translating each symbol to a text-only representation of the same symbol

code/drasil-lang/Language/Drasil/Chunk/Unitary.hs
samm82: You're right - fixed in 3e8bf7863

smiths:
As far as theta versus angle, I mentioned that in many scientific codes, the code authors use theta.  It isn't a bad name.  However, I do not have a strong feeling that theta is better than angle.  It just seemed like an opportunity for automation.  Having said that, I'm fine if we want to abandon this issue.

PR #1710, opened 2019-07-13
GameGS: Modified GS opening statement.

oluowoj:
@smiths, let me know if you are still good with this opening statement for the goal statement section(its the new update I made)

### DIFF
--- a/code/stable/gamephys/SRS/Chipmunk_SRS.tex
+++ b/code/stable/gamephys/SRS/Chipmunk_SRS.tex
@@ -304,7 +304,7 @@ This subsection provides a list of terms that are used in the subsequent section
 \end{itemize}
 \subsubsection{Goal Statements}
 \label{Sec:GoalStmt}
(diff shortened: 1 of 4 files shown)


code/stable/gamephys/SRS/Chipmunk_SRS.tex
```diff
@@ -304,7 +304,7 @@ \subsubsection{Terminology and Definitions}
 \end{itemize}
 \subsubsection{Goal Statements}
 \label{Sec:GoalStmt}
-Given the inputs, the goal statements are:
+Given the kinematic properties, and forces including any (collision forces) applied on a set of rigid bodies, the goal statements are:
```
smiths: Let us change `kinematic properties` to `current kinematic state`.  We should also add mass and gravitational field to the list of inputs.  These will all be nicely refined by the instance models.

smiths:
I've suggested a change to the introduction to the goal statements.

PR #1714, opened 2019-07-16
Add Variables to GOOL

### LINKED ISSUE #1689: Add "Variables" to GOOL
Currently GOOL has a single type `Value` used for both values and variables. This means that any value can be passed in places where really what is wanted is a variable. For example, a `Parameter` can be made from an expression (which is a `Value`), which doesn't make sense -- a `Parameter` is just a variable. We should therefore separate `Value` into two types: `Variable` and `Value`.

A `Variable` would be equivalent to some `VarData` type that holds a (string) name and a type.
A `Value` would be equivalent to the current `ValData` type, except the `valName` field can be removed.

Then, `Variable`s should be used instead of `Value`s where appropriate:
- `Parameter`s should be made from `Variable`s, not `Value`s
- Assignments should take a `Variable` (being assigned to) and a `Value` (being assigned), instead of 2 `Value`s
- Every instance of a `Value` being passed in GOOL should be evaluated to determine whether it should instead be a `Variable`. 

(This issue based on discussion in #1680)

bmaclach:
This PR closes #1689 by adding a `Variable` type to GOOL, and updating functions that previously worked on `Value`s to instead work on `Variable`s, where appropriate.

I haven't introduced lvalues yet, though I think I see where they will be useful. I'll create a separate issue for that.


JacquesCarette:
So this code is great, with the exception of the thing about `switch` and `Eq`. Once those are addressed, this can be merged.

JacquesCarette:
Hmm, I would have merged this in, pre-`$!` changes... maybe revert that while we figure out a reasonable syntax, and then make a new PR for that part?

bmaclach:
Sounds good, reverted.

JacquesCarette:
Let's stick with `valueOf` for now.

PR #1721, opened 2019-07-16
[Small PR] Fixed Line Spacing in LaTeX

samm82:
Accomplished by adding a new line between any two consecutive `Paragraph`s.

_Closes #1691_


code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -181,33 +180,31 @@ sec :: Int -> D -> D
...
 ddref   = comm "ddref"           "MG\\ref{#1}"       (Just "1")
 colAw   = comm "colAwidth"       "0.2\\textwidth"    Nothing
 colBw   = comm "colBwidth"       "0.73\\textwidth"   Nothing
-
-modnum    = comm "mthemodnum"        "M\\themodnum"        Nothing
+modnum  = comm "mthemodnum"      "M\\themodnum"      Nothing
 
 arrayS  = renewcomm "arraystretch" "1.2"
 
+newline = pure $ text ""
```
JacquesCarette: So this seems wrong now. I'm guessing that this doesn't act as a newline anymore!  We likely just want to remove calls to this fake `newline`.

code/drasil-printers/Language/Drasil/TeX/Print.hs
```diff
@@ -56,7 +56,7 @@ buildStd sm (Document t a c) =
 lo :: LayoutObj -> PrintingInformation -> D
 lo (Header d t l)       _  = sec d (spec t) %% label (spec l)
 lo (HDiv _ con _)       sm = print sm con -- FIXME ignoring 2 arguments?
-lo (Paragraph contents) _  = toText $ spec contents
+lo (Paragraph contents) _  = toText (spec contents) %% newline
```
JacquesCarette: I'm guessing it's the `%%` which adds the newline, not the `newline` itself. This is going to be confusing for others!

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: This is actually what makes this PR work. Before, I added a `Paragraph (S "")` in between consecutive `Paragraph`s, which generated an empty line in the LaTeX, leading to a new line in the generated PDF. This is essentially doing the same thing. As per your other comment, yes, in essence the `%%` is adding a newline, because it concatenates two `D`s with a newline character, and the second `D` is an empty line. As mentioned before, I tried using `empty`, but that is the identity element, so it doesn't actually do anything. Is there an alternative that would be less ambiguous?

JacquesCarette: I think there is a `newline` type of command in the HughesPJ package, that would be the right way to do this.

samm82: There isn't a `newline`-esque command in the `Text.PrettyPrint` package. The closest thing is `($+$)`, which adds two `Doc`s together on separate lines. This gives us one new line, but we need another empty line to actually have the PDF have a new line.

Related question; why do we have both of these functions? I've tested it out, and it seems that they are completely interchangeable:
```haskell
(%%) :: D -> D -> D
(%%) = mappend

($+$) :: D -> D -> D
($+$) = liftA2 (TP.$+$)
```

JacquesCarette: First part: Ok.  But we still need to do things differently, because 
```haskell
pure $ text ""
```
isn't a "newline", it just happens to work after a `(%%)`.  If the type was more like
```haskell
-- add newline
newline :: D -> D
newline s = s %% (pure $ text "")
```
then that couldn't be mis-used.

Second part: no idea! If `mappend` in `Text.PrettyPrint` is defined as `$+$` (or the other way around), then indeed they are equivalent, and one should be removed.  But I need to know with more certainty first.

samm82: I revised `newline` in 3dd9220388b5e45a204b81282f37ea9c7713ea0a.

Do you think the `(%%)` vs. `($+$)` discussion is worth having an issue for and discussing it there?

code/drasil-printers/Language/Drasil/TeX/Print.hs
```diff
@@ -324,8 +324,8 @@ makeDefn _  [] _ = error "Empty definition"
 makeDefn sm ps l = beginDefn %% makeDefTable sm ps l %% endDefn
 
 beginDefn :: D
-beginDefn = newline
-  %% pure (text "\\noindent \\begin{minipage}{\\textwidth}")
+beginDefn = commandD "vspace" (command0 "baselineskip") %% command0 "noindent" %%
```
JacquesCarette: So why was this vspace added?

samm82: The vspace was added to add a newline before the `minipage` environment, as per [this StackOverflow thread](https://tex.stackexchange.com/questions/136612/space-just-before-minipage). For reasons unknown to me, newlines aren't added before the minipage environment, so the `vspace` does what the `newline` function does otherwise.

Now that I'm thinking about this, I think we should have a `minipage` function in `Tex.Helpers` that automatically adds the `vspace`.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: I looked into it, and found the key difference between the two functions. Issue created at #1758.

code/drasil-printers/Language/Drasil/TeX/Print.hs
JacquesCarette: Yes, abstracting that out is a good idea.

JacquesCarette: Can be done in a new PR, as I'll pull this one in now.

samm82: Done in 8d45d8266

PR #1724, opened 2019-07-17
PR - Web icon fix

### LINKED ISSUE #1633: Icon for website
Do we want an icon for our site?

I have made the change required for adding an icon in d429c515b16bf940a4b0e1632becb2c3d4bc2809 as a template.

Screen Shot 2019-06-27 at 9 45 35 AM

### LINKED ISSUE #1712: Website icon not displaying
On the webpage, the tree icon is not displaying.
https://jacquescarette.github.io/Drasil/

It seems that the image icon is stored in a different directory than it should be.

danscime:
Closes #1712. Closes #1633.
Website icon was not displaying due to some problems with the way the images folder was being copied.


danscime:
Okay so 2 problems, the first being that the way I had my copy statement, the images folder was being copied into an images folder, hence `images/images`.

Second problem being that my icon does not seem to be under version control, I will look into why.

Once the image is pushed, I like the new way I have the `copy_images` function, this way we remove all images then regenerate them, that way, no old images will be stored.

JacquesCarette:
`images/images` is definitely bad. Regenerating everything does seem like the Drasil way to do things.

danscime:
Upon further inspection, it could possibly be due to different systems treating
`cp -r "$CUR_DIR"website/images/ "$CUR_DIR"deploy/images`
like
`cp -r "$CUR_DIR"website/images "$CUR_DIR"deploy/images`

the first images has a `/` afterwards, which could be causing a discrepancy between how my computer compiles vs. whatever system the CI uses.

Either way, I think my newer version is better as well.

danscime:
Weird that the image wasn't being tracked online, but on my system, it said it was.
Github shows the deletion, but it never showed the original in master.
Very weird.

The new image seems to be appearing.

danscime:
Will close #1633.

Mornix:
@danscime's conclusions are generally correct. The way the `copy_images` was written the directory was being copied inside another `images` directory leading to `images/images/...`. The reason the image itself wasn't being copied, was the issue that seems to pop up occasionally relating to case-sensitivity of different filesystems. That was the case here. In the [website folder](https://github.com/JacquesCarette/Drasil/tree/4528e776df7bc7133e8046640ebca8280681222e/code/website) there are two directories named `image`, one with a capital `I` and one without. This PR should fix both of the issues relating to the absence of a favicon.

JacquesCarette:
Ok, I'll take this as a positive code review -- merged.

PR #1728, opened 2019-07-18
Generate Doxygen Configs

### LINKED ISSUE #1726: Generate Doxygen config file
Drasil should generate a Doxygen config file for examples that have doxygen comments enabled in the generated code. The file output of running `doxygen -g` in the command line can be used as a starting point for our template. The `PROJECT_NAME` and `INPUT` fields of the config will have to change between examples.

bmaclach:
This PR closes #1726. Doxygen config files now get generated if doxygen comments are enabled for the example. The config is the default config with the exception of 3 fields:
- PROJECT_NAME is updated with the program name
- INPUT is updated with the list of source files
- OPTIMIZE_OUTPUT_JAVA is set to YES for Java and Python, as suggested by the comments preceding that field in the config


smiths:
@bmaclach - great to see the doxygen comments in the repo, along with the doxConfig file.  I do have a question about the doxConfig file.  I see that it is essentially hard-coded into the Import.hs file.  I haven't thought about it deeply, but I thought `doxygen -g` would be called when the code is generated, and then the results would be modified (or maybe parameters can be provided to doxygen configuration creation)?  This explicit calling of doxygen might help as versions of doxygen change, but, of course, it would be a problem if the user did not have doxygen installed.  Actually, I could see doxygen versions being a problem, since we do not say what version of doxygen we are using.  We also don't say which version of each programming language.  This kind of information should probably be available in the future.

I'm probably fine with the approach that you use for generating the doxConfig file, but I would feel better if you explained the rationale for this approach.  To be honest, it just seems odd to me, but @JacquesCarette mentioned that we did something similar with the css files, so this could be a very standard approach.

bmaclach:
@smiths I was actually thinking the same thing - to generate the config using `doxygen -g` and then edit it as part of the build. But I asked @JacquesCarette about it after the meeting and he said we shouldn't do that, and should instead do it this way (the way I did it in this PR). He noted that it means we might get out of sync with Doxygen, but he was okay with that.

I didn't ask him for his rationale at the time, though I wished I had afterward. So thanks for bringing it up!

smiths:
@bmaclach if @JacquesCarette said this is the way to do it, I'm good with that.  However, I'm also curious about the rationale.  😄

JacquesCarette:
This way we can generate a doxygen config file even if the user's environment is wonky and their doxygen is broken. This means that Drasil will "succeed more".

But most importantly, editing the doxygen-generated template is quite fragile: if they make some changes to their template, we might fail to edit it properly.

JacquesCarette:
[Also, this needs a master merge]

smiths:
Your explanation makes sense.  Thank you @JacquesCarette.

bmaclach:
Reopening this PR as I have added infrastructure for generating non-source code files as part of a GOOL `Package` and used this to generate the doxygen config file for each language.

PR #1741, opened 2019-07-22
Update Case expressions with Completeness

### LINKED ISSUE #1704: Revisit Else Case in Cases
As mentioned in #1664, the current fix for else cases in case statements might want to be revised.
The relevant proposals have been pasted below (I think I got all of it?)

>We could replace the last else if branch with else, or add a 4th else branch with some dummy return, or initialize a variable with some placeholder value, then assign to the variable in each branch of the if, then return the variable. I prefer the first option because it doesn't require any dummy or placeholder assignments, but it assumes that every Case Expr in Drasil is set up properly so that the last case is equivalent to an "else" statement.
@bmaclach 

> With respect to the case statements, the simplest solution is to add an else statement at the end that returns an error message.
@smiths 

>Completeness of piecewise-defined functions is a semantic property that can be arbitrarily hard to verify -- even though all the ones in our current examples are trivial. What this means is that we need to know a priori whether our "cases" are complete or not; when they are, we can generate code straightforwardly by dropping the last condition and using else. When they are not, or not marked complete, we should generate code with an extra else that throws an error.
@JacquesCarette

bmaclach:
This PR closes #1704 by tagging `Case` expressions with either `Complete` or `Incomplete`, which is used by the code generator to either generate an `else` that throws an error (if `Incomplete`) or replace the final case with an `else` (if `Complete`).


bmaclach:
I'm guessing the failed push build here was due to Github going down briefly... The PR build passed so I think this should be fine to merge.

PR #1742, opened 2019-07-22
Fixed incorrect case in SSP

bmaclach:
I noticed a small mistake in SSP with a case expression, which this PR fixes.


smiths:
Yes, this change is correct.  Good catch @bmaclach.

PR #1746, opened 2019-07-23
replaced nonNegativeConstraint function with generic gtZeroConstr fxn

oluowoj:
Used generic function to display physical constraints


JacquesCarette:
This changes some constraints from `>= 0` to `> 0`, as well as using a generic combinator instead of a custom one (good).  Is that change correct? @smiths ?

smiths:
I think I'm fine with the change to `> 0`.  In any real problem the physical quantities of length, mass and moment of inertia are not going to be exactly equal to zero.  If they are zero, the problem cannot be well defined.  Having said that, I still usually allow things to equal zero, just to see if the degenerate case falls apart properly.  I cannot think of any interesting test case for game physics though.  If we set I = 0, then that means that any torque we have an angular acceleration that approaches infinity.  We can't really run that simulation.  :-)

In thinking about it now, I'm okay with the `<0` constraint.  However, I do think there are times, for other degenerate cases in other problems, when we want to let the `=0` through, just to see what happens.

oluowoj:
My bad, I could use a new constructor to have the >= 0, if we need this

smiths:
For the examples given in this PR, I'm fine with >0.  The equal 0 case would often lead to divide by zero errors for this example.

oluowoj:
Okay.

PR #1749, opened 2019-07-23
Fixed type of for each variable in C# and java

bmaclach:
I noticed that the iterating variable in for each loops in C# and Java was being mistyped as a list instead of an element of the list. This PR fixes this typing.


JacquesCarette:
Good - but needs a merge.

PR #1750, opened 2019-07-23
Improved Variable Display

samm82:
Defined a `Variable` and `Label` type for symbols and used them in the examples

_Closes #1639_


smiths:
I'll respond to the places where I was atted above:

> So I'm wondering about 0, 1 and 2 as labels. They seem of a different nature - and perhaps shouldn't be displayed as text, but in math font instead -- @smiths ?

I'm not sure where the 0, 1 and 2 came up.  My instinct though is that these are labels.  They can't be variables, since they can't vary their values.  

> These are much better internal names. But why the change from `Sat` to `sat` ?

I'm not surprised that a consistency mistake was made.  All of the `saturated` labels should use the same short form, and case.  The inconsistency is a mistake.  Making them all `sat` makes sense to me.

JacquesCarette:
What I mean with the numbers is, it should likely be `$a_{1}$` in the output, and not `$a_{\text{1}}$`.  In that `a_1` is actually the "full name" of the variable (?), and not somehow `a` "of" `1` (the same way `$a_{\text{min}}$` is 'the minimum a').

smiths:
Yes, @JacquesCarette, I see what you are getting at.  If we called the variable a1, we wouldn't think twice about it being in italic font.  I'm confident that in 90%+ of papers, it would be $a_1$.  The other option seems odd.

I don't think we are going to find a completely unambiguous rule of symbols.  I don't think such a rule exists outside of Drasil.  After talking to some colleagues, it appears that people develop their own conventions.  We probably don't want to go too far down this "rabbit hole" at the moment.

Given that others would write $a_1$, I think we should do the same thing.

samm82:
Some smaller questions about using numbers in symbols:
1) Should we always wrap numbers in the curly braces, or only if the number is more than one digit?
2) Should we allow for negative numbers in symbols?
3) Should we not allow for numbers in superscripts (so that they don't get confused for exponents)?

smiths:
@samm82, I'll try to answer your questions below:

> Some smaller questions about using numbers in symbols:
> 
> 1. Should we always wrap numbers in the curly braces, or only if the number is more than one digit?

To a small extent I prefer to not include the curly braces if they aren't necessary, like the case where there is one digit (character).  It is just a bit easier for a human to read.  Given that humans will not routinely read the generated tex, I'm okay if this special case is too much work to handle.

> 2. Should we allow for negative numbers in symbols?

Yes, it is common to use the subscript as an index that can be negative.  At times the index is relative and the negative sign gives a direction.

> 3. Should we not allow for numbers in superscripts (so that they don't get confused for exponents)?

There are cases where people use numbers in superscripts where they don't mean exponents.  It isn't common, but I've seen it.  For instance, combination of super and sub scripts are used with contravariant and covariant tensors and vectors:

https://en.wikipedia.org/wiki/Covariance_and_contravariance_of_vectors

You'll also see an example of a negative subscript on that web-page.  😄

JacquesCarette:
And this now needs a merge before it goes anywhere.

samm82:
Rendering the braces and the numbers for symbols occur at different stages, so at least for now I think it's smart to just wrap all numbers in braces and not get distracted by making the LaTeX the prettiest.

smiths:
Sounds reasonable to me.

JacquesCarette:
Another master merge needed. And is this otherwise ready to go?  I've lost track...

samm82:
Yup, this is ready to go!

PR #1753, opened 2019-07-23
Spec Length

samm82:
- Fixed `specLength`
- Minor improvements to LaTeX generation

_Closes #1717_


code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -46,14 +46,17 @@ commandD s c = pure (H.bslash TP.<> text s) <> br c
 command1o :: String -> Maybe String -> String -> D
 command1o s o c = pure $ (H.bslash TP.<> text s) TP.<> maybe TP.empty H.sq o TP.<> H.br c
 
--- no braces!
 command1oD :: String -> Maybe D -> D -> D
-command1oD s o c = pure (H.bslash TP.<> text s) <> maybe empty sq o <> c
+command1oD s o c = pure (H.bslash TP.<> text s) <> maybe empty sq o <> br c
```
JacquesCarette: So this function was explicitly documented to produce no braces -- you removed the comment and the braces. Why? [I'm not saying it is necessarily wrong, just that it needs justification]

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -105,26 +108,18 @@ genSec d
 -- For references
 ref, sref, hyperref, externalref, snref :: String -> D -> D
 sref         = if numberedSections then ref else hyperref
-ref      t   = custRef (t ++ "~\\ref")
-hyperref t x = command0 "hyperref" <> sq x <> br (pure (text (t ++ "~")) <> x)
+ref      t x = pure (text $ t ++ "~\\ref") <> br x
+hyperref t x = command1oD "hyperref" (Just x) (pure (text (t ++ "~")) <> x)
```
JacquesCarette: This use of `command1oD` (and the next) both seem a little silly, since they hard-code a `Just`. That function is really meant to be used when you have a dynamic `Maybe`. There ought to be a more direct combinator that does the same job that should be used here (and in `snref`).

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -157,7 +150,7 @@ item' bull = command1oD "item" (Just bull)
 maketitle, maketoc, newline, newpage, centering :: D
 maketitle = command0 "maketitle"
 maketoc   = command0 "tableofcontents"
-newline   = command0 "par" <> pure (text "~\n")
+newline   = command0 "par~\n"
```
JacquesCarette: So is this now obsolete?

code/drasil-printers/Language/Drasil/TeX/Print.hs
```diff
@@ -321,20 +326,13 @@ pUnit (UDiv n d) = toMath $
...
 makeDefn _  [] _ = error "Empty definition"
-makeDefn sm ps l = beginDefn %% makeDefTable sm ps l %% endDefn
-
-beginDefn :: D
-beginDefn = newline
-  %% pure (text "\\noindent \\begin{minipage}{\\textwidth}")
-
-endDefn :: D
-endDefn = pure (text "\\end{minipage}")
+makeDefn sm ps l = newline %% command0 "noindent" %% mkEnvArgs "minipage" "\\textwidth" (makeDefTable sm ps l)
```
JacquesCarette: I think I had asked for the `newline` here (or the `%%` after noindent) because of weird indenting inside the minipage environment; this doesn't seem to be the case anymore. So as long as what's use doesn't produce indent, there is no need for a newline after the noindent.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: The `newline` fix is in a separate PR, but I'll remove this one when the merge conflict between them arises. 😂

code/drasil-printers/Language/Drasil/TeX/Print.hs
samm82: Fixed in #1721- should I do something in this PR about it as well?

JacquesCarette: No. Though there might be interesting merge conflicts arising... we'll see.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: Fixed in dac98cdacbf2ba208917c12c07abd48f238b8d7b

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: It was only used in `item'`, and its counterpart `item` (the same without an extra argument) used the braces, so it made sense to make this outlier consistent with the rest of the family of `command` functions.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -44,14 +48,19 @@ commandD s c = pure (H.bslash TP.<> text s) <> br c
 
 -- 1-argument command, with optional argument
 command1o :: String -> Maybe String -> String -> D
-command1o s o c = pure $ (H.bslash TP.<> text s) TP.<> maybe TP.empty H.sq o TP.<> H.br c
+command1o s Nothing  = command s
```
JacquesCarette:
```haskell
command1o s = maybe (command s) (command1p s)
```
I do believe.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
```diff
@@ -44,14 +48,19 @@ commandD s c = pure (H.bslash TP.<> text s) <> br c
...
+command1o s (Just o) = command1p s o
 
 command1oD :: String -> Maybe D -> D -> D
-command1oD s o c = pure (H.bslash TP.<> text s) <> maybe empty sq o <> br c
+command1oD s Nothing  = commandD s
```
JacquesCarette: same idea here.

JacquesCarette:
and a big master merge too.

code/drasil-printers/Language/Drasil/TeX/Helpers.hs
samm82: f0536618d

samm82:
For some reason, my latest commit f0536618d isn't showing up in the PR.

Issue #1743, opened 2019-07-22
GOOL input/output functions should be input/output/both functions


bmaclach:
Based on discussion started here: https://github.com/JacquesCarette/Drasil/commit/532119745fdb49e1c588229009bd4eb425c0eb21#r34335773.

Currently the implementations for `inOutFunc` and `inOutCall` in GOOL in some languages check to see which inputs are also outputs. Instead, these functions should take an additional parameter of things that are both inputs and outputs. So the functions should take a list of inputs, a list of outputs, and a list of things that are both, removing the need to check for equality between the inputs and outputs (and allowing us to remove the derived `Eq` instance on `ValData`). `inOutFunc` should then check that the resulting generated parameters for the generated functions all have unique names, and throw an error if not.

bmaclach:
Closed by #1761

PR #1761, opened 2019-07-25
Remove derived Eq instances from underlying GOOL types

bmaclach:
This PR removes derived `Eq` instances from `ParamData`, `TypeData`, and `ValData`. For `ParamData` and `TypeData`, it replaces them with hand-made instances where the `Doc` field is not compared. For `ValData`, comparisons were only made in the functions `inOutFunc` and `inOutCall`, so these functions were updated so as not to need to compare `Value`s for equality, closing #1743.

As a bonus, these changes fixed an undiscovered bug where parameters were being left out of the Doxygen comments (you can see that they've now been added in the changes to stable).


JacquesCarette:
Needs yet another master merge.

Issue #1061, opened 2018-11-26
SWHS: Difference b/w Stable and Manual: R (Aspect Ratio)


smiths:
Issue #523 has far too much going on to keep track of all of the details. This issue (and several others) highlights one specific difference between Stable and Manual.

In the manual version for Drasil we use the aspect ratio R, but it is missing from the stable (Drasil) version.

Manual:
![image](https://user-images.githubusercontent.com/35268928/41620979-5a9e9c10-73d9-11e8-810b-ebca7cbac6de.png)

The aspect ratio (R) was discussed in smiths/caseStudies#26. Changes were made to the manual version, but not propagated to Drasil. The LaTex for adding R was given in commit smiths/swhs@f68dc8e.  @vajihehm, Drasil needs to be modified to reflect these changes.

This requested change is related to updating the Drasil version of the SRS, so it can count as part of the make-up assignment for the MIS assignment in CAS 741.

smiths:
@samm82, can you also have a look at this issue.  We realized that we needed to add aspect ratio in a discussion given at:

https://github.com/smiths/caseStudies/issues/26

You might need to add a data definition for R (aspect ratio).

samm82:
I'm going to track all questions here instead of https://github.com/smiths/caseStudies/issues/26.

@smiths
1) What are the values of _R<sub>max</sub>_ and _R<sub>min</sub>_?
2) Should the manual be updated for the new data definition or are we not maintaining it anymore? (I know we're not maintaining the caseStudies repo, but I'm not sure about swhs.)

samm82:
3) Should I call it AR for consistency with GlassBR (and so R isn't used for Requirements and Aspect Ratio), and if so, (related to (2)) above, should I update manual as well? @smiths

smiths:
@samm82, we shouldn't have to worry about consistency of symbols between examples.  This is too high a bar.  There will be more concepts than symbols.  Having said that, you can make this change.  Since we are able to have consistency in this case, we might as well.  😄 

With respect to minimum and maximum aspect ratios, I suggest 0.01 to 100.  I don't have any special reason for those numbers, other than my common sense understanding that a very tall and narrow tank, or a short and wide tank, wouldn't not easily fit beside someone's home.

samm82:
Should I update the manual SWHS repo?

smiths:
No, you don't need to update the manual SWHS.  We should focus our efforts on the Drasil version now.

samm82:
[...]

PR #1762, opened 2019-07-25
Updated DDs

### LINKED ISSUE #1483: SWHS: Additional Information about Quantities in DD
In manual, this the symbol description contain more information about the quantities than Drasil does:

manual

Drasil:
drasil

This could potentially be more difficult, since I'm guessing these descriptions come from the concepts defined in `drasil-data`, but should this additional information be added to Drasil?

samm82:
Added missing DDs in SWHS, and improved existing ones.

_Closes #1061, closes #1483, and closes #1484_


JacquesCarette:
master merge needed.

Comment for @bmaclach : the generated code has errors with just "constraint violated" in it; while better than no checking, we should definitely improve those error messages to be more informative! At the very minimum, the variable name should be mentioned. Better still would be the expected range and the value that violates it, and the name of the constraint from which this comes.

JacquesCarette:
Another master merge needed... we'll eventually get this one in!

JacquesCarette:
and yet another merge needed...

PR #1786, opened 2019-07-31
Used punctuate in TeX.Print

samm82:
Fixed alignment of matrices and case statements in the generated LaTeX.

_Closes #1774_


smiths:
The generated aligned tex files look good to me.

JacquesCarette:
Needs a master merge, but is otherwise ready to go.

PR #1793, opened 2019-08-02
Add authors to Doxygen comments

### LINKED ISSUE #1781: Add author to doxygen comments for modules
The module-level doxygen comments in generated code could be enhanced with an author field. Author information is present in `SystemInformation`. It will need to be carried into `CodeSystInfo` for use in the generator.

bmaclach:
This PR closes #1781 by adding author fields to the module-level Doxygen comments.


smiths:
@bmaclach, I had a look at the changes to stable.  I was going to write that the code authors are not the same people as the SRS authors, and then I remember that the code is generated.  I was then going to say, that you have the right behaviour, but then I thought some more.  :-)  Right now our code generation step is fairly simple, but in the long run there will be a need to make design decisions, by selecting algorithms, data structures, file formats etc.  The logical possibility exists that the "authors" of the generated code will not be the same people as the authors of the SRS.  The SRS authors will be domain experts.  The code authors will be design, and possibly numerical analysis, experts.  I think we should allow for the possibility that these are different people.

JacquesCarette:
@bmaclach I would open a new issue with @smiths 's comments. The underlying thought processes in that comment itself show how this is going to be a somewhat thorny issue that we won't fix in just a single day.
[And there's a conflict that needs resolved before a merge can happen]

bmaclach:
Requested comments added, master merged, and #1804 created. This should be good to go.

PR #1806, opened 2019-08-06
Resolve

samm82:
While trying to fix the bug from f073010978739f137bbc4ed0138f0275a0a3ca24, I saw the potential for some code cleanup in `docLang`. Some of the lookup functions were only used to search for a UID in the corresponding table, so I only exported a new `resolve` function when applicable.

I figured it was a good idea to do this in a separate branch 😁


samm82:
The merge conflicts have been ... `resolve`d 😎

PR #1810, opened 2019-08-07
Add date to Doxygen comments

### LINKED ISSUE #1782: Add date to doxygen comments for modules
The module-level doxygen comments in generated code could be enhanced with a date field. There are Haskell libraries that can be used to get the current date (http://hackage.haskell.org/package/time-1.9.3/docs/Data-Time-Calendar.html and http://hackage.haskell.org/package/time-1.9.3/docs/Data-Time-Clock.html). The date will likely need to be requested within the `genCode` function because it is an IO action, and then passed along to the generator from there.

bmaclach:
This PR closes #1782 by adding a date field to the module-level doxygen comments.

#1793  should be merged first.


Mornix:
Dates enabled for the examples in stable will cause many issues when diffing after today's date (August 7th) with the current changes to stable.

bmaclach:
@Mornix Good point. I can't independently turn off the dates without turning off the module-level comments altogether... but I could add an additional `Choice` for turning dates on/off. Should I do that @JacquesCarette @smiths? Or any other ideas?

smiths:
@bmaclach, I like the addition of dates to the comments, but @Mornix is right about the problem with our diffs.  Your solution of a choice for turning the dates off sounds good to me.

bmaclach:
Added the choice and turned off dates for our examples.

JacquesCarette:
Needs a master merge, but otherwise ready to go.

PR #1812, opened 2019-08-08
Add return values to Doxygen documentation

### LINKED ISSUE #1784: Add return fields to Doxygen comments for functions
This issue stems from #1771. We should be able to "tag" values as return values, which would trigger our generator to generate a `return` field with a description of the returned value in the Doxygen comments for the corresponding function, assuming Doxygen comments are enabled. Whether to generate this return field could also be subject to a global option.

bmaclach:
This PR closes #1784 by documenting function return values in Doxygen comments for the generated code.

#1807 should be merged first.


JacquesCarette:
Needs a merge, but otherwise ready to go.

JacquesCarette:
and the other merge now created a couple more conflicts...

PR #1813, opened 2019-08-08
Fix Derived Values function for Unbundled case

bmaclach:
I noticed I hadn't handled the case where `Unbundled` was chosen and there are derived values for an example. None of our current examples meet these criteria, so they were all fine, but potential future code generation examples could have been broken. This PR fixes that.

#1807 should be merged first.


JacquesCarette:
Changes seem fine - but Travis failed?

bmaclach:
Fail was due to a connection time out... I think it's okay because the push build passed.

JacquesCarette:
Indeed - and #1807 is merged, so in this goes!

PR #1816, opened 2019-08-09
Use mkVar throughout generator

bmaclach:
This PR cleans up/improves our generator. I noticed the pattern `variable (codeName x) (convType $ codeType x)` was used many times throughout the generator to make a variable, so this PR pulls out that pattern into a function `mkVar` and then uses it everywhere. This included fixing some places where variable names and types were being hardcoded wherever used instead of being read from the appropriate `CodeChunk`. Doing this clean-up helped me notice a bug where variables with the same name (but different scope) were mistakenly being given the same description in the doxygen comments because our `VarMap` and `FuncMap` were keyed by variable names rather than UIDs. I fixed the maps to use UIDs as keys, which fixed the descriptions as seen in the stable changes.

#1812 should be merged first.


JacquesCarette:
And shoot, there's a conflict.

Issue #1815, opened 2019-08-09
[GamePhysics] Duplicated Chasles Theorem


samm82:
The DD below should be updated with the correct label/refname and the source from Wikipedia, and the TM should be removed.

![tm](https://user-images.githubusercontent.com/35857611/62805939-6e168400-babf-11e9-9666-c0c9751d0ad2.png)

![dd](https://user-images.githubusercontent.com/35857611/62805900-55a66980-babf-11e9-804b-003f35259e60.png)

_While updated the Notes section in the examples. I noticed that Chasles' theorem appears twice, once as a TM and once as a DD. The only differences I see are that 
a) the labels are different
b) the DD has more assumptions

Is the DD a refinement of the TM because of the assumptions, or should only one be kept? If so, which one?_

oluowoj:
@samm82, yes, I actually moved it from a TM to a DD as per @smiths and have not removed the one in TM, I just wanted to be more convinced its a DD. @smiths I know you wanted to make this a DD, is it still the case? @samm82  what do you think? 

@samm82, as per label, I want something more descriptive than Chasles' , unless @smiths thinks otherwise. The definition relates to the velocity of an arbitrary point on a rotating and translating body. I have not found a lot of good info. on Chasles theorem online, so not sure about this label. If you find anything please post here. But let's see what @smiths thinks

Thanks for spotting this

smiths:
@oluowoj, can you tell me your opinion on whether this is a TM or a DD, and why?  I don't want to be the only one that has an opinion on this.  :-)  I think over the course of the summer we have come to a fairly reasonable distinction between TMs and DDs, but if I cannot communicate the distinction to others, than it isn't as good as I hope.  :-) 

We've discussed it before, but let me recap again the distinction that we are using as a working model.  TMs are refined to create new models, that are called GMs or IMs.  DDs are not refined; they are just used.  GDs and IMs are derived, or refined, from other models.  DDs are not derived; they are just given.  If a DD includes a derivation, then that means it is refining other models, which would make it a GD or an IM.

I've been using the above mental model for most of the summer and it seems to be working.  Of course, I'm open to refinement or change.  If there is an inconsistency, that would be good to know and then we can fix it.  I think this is what is meant by grounded theory.  :-)

I'm not trying to test you @oluowoj.  I just want you to apply the above mental model to Chasles' theorem and see what decision you make.  Whatever you say is useful information.  Much more useful than me just saying what to do.

[...]

PR #1825, opened 2019-08-13
[GamePhysics] Fixed Chasles' Theorem DD

samm82:
- Updated the Chasles' Theorem DD with the correct label/refname 
- Added Wikipedia source to Chasles' Theorem DD
- Removed Chasles' Theorem TM

_Closes #1815_


smiths:
@samm82, the move of Chasles' theorem looks good, except for some reason two assumptions have been added that were not there before:

- A: objectDimension
- A: dampingInvolvement

Chasles' theorem does not depend on these assumptions.  They shouldn't be referenced.  Moreover, if we do use assumptions, we shouldn't just list them, there should be text to justify/explain the assumption.

smiths:
The changes to stable look good to me.  Thanks for removing the unnecessary assumptions @samm82.

code/stable/gamephysics/Website/GamePhysics_SRS.html
```diff
@@ -1525,7 +1525,7 @@ <h3>Data Definitions</h3>
                       <th>Notes</th>
                       <td>
                         <p class="paragraph">
-                          The linear velocity <em><b>v</b><sub>B</sub></em> (m/s) of any point B in a rigid body (from <a href=#assumpOT>A: objectTy</a>) is the sum of the linear velocity <em><b>v</b><sub>O</sub></em> (m/s) of the rigid body at the origin (axis of rotation) and the resultant vector from the cross product of the rigid body's angular velocity <em>ω</em> (rad/s) and the displacement vector between the origin and point B, <em><b>r</b><sub>OB</sub></em> (m).
+                          The linear velocity <em><b>v</b><sub>B</sub></em> of any point B in a rigid body (from <a href=#assumpOT>A: objectTy</a>) is the sum of the linear velocity <em><b>v</b><sub>O</sub></em> of the rigid body at the origin (axis of rotation) and the resultant vector from the cross product of the rigid body's angular velocity <em>ω</em> and the displacement vector between the origin and point B <em><b>r</b><sub>OB</sub></em>.
```
samm82: The Notes section here essentially restates the equation. Is this desired? How much redundancy is bad? @smiths

smiths: Restating the equation isn't really necessary, but it doesn't bother me.  The words are sometimes helpful to readers, especially those that aren't strong with mathematical notation.  I don't think we always want the word version, but I also don't think we should delete all instances.  We can leave the Notes field for this DD as it is.

samm82:
Based on the conversation at #1815, I think reverting this change would be the easiest way to go about fixing the refinement issue. How should we go about doing that? @JacquesCarette

JacquesCarette:
If the change was done in a single commit, `git revert` can do that.  Otherwise it's more complicated, and you'll have to google it.

Issue #1783, opened 2019-07-31
Add definitions to doxygen comments for parameters


bmaclach:
This issue stems from #1771. We should be able to "tag" parameters with additional information (i.e. a definition) which can then be displayed as part of the Doxygen comment for that parameter in the generated code. Whether or not to include the definition in the comment could be subject to  a global option.

bmaclach:
@JacquesCarette I still have questions about this one... 

Generated function parameters come from `Quantity`s (or `CodeQuantity`s) in Drasil. For the most part, "users" of Drasil don't interact with parameters directly, but they do interact with the `Quantity`s they come from. So if we want to tag parameters with more information, and if we want the author to control how that information gets tagged, the `Quantity`s need to contain 1. the information to tag and 2. information about whether or not it _should_ be tagged.

Part 1 is already done -- we have `DefinedQuantityDict`s that are tagged with additional information (i.e. the definition). 

Part 2 is where I have questions. I don't think we just want to tag every parameter that has a definition with the definition. We want the author to be able to decide on a per-parameter basis whether the definition is important enough to show in the doxygen comments. We could add an additional field to `DefinedQuantityDict` with some show/hide data type. But then for `DefinedQuantityDict`s in drasil-data, we'd need to decide on either show/hide ourselves, and then provide a function to switch between show/hide so the user can make their own decision?

This also reminds me of talk we've had about using characteristics of the intended reader to determine what information gets displayed, where we can assume that readers know a certain amount and not show definitions if they fall into the categories that the intended readers know.

JacquesCarette:
Excellent analogy with "characteristics of the intended reader".  And yes, `DefinedQuantityDict` is indeed exactly one kind of 'tagged' `Quantity`. Though it is important to note that it's neither an arbitrary tag, nor a dynamically discoverable tag (we had those for a while, called them 'attributes', and they became a horrible mess).

I think we want to use the same process as with the rest of Drasil to move 'tags' from one stage to another: we do an explicit 'query' to populate the information. That way the choice of what to display becomes part of the recipe. And, to go along with the above analogy, the query could be "add all sufficiently interesting definitions". [Of course, we're far from that now, but we need to start providing infrastructure that will move us in that direction].

smiths:
Great discussion!  As a starting point, I think the default for the doxygen comments should be to show all definitions.  That is what we do with the SRS.

JacquesCarette:
Does #1826 close this issue?

[...]

PR #1826, opened 2019-08-13
Add definitions to Doxygen comments

bmaclach:
This PR contributes to #1783 by adding definitions to the Doxygen documentation in the generated code. For now, we show all definitions. 

At first when I enabled definitions, most of GlassBR's definitions were not showing. I realized it was because there were many quantities in GlassBR that were defined completely independently from their corresponding concepts. In other words, the same "thing" was being defined twice from scratch, once as a concept and once as a quantity, with different UIDs. I attempted to fix this by tying together quantities with their definitions, but it's possible I made mistakes and matched a quantity with an inappropriate definition. Also, in many cases the term for the concept was slightly different from the term for the quantity. The term for the concept is now the one that is used everywhere, which is why there were some changes to the stable SRS. So @smiths when you review stable changes be on the lookout for bad definitions or insufficient terms. If I made mistakes or you think any of the terms/definitions should be changed, let me know.


code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -75,15 +75,15 @@ \subsection{Table of Symbols}
 \\
 $b$ & Plate width (short dimension) & m
 \\
-$capacity$ & Capacity or load resistance & Pa
+$capacity$ & Capacity & Pa
```
smiths: I prefer the old definition of capacity.  The new one does not say anything that the variable name itself does not already convey.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -109,7 +109,7 @@ \subsection{Table of Symbols}
 \\
 $LDF$ & Load duration factor & --
 \\
-$Load$ & Applied load (demand) or pressure & Pa
+$Load$ & Load & Pa
```
smiths: As for capacity, I prefer the old definition for Load.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -127,7 +127,7 @@ \subsection{Table of Symbols}
 \\
 ${P_{\text{f}\text{tol}}}$ & Tolerable probability of failure & --
 \\
-$q$ & Applied load (demand) & Pa
+$q$ & Demand & Pa
```
smiths: I prefer the old definition for q.

bmaclach:
@smiths I made the requested changes in d9e1f1da5bcdd8443e92a06d0c9d22c44471454d. There was a small ripple effect to other parts of the SRS, so you probably want to review the new stable changes and confirm that this is what we want.

Issue #1586, opened 2019-06-19
Constrain GlassBR Probabilities


samm82:
From #1478, the probabilities from the TMs and IMs in GlassBR should be constrained to be between 0 and 1 (inclusive).

PR #1827, opened 2019-08-13
[GlassBR] Minor TM Touch Ups

samm82:
- Renamed GlassBR safety requirements (migrated from the `swhsTMs` branch)
- Updated TM Notes
- Fixed constraints on probabilities

_Closes #1586 and closes #1587_

### DIFF
--- a/code/stable/glassbr/SRS/GlassBR_SRS.tex
+++ b/code/stable/glassbr/SRS/GlassBR_SRS.tex
@@ -97,7 +97,7 @@ $is-safeLoad$ & Load resistance safety requirement & --
 \\
 $is-safeLR$ & Load resistance safety requirement & --
 \\
-$is-safePb$ & Probability of breakage safety requirement & --
+$is-safePb$ & Probability of glass breakage safety requirement & --
 \\
 $is-safeProb$ & Probability of failure safety requirement & --
 \\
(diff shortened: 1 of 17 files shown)


code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -93,13 +93,13 @@ \subsection{Table of Symbols}
 \\
 $interpZ$ & InterpZ & --
 \\
-$is-safeLoad$ & Variable that is assigned true when load resistance (capacity) is greater than applied load (demand) & --
+$is-safeLoad$ & Load resistance safety requirement & --
 \\
-$is-safeLR$ & Variable that is assigned true when load resistance (capacity) is greater than load (demand) & --
+$is-safeLR$ & Load resistance safety requirement & --
```
smiths: The description of `is-safeLoad` and `is-safeLR` are the same.  We shouldn't really have symbols with the same definition.  My guess is that this due to the confusion of the general theory of safety and the specific instance of safety for GlassBR.  @samm82 I can't remember which is which, but please add wording that shows which one is the more general notion and which is specific to GlassBR.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -93,13 +93,13 @@ \subsection{Table of Symbols}
...
 \\
-$is-safePb$ & Variable that is assigned true when calculated probability is less than tolerable probability & --
+$is-safePb$ & Probability of breakage safety requirement & --
```
smiths: Can we change this to "Probability of glass breakage ..."?  As mentioned above, it is nice to distinguish the general notion from the GlassBR specific.

code/stable/glassbr/SRS/GlassBR_SRS.tex
samm82: For the GlassBR specific quantity, how about "applied load resistance safety requirement", "demand resistance safety requirement", or "3 second load equivalent resistance safety requirement"? I'm trying to think of how to indicate that the load is the specific instance from GlassBR.

smiths: @samm82, I like your last option "3 second ...".  Let us go with that one.  Thank you!

smiths:
The changes to stable look good, but @JacquesCarette should check the changes to the Drasil code files.

Issue #1824, opened 2019-08-13
Use better MathJax delimiters.


danscime:
First I will work to close #479 as MathJax is close to fully being implemented.

And then after, change the delimiters to better match the [MathJax Docs](http://docs.mathjax.org/en/latest/tex.html).

danscime:
Changing `$$` to `\[` and `\]` for displayed math in 8bb660558773246b9ed4dcd91abb7fe526763375

danscime:
Changing `$` to `\(` and `\)` for inline math in 6612748a629ec56992a56c145b461f41863fa752

PR #1832, opened 2019-08-15
PR - Implementing MathJax

### LINKED ISSUE #479: Should move from raw HTML to MathJax for embedded math
See #478 for most of the discussion on this topic.

### LINKED ISSUE #1251: HTML fraction formatting
I have noticed at least 2 fractions are improperly formatted in the HTML output of the SWHS example. These fractions appear correctly in the Latex output, just not the HTML. Might be the same issue as #1067

Both occur in the notes sections of the first 2 instance models.

I have included an image showing an example fraction.

Screen Shot 2019-05-02 at 10 52 37 AM

danscime:
- Implements MathJax into HTML. Closes #479.
- Fixes improper fraction formatting. Closes #1251.

MathJax does not appear everywhere in the HTML. Currently it is set up to work on:
- displayed math equations
- fractions
- square roots
- matrices
- cases

In another PR, better delimiters will be looked into, as per #1824.


People/DanScime/Print/HTML-Print.hs
```diff
@@ -0,0 +1,491 @@
+module Language.Drasil.HTML.Print(genHTML) where
```
JacquesCarette: Do you really want this file in this PR?

danscime: Don't need anymore. Will remove.

PR #1834, opened 2019-08-15
[GamePhysics] Improved Notes

samm82:
Improved the notes sections in the GamePhysics example.

_Part of #1476_


code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -438,7 +439,9 @@ \subsubsection{Theoretical Models}
...
               \end{symbDescription}
 \\ \midrule \\
-Notes & Two rigid bodies in the universe attract each other with a force $\mathbf{F}$ (N) that is directly proportional to the product of their masses, ${m_{1}}$ and ${m_{2}}$ (kg), and inversely proportional to the squared distance ${\text{||}\mathbf{r}\text{||}^{2}}$ ($\text{m}^{2}$) between them. The vector $\mathbf{r}$ (m) is the displacement between the centres of the rigid bodies and $\text{||}\mathbf{r}\text{||}$ (m) represents the Euclidean norm of the displacement, or absolute distance between the two. $\mathbf{\hat{r}}$ denotes the displacement unit vector, equivalent to the displacement divided by the Euclidean norm of the displacement, as shown above. Finally, $G$ is the empirical physical constant used to show the force between two objects caused by gravity ($\frac{\text{m}^{3}}{(\text{kg}\text{s}^{2})}$).
+Notes & Two rigid bodies in the universe attract each other with a force $\mathbf{F}$ that is directly proportional to the product of their masses, ${m_{1}}$ and ${m_{2}}$, and inversely proportional to the squared distance ${\text{||}\mathbf{r}\text{||}^{2}}$ between them.
+
```
smiths: You have removed the fact from the notes that the displacement vector is the distance between the two bodies.  I can see why you made your other simplifications, and I'm fine with them, but the word displacement does not immediately suggest to me the physical meaning of the vector r.  Since its description as a displacement vector is generated, I think we should keep the note that clarifies that it represents distance in this case.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -549,9 +556,7 @@ \subsubsection{Data Definitions}
               \item{${m_{T}}$ is the total mass of the rigid body (kg)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
-        
-        \hyperref[assumpOD]{A: objectDimension}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: We can remove the assumption that the body is 2D.  The equation for centre of mass is the same for 2D as it is for 3D.  I just verified that the original case study version makes this same mistake.  Now that I noticed it though, we should remove it.  😄

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -590,11 +595,9 @@ \subsubsection{Data Definitions}
               \item{$\mathbf{p}$ is the position (m)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
-        
-        \hyperref[assumpOD]{A: objectDimension}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: Again, we can remove the 2D assumption.  The same equation is valid in 3D.

smiths: We can also remove the assumption about damping.  If there is friction its effect is elsewhere.  This definition is unaffected.  (I'm not 100% sure about this definition overall, but now isn't the time to get into that.)

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -633,11 +636,9 @@ \subsubsection{Data Definitions}
               \item{$\mathbf{r}$ is the displacement (m)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: Again, we can remove the assumptions related to 2D and damping.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -676,11 +677,9 @@ \subsubsection{Data Definitions}
               \item{$\mathbf{v}$ is the velocity ($\frac{\text{m}}{\text{s}}$)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
-        
-        \hyperref[assumpOD]{A: objectDimension}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: We can remove the 2D and damping assumptions.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -719,11 +718,9 @@ \subsubsection{Data Definitions}
               \item{$ϕ$ is the orientation (rad)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: We do not want to remove the 2D assumption here, since we really only focus on one angle.  (The notation suggests that we have a more generalized notion of rotation, but that isn't really true.)  We can remove the reference to damping though.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -762,11 +759,9 @@ \subsubsection{Data Definitions}
               \item{$θ$ is the angular displacement (rad)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
-        
-        \hyperref[assumpOD]{A: objectDimension}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
         
-        \hyperref[assumpDI]{A: dampingInvolvement}
+        No damping occurs during the simulation (from \hyperref[assumpDI]{A: dampingInvolvement}).
```
smiths: We can remove the damping assumption.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -805,11 +800,9 @@ \subsubsection{Data Definitions}
               \item{$ω$ is the angular velocity ($\frac{\text{rad}}{\text{s}}$)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
         
-        \hyperref[assumpOD]{A: objectDimension}
-        
-        \hyperref[assumpDI]{A: dampingInvolvement}
+        No damping occurs during the simulation (from \hyperref[assumpDI]{A: dampingInvolvement}).
```
smiths: We can remove the damping assumption.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -856,13 +849,11 @@ \subsubsection{Data Definitions}
               \item{${\mathbf{I}_{\text{B}}}$ is the moment of inertia of rigid body B (kg$\text{m}^{2}$)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpOT]{A: objectTy}
-        
-        \hyperref[assumpOD]{A: objectDimension}
+Notes & All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: I would have to go through the details of the derivation to see if it generalizes to 3D, so for now, let us leave in the assumption that it is 2D.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -994,13 +987,11 @@ \subsubsection{Data Definitions}
               \item{$\mathbf{v}$ is the velocity ($\frac{\text{m}}{\text{s}}$)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & The kinetic energy of an object is the energy it possess due to its motion.
-        
-        \hyperref[assumpOT]{A: objectTy}
+Notes & Kinetic energy is the measure of the energy a body possesses due to its motion.
         
-        \hyperref[assumpOD]{A: objectDimension}
+        All bodies are assumed to be rigid (from \hyperref[assumpOT]{A: objectTy}) and two-dimensional (from \hyperref[assumpOD]{A: objectDimension}).
```
smiths: We can keep these assumptions, although kinetic energy is already separate from strain energy.  We probably do not need the rigid body assumption, but I don't have time right now to think about this too deeply.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -994,13 +987,11 @@ \subsubsection{Data Definitions}
...
         
-        \hyperref[assumpDI]{A: dampingInvolvement}
+        No damping occurs during the simulation (from \hyperref[assumpDI]{A: dampingInvolvement}).
```
smiths: We should keep damping here, to remind us that the energy balance changes if energy is converted into heat through friction.  Technically this would come up somewhere else, but that can wait for the future overhaul of this example.

smiths:
Looks good to me @samm82.  We'll wait for the approval of @JacquesCarette before merging though.

PR #1838, opened 2019-08-19
Move 'Impulse for Collision' from DD to GD

oluowoj:
Moved DD to GD.
Derivation to come in a separate PR.


oluowoj:
> @oluowoj, please add the fixes to stable to your PR.  Your PR will not pass through the Travis checks without these updates.  Moreover, rather than read the code to see the changes, my approach is to first review stable from the physics/documentation point of view.

Oops! sorry, I just did. Thanks

oluowoj:
Yes @smiths, I will add the derivation in a separate PR. Thanks
image

oluowoj:
@bmaclach, HasSymbol is not defined as an instance in Gendef.hs(Theory), and that's the reason for the error I am getting when I do a merge.
image
impulseDD is now impulseGD but definedIn will not work for GenDef.hs. Should I be defining a new function like definedIm without HasSymbol? What do you think?

oluowoj:
Here is the actual error:
image

bmaclach:
@oluowoj Looks like that `definedIn` function was added specifically for referring to DDs (I think they are the only type of "theory" that are an instance of `HasSymbol`). So yes, you could write an alternative version of that function where the `Referable` does not also need `HasSymbol`.

oluowoj:
Added a new function definedIn to fix the merge error on this branch due to impulse being moved from DD to GD. Maybe there is a better way to write this function but it works at the moment.

Ready for merge

Issue #1476, opened 2019-06-04
Redundant Information in Notes


samm82:
In most of the examples, there is a lot of redundant information that occurs in the Description and Notes.

- [x] GamePhysics (#1834)
- [x] GlassBR (#1850)
- [x] HGHC (only has two DDs without Notes sections)
- [ ] NoPCM
- [x] Projectile (done during its creation)
- [ ] SSP
- [ ] SWHS

![redundantNotes](https://user-images.githubusercontent.com/35857611/58897489-6bb23b00-86c6-11e9-8bb7-af4be88075f0.png)

In this example, the redundant information is **q**, _g_, _T_, _t_, and possibly _C_, _ρ_, and ∇. Should these be removed from the notes and stay captured in the description?

JacquesCarette:
My opinion: yes. Especially because this information is "linked together" and thus meant to be read as a single unit, it doesn't make sense to be quite so redundant.

smiths:
I agree with @JacquesCarette.  The notes field shouldn't just repeat the description field.  The description field provides the symbols and their definitions.  The purpose of the Notes field is to capture the information that was lost in the transition of replacing the manual version of the Description with the Drasil version.  Since Drasil could only do the list of variables, we added the Notes field.

For this specific example, you can remove the redundant information, but please keep the new information (the last sentence).

Ant13731:
I briefly looked through all the examples and I couldn't find any that were a direct copy of the description (some were close to defining the same thing twice, but the notes still added necessary context to the variables), so I think this issue could be closed.

smiths:
Sounds good @Ant13731.  I agree that we can close this issue.

PR #1850, opened 2019-08-21
Improved Notes in GlassBR

samm82:
As per #1476.

### DIFF
--- a/code/stable/glassbr/SRS/GlassBR_SRS.tex
+++ b/code/stable/glassbr/SRS/GlassBR_SRS.tex
@@ -492,13 +492,13 @@ Description & \begin{symbDescription}
               \item{$J$ is the stress distribution factor (Function) (Unitless)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & $a$ and $b$ are dimensions of the plate, where ($a\geq{}b$).
+Notes & the $a$ and $b$ are dimensions of the plate, where ($a\geq{}b$).
         
-        $h$ is the minimum thickness, which is based on the nominal thicknesses as shown in \hyperref[DD:minThick]{DD: minThick}.
+        $h$ is defined in \hyperref[DD:minThick]{DD: minThick} and is based on the nominal thicknesses.
         
-        $LDF$ is the load duration factor as defined by \hyperref[DD:loadDurFactor]{DD: loadDurFactor}.
+        $LDF$ is defined in \hyperref[DD:loadDurFactor]{DD: loadDurFactor}.
         
-        $J$ is the stress distribution factor (Function), as defined in \hyperref[DD:stressDistFac]{DD: stressDistFac}.
+        $J$ is defined in \hyperref[DD:stressDistFac]{DD: stressDistFac}.
         
 \\ \midrule \\
 Source & \cite{astm2009}, \cite[(Eqs. 4-5)]{beasonEtAl1998}, and \cite[(Eq. 14)]{campidelli}
@@ -588,9 +588,9 @@ Description & \begin{symbDescription}
               \item{$m$ is the surface flaw parameter ($\frac{\text{m}^{12}}{\text{N}^{7}}$)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & \hyperref[assumpSV]{A: standardValues}
+Notes & ${t_{\text{d}}}$ and $m$ come from \hyperref[assumpSV]{A: standardValues}.
         
-        \hyperref[assumpLDFC]{A: ldfConstant}
+        $LDF$ is assumed to be constant (from \hyperref[assumpLDFC]{A: ldfConstant}).
         
 \\ \midrule \\
 Source & \cite{astm2009}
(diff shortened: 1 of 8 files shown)


code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -492,13 +492,13 @@ \subsubsection{Data Definitions}
               \item{$J$ is the stress distribution factor (Function) (Unitless)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & $a$ and $b$ are dimensions of the plate, where ($a\geq{}b$).
+Notes & the $a$ and $b$ are dimensions of the plate, where ($a\geq{}b$).
```
smiths: I think we should remove the "the" here; it sounds awkward.

samm82: Good catch! It's supposed to be "$a$ and $b$ are the dimensions of the plate, where ($a\geq{}b$)."

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -676,13 +676,13 @@ \subsubsection{Data Definitions}
...
+Notes & ${\hat{q}_{\text{tol}}}$ is defined in \hyperref[DD:tolLoad]{DD: tolLoad}.
         
-        $h$ is the minimum thickness, which is based on the nominal thicknesses as shown in \hyperref[DD:minThick]{DD: minThick}.
+        $E$ comes from \hyperref[assumpSV]{A: standardValues}.
         
-        ${\hat{q}_{\text{tol}}}$ is the tolerable load defined in \hyperref[DD:tolLoad]{DD: tolLoad}.
+        $h$ is defined in \hyperref[DD:minThick]{DD: minThick} and is based on the nominal thicknesses.
         
-        \hyperref[assumpSV]{A: standardValues}
+        the $a$ and $b$ are dimensions of the plate, where ($a\geq{}b$).
```
smiths: I wrote about the "the" previously.  Deleting the "the" applies for all occurrences of this text.

code/stable/glassbr/SRS/GlassBR_SRS.tex
samm82: Fixed in 5b307d947.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -771,17 +771,17 @@ \subsubsection{Data Definitions}
               \item{$GTF$ is the glass type factor (Unitless)}
               \end{symbDescription}
 \\ \midrule \\
-Notes & $q$ is the 3 second equivalent pressure, as given in \hyperref[DD:calofDemand]{DD: calofDemand}.
+Notes & The glass is considered to be a single lite (from \hyperref[assumpGL]{A: glassLite}).
```
smiths: The relevance of this statement is not clear.  In the case study (manual) version, no reference is made to this assumption.  Where did it come from here?  I'm not saying it is wrong, maybe the derivation of qhat depends on this assumption?

samm82: It was added in #932. Looking at it in more detail now, I don't see the relevance of it either, and the DD in manual doesn't have a derivation that might use this. Should we remove it? I don't think that the change in #932 is justified in adding this assumption.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -1066,13 +1068,13 @@ \subsubsection{Data Definitions}
...
+Notes & $LR$ is also called capacity.
         
-        \hyperref[assumpGL]{A: glassLite}
+        $NFL$ is defined in \hyperref[DD:nFL]{DD: nFL}.
         
-        \hyperref[DD:gTF]{DD: gTF}
+        $GTF$ is defined in \hyperref[DD:gTF]{DD: gTF}.
         
-        \hyperref[DD:nFL]{DD: nFL}
+        The glass is considered to be a single lite (from \hyperref[assumpGL]{A: glassLite}).
```
smiths: Where did this come from?  It is not in the previous version, and it is not in the case study.  Moreover, this equation has LSF as part of its definition.  The equation does not assume one lite of glass, even though for the overall problem we can make this assumption.  The assumption is not true for this equation.  The assumption is applied elsewhere.

samm82: The assumption says that LSF = 1 for all equations, which is where its value comes from. Should the note be modified to make this more clear? Something along the lines of "LSF comes from [A: glassLite]"?

smiths: @samm82, you are missing the point.  This equation does not assume that LSF = 1.  If it did, the equation would be LR = NFL*GTF.  The equation shows LR = NFL*GTF*LSF because in the scope of this definition, LSF can have values other than 1.  When this DD is applied elsewhere, the value of LSF = 1 is substituted in, but that hasn't happened yet.  This should be more clear in the future when we add the derivations.

samm82: Yup, you're right. I had this thought, but somewhere between thinking "oh yeah, that makes sense" and typing up a response, I confused myself. Should I just remove the assumption? I don't think I should be adding derivations at this stage of the summer. 😂

smiths: Yes, please remove the assumption from the definition for LR.  The derivations were more something I was thinking about for the future, not this summer.  :-)

code/stable/glassbr/SRS/GlassBR_SRS.tex
smiths: Yes, please remove the assumption here.  Thank you.

code/stable/glassbr/SRS/GlassBR_SRS.tex
smiths: 👍

code/stable/glassbr/SRS/GlassBR_SRS.tex
samm82: Done in 4130d37c1.

code/stable/glassbr/SRS/GlassBR_SRS.tex
samm82: Done in 4130d37c1.

PR #1860, opened 2019-09-01
Doxygen Build Improvements

Mornix:
This PR improves some Doxygen stuff from a build perspective.

This PR addresses Doxygen 404ing on deploy's due to the full path (i.e. /home/travis/JacquesCarette/Drasil/...) and messing with the path when combined with a web URL.

This PR adds the Makefile target `doxygen` to build the doxygen of all examples. We also build the Doxygen in all CI builds now. 

Finally, I've modified the Doxygen config to make the Doxygen output less verbose (@bmaclach) as we compile a lot of Doxygen and I'm trying to keep the number of lines for a full build in Travis <10k to avoid truncation.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs
@@ -765,7 +765,7 @@ makeDoxConfig prog opt p =
   text "# messages are off.",
   text "# The default value is: NO.",
   blank,
-  text "QUIET                  = NO",
+  text "QUIET                  = YES",
   blank,
   text "# The WARNINGS tag can be used to turn on/off the warning messages that are",
   text "# generated to standard error (stderr) by doxygen. If WARNINGS is set to YES",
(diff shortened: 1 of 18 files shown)


code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs
```diff
@@ -765,7 +765,7 @@ makeDoxConfig prog opt p =
   text "# messages are off.",
   text "# The default value is: NO.",
   blank,
-  text "QUIET                  = NO",
+  text "QUIET                  = YES",
```
bmaclach: @JacquesCarette @smiths Is this something we think users might care enough about that we should make it toggle-able with a `Choice`? Would be easy to do.

smiths: My vote is Yes - we should make this choice toggle-able.  I think it is the right thing to do, and it gives us another example of user customization, and it is easy to do.  :-)

JacquesCarette: But should be done 'later', so that an issue should be filed rather than doing it right now.

bmaclach: #1871 created.

PR #1872, opened 2019-09-05
GOOL Paper Intro

bmaclach:
I wrote a first draft of an intro for the GOOL paper, as well as outlined my ideas for section headings/topics for the remainder of the paper. 

We didn't discuss how we wanted the feedback loop to be for this -- how often you want to review what I write, etc. So I'm creating a PR now but let me know if you would rather review once I have more written. 

In any case, it's probably best if we keep this branch going until the paper is finished, since PRs will almost always be dependent on the previous.


smiths:
@bmaclach, I like your writing.  Good start.  @JacquesCarette might feel otherwise, but I think we need a meeting to discuss the focus of the paper.  @JacquesCarette is the expert, but I think the message of the paper might be a little more nuanced than the current start.  I would be surprised if GOOL is the first example of a code generator with multiple language back-ends.  We need to focus on what makes GOOL special.  I don't know exactly what that is, but @JacquesCarette will know.  😄  We might also want to agree on the structure of the paper (section headings) before we start writing in earnest.

bmaclach:
The most recent commit (f1b4557) expands the intro based on our discussion during the Sept. 12th, 2019 meeting.

bmaclach:
A first draft of a Related Works section has now been added.

JacquesCarette:
I've just pushed a whole bunch of edits.  Once Travis comes back clean, I'll merge this.
The biggest thing is that we need to have a 'requirements' section. In other words, we need to describe in a reasonable amount of detail what problem we're solving.

bmaclach:
@JacquesCarette Thanks for the edits! The requirements to discuss in the Requirements section would be the same requirements that are enumerated in the intro, correct? So we just want to expand on each of those?

JacquesCarette:
I'm not going to delete the branch, as I'm guessing we'll keep working on it rather than create new ones for this purpose.

bmaclach:
@JacquesCarette I think https://github.com/JacquesCarette/Drasil/pull/1872#issuecomment-533306570 slipped through the cracks... can you take a look?

JacquesCarette:
Definitely at least those need to be discussed. As we expand, we may realize more need to be added.

PR #1874, opened 2019-09-06
GOOL inOutFunc considers state change as output

### LINKED ISSUE #1858: Objects whose state get modified should be passed as outputs to GOOL's inOutFunc
We recently added `inOutFunc` to GOOL. `inOutFunc` generates code for a function based on a list of inputs, a list of outputs, and a list of things that are both input and output. Our generator uses `inOutFunc` to generate the function for reading inputs. The `InputParameters` object is passed as an input to `inOutFunc`, but we pass an empty list of outputs. I did it this way because I was thinking of outputs as things that the function returns, and since the function does not return the `InputParameters` object (it just modifies the state), I didn't think of it as an output.

Since then I've wondered if that was the right way of thinking, or if the `InputParameters` object should be considered an output since it undergoes a transformation. If so, we would need to update `inOutFunc` to check if each output is an object, and know not to return it if so.

bmaclach:
Closes #1858. Updates our generator to pass the `InputParameters` object as something that is both an input and output to GOOL's `inOutFunc`, and updates GOOL to handle that scenario.

The changes to stable are just swapping the order of parameters, which is just a consequence of `inParams` changing from an input to an input/output. But I'll add @smiths as a reviewer just in case.


JacquesCarette:
conflicts...

bmaclach:
Merged.

PR #1878, opened 2019-09-13
drasil-gool subpackage

bmaclach:
This PR separates GOOL into its own subpackage. The very top-level typeclasses for generating auxiliary files and the full package remain in drasil-code because they are dependent on other parts of Drasil.

The build is failing due to orphaned instance warnings. Orphaned instances are instances defined in a different module from both the typeclass definition and the type definition for which the instance is being written. The instances for `AuxiliarySym` and `PackageSym` are orphaned because the types are defined in drasil-gool and the typeclasses are defined in drasil-code's Symantics.hs and the instances are defined in different files for each language in drasil-code. I did it this way because I was trying to mimic the drasil-gool directory structure (a Symantics.hs and then different files for each language renderer) in drasil-code. I could remove the warnings by combining all of these instances into Symantics.hs (in drasil-code). I don't like losing that separation, but I suppose the only other option would be to suppress the warnings. What do you think about this @JacquesCarette?


Mornix:
Note: `gool` should be added to the `PACKAGES` line in the Makefile.

bmaclach:
Thanks @Mornix, added.

bmaclach:
@JacquesCarette I explained it in the original comment for this PR. There are two routes to fixing it, and I don't know which is better.

JacquesCarette:
Oops!

The fact that the class is defined elsewhere is not a problem. What is a problem is that the data definition and the instance are not together.  So it's odd that the data type is defined once globally, but there are separate instances, depending on the language. Are they really separate?

bmaclach:
@JacquesCarette I'm confused by that question. There have to be separate instances for each language so that we can generate different code/text for each language. And it's a different data type for each language anyway. The data types `PythonCode`, `JavaCode`, etc. are defined in drasil-gool along with _most_ of the instances, but `the AuxiliarySym` and `PackageSym` typeclasses specifically are still in drasil-gool because they are dependent on too much of the rest of Drasil. And we want to write instances of these two typeclasses for `PythonCode`, `JavaCode`, etc. So that's where the separation is.

bmaclach:
(And the instances really are different depending on the language, which is maybe what you were asking)

JacquesCarette:
I think I'm just confused and I don't understand what the real problem is. Can you point me to the exact lines of source code that cause the problem?

bmaclach:
Using Python-related code to demonstrate the issue, but the problem is true for all the languages:

We define `PythonCode` in drasil-gool:
https://github.com/JacquesCarette/Drasil/blob/ffc6b5ee23285661f63d8cf4f788225fcd391264/code/drasil-gool/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs#L60

In drasil-code's Symantics.hs we define `AuxiliarySym` and `PackageSym`:
https://github.com/JacquesCarette/Drasil/blob/ffc6b5ee23285661f63d8cf4f788225fcd391264/code/drasil-code/Language/Drasil/Code/Imperative/GOOL/Symantics.hs#L15-L27

And in drasil-code's PythonRenderer.hs, we import `PythonCode` from drasil-gool and write the instances for `AuxiliarySym` and `PackageSym`:
https://github.com/JacquesCarette/Drasil/blob/ffc6b5ee23285661f63d8cf4f788225fcd391264/code/drasil-code/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/PythonRenderer.hs#L18
https://github.com/JacquesCarette/Drasil/blob/ffc6b5ee23285661f63d8cf4f788225fcd391264/code/drasil-code/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/PythonRenderer.hs#L24-L36

And these instances throw "orphaned instance" warnings because they are neither defined in the same file as `PythonCode` nor the same file as `AuxiliarySym` and `PackageSym` typeclasses.

JacquesCarette:
Why are these special though?  And why isn't `PythonCode` defined in `PhythonRenderer` ?

I understand why the warnings are issued, I'm not understanding why this only happens with those 2 classes and not any others.

bmaclach:
The confusion is probably stemming from there being two "PythonRenderer.hs" and two "Symantics.hs" -- one each in `drasil-gool` and `drasil-code`.

`PythonCode` is defined in "PythonRenderer.hs" in drasil-gool, along with instances for all of the typeclasses except for `AuxiliarySym` and `PackageSym`.

`AuxiliarySym` and `PackageSym` are special because they are defined in drasil-code's "Symantics.hs", not drasil-gool's. And the instances are in drasil-code's "PythonRenderer.hs", not drasil-gool's. I had to do it this way because the functions for generating certain auxiliary files are dependent on drasil-build, drasil-lang, drasil-database, and other packages, and we wanted drasil-gool to be independent.

JacquesCarette:
It feels to me that perhaps `AuxiliarySym` should work on (say) `PythonProject`, a new type, rather than `PythonCode`. That could be key to breaking this impasse. [And `PythonProject` could 'delegate' some work to `PythonCode`].

bmaclach:
Hmm. Okay, let me see if I can work something out that way. I'll close the PR until then.

PR #1887, opened 2019-09-16
Make gravitational acceleration a vector quantity

oluowoj:
Updated Gravitational Acceleration to vector.


oluowoj:
Travis build failed, everything seems up to date, including stable. @bmaclach , any clue?

bmaclach:
@oluowoj The Travis build log is showing a diff in stable for projectile. Probably what you changed is used in Projectile as well as GamePhysics. You should be able to see the diff locally if you run `make` (so that it makes everything, not just gamephysics).

oluowoj:
> @oluowoj The Travis build log is showing a diff in stable for projectile. Probably what you changed is used in Projectile as well as GamePhysics. You should be able to see the diff locally if you run make (so that it makes everything, not just gamephysics).


You are right @bmaclach, I updated gravitational acceleration to vector, but for the projectile example it looks like issue is from the src , I updated stable for the .html and .tex, which fixed the issue partly, but I still have issues with the files in src.

oluowoj:
> @oluowoj Sorry, didn't notice the diff was on the src files at first. Travis would have caught it if the generated code didn't compile after the change, but just to be safe I checked out your branch and ran the Projectile examples -- everything works so feel free to update the src folder in stable.
> The reason for the changes is because of how the code generator translates a symbol to a variable name. Vector symbols get _vect appended to the variable name, so the variable name for the gravitational acceleration changed from g to g_vect.

Thank you Brooks

oluowoj:
> There is an assumption in the derivation that the positive direction is from the small mass to the larger mass.  If this is the case, then the acceleration will be downward, in the opposite direction to the positive axis.  If we assume that "down" is positive, then we do not need the negative sign

Noted

oluowoj:
> @oluowoj Sorry, didn't notice the diff was on the src files at first. Travis would have caught it if the generated code didn't compile after the change, but just to be safe I checked out your branch and ran the Projectile examples -- everything works so feel free to update the src folder in stable.
> The reason for the changes is because of how the code generator translates a symbol to a variable name. Vector symbols get _vect appended to the variable name, so the variable name for the gravitational acceleration changed from g to g_vect.

@bmaclach, the src folder in build does not look like there was any recent update, so not sure if it will make any difference.
image

oluowoj:
I just updated, looks like it worked, the log is empty now. thanks

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -510,16 +510,19 @@ \subsubsection{General Definitions}
...
            \end{displaymath}
 \\ \midrule \\
 Description & \begin{symbDescription}
               \item{$\mathbf{g}$ is the gravitational acceleration ($\frac{\text{m}}{\text{s}^{2}}$)}
               \item{$G$ is the gravitational constant ($\frac{\text{m}^{3}}{(\text{kg}\text{s}^{2})}$)}
               \item{$M$ is the mass of the larger rigid body (kg)}
-              \item{$\text{||}\mathbf{r}\text{||}$ is the Euclidean norm of the displacement (m)}
-              \item{$\mathbf{\hat{r}}$ is the displacement unit vector (m)}
+              \item{$\mathbf{r}$ is the displacement (m)}
+              \item{$\mathbf{\hat{r}}$ is the unit vector directed from the large mass to the smaller mass (m)}
```
smiths: The text should say from the centre of the large mass to the centre of the smaller mass.

oluowoj: I used the definition of r from the wiki source as is, but I have updated as per your request.

smiths: @oluowoj, once you push this change to the definition of r, I can approve the pull request from the point of view of stable.

oluowoj:
i need to merge master now

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
oluowoj: @smiths, I pushed  the updated definition, build is still failing though

oluowoj:
Yes, I think there are some files in the src folder that i might have included from an old build as per @bmaclach. I am looking into it.

oluowoj:
Travis build error fixed.  @bmaclach  removing those old files fixed the issue. Thanks

Issue #1892, opened 2019-09-17
Inconsistent phrases in 2. Introduction - SRS - all projects


Ao99:
@smiths 

In all SRS documents, at the end of ‘2. Introduction’ and just before ‘2.1 Purpose of Document’:

We have these words ‘This section explains the purpose of this document, **the scope of the system**, the characteristics of the intended reader, and the organization of the document.’

However, 2.2 is ‘**Scope of Requirements**’. This inconsistency may cause some confusion. It might be better to choose the more appropriate one between the two, and use it at both of the two places.

smiths:
Thank you @Ao99.  I agree with your assessment.  @bmaclach, can  you please change the phrase "the scope of the system" to "the scope of the requirements."  (@Ao99 found this inconsistency as part of an exercise being done by the 741 class.)

bmaclach:
Closed by #1898.

PR #1898, opened 2019-09-17
Consistently say "scope of the requirements"

bmaclach:
This updates the intro text that referred to "scope of the system" to instead refer to "scope of the requirements", closing #1892.


smiths:
As simple as this, it is nice to see the change completed so quickly and instantly reflected in all of our case studies.  It isn't a world changing example, but it still might be something for @szymczdm to consider in the paper he is working on.  Fixing all of the case studies manually would have been too tedious.  We just wouldn't have made this correction if we had to do it manually.

PR #1907, opened 2019-09-18
Added assumption for value of gravitational acceleration to Projectile

### LINKED ISSUE #1884: Projectile Case Study SRS - g assumption
@smiths 

Does gravitational acceleration need to be further specified in the assumption section to a lunch from Earths surface and at a specific location (at sea level, at the equator, at the poles, or specific gravity)? Or if not, does it need to be stated that the differences in the case study are negligible - similar to how neglectCurv negates the curvature of the earth? 

I see that 9.8 m/s^2 is indeed listed in Table 10: Auxiliary Constants, but does that fact - that this case study will negate the few % variance in gravitational pull across the Earths surface - also need to be stated in the assumption section?

bmaclach:
This PR closes #1884 by adding an assumption for the value of the gravitational acceleration, and referencing the assumption in the appropriate IMs.


JacquesCarette:
Do you know why both builds failed?

bmaclach:
No... looks like the builds failed before they even ran anything. Luckily now there's a conflict, so I'll have to push another commit which will re-trigger the builds. Hopefully they'll pass.

PR #1913, opened 2019-09-19
Moment and axis added to Drasil

oluowoj:
Added Moment.hs in physics.hs
In this PR, I also added 'axis' in maths.hs as I noticed it's not defined anywhere in Drasil.

@bmaclach, you can use 'moment' after this PR is merged.

 Torque = Moment?


smiths:
Yes torque and moment are the same thing mathematically (force times distance), but there are some conventions on how people use the terms.  As you can see from the discussion on Wikipedia (https://en.wikipedia.org/wiki/Torque), under Defining Terminology, the terminology is not quite standardized.

PR #1914, opened 2019-09-19
FIXMEREF text in User Responsibilities section updated

oluowoj:
Update FIXMEREF text in User Responsibilities section.
Replaced the FIXMEREF text with reference to Assumption section


oluowoj:
oh yes, I tried to see if the error I was getting would be resolved, so I commented it and forgot to remove it afterwards. Will remove

Issue #1916, opened 2019-09-20
Detect and complain if the same symbol is used for two or more concepts within the same document


smiths:
This issue came up during the discussion in #1901.

Symbols are routinely used for different meanings in science and engineering.  However, if the same symbol has more than one meaning in the same document, this is a problem.  Drasil should prevent the user from making this mistake by automatically detecting if a symbol is use for more than one purpose within the same document.

smiths:
I'm sorry @muhammadaliog3, but I do not know the answer to your question.  Maybe you can find it by looking at the code documentation:

https://jacquescarette.github.io/Drasil/docs/index.html

Alternatively, this is likely a question that @JacquesCarette or @bmaclach could answer.

JacquesCarette:
First, you should really investigate if that `Sentence` is crucial, i.e. if it is inevitable that the only information you have, when you're ready to throw that error, is a `Sentence`. Often, it is not, and then printing is simplified.

If it is, then there are various 'rendering' functions in `drasil-printer`. You'll need to use the `plain` printer.

If you tell us the context more precisely, we can direct you to the proper functions. Often, wishing to print a `Sentence` is actually a sign that you're not quite doing the right thing. [And sometimes, it's a design bug in Drasil.]

muhammadaliog3:
I guess this is a good time to present my solution.
Say we have the the following rows in a symbols table
[(sigma, description_1),(sigma, description_2),(E, discription_3),(E, description_4)]
Here two variables have the conflicting symbol of sigma and 2 have conflicting symbol of E. 

I do not believe that in the error we should simply print "Duplicate symbols found", rather we should say "Duplicate symbols found. See the following descriptions: (description_1,description_2,description_3,description_4)".

These descriptions are stored as a list of sentences. I can fold the sentences into one sentence, but to print the error I still need to convert that one sentence to a string. This is the code by the way, the line with ----HERE------------ is where I print the error. 



```haskell
table :: (Quantity s, MayHaveUnit s) => Stage -> [s] -> (s -> Sentence) -> LabelledContent
table st ls f
    |noDuplicate = llcc symbTableRef $
      Table [atStart symbol_, atStart description, atStart' unit_]
      (mkTable [P . (`symbol` st), f, toSentence]
      $ filteredChunks)
      (titleize tOfSymb) True
    |otherwise = error . show $ S "Duplicate Symbols" +:+ foldlSentCol descDuplicates ---HERE-----------
    where 
[...]

PR #1922, opened 2019-09-24
Acceleration due to gravity derivation.

oluowoj:
Added derivation for acceleration due to gravity.


code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -511,14 +514,14 @@ \subsubsection{General Definitions}
...
-           \mathbf{g}=-\frac{G M}{\mathbf{r}^{2}} \mathbf{\hat{r}}
+           \mathbf{g}=-\frac{G M}{\text{||}\mathbf{r}\text{||}^{2}} \mathbf{\hat{r}}
            \end{displaymath}
 \\ \midrule \\
 Description & \begin{symbDescription}
               \item{$\mathbf{g}$ is the gravitational acceleration ($\frac{\text{m}}{\text{s}^{2}}$)}
               \item{$G$ is the gravitational constant ($\frac{\text{m}^{3}}{(\text{kg}\text{s}^{2})}$)}
               \item{$M$ is the mass of the larger rigid body (kg)}
-              \item{$\mathbf{r}$ is the displacement (m)}
+              \item{$\text{||}\mathbf{r}\text{||}$ is the Euclidean norm of the displacement (m)}
```
smiths: Why is r called displacement?  I believe we have discussed this before, but I do not remember the explanation.

oluowoj:
@smiths, here is your explanation on the definition of r. 
image

Since we need to generalize the definition of r (vector), can we define it as :
...an  Euclidean vector that represents the position of a point  in space in relation to an arbitrary reference. I can then make a note in the description field with your comment about r for this specific GD?


Is this okay?

oluowoj:
> @smiths, here is your explanation on the definition of r.
> 
> Since we need to generalize the definition of r (vector), can we define it as :
> ...an  Euclidean vector that represents the position of a point  in space in relation to an arbitrary reference. I can then make a note in the description field with your comment about r for this specific GD?
> Is this okay?

or " A fixed vector which locates a point in space relative to another point.

smiths:
Why do we need to generalize the definition of the vector r?  As you are using it here, the vector r is the vector that connects the centre of the two masses.  Why do you need a more general definition?  You use this definition for rhat (the unit vector), so why not for r?

Are you thinking of introducing a data definition for the norm of a vector?  We would eventually like to introduce the idea of norm (or magnitude) as a separate concept, which could be applied to any vector.  The infrastructure for Drasil isn't quite ready for that yet though.

oluowoj:
@smiths, we need a more general definition because r vector will be used in other places e.g in DD - Torque;  for the unit vector, I will need to review this as well, to have a more general meaning because these parameters are defined globally, the meaning should make sense in every place it appears.

oluowoj:
For the definition of norm, I can open a separate issue for this

oluowoj:
Yes I need to redefine unit vector, .....not just specific to the mass of 2 bodies.

smiths:
If you need to refine the concept of unit vector, you cannot do that within the acceleration example.  You should add a data definition for the generic concept of a unit vector, and then we can invoke that concept in the acceleration example.

oluowoj:
> If you need to refine the concept of unit vector, you cannot do that within the acceleration example.  You should add a data definition for the generic concept of a unit vector, and then we can invoke that concept in the acceleration example.

Agreed, I am not going to do this within acceleration example, unit vector is defined in drasil-data and this information is pulling from there, I was thinking of modifying the definition from there.  Although it's also a good idea to have unit vector as a DD which will pull from information from the same source(drasil-data). Hope you get my point.

oluowoj:
I can update the PR, so you can review and see if it makes sense.

smiths:
I think you might have misunderstood my comment @oluowoj.  I do not think we should be using a generic vector in the acceleration chunk.  It isn't generic.  It is the vector that connects the centres of the masses.  If you read the Wikipedia entry defining Newton's Law of Gravitation and they said that r was the displacement vector, don't you think that would be confusing?  If the problem comes because you want the vector r to be in drasil-data, then I think r is in the wrong spot.  The concept of a unit vector is generic; it could appear in many different documents and contexts.  The concept of a vector connecting the centre of two masses is specific to Newton's law of gravitation.

oluowoj:
Okay, thanks @smiths, I think I need a more in-depth understanding of how this is currently being used in Drasil. I will get more info. on this and make sure I understand how I can put this in Drasil.

smiths:
Sounds good @oluowoj.

oluowoj:
@smiths one issue  that might arise with the game physics example is that 'r' as a symbol is also in Torque equation i.e r(vec) * F, using the same symbol in the same document with different meanings might be an issue down the road in Drasil with issue #1916.

Please Note: The 'r' in torque, is also not displacement but position vector...(just an FYI), that needs to be updated.

My thought is that we can use another symbol for displacement instead of r,  but there will still be a conflict of symbols, which is not an issue now in Drasil but will be later.

image

oluowoj:
We might need to use a different symbol for the 'r'(position vector) in torque maybe

smiths:
@oluowoj, you are correct that within the same document we do not want to use the same symbol with a different meaning.  We should change one of the rs to be something else.  Even if Drasil isn't complaining, it isn't a good document if we are manually violating the rule of one meaning per symbol.  Although r is common for the moment arm in the torque equation, other symbols are also used, like d.  Alternative we could change Newton's law to use d instead of r.

oluowoj:
@smith, the issue with conflicting symbols has been addressed. I have used the symbol 'd(vec)' to represent the distance between the center of masses instead of r, since r has been taken. 
Here is what the equation looks like now:
image

I also think that the highlighted part of the  note section is a duplication of the description section highlighted in yellow. Do you want to keep it still?

smiths:
The use of the d vector here looks good to me.  I'm fine with you removing the redundant material in the Notes field.

oluowoj:
@JacquesCarette this is ready to be merged.

JacquesCarette:
I can't merge because @smiths has not cleared his change request.

PR #1923, opened 2019-09-25
More minor GOOL improvements

bmaclach:
In this PR:

- Renaming more things. A lot of GOOL functions/types had "state" in the name which I never really understood before but I now think was a relic of GOOL's original focus on state machines, and so I am now giving better names to these things
- Updating functions for generating methods/functions to take a `Type` instead of a `MethodType`. The `MethodType` distinction is useful, but only internally, so there's no need to expose it and complicate the user-facing interface of GOOL.


JacquesCarette:
In part it is indeed because it was modeling state machines. Some of it might be from OO terminology where the variables in an object are its 'state'. In any case, all this renaming is fine.

PR #1927, opened 2019-09-27
Updated torque to vector

oluowoj:
Torque has been updated to vector


JacquesCarette:
@smiths you should open an issue about units of vectors. Something interesting to think about.

Issue #1868, opened 2019-09-03
Factor out more common patterns in GOOL renderers


bmaclach:
As discussed in the meeting on Aug 29th, 2019, it would be nice to clean up the GOOL renderers more by factoring out more common code in advance of the GOOL paper, but we should be selective in which instances of code duplication we tackle.

We want to be able to quantify the "sharedness" of two renderers in GOOL, so any common code we factor out should be low hanging fruit that makes this quantity more impressive. For example, we want to maximize the number of instances of class functions that are just dispatches to the common functions in LanguageRenderer.hs.

bmaclach:
@JacquesCarette I've started working on this. The following GOOL symantic function:
```
(#~) :: repr (Value repr) -> repr (Value repr)
```
Is defined in all 4 renderers the exact same way:
```
(#~) = liftA2 unExpr' negateOp
```
Is it worth it to factor out the pattern of `liftAx someFunction param1 param2 ...`? This sort of pattern is extremely common in GOOL.

The pro to doing this would mean the function definition would just be
```
(#~) = negateExpr
```
in all of the renderers, and then we could truly say that every renderer defines this function (and others like it) simply by dispatching to a common function. And then for the paper I could count up all functions with that property between Java and C# and use that as a metric of our re-use between the renderers.

The con to doing this is that there's actually a lot of tedious work that has to be done to make that change. Since `negateExpr` would have to be written in the context of the generic `repr`, we would need to add symantic functions to GOOL for accessing the `Doc` of a `UnaryOp`. And then if we are factoring out this pattern all throughout GOOL, we will need to write these `Doc` accessor functions for probably every one of GOOL's current types. And some of the "bigger" types will need accessors for more than just `Doc`.

An intermediate solution could be to not factor out the lifting, but still factor out the parameter passing, so the function would be defined like:
```
(#~) = fmap negateExpr
```
Then we wouldn't need to write all of the accessors, but we could still say that every renderer defines this function (and others like it) by dispatching to a common function _plus some lifting_. And then that would be the property I would use when counting up functions for our re-use metric. I'm leaning towards this option but want to know your thoughts.

bmaclach:
@JacquesCarette Scratch my intermediate solution -- I wasn't thinking right. The operator parameter would need to be passed regardless, because it comes from a symantic function, so a common function written outside of the context of a `repr` wouldn't know what it was unless we pass it explicitly.  

[...]

PR #1930, opened 2019-09-30
Reduce GOOL code duplication

bmaclach:
First batch of factoring out code for functions defined the same between renderers in GOOL, contributing to #1868.


code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
```diff
@@ -163,6 +170,9 @@ bodyDocD bs = vibcat blocks
   where blocks = filter notNullBlock bs
         notNullBlock b = not $ isEmpty b
 
+oneLinerD :: (RenderSym repr) => repr (Statement repr) -> repr (Body repr)
```
JacquesCarette: One of the things I meant to mention on issue #1868 is that one way to factor things out is to provide `repr`-generic functions (like this one) which works in all cases, by dispatching its work to the right lower-level combinators.

code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
```diff
@@ -702,62 +745,225 @@ classVarCheckStatic v = classVarCS (variableBind v)
...
-notNullDocD :: OpData -> ValData -> ValData -> Doc
-notNullDocD op v1 v2 = binOpDocD (opDoc op) (valDoc v1) (valDoc v2)
+newObjDocD' :: (RenderSym repr) => repr (Type repr) -> Doc -> Doc
+newObjDocD' st vs = getTypeDoc st <> parens vs
 
 listIndexExistsDocD :: OpData -> ValData -> ValData -> Doc
 listIndexExistsDocD greater lst index = parens (valDoc lst <> 
   text ".Length" <+> opDoc greater <+> valDoc index) 
 
+varD :: (RenderSym repr) => Label -> repr (Type repr) -> repr (Variable repr)
```
JacquesCarette: It used to make sense to have all of these being named with a post-fix D, as they used to all operate on a `Doc`. Now that postfix D doesn't make as much sense...  I would seriously consider either dropping the D, or putting all of these in some kind of `.Generic` sub-module meant to be imported qualified. [But that can be done later]

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CSharpRenderer.hs
```diff
@@ -245,20 +230,20 @@ instance VariableSym CSharpCode where
...
 instance ValueSym CSharpCode where
   type Value CSharpCode = ValData
-  litTrue = liftA2 mkVal bool (return litTrueD)
-  litFalse = liftA2 mkVal bool (return litFalseD)
-  litChar c = liftA2 mkVal char (return $ litCharD c)
-  litFloat v = liftA2 mkVal float (return $ litFloatD v)
-  litInt v = liftA2 mkVal int (return $ litIntD v)
-  litString s = liftA2 mkVal string (return $ litStringD s)
+  litTrue = litTrueD
+  litFalse = litFalseD
```
JacquesCarette: one of the things to look at, in the future, will be if all `instance`s of (say) `litFalse` all dispatch to the generic one -- and then maybe remove it from the `ValueSym` class altogether, since it is never overloaded.

in some case, we'll want to keep them anyways. But then a 'better' mechanism might be to use default in the class declaration itself, so that the actual instances can be made much shorter.

JacquesCarette:
I think the failures are due to `hlint`'s new version is pickier.  I'll close and re-open this PR to trigger a recompile (after I merge in @Mornix 's changes).

code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
bmaclach: I intended the `D` here to stand for `Default` (whereas the functions for working with `Doc` end in `DocD`). I like the idea of putting them in a separate module to be imported qualified though. That would be less confusing.

JacquesCarette:
Never mind, there are now conflicts to resolve, so that will need fixed and will trigger a recompile too.

code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
JacquesCarette: Indeed, I'd forgotten about `DocD`.  Such a move can be done incrementally, whenever you have to 'touch' some code anyways, then you can refactor those bits. Or if you're bored and looking for something straightforward to do 😄

PR #1933, opened 2019-10-03
Gool paper first draft

bmaclach:
I've completed a first draft of the paper, excluding the abstract. 

We are less than half a page off from the page limit (12, not including the bibliography), and it still needs an abstract, so keep that in mind with edits or suggestions for changes. Though I'm sure we could trim what I've written if need be, as I think erred on the side of including too many examples/too much detail, especially when describing GOOL's syntax.

Deadline for submission is Oct 18th, two weeks from tomorrow!


Papers/GOOL/GOOL.tex
```diff
@@ -73,7 +73,15 @@
 
 %%%%%%%%%%%%%%%%%%%%%
 %% Useful abbreviations
-\newcommand{\Csharp}{C$^{sharp}$}
+\newcommand{\Csharp}{C$^{sharp}$} %Is this really what we want?
```
JacquesCarette: Oops - \sharp !

JacquesCarette: Will fix.

PR #1934, opened 2019-10-03
displacement symbol changed to u

dusear4la:
Updated displacement and distance symbols.

@smiths, in Drasil, we have both displacement and linear displacement. Not sure if we need both.
image.

Here is the data definition of linear displacement

image.

All other related issues with symbol conflict will be fixed in the next PR. All I did here was to update the symbol of displacement as we discussed.


dusear4la:
Looks like SSP is using displacement, maybe other examples also use it

oluowoj:
@smiths, could you please review so it can be merged, I need this in my next PR update. Thanks

smiths:
@oluowoj, the linear part of "linear displacement" is there to distinguish this displacement from the angular displacement.  It isn't strictly necessary to have the word linear with displacement, but that decision runs throughout the game physics example.  If you have a look at acceleration and velocity, the word linear is used for these as well.  Again, the idea is to distinguish angular from linear terms.  It seems like the displacement without the word linear is actually the odd one out.  There isn't a data definition for that displacement.  I believe we should be able to use linear displacement in place of all of the occurrences of displacement.

JacquesCarette:
Who is @dusear4la ?

smiths:
I had the same question.  My guess is that @dusear4la is @oluowoj, based on the discussion comments.

oluowoj:
@

> Who is @dusear4la ?
Its me. I was logged in to my old account and forgot to log out before opening Drasil page.

oluowoj:
@JacquesCarette , please merge.

PR #1935, opened 2019-10-04
GOOL bugs

bmaclach:
This PR fixes a few small bugs I noticed with GOOL:

- Removed some random characters that must have been mistakenly added to print statements
- Fixed C# to use `intMethod` instead of defining its functionality twice
- Made get and set methods start with a lowercase
- Fixed get and set methods in C++, previously weren't correctly referencing variables on `this`


bmaclach:
@smiths I honestly think it is pure coincidence that \f actually means something. I traced that back to a commit where I apparently added it, but I didn't know what it meant and I don't remember adding it! I have to assume I accidentally typed those characters thinking I was typing somewhere else, and didn't notice.

smiths:
Thank you for the explanation @bmaclach.  That makes sense.  😄

PR #1937, opened 2019-10-10
More GOOL code duplication reduction

bmaclach:
This PR has the final batch of functions that factor out common code between GOOL's language renderers.

Did some minor refactoring as I went, which lead to the removal of some trailing blank lines at the end of the generated code files, thus the changes to stable.


JacquesCarette:
It all seems fine - but I'll wait to merge until that one name has been figured out.

bmaclach:
@JacquesCarette I fixed the name in the latest commit, this should be good to go.

PR #1947, opened 2019-10-23
GOOL Paper Improvements

bmaclach:
This PR proposes changes related to @JacquesCarette's point 1 in his email, making it more clear that GOOL abstracts OO programs, not languages.

Edit: Commit 2b71639454858ec84eabced15657528689271f9d added to address point 2 of the email.

Edit Finished changes for point 3 of the email as of 43a7432df93cfa142ab18e758ea67311be4df439.

Edit: Points 4, 5, and 6 have been addressed (4 only minimally, as there wasn't much to say).


bmaclach:
I think it makes sense to continually add to this PR, since separate PRs would be dependent anyways and would likely have conflicts when it comes time to merge. But let me know if you disagree @JacquesCarette.

JacquesCarette:
For adding the graph: do you know roughly how many lines you need?  There's 11 spare lines now.  I'm sure I can give you 20 more without difficulty.

bmaclach:
@JacquesCarette 20 lines total for the graph and some text referencing the graph should be enough.

bmaclach:
I added the graph -- everything fits, with a few lines to spare.

Interestingly, the graph tells a different story than our lines-of-code analysis. From the graph, it is clear that Python has the least in common with the other languages, whereas the lines of code suggested it was C++ (source code) that was most different. Both agree that C# has the most in common with the others. How best to address this? Remove the lines of code analysis, and replace it with similar ideas but read from the graph instead of the lines of code? @JacquesCarette @smiths

JacquesCarette:
I think we should read from the graph - the LOC was too crude.  Can you fix the text?

I think I am done with my edits.

bmaclach:
Yeah, I'll fix that. And sorry for pushing -- didn't realize you were still editing.

JacquesCarette:
I won't delete the branch yet - if the paper gets accepted, we'll likely have changes to do, and we may as well keep them on this branch.

Issue #1880, opened 2019-09-13
Use State monad in GOOL language renderers


bmaclach:
In the GOOL renderers, it makes sense to store some "extra" information to be used later on to generate better code. An example is storing a list of libraries that are used so that, at the module-level, GOOL can generate only those import statements that are actually needed. For examples like this one, the list would originate at the level of `Value`s, and accumulate as the renderer worked its way up to functions, classes, and finally the module. This accumulating/modifying behaviour can be cleanly implemented with the State monad. 

Note that, if possible, it would be best to do this in small pieces. So we should look for places where the State monad will be useful near the top of the hierarchy, so that we can implement the State monad only for those types near the top, and then slowly spread it down to the lower-levels (as opposed to the example given above where we would need to make everything from `Value`s all the way up to `Module`s use the State monad all at once).

This was discussed at the Sept. 12th, 2019 meeting.

JacquesCarette:
This is done, right @bmaclach ?

bmaclach:
Yes, we now use the `State` monad throughout GOOL. This can be closed.

PR #1955, opened 2019-11-06
Start to use State monad in GOOL

bmaclach:
This PR updates GOOL to start using the State monad. For now only the top of the GOOL hierarchy uses it (`File`s and `Program`s). I will incrementally work my way down the GOOL hierarchy, making each level use the State monad and adding more to the State as I go, in future PRs. 

For now, the `State` holds lists of file paths, which I use to add a new check that throws an error if someone attempts to give 2 different modules the same name. 

(contributes to #1880)

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Build/Import.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Build/Import.hs
@@ -8,27 +8,30 @@ import Language.Drasil.Code.Imperative.Build.AST (BuildConfig(BuildConfig),
   Runnable(Runnable), BuildName(..), RunType(..))
 import Language.Drasil.Code.Imperative.GOOL.LanguageRenderer (doxConfigName)
 
-import GOOL.Drasil (FileData(..), isSource, isHeader, ModData(..), ProgData(..))
+import GOOL.Drasil (FileData(..), isHeader, ModData(..), ProgData(..), 
+  GOOLState(..), initialState, sources)
 
 import Build.Drasil ((+:+), genMake, makeS, MakeString, mkFile, mkRule,
   mkCheckedCommand, mkFreeVar, RuleTransformer(makeRule))
 
 import Control.Applicative (liftA2)
+import Control.Monad.State (State, evalState, execState)
+import Control.Lens ((^.))
 import Data.Maybe (maybe, maybeToList)
 import System.FilePath.Posix (takeExtension, takeBaseName)
 import Text.PrettyPrint.HughesPJ (Doc)
 
-data CodeHarness = Ch (Maybe BuildConfig) Runnable ProgData [Comments]
+data CodeHarness = Ch (Maybe BuildConfig) Runnable ProgData GOOLState [Comments]
 
 instance RuleTransformer CodeHarness where
-  makeRule (Ch b r m cms) = [
+  makeRule (Ch b r m s cms) = [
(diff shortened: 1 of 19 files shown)


code/drasil-code/Language/Drasil/Code/Imperative/Build/Import.hs
```diff
@@ -8,27 +8,30 @@ import Language.Drasil.Code.Imperative.Build.AST (BuildConfig(BuildConfig),
...
 
 import Control.Applicative (liftA2)
+import Control.Monad.State (State, evalState, execState)
+import Control.Lens ((^.))
 import Data.Maybe (maybe, maybeToList)
 import System.FilePath.Posix (takeExtension, takeBaseName)
 import Text.PrettyPrint.HughesPJ (Doc)
 
-data CodeHarness = Ch (Maybe BuildConfig) Runnable ProgData [Comments]
+data CodeHarness = Ch (Maybe BuildConfig) Runnable ProgData GOOLState [Comments]
```
JacquesCarette: You should make `CodeHarness` into a record instead of a (named) 5-tuple.  Then you can use either pattern-matching or accessors, whenever each makes the code cleaner.  I won't make this a blocking issue on merge though.

code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs
```diff
@@ -22,10 +24,11 @@ verbosityToDoc :: Verbosity -> Doc
 verbosityToDoc Verbose = no
 verbosityToDoc Quiet = yes
 
-makeDoxConfig :: ProjName -> ProgData -> OptimizeChoice -> Verbosity -> Doc
+makeDoxConfig :: ProjName -> State GOOLState ProgData -> OptimizeChoice -> Verbosity -> Doc
 makeDoxConfig prog p opt v = 
   let fs = map filePath (filter (\f -> not (isEmpty $ modDoc $ fileMod f) && 
-             (isMainMod (fileMod f) || isHeader f)) (progMods p))
+             (isMainMod (fileMod f) || isHeader f)) (progMods (evalState p
```
JacquesCarette: It's usually not a good sign if you're using `evalState` in many places. It might be that really you do want `ProgData` here, and make it the responsibility of the caller to do the `evalState`.

code/drasil-code/Language/Drasil/Code/Imperative/GOOL/Symantics.hs
```diff
@@ -10,23 +10,26 @@ import Database.Drasil (ChunkDB)
...
-import GOOL.Drasil (ProgData)
+import GOOL.Drasil (ProgData, GOOLState)
 
+import Control.Monad.State (State)
 import Text.PrettyPrint.HughesPJ (Doc)
 
 class (AuxiliarySym repr) => PackageSym repr where
   type Package repr 
-  package :: ProgData -> [repr (Auxiliary repr)] -> repr (Package repr)
+  package :: State GOOLState ProgData -> [repr (Auxiliary repr)] ->
```
JacquesCarette: Likewise, passing in a `State` like this doesn't feel right.  If `package` was all in the `State` monad, that would be different.

bmaclach:
@JacquesCarette I've addressed all feedback in 1466d2416cb0663badc61daae210edbfc5e23d8d

PR #1956, opened 2019-11-08
Adding State to Modules, Classes, StateVars, Methods

bmaclach:
This PR adds State to some more of GOOL's types, and keeps track of the main module in the State, including ensuring there is only one main function defined.


bmaclach:
@JacquesCarette The majority of the places where I've been calling `evalState` are in the functions that extract a bit of information from the monad. `moduleDoc`, `moduleName`, and `getFilePath` are examples. I'm not sure of a better way... maybe these functions should have the return type still wrapped in `State`? So `moduleDoc` instead of returning a `Doc` would return a `State GOOLState Doc`?

JacquesCarette:
Yes, I think you need to make things more stateful, as you describe.  Then you'd have few calls to `evalState`, all at the very top.

bmaclach:
Okay, thanks, I'll do that.

PR #1957, opened 2019-11-12
Start to make accessor functions stateful

bmaclach:
@JacquesCarette I'm looking for feedback on what I did here, specifically in 323e4ef, before I continue down this path.

I'm trying to do what we talked about here: https://github.com/JacquesCarette/Drasil/pull/1956#issuecomment-552482087

When I updated `getFilePath` to return `State GOOLState FilePath` instead of just `FilePath`, I started getting errors saying that multiple main functions were defined, even though that wasn't the case.

I eventually figured out this was happening because the single `mainFunction` was essentially getting run twice, where the second time was happening in `getFilePath`. It's complicated, but basically since `getFilePath` needed to return the `FilePath` which it gets from reading the passed `FileData`, that required an extra evaluation of the entire passed `Module`. If that doesn't make much sense, here is some Haskell-style pseudocode I put together that shows how the `State` evolves from methods/classes to program, as if it were happening all in one function:
```
theGoolProgram = do
  files <- sequence (
    cmt <- (
      lns <- (
        fp <- (
          file' <- (                           -- ****first evaluation of the passed file****
            mod <- (
              methods <- sequence ()
              classes <- sequence ()
              return $ md n m d
            )
            updatedMod <- return $ md n m d
            s <- get
            put $ setMainMod s mod
            return $ vf mod
          )
          return $ filePath file'
        )
        return $ moduleDox fp
      )
      return $ pyDocComment lns
    )
    file <- (                              -- ****second evaluation of the passed file****
      mod <- (
        methods <- sequence ()
        classes <- sequence ()
        return $ md n m d
      )
      updatedMod <- return $ md n m d
      s <- get
      put $ setMainMod s mod
      return $ vf mod
    )
    return $ commentedModD cmt mod
  )
  return $ progD n files
```

This problem would only get worse as I changed more accessors to be stateful -- each time one was used, it would cause an extra evaluation of the passed value.

To fix this, I decided I should store directly in the `State` any information that we have an accessor for. So, for example, in 323e4ef, I added the "current file path" to the State. Then `getFilePath`, instead of being part of a GOOL typeclass, is just a regular function that reads the current file path from the `State` and returns the result also in `State` (This means that we can only get the current file path instead of getting the file path for any specific `File` we pass in, but our only use case for this function was to get the current file path, so I think that's okay). Then I had to switch the order of the parameters to `commentedMod` so that the `File` came first and the `BlockComment` second -- that way, the evaluation of the `File` updates the current file path in the state, and then the evaluation of the `BlockComment` can read the (up to date) current file path and include it in the comment. It seems unexpectedly fragile that the order of the parameters has such an impact, but maybe that's just something we have to live with when using the `State` monad?

I would do something similar for all of the other accessors, if you think this is the right way to move forward.

### DIFF
--- a/code/drasil-gool/GOOL/Drasil/Data.hs
+++ b/code/drasil-gool/GOOL/Drasil/Data.hs
@@ -55,8 +55,8 @@ data ModData = MD {name :: String, isMainMod :: Bool, modDoc :: Doc}
 md :: String -> Bool -> Doc -> ModData
 md = MD
 
-updateModDoc :: Doc -> ModData -> ModData
-updateModDoc d m = md (name m) (isMainMod m) d
+updateModDoc :: (Doc -> Doc) -> ModData -> ModData
+updateModDoc f m = md (name m) (isMainMod m) (f $ modDoc m)
 
 data MethodData = MthD {isMainMthd :: Bool, mthdParams :: [ParamData], 
   mthdDoc :: Doc}
--- a/code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
+++ b/code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs
@@ -69,10 +69,8 @@ import GOOL.Drasil.Data (Terminator(..), FileData(..), fileD, updateFileMod,
   ValData(..), vd, Binding(..), VarData(..), vard)
 import GOOL.Drasil.Helpers (angles, doubleQuotedText, hicat, vibcat, vmap, 
   emptyIfEmpty, emptyIfNull, getInnerType, getNestDegree, convType)
-import GOOL.Drasil.State (GOOLState)
 
 import Control.Applicative ((<|>))
-import Control.Monad.State (State)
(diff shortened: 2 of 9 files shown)


bmaclach:
@JacquesCarette I guess I should also mention some other alternatives to get around this problem:

1. Change the way I'm using state such that no problems are caused if the same things get evaluated more than once.
2. Keep the accessors the same as they were before, which would mean keeping all of those uses of `evalState`

JacquesCarette:
I'm going to look at the actual code now.  But I can comment on the alternatives:

1. is basically not viable. If it was `Reader`, sure, but if we're going to use `State`, let's indeed be stateful. And state, like in imperative programs, can be a real pain.
2. this is less bad than 1, but doesn't feel "right".

I do understand the question of fragility, and that too seems sub-optimal.  But here, the details will matter.

bmaclach:
Build failed during TeX building for some reason... anyway, I'll close this PR until I make the discussed changes.

PR #1960, opened 2019-11-21
Zoom into and out of MethodState

bmaclach:
This PR splits `MethodState` off of `GOOLState`, where `MethodState` holds state that is local to `Method`s. We use `zoom` to lift `State GOOLState` type values to `State (GOOLState, MethodState)` type values, and also to lift `State (GOOLState, MethodState)` type values back into `State GOOLState` type values. Which means I had to build lenses going both ways, and to get from `State GOOLState` to `State (GOOLState, MethodState)` the lens is sort of backwards -- it "focuses in" on something even bigger than what it starts with, which sounds impossible, but works because when we need to lift "something" without `MethodState` into a computation with `MethodState` we can just initialize the `MethodState` for the "something" however we want, because it doesn't actually care what its `MethodState` is, only that its type does include `MethodState`! I'm kind of amazed that this works! 

Another thing that happened was I initially switched all `Method` types to use `MethodState`, and left all other types to just use `GOOLState`. But the `BlockComment` passed to `commentedFunc` needs to read `MethodState` to get the parameter names, and since I had left it only with `GOOLState` it couldn't do that, and the documentation for parameters disappeared. I was hesitant to switch the `BlockComment` to use `MethodState` to fix this problem because `BlockComment`s in general aren't specifically related to `Method`s. But then I realized that, since we switched the order of the monads, I can make the `BlockComment` passed to `commentedFunc` use `MethodState` while keeping `BlockComment`s passed to other functions using `GOOLState`! I just had to make the constructor for the `BlockComment` generic on the type of State it has, which worked like a charm. In other words, I'm really happy with how nicely all these changes worked out :)

I'll be curious to see how this scales up, if in the future we have more types of local state beyond just `MethodState` (even now, it probably makes sense to add `ModuleState` for the `currModName` field and/or `FileState` for the `currFilePath` field). I think we will still be able to make everything work, though we will likely have to build a whole bunch of lenses for transitioning between all the different types of State.

This PR depends on #1959, so that should be merged first. The new commits for this PR start with 
60398ab.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Comments.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Comments.hs
@@ -4,7 +4,7 @@ module Language.Drasil.Code.Imperative.Comments (
 
 import Language.Drasil
 import Database.Drasil (defTable)
-import Language.Drasil.Code.Imperative.State (State(..))
+import Language.Drasil.Code.Imperative.State (DrasilState(..))
 import Language.Drasil.CodeSpec (CodeSpec(..), CodeSystInfo(..))
 import Language.Drasil.Printers (Linearity(Linear), sentenceDoc, unitDoc)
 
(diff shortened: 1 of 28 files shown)


JacquesCarette:
Yes, this is definitely really nice.  And I too am hopeful this will generalize.

PR #1965, opened 2019-12-04
Add State to Parameters in GOOL

bmaclach:
This PR makes GOOL `Parameter`s stateful, where during the state transition an error is thrown if multiple parameters for a given function are the same. This was previously handled by a `checkParams` function, but now that it is handled by State, it enabled some simplifications to the messy `intMethod` and `intFunc` function implementations in the renderers.

@JacquesCarette I discovered while working on this that if the source and header instances of a GOOL function just call other GOOL functions exactly the same way, our `Pair` instance for that function can just do those calls directly, instead of breaking down the pair, dispatching to the two instances, and then putting the results back together in a pair. For example, both instances of `docInOutFunc` are defined as `docInOutFunc n = G.docInOutFunc (inOutFunc n)`. So, whereas the `Pair` instance is currently:
```
docInOutFunc n s p desc is os bs b = on2StateValues pair (docInOutFunc n 
    (pfst s) (pfst p) desc (map (mapPairSnd pfst) is) (map (mapPairSnd pfst) os)
    (map (mapPairSnd pfst) bs) (pfst b)) (docInOutFunc n (psnd s) (psnd p) desc 
    (map (mapPairSnd psnd) is) (map (mapPairSnd psnd) os) (map (mapPairSnd psnd)
    bs) (psnd b))
```
We can instead make it the same as the other instances: `docInOutFunc n = G.docInOutFunc (inOutFunc n)`.

Do you think we should do this? Though it would simplify the `Pair` instance, I can think of two reasons not to do this:
1. Even if we do it, we still need to duplicate the instances in `CppSrcCode` and `CppHdrCode` because we can't just leave those instances blank (would get warnings or errors if we did).
2. This would make the `Pair` instance less re-usable. Currently, if we ever implement another language renderer that needs 2 files for every module, we can just copy and paste the same `Pair` instance we already have and just find and replace `CppSrcCode` and `CppHdrCode` with the right types. But if we made this change, then these instances may also need to be re-defined for the new language.

---

Nobody is obligated to read the rest of this comment, but it is a story about a big roadblock I ran into during this PR that I wanted to record because it is the single most confusing thing that I've gone through since I started programming. 

After adding `State` to the parameters, I was hitting the error in C++ saying that a function had duplicate parameters. Since none of our functions in our examples actually have duplicate parameters (and since the other target languages were not hitting this error), the only explanation was that in C++ for some reason the state transitions for adding the parameters to the state were happening twice. Just to test, I changed it so only the source code half of the C++ renderer would be used, and the error went away. So something in the `Pair` instance was causing the error. My first thought was that this was happening because the `CppSrcCode` instance and `CppHdrCode` instance were both doing the state transitions. I had originally hoped that the two instances would be able to do state transitions in parallel, but upon further investigation it became clear why that wasn't happening, and that they needed to be done in series. The solution then seemed clear: I had to update `CppHdrCode` instances to not repeat the state transitions, since they would already be done by `CppSrcCode`. So that's what I did.

Only I was still getting the same error. At this point I was very confused. Since the `CppHdrCode` instance was not doing the state transitions anymore, the only way this could be happening was if the `CppSrcCode` state transitions were running twice. But I had already tested running the `CppSrcCode` half alone, and everything had worked! So the `Pair` instance was somehow causing the `CppSrcCode` instances to run twice. I investigated further and uncovered more details: The error was only happening for functions defined with `inOutFunc`, and only happening when comments were turned on. So `inOutFunc` and `docFunc` worked fine, but `docInOutFunc` hit the error. I continued, and things got even more confusing. Even though the state transitions had already been removed from the `CppHdrCode` instance, I found that if I updated the `CppHdrCode` instance of `docInOutFunc` to just do nothing (i.e. return an empty function), the error went away. So it was as if the `CppHdrCode` instance had somehow been calling the `CppSrcCode` instance, though this made no sense as that would have surely led to type errors if it were happening. And indeed, I hardcoded some errors into functions in both instances and found that the `CppHdrCode` ones were being called as expected, and the `CppSrcCode` ones were not being called by the `CppHdrCode` instance (which I confirmed by updating the `CppSrcCode` instance of `docInOutFunc` to do nothing). So everything seemed to be working normally, but I was still hitting the duplicate parameters error. Somehow, the mere presence of a meaningful `CppHdrCode` instance for `docInOutFunc`, even one that did not perform state transitions, was causing the `Pair` instance to run the `CppSrcCode` half twice.

So many of my observations were contradicting each other, and I was as confused as I ever remember being since I started programming. I finally stumbled upon a potential solution -- I noticed the `CppSrcCode` and `CppHdrCode` instances for `docInOutFunc` were the same, in that they each called (their corresponding instance of) `inOutFunc` and then `commentedFunc`. It occurred to me that for higher-level functions where both instances are formed by calling lower-level functions in the same way, couldn't I just do that behaviour directly in the `Pair` instance, instead of extracting each half of the pair, feeding it to its corresponding instance, and then re-combining the results into a pair? Turned out the answer was yes. Updating the `Pair` instance of `docInOutFunc` to just do what the other instances were doing directly (and with some minor modifications to related functions), the error finally went away.

But I wasn't satisfied, because I still didn't understand what had caused the error in the first place, and my solution was more of a workaround than a real fix. So I reverted that fix, and went through the code, following the path `docInOutFunc` would take with a fine-toothed comb, and comparing it to the (working) `docFunc`. Finally, I found an oddity. In `cppCommentedFunc`, we had:
https://github.com/JacquesCarette/Drasil/blob/d642ccbdb7c2e5f28697d8f1868a34948dd2cc95/code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs#L2067-L2080
The line `f <- fn` is what was supposed to trigger the state transitions for the passed `Method`, including adding the parameters to the state, but if you look at the `if-then-else` conditions you see that there are some cases where `cppCommentedFunc` just returns `fn`. In those cases, `f` is never used, so, if I understood Haskell's laziness correctly, that would mean those state transitions would never truly happen. So, even though my bug was that the state transitions were happening twice, I had found a reason why they would be happening 0 times. Another contradiction, but this was a bug I should fix nonetheless, so I updated `cppCommentedFunc` to return `toState f` instead of `fn` in those cases, so that `f` would actually be used and the associated state transitions would be forced to happen. And, lo and behold, the error went away. My best guess is that the fact that the state transitions were not being run at this point was leading to a scenario higher up where the state was expected to be further along than it really was. I still don't have a full understanding of why I was getting the error, and especially why my test where I updated the `CppHdrCode` instance to do nothing miraculously made the error not get hit despite having seemingly nothing to do with it, but I am at least confident I found and fixed a bug that was causing it, so I can move on.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/GenerateGOOL.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/GenerateGOOL.hs
@@ -72,7 +72,8 @@ fAppInOut m n ins outs both = do
     (eMap $ codeSpec g) == Just cm then inOutCall n ins outs both else 
     selfInOutCall m n ins outs both
 
-mkParam :: (RenderSym repr) => repr (Variable repr) -> repr (Parameter repr)
+mkParam :: (RenderSym repr) => repr (Variable repr) -> 
+  MS (repr (Parameter repr))
 mkParam v = paramFunc (getType $ variableType v) v
   where paramFunc (List _) = pointerParam
         paramFunc (Object _) = pointerParam
--- a/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
@@ -141,8 +141,8 @@ genConstructor :: (RenderSym repr, HasUID c, HasCodeType c, CodeIdea c) =>
 genConstructor n desc p = genMethod (constructor n) n desc p Nothing
 
 genMethod :: (RenderSym repr, HasUID c, HasCodeType c, CodeIdea c) => 
-  ([repr (Parameter repr)] -> repr (Body repr) -> MS (repr (Method repr))) -> 
-  Label -> String -> [c] -> Maybe String -> [repr (Block repr)] -> 
+  ([MS (repr (Parameter repr))] -> repr (Body repr) -> MS (repr (Method repr))) 
+  -> Label -> String -> [c] -> Maybe String -> [repr (Block repr)] -> 
   Reader DrasilState (MS (repr (Method repr)))
 genMethod f n desc p r b = do
   g <- ask
--- a/code/drasil-gool/GOOL/Drasil/Helpers.hs
+++ b/code/drasil-gool/GOOL/Drasil/Helpers.hs
@@ -2,8 +2,8 @@ module GOOL.Drasil.Helpers (verticalComma, angles, doubleQuotedText, himap,
   hicat, vicat, vibcat, vmap, vimap, vibmap, emptyIfEmpty, emptyIfNull, 
   mapPairFst, mapPairSnd, toCode, toState, onCodeValue, onStateValue, 
   on2CodeValues, on2StateValues, on3CodeValues, on3StateValues, on4CodeValues, 
-  on5CodeValues, onCodeList, onStateList, on2StateLists, on1CodeValue1List, 
-  on1StateValue1List, getInnerType, getNestDegree, convType
+  onCodeList, onStateList, on2StateLists, on1CodeValue1List, on1StateValue1List,
+  getInnerType, getNestDegree, convType
 ) where
 
 import Utils.Drasil (blank)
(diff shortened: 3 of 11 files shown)


smiths:
@bmaclach, thank you for your explanation of the bug you found and the steps  you took to diagnose it.  You have my sympathy.  That sounds truly frustrating!  I think it is important that you recorded your experience here and that you gave your steps for investigation.  Although it isn't as clean/linear/satisfying a story as we would like, I think it has a good lesson.  You did things in a systematic way and you found a bug to fix.  Not exactly the bug that explains the problem, but it was a bug that needed to be fixed.  I'm not sure if it will fit the narrative of your thesis document, but I could see including this anecdote in an appendix.  Future students could benefit from your explanation.

JacquesCarette:
We should sit down and look at `Pair` together. Some simplifications might be worth it, while others not, and it's not going to be straightforward to tell.
[And now on to reading the saga...]

JacquesCarette:
That is quite the saga! Your reasoning seems sound though.

bmaclach:
@JacquesCarette For sitting down to look at `Pair`, does Monday work for you (the 9th)? 1:30 PM?

PR #1968, opened 2019-12-13
Stateful Values, Functions, Variables, Types 

bmaclach:
This PR updates GOOL to have State with `Value`s, `Function`s, `Variable`s, and `Type`s. 

This PR is rather large but very mechanical -- it "just" adds state to type signatures and then makes it compile. It doesn't yet do anything meaningful with the state. 

One thing I want to bring up is that I need to filter a list of variables to remove those that are objects. Now that variables have state, this task is filtering a list of stateful things. After some googling, it seems that filtering a list of monads isn't possible in general, unless you want the return type to be a monadic list, which doesn't work for my purposes (i.e. you can filter from `[m a] -> m [a]` but not `[m a] -> [m a]`). The only way I could think to get around this was to use an `evalState` in my filter function. Is this okay, @JacquesCarette? The relevant code is here:
https://github.com/JacquesCarette/Drasil/blob/b8638cda56645567bae6e33492a95d6a6a39385b/code/drasil-gool/GOOL/Drasil/LanguageRenderer.hs#L707-L710


JacquesCarette:
Are you intending to do the changes as discussed above for this PR, or wait for another?  I'm fine either way. The current code isn't wrong.

bmaclach:
@JacquesCarette Go ahead an merge this one, I'll create a new one for the changes.

JacquesCarette:
Well, except maybe for `filterOutObjs`.  I think the current code seems wrong. Obviously, it works, but it feels like if things really were quite 'effectful', then you might be running the effects prematurely or, much worse, twice.

I think the root of the problem might be that you're applying the filter in the 'wrong' place - I don't know if it's too late or too early though. There ought to be a time in the code where you really do have a list of `Variable`s, and there is the correct place to put the filter.

JacquesCarette:
Ok, I'll merge, even with the problem with `filterOutObjs`.

bmaclach:
> Well, except maybe for `filterOutObjs`. I think the current code seems wrong. Obviously, it works, but it feels like if things really were quite 'effectful', then you might be running the effects prematurely or, much worse, twice.

I think you're right that in this case it would be running the effects twice. Given our current plan for how `Variable`s will modify state, which is to keep a list of needed imports, this isn't a problem functionally, because even with everything only running once, we will need to have logic to not add the same import twice, since two different variables might require the same import. 

In general, I think at this low level (`Variable`s, `Value`s, etc.), the goal of only running state transitions once might be incredibly difficult to hit. For example if the same variable/value is used in multiple different places, the state transitions caused by that variable/value must happen multiple times. I guess the problem you're getting at is that in this case with `filterOutObjs` we would be running the effects twice _for a single use case_, which definitely doesn't seem ideal.

> I think the root of the problem might be that you're applying the filter in the 'wrong' place - I don't know if it's too late or too early though. There ought to be a time in the code where you really do have a list of `Variable`s, and there is the correct place to put the filter.

The constructors for `Variable`s immediately add `State`, so there really never is a time when I have a list of plain, unstateful `Variable`s that I can filter. For context, the reason I need to do this filter is when using `inOutFunc`, if an output is an object, we don't want to explicitly return that object in the generated code, since an object has state that will be modified. So this filter isn't strictly necessary, because I don't think things would break if we returned objects, but it would be unidiomatic. Maybe this is a sign that it was wrong to say that an object is an output of a function if the function modifies the object's state? Could state transitions fall into a fourth category of function parameters (neither input, output, nor both)?

JacquesCarette:
Yes, it is wrong to say that a function returns an object if it doesn't return that object... Modifying an object 'in place' as it were, is a different effect then returning. So yes, this is another case. However, it might be a better design to use modifiers on in/out/both rather than explode the number of cases.

By default, an input parameter should be considered read-only, output write-only, and both, well, both.  But an input parameter could be 'marked' as writeable, to enable state transitions; similarly, one could think of being able to read an 'out' variable [this might help eliminate useless temporaries].

bmaclach:
@JacquesCarette Okay, so updating `inOutFunc` to not expect state transitions to be included as outputs should be easy and will remove the need for `filterOutObjs` and its associated problems.

The other part of your comment, about the modifiers, I don't think we need yet given how we currently have things set up. `inOutFunc` gets passed a function body, so it can't enforce that an input is only used as read-only in that body. To be able to enforce that, we would either have to use `State` to throw an error if someone tries to write to a read-only variable, or constrain `inOutFunc` for more specific use cases (or more likely, abstract more specific patterns that fall within the umbrella of the input-output-functions pattern). As things currently stand, a writable input would not be treated any differently by `inOutFunc`, so if we marked it as writable we wouldn't actually be using that information. If a function performs a state transition on some object, passing the object to the Inputs list of `inOutFunc` (rather than the `both` list like we do currently) is enough to achieve the behaviour we need right now.

JacquesCarette:
If we don't need a feature yet, then it shouldn't be implemented (yet). We don't want to implement too fancy a type-system in Drasil, at least not until it is needed.

PR #1970, opened 2019-12-19
Information collection pass in GOOL

bmaclach:
While implementing the automation of the generation of import statements, I discovered I needed an information collection pass for C++ so that we can determine which objects have a type that needs to be imported from another module.

This PR adds the infrastructure for an information collection pass in GOOL. Currently it just creates a map from module names to class names, but we can add to the `CodeInfo` interpretation to store more information as we need it. The pass is also written generically, so the collected information is available to all language renderers.

The actual implementation of automated import statements isn't here, that will be in the next PR.


bmaclach:
Push build failed due to connection errors, but PR build passed so this should be good to go.

PR #1971, opened 2019-12-20
Automate imports in GOOL

bmaclach:
This PR makes GOOL generate import statements automatically based on what is used in the file, removing unnecessary imports from the generated code. 

@JacquesCarette @smiths Some of the imports look messy due to their order (ex. the import for `System.Collections.Generic` appearing before the import for `System` in C#). Also the order is inconsistent between files. What would be an appropriate way to sort the imports? Alphabetically?

#1970 needs to be merged first, ac7ab1b is the first new commit.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code/Imperative/ConceptMatch.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/ConceptMatch.hs
@@ -5,7 +5,7 @@ module Language.Drasil.Code.Imperative.ConceptMatch (
 import Language.Drasil.CodeSpec (Choices(..), CodeConcept(..), 
   MatchedConceptMap)
 
-import GOOL.Drasil (ProgramSym, ValueSym(..), GS)
+import GOOL.Drasil (ProgramSym, ValueSym(..), FS)
 
 import Prelude hiding (pi)
 import qualified Data.Map as Map (map)
@@ -17,6 +17,6 @@ chooseConcept chs = Map.map (chooseConcept' chs) (conceptMatch chs)
           "ConceptMatchMap"
         chooseConcept' _ cs = head cs
 
-conceptToGOOL :: (ProgramSym repr) => CodeConcept -> GS (repr (Value repr))
+conceptToGOOL :: (ProgramSym repr) => CodeConcept -> FS (repr (Value repr))
 conceptToGOOL Pi = pi
 
--- a/code/drasil-code/Language/Drasil/Code/Imperative/ConceptMatch.hs
+++ b/code/drasil-code/Language/Drasil/Code/Imperative/ConceptMatch.hs
@@ -5,7 +5,7 @@ module Language.Drasil.Code.Imperative.ConceptMatch (
 import Language.Drasil.CodeSpec (Choices(..), CodeConcept(..), 
   MatchedConceptMap)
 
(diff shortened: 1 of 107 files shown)


JacquesCarette:
Our collection of known imports should satisfy `Ord`, and that can be used to sort.  What I mean is that the 'symbolic' list should be sorted, not the rendered version.

You might also consider keeping the list always sorted as you accumulate, rather than sorting afterwards.  You can even thinking of keeping the 'list' as a bit-vector, which definitely would take care of all of that, as there would be no other choice.

smiths:
@bmaclach, for the ordering of the imports, I think we should follow an appropriate style guide, or at least do something consistent with standard practice.  For Python Pep8 is the generally accepted style guide.  Import ordering is discussed at:
https://www.python.org/dev/peps/pep-0008/#imports
The order is
1. Standard library
2. Related third party
3. Local application/library specific

The standard seems to be silent on the ordering within those categories, but a little googling suggests, that alphabetical within each category is common advice.  For instance, this question is discussed at:

https://softwareengineering.stackexchange.com/questions/377115/c-namespace-ordering-coding-convention

@bmaclach, it probably would be worth a few minutes of your time to google around for the other languages style guides.  I found coding conventions for C# at:

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions

However, I didn't see anything here that specifically addresses import order.  The stack exchange link above suggests there is something though, so you might want to do a deeper search than I have done.

It is far from a priority, but it would be an interesting exercise to run our generated code through linters that measure against coding standards.  For instance, how would our generated python code do if we ran it through flake8.  It would be a cool claim to be able to say that we generated pep8 compliant Python code.  If you think that idea is worth pursuing, please create a separate issue so that we don't lose track of that thought.  :-)

bmaclach:
Looking at some style guides for our target languages, most suggest a particular order for "groups" of imports, like the one for Python that @smiths showed above. Within groups, they don't always specify, but those that do (Microsoft's guide for C#, Google's guide for Java) suggest alphabetical. 

I can already do the inter-group ordering because I store language imports and module (i.e. local) imports separately in the state. We don't have any "third-party" type imports yet, but when we do (for ode solvers, for example) I can similarly store those in a different field, allowing each language to place that group in the correct spot in the order for that language.

For ordering within groups, since all of our current languages suggest alphabetical ordering, and since that is Haskell's built-in ordering for strings, does it still make sense to define custom data types with custom instances of `Ord`, @JacquesCarette? Doing so comes with its own challenges: We would ideally want a single type for imports, so that we don't need to store a separate field in the `State` for each language, but the ordering of imports may be different per-language, which I guess means our type would have to have different elements for each language. Something like:

`data Imports = CppMath | JavaMath | PythonMath | CppIO ...`

I don't know where that would rate on a "good Haskell" scale, but it doesn't seem quite right to me. To me, a simple sort on the strings is the cleanest way to do this, at least until we encounter a case where a different, non-alphabetical ordering is needed. Thoughts?

JacquesCarette:
I was originally thinking we'd have a "global map" of the different concepts, for all of Drasil, and then each language could map that to its own needs (i.e. which imports that corresponds to).  It's still not a bad idea, but probably overkill.  Furthermore, it seems to go against the grain of obeying the per-language recommendations for sorting.

So let's stick with the current design, as it accommodates what @smiths was pointing out more readily.

PR #1974, opened 2019-12-20
mkQuantDef fix

bmaclach:
I noticed one of GlassBR's variables had type `double` when it was supposed to be `int`. Traced it back to `mkQuantDef`, which was overwriting all types to be `Real`.


bmaclach:
It definitely took some digging!

PR #1976, opened 2019-12-20
Fix bug in Java renderer

bmaclach:
This PR uses State to fix a small bug in Java that would lead to errors if `inOutCall` was used twice in the same scope.


bmaclach:
Last fail was due to a connection failure, PR build passed though,

PR #1978, opened 2019-12-23
GOOL Class state

bmaclach:
This PR adds `ClassState` to GOOL, which stores the current class name. A whole bunch of functions required the class name to be passed as a parameter, but now it can just be read from the state.


JacquesCarette:
This is now causing the build to fail (hlint issues).  My guess is that hlint got upgraded, so that there are new warnings.  Could you look into this and fix please?

PR #1980, opened 2020-01-12
Fix diff tool when used on machines that use CRLF line endings

szymczdm:
On windows machines, depending on settings, building will generate a lot of non-empty log files because of the additional carriage return at the end of lines.


JacquesCarette:
The lint problem was separate - and should be fixed.  I'll close and reopen to get the builds redone.

PR #1981, opened 2020-01-14
GOOL presentation first draft

bmaclach:
I've put together a first draft of the slides for presenting at PEPM. This draft is more wordy than the final draft will likely be, as we discussed.

I've included comments/questions in red throughout.

I couldn't find on the PEPM website or in the emails we were sent a precise specification of presentation length. On the program for the conference, it is scheduled for 20 minutes, but there is no time between the scheduled end of one presentation and the start of the next, so I'm guessing this is more like 15 minutes and then questions? You would probably know better from experience.

The line about the live demo, from the PEPM website, is "In the case of tool demonstration papers, a live demonstration of the described tool is expected." I included some ideas for demos in the red comments in the slides.

When it comes to making the slides less wordy, should I in general aim for a large number of slides (so that some slides may only justify a few seconds of presentation time) or few? For example if we do a large number of slides, we might have a new slide for each requirement in the requirements section, even though most can be explained very quickly. Whereas if we aim for fewer slides, we'd have one slide that encompasses all of the requirements. 

@smiths I don't know if you were planning on reviewing this, but I'll add you as a reviewer just in case.


JacquesCarette:
Yes, assume 15+5.

bmaclach:
I've addressed the feedback. There are a couple more comments (Patterns slide, Future slide).

bmaclach:
With regards to slide numbers, I've heard the advice to avoid them because they get the audience thinking about how far along the presentation is/when it's going to end. But I can add them back if @JacquesCarette likes them too.

smiths:
Yes, I agree that the audience can get fixated on the slide number.  To be honest, I like them as a presenter so that I can gauge how many slides I have left to present in the time available.  If @JacquesCarette does not need that "crutch," then leaving them out makes sense.

JacquesCarette:
No slide numbers.  Real comments will come tomorrow morning.

JacquesCarette:
Afternoon, not morning, apparently...

JacquesCarette:
Re: the requirements slide, I'm the one who asked for the reveal. It will allow me to talk about each one in turn without having the audience jump ahead.

JacquesCarette:
For the 'Blocks' slide -- some example code would really help on this slide.

bmaclach:
> For the 'Blocks' slide -- some example code would really help on this slide.

GOOL or target or both?

JacquesCarette:
For Blocks: one target code, with comments, is sufficient, thanks.

bmaclach:
@JacquesCarette Added some Python code to the Blocks slide.

JacquesCarette:
Thanks - exactly what I was looking for.

PR #1982, opened 2020-01-16
Automate "throws Exception" in Java

bmaclach:
This PR updates GOOL's Java renderer so that we no longer generate "throws Exception" for every single method. Instead we use state to keep track of which exceptions are thrown by a given function, and declare only those that we need.

This change was necessary for my work on ODE solvers to compile, which is why I'm doing it now.


JacquesCarette:
This does feel like progress.  Too bad there are some generic `Exception` still thrown, but one step at a time!

bmaclach:
@JacquesCarette Just for the record, the reason we still throw generic `Exception`s is because GOOL's `throw` function maps to those generic `Exception`s in Java.

JacquesCarette:
Ah indeed.  We'll need to fix that, in time.

PR #1997, opened 2020-01-28
Re-implement initializer lists in GOOL

### LINKED ISSUE #1991: Implement initializer list in GOOL
The ODE solver in C++ requires an initializer list in a constructor definition. Initializer lists aren't available in our other target languages, and are really just an alternative way of writing assignment statements in a constructor. So I think that rather than implementing a GOOL method for initializer lists, the right way to do this is probably to have the C++ renderer automatically use an initializer list if a constructor body has direct assignments from parameters to state variables (with the help of `State`).

bmaclach:
GOOL's `constructor` function now takes a list of initializers as a parameter. Also added smart constructors for the no-initializer case and no-body case. Initializer lists in C++ are now generated based on this parameter, instead of automatically using `State`.

Closes #1991.


JacquesCarette:
Can you deal with `hlint` and the conflicts?

bmaclach:
Hlint fixed and conflicts also fixed.

Issue #1999, opened 2020-01-29
Implement "unmodular" option for code generation


bmaclach:
This was talked about at the meeting on January 16th. We should provide a `Choice` for code generation to have the generated program not be modularized at all -- to generate a "big blob", as we called it.

bmaclach:
Some clarification questions for @JacquesCarette and/or @smiths:
1. For the unmodular option, should we still keep different functions (for reading input, checking constraints, doing calculations, etc.) and just put them all in the same file? Or would there only be one function (`main`) and all of the functionality that is currently in other functions would be inlined?
2. Related to 1., if we are inlining everything into a single main function, then the `Bundled` input choice would be incompatible with the `Unmodular` choice, right? Because bundled input requires writing a class for storing inputs, which would be a different module.

smiths:
I think we want to select whatever option is easiest.  At this point we don't need to have the best options for design.  The most important thing is to have options that we can show off.  My guess is that if you keep functions that will make it easier.  You can put all the functions and everything in one file, but you don't need to put everything in separate files.  That would let you keep the Bundles option, I believe.  The class doesn't have to be in its own file.

At some point we should do an analysis of our design options and the relationships between them.  I do think that we already have cases that are incompatible.  It would be nice if we had a way to document this.

Best wishes.

JacquesCarette:
I like the idea of virtual knobs you can fiddle with, that could go from very modular to maximally unmodular (one giant function). And yes, some options will be incompatible/irrelevant for some configurations.

Right now, whatever's easy to do.

bmaclach:
I implemented a version of an unmodular design where there are still multiple functions/classes, but they are all in one file (#2005).

Still to do: the "fully" unmodular design choice, where all that is generated is a single main function that does everything required for the program.

balacij:
Pulling in information from #3247 

> It would be "nice" to be able to also obtain this version of projectile:
> 
> ```python
> import math
> 
> v, t, p1 = 5.1, 43.0, 12.89
> d = 2.0 * v ** 2.0 * math.sin(t) * math.cos(t) / 9.8 - p1
> if (math.fabs(d / p1) < 2.0e-2) :
>     return "The target was hit."
> elif (d < 0.0) :
>     return "The projectile fell short."
> else :
>     return "The projectile went long."
> ```
> 
[...]

PR #2005, opened 2020-02-06
Unmodular choice for code generation

bmaclach:
This PR adds the `Unmodular` choice for code generation. After my first attempt at implementing it, the code it was generating wasn't right, exposing a bunch of weaknesses throughout the generator, mostly because the generator assumed there would be a certain degree of modularity and wasn't prepared for everything to be in one file. So it took more work than expected to fix up the generator (and GOOL, in some cases) and get things working, but I think the generator is in much better shape now.

This contributes to #1999. The option I've implemented keeps everything in separate functions/classes, just in one file. We can later add an even more unmodular option that just uses a single function. A lot of the problems I ran into would have come up even if I had tried to do it the even more unmodular way to start, so hopefully it will be not so bad to implement that now.

The changes to stable are a result of a fixed bug where exceptions as a result of transitive function calls weren't being declared in Java.

### DIFF
--- a/code/drasil-code/Language/Drasil/Code.hs
+++ b/code/drasil-code/Language/Drasil/Code.hs
@@ -6,8 +6,8 @@ module Language.Drasil.Code (
   ($:=), Choices(..), CodeSpec(..), CodeSystInfo(..), Comments(..), 
   Verbosity(..), ConstraintBehaviour(..), Func, FuncStmt(..), 
   ImplementationType(..), Lang(..), Logging(LogNone, LogAll), Mod(Mod), 
-  Structure(..), ConstantStructure(..), ConstantRepr(..), InputModule(..), 
-  CodeConcept(..), matchConcepts, AuxFile(..), Visibility(..),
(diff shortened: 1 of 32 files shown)


bmaclach:
I should also mention that I didn't choose to use `Unmodular` for any of our examples in stable. I tested the new option locally though, with different combinations of the other options, so I'm confident it works.

The reason I'm not using the option in stable is because our 3 examples already test 3 quite different combinations of options, so changing any of them isn't really giving any more test coverage, just different test coverage. We are at a point where it would be very useful to have more tests for all the combinations of choices. Probably choose 1 case study (Projectile?) and have a whole bunch of versions of it in stable? @JacquesCarette @smiths

JacquesCarette:
[I'll merge anyways, but you might have to patch this depending on the answer to the comment]

JacquesCarette:
And as to your question: yes, it makes sense to do exactly as you propose, have a bunch of different versions of Projectile in stable.

Issue #1993, opened 2020-01-22
External Library Design Proposal


bmaclach:
Preface: This issue evolved as I wrote it. I originally was just planning to summarize my commonality analysis but as I wrote things out I got a pretty clear idea of a design for this, so in the "General thoughts/analysis" section I start laying out those thoughts, then in the "Proposal for the types" section I wrote some code for the ASTs I have in mind. Many of the observations in the commonality analysis are probably too ODE-solver-specific, but hopefully I've done a decent job of generalizing things in the actual analysis/design sections.

---

As a first step towards designing a language for defining external libraries, I'll lay out the results of my commonality analysis for calling the 4 ODE-solving libraries I've been looking at (scipy (Python), Apache commons math (Java), oslo (C#), odeint (C++)).

## Commonality: Complex Input Data

Each library requires defining the input data. In the context of a solver, this is the "thing to solve". In the context of ODE solvers, this is the (system of) ODE(s). By "complex", I mean that the data cannot be expressed with basic types. For example, the data is not just a float or a Boolean.

Variabilities:
- Is the data expressed as a function or a class?
  - If it is a function, 2 cases:
    - parameters are dependent and independent variable (in that order), return a vector/list of the RHSs of the system
    - parameters are dependent, independent, and "result" variable (in some order), function body "sets" the result variable.
  - If a class, contains above function and a constructor that sets the state variables. The state variables are the variables used in the expression of the thing to be solved, other than the dependent and independent variables. The class might also contain more methods, such as `getDimension` for Java.
  - The class may or may not implement an interface

## Commonality: Set up solver object

Each ODE solver I looked at requires defining some sort of solver object of a class provided by the library. For ODE solvers, this is usually referred to as the "integrator". Setting up the integrator always involves specification of the numerical method to use, the absolute tolerance, and the relative tolerance, and sometimes also requires other parameters to pass.

Variabilities:
- The name of the constructor to call. Sometimes the constructor used corresponds to the chosen numerical method, and other times there is a single constructor provided by the library and the numerical method is passed as a parameter.
- The number of methods to call. Sometimes all of the set-up is done in the constructor call, but sometimes you must call additional methods on the object to set up certain properties.
- Are the tolerances passed in as individual parameters, or grouped together in an "Options" record?
- The parameters that need to be passed. Sometimes this set-up of the solver requires passing the time step, initial time, the data representing the thing to solve (ex. the ODE), the initial value, etc.

[...]

PR #2009, opened 2020-02-12
Start defining External Libraries in Drasil

bmaclach:
This PR implements the first of the two ASTs outlined in #1993, `ExternalLibrary` for defining external libraries. This AST is full of "holes" (represented by the AST nodes that take function parameters) which will be filled in by the 2nd AST, `ExternalLibraryCall`, which will be use-case specific and which I haven't implemented yet.

I also defined our 4 ODE libraries using this new AST, and refined my initial code for the AST as I went along.

What follows between the grey lines is a long thought-process that I want to keep record of, but isn't necessarily important for others to read. Feel free to skip to the tl;dr.

---

A question arose in doing this. My initial thinking was that each "step" in the process of using an external library would involve a call to some function or method provided by the library. From looking at my manual ODE-solving implementations, I knew this wasn't strictly true, but the steps that did not involve calling something from the library (like initializing the variable for holding the eventual solution, for example) seemed so use-case specific that I thought they would only show up in the 2nd AST, the one that will hold use-case specific information. However, in looking again at these, I realized that while a statement for the initialization of the solution variable is use-case specific because the name of the variable would be use-case specific, the initialization itself is mandatory. No matter what your specific use case is, you are going to need to declare a solution variable when using C#'s oslo library, for example. So I revised my thinking -- I now thought the `ExternalLibrary` AST should have support for these kinds of statements.

So I made the `Step` node in the AST have 2 possibilities -- `Call` for calling a library function/method, and `Statement` for doing other statements not directly related to interacting with the library. For `Statement`, I relied on our existing AST called `FuncStmt` as the backbone. `FuncStmt` allows us to build various types of statements from `Expr`, and there are functions already for converting both `FuncStmt` and `Expr` to GOOL code, thus I felt it made sense to use these instead of reinventing the wheel.

I ran into problems where `FuncStmt` and `Expr` were not expressive enough to handle all of the types of statements/values I needed in defining how these external libraries are used. The solution to this would of course be to expand these two types to handle the cases I need. For `FuncStmt`, that's just a matter of adding more AST nodes. For `Expr`, it's a little trickier because we don't want to pollute `Expr` with types of expression that are only meaningful in code -- for example, I need a way to have an expression for a constructor call. My planned solution for `Expr`, then, was the create a new type called `CodeExpr` which would extend `Expr` with more code-meaningful types of expression. But I've held off on doing this for now.

The extensions I'd have to make to `FuncStmt`, for things like declaring and defining a variable in a single statement, for example, are things that have already been defined in GOOL (GOOL provides `varDecDef`, for example). I began to feel like `FuncStmt` was just re-defining (a subset of) GOOL for use in Drasil, and I asked myself "why?". GOOL already has everything I need. Wouldn't it be better to bypass FuncStmt, and make `External Library`'s `Statement` step just use straight GOOL as the backbone?

Two drawbacks immediately stood out with that idea:
1. Since GOOL types all involve a type variable `repr`, all of the `ExternalLibrary` types would also need to be parameterized by a `repr`, which is messy.
2. Drasil users who wanted to encode their favourite library would need to know/use GOOL syntax as part of that process, which I don't think we want. From my understanding, part of the reason we want to define an `ExternalLibrary` type on the Drasil side of things, rather than having `solveODE` built into GOOL like how I did it originally, was to avoid forcing Drasil users to have to use GOOL syntax directly.

So if I want to avoid using `FuncStmt` but also avoid putting GOOL entities directly in an `ExternalLibrary` definition, the other alternative is to realize the different types of statement as different nodes in the `ExternalLibrary` AST. So instead of a `Step` having only 2 possibilities -- `Call` or `Statement` -- the `Statement` node would be broken down into many types of Statement -- `Declaration`, `Initialization`, `Return`, `Append`, `ListPopulation`, and more as needed. The advantage to this approach would be that the resulting statement language would be smaller and more focused than `FuncStmt`, including only the types of statements actually needed when interfacing with libraries. But the disadvantage is that doing this would just be replacing the need to translate from `FuncStmt` to GOOL with the need to translate from `ExternalLibrary` to GOOL. And even worse, whereas the functions for going from `FuncStmt` to GOOL have already been written, I would have to write new ones for `ExternalLibrary` to GOOL. So this definitely seems wrong.

So I'm going to stick with my original plan, to extend `FuncStmt` as needed and add `CodeExpr` to extend `Expr`. But in addition to that, I can add a smart constructors with external-library-focused names on top of `FuncStmt`, so that the syntax for building an `ExternalLibrary` is as natural as possible. (So, for example, instead of the generically-named `varDecDef`, I might name the constructor `initSolutionVar`).

---

### TL;DR

The `FuncStmt` and `Expr` types are not expressive enough in their current form to do everything I need them to do when interfacing with external libraries, so I'm going to extend them both. For `Expr`, since the extensions I need are code-specific, I'm going to create a new type called `CodeExpr` to house these extensions. To keep `ExternalLibrary` syntax natural for users of Drasil, I'll add some smart constructors on top of the `FuncStmt` AST with names that make sense in the context of external libraries.

(Note that this PR isn't broken, but has some hacks to make things work, which I plan on fixing in the short term. Just don't want this PR to get too big, so opening it now. I'll branch again for future work.)


JacquesCarette:
So I'm going to merge this in now, even though there's a couple of things I noted above. They are potentially larger changes, so I don't want to block progress on that. But they should be looked at seriously.

smiths:
@bmaclach, thank you for recording your thoughts in the introduction to this PR.  When you are writing your thesis, you'll want to look at these blurbs and mine them for thesis related content.

I agree with your thought process.  We don't want the users to write GOOL code.  Moreover, we want to capture the commonalities, like initializing variables before running the solver, and the hopefully capture the variabilities as Drasil options.  I'm sure this is easier said than done, but at the right abstraction, all ODE solvers are doing the same thing.

JacquesCarette:
Reading the PR text carefully, I have the following comments:

1. extending `Expr` with `CodeExpr` is wrong. `Expr` should be abstract and not concerned with execution details. Note that `Expr` may still not be perfect in this sense and need fixing.
2. there are times where repetition / redundancy is good. This is the 'micro pass' architecture of compilers, where each pass does just a little bit of work on the AST, so that in a very typed system, there is a lot of duplication between the types from stage to stage.  i.e. that `Statement` repeats things from GOOL is not necessarily bad - just annoying.
3. thus I think that a separate `ExternalLibrary` AST, with its own design and meaning that are higher-level (you list some really good ones).  I think the advantages of 'semantic clarity' outweigh the disadvantages.

Sorry if that throws a late wrench in your plans...

bmaclach:
@JacquesCarette A wrench, maybe, but not too bad of one. Thanks for the comments. With regards to point 1, I didn't end up doing `CodeExpr` (though what I did is still not great, which you've commented on the corresponding PR). And with regards to 2, that's good to know and addresses the concern I had with redundancy.

It seems like in 3 you're suggesting I add a lot of AST nodes for the different types of statements, instead of just one `Statement` node. The change wouldn't be too bad -- I've already written the smart constructors for `Statement` in #2018, and the new nodes would roughly correspond to those. But I do have one more question before doing this:

I don't plan on the `ExternalLibrary` AST being visible to users; I won't export it. I'll just export the smart constructors. And the smart constructors have the "semantic clarity". Doing it the way I have has the advantage that when I write the functions for going from `ExternalLibrary` to `FuncStmt` ASTs, I only need to write one pattern-match for translating all `Statement`s. If I switch to add more nodes to the AST, the "translation" functions will of course need some more work. But, that will come with the benefit of simpler implementations for the smart constructors. So my question is, given the AST won't be user-facing, is it still beneficial for the AST to have semantic clarity, and thus still worthwhile to make this change?

JacquesCarette:
The only advantage of a larger AST, when it isn't user-visible, is to retain more knowledge of the intent (as well as types and other invariants) to be used in getting more precise results further down the chain. If you don't need that, then there is no point to a larger AST.

PR #2012, opened 2020-02-18
External library imports

bmaclach:
This small PR adds to the `ExternalLibrary` AST to allow for specification of needed imports. It also allows imports to be specified at the module level in GOOL.


JacquesCarette:
I think it's worth changing the `Import` idea to `Requires` before merging this in.  Looking at the bundling opportunities can be done later.

Issue #2010, opened 2020-02-13
Adding more Projectiles to stable


bmaclach:
We've decided it would be a good idea to add more versions of Projectile to stable to get more coverage on the various `Choice` combinations. 

What I've done so far on the `manyProjectiles` branch is selected a subset of our `Choices` that have the biggest impact on the generated code, and therefore are most likely to cause issues. I set up the `main` function for Projectile to build `Choices` structures using every possible combination of the choices in the subset, and generate the code with each. This led to 96 projectile programs. 

I then updated our script for building the generated code to go through each of the 96 projectiles and build them all. This required hardcoding a check for if the current example was projectile, though I could make this a little nicer by having a Makefile variable store the names of any example for which there are multiple versions. 

I tested this step of the build locally. Prior to this change, when we were only building 1 projectile, 1 glassBR, and 1 NoPCM, this step only takes about 15 seconds. After the change, it unsurprisingly takes about 5 minutes. Adding 5 minutes to each build is probably not wanted, and especially unrealistic when we think about adding more choices, and how each time we add a choice the time for this step of the build will (at least) double. (I also don't know if Travis builds run slower/faster than locally for this step).

The other option is to hand-build some Choices structures for a handful of combinations, and only generate those, but I don't like this option either. For one, how would we decide which combinations were important enough to test? And as we add more choices, we'd need to hand-build more `Choices` structures and might eventually run into the same problem where we are adding too much time to the build.

I really like the idea of having a test suite of Projectile programs covering the entire spectrum of variations we offer. So here are some alternative ideas that could make this work:

- Have the standard build still only build 1 projectile, but have a once-daily build that will build them all. This has the drawback that if something breaks a certain choice combination, we won't know until the daily build after it gets merged.
- Store all versions of Projectile in stable, but still only build 1 projectile in our standard travis build. Then at least when we are working locally, we can see the diffs with all versions and the Travis build will fail if there is a diff with any version and stable (though from a developer's perspective, this diff would be very long and tedious to go through). We could also provide a rule in the Makefile that will build all versions of Projectile. We wouldn't use this rule in any Travis builds, but it would allow a developer to test them locally.

What are your thoughts @JacquesCarette and @smiths?

Mornix:
👀Popping in here. 

[...]

PR #2015, opened 2020-02-19
Add 5 Projectiles to stable

bmaclach:
This PR updates our Travis builds to build 5 different versions of Projectile covering a range of `Choice` combinations, and adds all 5 versions to stable. This is as discussed in #2010.


bmaclach:
The Choices that change between these examples are below. Beside each option, in brackets I've put a possible short form for that option.

Modularity - `Unmodular` [U], `Modular Combined` (one input module) [C], `Modular Separated` (multiple modules related to input) [S]

Logging - None [NoL] or All [L]

Comments - None [NoC] or All [C]

Input Structure - `Bundled` [B] or `Unbundled` [U]

Constant structure - `Bundled` [B] or `Unbundled` [U] or `WithInputs` [WI]

Constant representation - `Var` [V] or `Const` [C]

We can use underscores to separate the different choices, ending up with file names like: Projectile_U_NoL_NoC_U_WI_V. Would that be okay?

We have also talked in the past about documenting the chosen options for the examples in stable in some sort of table on the webpage. I can't find an open issue for it, but I remember it not being straightforward which is why it wasn't done (I think the idea we were thinking of was `grep`-ing the code to find the `Choices` structure for each example, then extracting the `Choices` from there, but that's a really fragile solution).

smiths:
@bmaclach, I like your proposal for naming the folders.  We should record the codes somewhere, so that someone can interpret the folder names.  Maybe this would make sense in the introduction blurb for the generated artifacts web-page (https://jacquescarette.github.io/Drasil/)?

Yes, we have discussed documenting the variability choices for the case studies.  I like the idea of automating this, so that if the choices change, so does the documentation.  However, we have intentionally selected these examples, as you did for Projectile, maybe for now we can just "hard code" the choices.  The choices could go in the introductory blurb for the web-page I mentioned above.  There is a maintainability concern, but until we have a more elegant solution, just recording the information somewhere makes sense to me.

bmaclach:
Okay @smiths, I'll add the table and short-form guide to the website intro blurb. I agree that it would be good to record it, even if that means hard-coding it for now.

smiths:
Sounds good @bmaclach.  Maybe you should also create an issue to remind us to circle back to this to find a more maintainable generative approach in the future?

bmaclach:
I just pushed a bunch of changes. First, I made the requested changes:
1. Changed folder names to encode design decisions
2. Added to website table showing design decisions for each example
3. Added to website guide for decoding the short forms for the design decisions

I then realized that because updating projectile to have multiple versions changed the directory structure for projectile in stable, the links to Projectile src code/doxygen pages were broken on the website. I updated our Makefiles/scripts/website code to handle the scenario where an example has multiple versions, and added some formatting to the website to display links for each version in a nice list:
image

This had the side effect of making a list even for examples with only one version, but I think this still looks nice:
image

I think I've made these changes in a generic enough way that, should we ever decide to host multiple versions of another case study, we need only add the corresponding directory to the `MULTI_SRC_DIRS` Makefile variable, and the build process and website should be automatically updated to build/show all versions of the example.

bmaclach:
Also created issue #2017 for automating the table of design decisions in the future.

PR #2016, opened 2020-02-19
Add ForEach to FuncStmt, remove Iterate node for ExternalLibrary

bmaclach:
This small PR first does some organization, moving some things from CodeSpec.hs to more appropriate places, then adds for-each loops to `FuncStmt` and removes the `Iterate` node from `ExternalLibrary`, since the `Statement` node can now be used instead.

#2012 should be merged first. afcf08f is the first new commit.


bmaclach:
The build isn't actually pending -- it passed. Seems like it wasn't reported back to GitHub for some reason.

PR #2019, opened 2020-02-21
More Expr/FuncStmt/ExternalLibrary Updates

bmaclach:
This PR does a set of improvements on `Expr`, `FuncStmt`, and `ExternalLibrary` which will help when generating code for external library calls:
- Remove `FProcCall` from `FuncStmt` because it isn't a statement and `Expr` already supports function calls.
- Allow a `CallStep` in an `ExternalLibrary` to be part of a return statement, update `Fn` node to define the function body as a step instead of the messier `[Expr] -> FuncStmt`
- Add `New` to `Expr` for calling constructors. I was originally planning on defining a new type, `CodeExpr`, for this, but in practice that wasn't working well -- we want to have all of the functions on `Expr` still work. So for now I've added it directly to `Expr` but only defined/exported smart constructors for the new node in the drasil-code package.
- Update `ExternalLibrary` `FunctionInterface`s to represent the function being called as a `CodeChunk` instead of just a name. This will make it easier to translate the definition to `Expr` for generation.

#2018 should be merged first, 1b35da9 is the first new commit.


bmaclach:
Updated `New` in latest commit to take a `UID` instead of `Expr`, which I think addresses the concerns so this PR (and the trailing chain of PRs) can finally be merged?

JacquesCarette:
Yes, it does, thanks.  And there's a route forward too.  So I'll start merging in.

PR #2023, opened 2020-02-25
Update type for representing named arguments from string to CodeChunk

bmaclach:
Small PR to improve `ExternalLibrary` by updating the type for representing the name of a named argument from string to a `CodeChunk` corresponding to the parameter-variable being referenced by name. This will make it easier to map to GOOL code for generating named parameters, and is less fragile than using strings.

#2019 and #2021 should be merged first. The only (non-merge) new commit is b7abee2.


JacquesCarette:
There are now conflicts - can you resolve please?  Otherwise, this looks good.

bmaclach:
Weirdly the merge succeeded without conflicts.

Anyway, merge is pushed but just a reminder that merging this will also implicitly merge #2019, so probably still want to hold off.

JacquesCarette:
Weird. And indeed, I will hold off.

bmaclach:
Again, the merges are resolving automatically locally even though GitHub is reporting conflicts...

Issue #2025, opened 2020-02-26
Expr extensions needed for ExternalLibrary


bmaclach:
My big picture plan for external libraries is to define `ExternalLibrary`s which can be paired with `ExternalLibraryCall`s to generate the statements and additional classes needed for the call. These statements and classes will be represented by Drasil's existing AST for representing functions/classes, primarily the statement language `FuncStmt`. There are many reasons for doing this:
1. The functions for translating from `FuncStmt` to GOOL code are already defined
2. These functions are built into Drasil's generators, so they use the generator's `DrasilState` to generate correct code
3. By converting the `ExternalLibraryCall` to `FuncStmt` prior to generation, we can do some analysis to properly set the `DrasilState` `Reader` for generation (for example, adding the appropriate functions to the module export map and class export map). Again, we already do this analysis for other modules that are represented by `FuncStmt`, so the functions for doing this already exist.

The drawback to this is that `FuncStmt` relies on `Expr` as the representation for the values involved in the statements. So if we want to use `FuncStmt` as an intermediate representation between `ExternalLibrary`/`ExternalLibraryCall` and GOOL, every value needs to be expressible by `Expr`. 

This has already lead to problems in PR #2019, where I needed to represent a call to a constructor, so I added a `New` node to `Expr` to represent this. This is not ideal because it pollutes `Expr` with expressions that are only meaningful in code. My idea to make this less bad was to not export any smart constructors to this node or any other code-specific nodes I may need to add, so that a Drasil user would not use these nodes. 

The reason I'm writing an issue for this is to have a place for discussion, as well as to provide context for the issue that may not have been apparent just from looking at #2019. I'm also writing this issue because constructor calls are not the only type of expression needed for external libraries that aren't currently expressible in `Expr`. So I am putting together a list of needed but currently unexpressible expressions as I come across them while working on my external library implementation:
- Constructor calls
- Method calls (and object access in general)
- Named expressions
- Function-valued expressions

My thinking is that of those, only function-valued expressions are a mathematical concept that probably deserves a place in `Expr`. 

[...]

PR #2026, opened 2020-02-26
ExternalLibrary CodeChunks instead of Names

bmaclach:
Continuing the trend of replacing Strings in the `ExternalLibrary` AST with `CodeChunk`s. Now there are no strings left.

#2024 should be merged first. ce90367 is the first new commit.


JacquesCarette:
But I shouldn't merge yet, as this implies merging in a whole chain, some parts of which I'm not happy with, right?

bmaclach:
#2019 is the one holding everything up. I'm not sure exactly what needs to change though. (Depends on your answer to https://github.com/JacquesCarette/Drasil/issues/2025#issuecomment-591528781)

PR #2033, opened 2020-03-05
Allow users to choose how Spaces maps to CodeTypes -- Floats/Doubles choice

### LINKED ISSUE #1863: Code variability: double versus float
A "low hanging fruit" code variability is to switch the computational representation of the mathematical concept of real between double and float, for languages that make the distinction.  Double will have twice as many bits dedicated to representing floating point numbers.  We (@bmaclach, @JacquesCarette and myself) discussed this issue on August 29, 2019.

bmaclach:
This PR closes #1863 by implementing the choice between floats and doubles. It is done in a generic way, where the user specifies a mapping between `Space` and `[CodeType]`, where the list is in their order of preference. The mapping is in the form of a function, rather than a Haskell `Map` like originally planned, since `Space` is a recursive type so it's impossible to exhaustively cover it with a `Map`. 

I had to add a few new `Space`s since `CodeChunk`s now store their type as a `Space` instead of a `CodeChunk`, but I'm not sure if some of them really belong as `Space`s  (`Array`? `Void`?). If not, we likely need to go back to the idea of storing a `CodeChunk`s type as `SpaceMatch -> CodeType` because, for example, external libraries can use a combination of arrays and lists, so there needs to be a way to distinguish between them. 

I also changed 2 of our Projectile versions to use floats instead of doubles. Interestingly, if you try to use floats in GlassBR the generated code will not compile because some values are too small to represent as floats (ex. the surface flaw parameter `k` is 2.86e-53). Not sure if that's something we would want to detect prior to generation...

For reviewing changes to stable, just look at f11f118 (if you look at the changes across the whole PR, you will see entire files in the diff because the names of Projectile's directories were changed to encode the new choice).


JacquesCarette:
There are some conflicts now - can you resolve them please?

PR #2040, opened 2020-03-20
Implement library vs executable choice

### LINKED ISSUE #443: Fix implementation choice for library vs. executable
There is currently a choice option for generating a library or a executable, but the choice doesn't actually do anything.  Right now the generated code is always for an executable program.

This will require additions to the makefile generator so that we can generate makefiles that either build a library or executable depending on the choice.

bmaclach:
This PR implements the choice between generating a program or a library (the choice existed previously but was not used at all in the generator, so this PR adds the functionality for actually reading the choice in the generator).

This closes #443. The issue comments discuss the idea of compiling a `.whl` for Python libraries, which I didn't do in this PR (instead I just do nothing for python libraries, leaving them as `.py` files). Compiling a `.whl` is less straightforward, requiring a whole new file to be generated, and I don't think it is outright wrong to represent a Python library as regular Python code files, which is why I didn't do the wheels here. But they are still a good idea, so I will open a new issue for that.

I updated one of our Projectiles to be generated as a `Library` instead of a `Program`, so there are changes to stable. I updated the website, too, to explain the filename-encoding for the new choice.


JacquesCarette:
(The builds have completed and passed, but that somehow wasn't communicated back. Merging)

PR #2051, opened 2020-04-02
Generator improvements

bmaclach:
For some reason the Calculations module was generated by first defining a `Mod` containing various calculation functions, whereas most of our other modules do not have this intermediate `Mod` representation. I couldn't find a good reason why the Calculations module was treated differently, and in fact treating it this way makes things less flexible (it forces each calculation to be in its own function, so it wouldn't be possible to later have a fully unmodular choice where everything is all in one function). So I fixed this so that Calculations is treated like the other modules we generated, then removed the `FCD` node from `Func` altogether, because it is no longer used or needed.

I also fixed a minor bug where the class definition map and module export map weren't being completely defined and some classes were not getting generated when Unmodular is chosen.


bmaclach:
(Minor stable update because bring Calculations module into the generator means its now generated before the output module, show it shows up first whenever the modules are listed out like in the Makefile and doxConfig file)

PR #2059, opened 2020-04-08
Fix for HLint errors in master

bmaclach:
Some HLint violations got through to master. This fixes them.

On a related note, anyone know why Travis has stopped reporting build statuses here?

PR #2060, opened 2020-04-09
Rearrange GOOL class constraints

bmaclach:
Some class constraints in GOOL were no longer required given changes to GOOL that have occurred since they were written. This PR fixes up all of the constraints to be accurate and sensical.

The HLint fix #2059 should be merged first.


JacquesCarette:
Note that there's a conflict to fix.

And Travis is actually running, as I get email notifies.  Just not linking back.

Issue #2061, opened 2020-04-09
Travis build statuses no longer reported to GitHub


bmaclach:
Apparently this is a rather widespread problem. 

I found some discussion from the Travis team about this, where they started receiving reports of this issue around the same time build statuses disappeared for us: https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

Apparently they are working to fix it, so one option is to just wait.

The other option is to migrate to GitHub Apps. Apparently this issue only affects repositories using legacy GitHub Services integration. (I know we are, because our travis page is at travis-ci.org and not travis-ci.com).

If we migrate to GitHub Apps integration, the issue should be fixed. Instructions for doing that are at: https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/

It does not seem too difficult, but I don't have the permissions to do it myself. So I'll assign this to @JacquesCarette, but if you want to give me permissions and re-assign to me, go ahead.

JacquesCarette:
I think I've done what's needed, but I'm not 100% sure that it all worked.  Also, I did it as me (as the owner of the repo), but the builds are done by drasil-bot, so maybe it needs migrated too?

bmaclach:
The links to our builds still go to travis-ci.org. I expected that to change to travis-ci.com when we were migrated. So maybe you are right, and it needs to be migrated as drasil-bot.

JacquesCarette:
As far as I can tell, whatever I just did, fixed it. The build is now on travis-ci.com.

PR #2062, opened 2020-04-09
Renamed pubGVar, pubMVar, etc. to use S and D instead of G and M

bmaclach:
I'm not sure what G and M stood for, but regardless I think S and D are more understandable because they align with GOOL's syntax for binding time (`static`/`dynamic`)


bmaclach:
Looks like Travis CI is reporting again! If it continues to work, we can close #2061.

JacquesCarette:
We may as well migrate away from the obsolete stuff.  Leave #2061 open (and assigned to me) for now.

bmaclach:
Nope, didn't mean to re-add Symantics.hs. Thanks for catching that. Removed now.

PR #2064, opened 2020-04-14
Remove some constraints from GOOL class hierarchy

bmaclach:
@JacquesCarette I've taken a  half-measure here because I have questions.

Removing constraints from the GOOL class hierarchy like we talked about in today's code review meeting will have the consequence of needing to specify more constraints where the methods are used. For example, if `AssignStatement` is no longer a constraint in the GOOL class hierarchy, it would mean that throughout Drasil's generator, we would have to add the `AssignStatement repr` constraint to any function that does an assignment. Currently the only constraint we need is `ProgramSym repr`, which is essentially a catch-all constraint, but with this change the generator would become cluttered with constraints.

So the half-measure I've taken is to remove unnecessary constraints in the hierarchy, but then add them all as constraints on the `FileSym` class, so `ProgramSym repr` and `RenderSym repr` can still act as catch-all constraints.

I doubt that is what we actually want. A perhaps better option would be to define, in Drasil rather than in GOOL, a new type class, say `DrasilProg`, which will be constrained by all of the GOOL type classes (and have no methods of its own), and then `DrasilProg` can be used in the generator as a catch-all constraint. Does that sound okay? Or do you disagree with the idea of a catch-all constraint altogether?

### DIFF
--- a/code/drasil-gool/GOOL/Drasil/ClassInterface.hs
+++ b/code/drasil-gool/GOOL/Drasil/ClassInterface.hs
@@ -39,7 +39,10 @@ class (FileSym repr) => ProgramSym repr where
 
 type SFile a = FS (a (RenderFile a))
 
-class (ModuleSym repr) => FileSym repr where 
+class (ModuleSym repr, ControlBlock repr, AssignStatement repr, 
+  DeclStatement repr, IOStatement repr, FuncAppStatement repr,
+  MiscStatement repr, ControlStatement repr, InternalControlBlock repr) => 
+  FileSym repr where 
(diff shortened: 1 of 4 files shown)


JacquesCarette:
As I was reading your comment, I was going to suggest essentially what you did!

I would be tempted to create more than 1 catch-all class though, with various sub-languages.  For example, could we have `ProceduralProg` and `OOProg` ?  That would be nice.

In some use-cases, I would be happy to specify all the constraints at the use site. But we're not ready for that now, so your suggestion is the better route.

PR #2066, opened 2020-04-14
Abbreviations throughout GOOL

bmaclach:
This PR abbreviates GOOL in various ways:
- Introducing type synonyms for repeated type patterns
- Shortening `repr` type variable to `r`
- Using `>>` instead of `_ <-`

Based on discussion from the April 14th GOOL code review meeting.


bmaclach:
> I think I would have done `NamedArgs` and swallowed the list in too.

Done, and did same with `Initializer`.

PR #2070, opened 2020-04-16
Make CodeInfo fully generic

bmaclach:
CodeInfo had some Java-specific exception definitions. This PR adds an `ExceptionType` type to represent exceptions in a language-independent way, and a class for mapping an `ExceptionType` to the language specific `Exception`. The `CodeInfo` pass then uses `Exception` type so it is fully language-generic, and the Java renderer has an instance of the new class to translate that to concrete exception declarations in Java.


JacquesCarette:
Wow, that was surprisingly easy to do.  Very good.

PR #2074, opened 2020-04-17
Minor generator improvements

bmaclach:
This PR improves the generator in 3 minor ways:
- Re-orders parameters to some functions to be more consistent with other functions
- Set the `currentClass` field of the `DrasilState` in the generic `mkClass` function, instead of having to set it in each generator for a specific class
- Use `Name`, `Description`, and `Import` type synonyms for `String` more consistently.

#2072 should be merged first. cc266cb is the first new commit.


JacquesCarette:
There's still a conflict?

bmaclach:
My bad, fixed now.

Issue #2076, opened 2020-04-20
Make GOOL's "FromData" functions stateful


bmaclach:
GOOL has many methods ending in `FromData`, some of which, like `stmtFromData` for example, are not stateful:
https://github.com/JacquesCarette/Drasil/blob/0d1fc1a1aa59a44e5b5f8fb60b10f6b8b2c8c32f/code/drasil-gool/GOOL/Drasil/RendererClasses.hs#L213

These should be updated so that the return type is stateful, so that they are more consistent with other methods in the same class. For example, `stmtFromData`'s signature should be:
```
stmtFromData :: Doc -> Terminator -> MSStatement r
```

Note that currently where these functions are used, they are sometimes lifted into the state context. After updating them to be stateful themselves, if they get lifted into state then there will be two `State` monad layers. `join` can be used to combine the layers in these cases. (`join` is already used in some places in GOOL, so those can be used as examples).

JacquesCarette:
Does #2077 close this @bmaclach ?

bmaclach:
No, #2077 is unrelated... I must have accidentally linked the wrong issue to that PR. This still needs doing.

muhammadaliog3:
Quick question, currently I am using dollar signs more often then brackets, for example
```haskell
toState $ toCode (d, t)
```
if anyone wants I could change them all to 
```haskell
toState (toCode (d, t))
```
So if anyone has any preferences for dollar signs or brackets then reply to this comment.

JacquesCarette:
Haskell style is to use `$`.

muhammadaliog3:
The latest commit now compiles, but I do not know about HLint. Now I will try to make the code use less lines/make it more modular. I could make this a new issue called "Modularize FromData functions", but I will do it here in this issue instead. 

The reason I am commenting is if any knows how to do the following without do statements: (a->b-> m c)->m a->m b-> m c, then please comment below.

bmaclach:
@muhammadaliog3 The only way I know of to do that without do notation is `join $ liftM2 f ma mb` where `f :: a -> b -> m c`, `ma :: m a`, `mb :: m b`. I remember recently seeing a comment from @JacquesCarette about using `join` though, so he likely has insight here. Also, do notation might be preferable, because it's usually more readable.

JacquesCarette:
With `do`, it is easy:
```haskell
foo :: Monad m => (a-> b-> m c)-> m a-> m b-> m c
foo f a' b' = do
[...]

PR #2077, opened 2020-04-20
Rename Doc functions

bmaclach:
This PR renames all the renderer functions for extracting `Doc`s, by removing the word `Doc` from the function names. As discussed in the April 20th GOOL code review meeting.


bmaclach:
#2076 should be merged first, 4d300da is the first new commit.

PR #2078, opened 2020-04-20
Move GOOL smart constructors to new file

bmaclach:
Some smart constructors for the GOOL renderers, like `mkSt`, were scattered throughout files where they really didn't fit. This PR adds a new file for constructors and moves them all there. It also renames `mkSt` to `mkStmt`, which makes its purpose more clear.

This move was discussed at the April 14th meeting (I held off until now because I thought the April 20th meeting might result in removing these functions, but it didn't).

#2077 should be merged first. 29a4b20 is the first new commit.


bmaclach:
I pushed another commit to this branch (d1b8e0c) but for some reason it's not showing up in this PR. Usually they get added automatically... this may be related to the problem's GitHub seems to be having today. In any case, the commit is showing up in the next PR (#2079), so it can get merged in with that one if not with this one.

JacquesCarette:
It seems to be in now.

PR #2079, opened 2020-04-21
Remove some internal GOOL constructors

bmaclach:
This PR removes some of GOOL's internal constructors for various types. Instead of using these constructors, I updated the generic functions to return the underlying GOOL type. For example, whereas before the generic `multiBody` function in `LanguagePolymorphic.hs` returned a `MSBody r`, now it returns an `MS (r Doc)`, because a body in all languages is really just a `Doc`. Previously the internal constructor was needed because the generic function has no guarantee that a `Body r` is equivalent to a `Doc`, but now that we've concretized the return type to `Doc`, the internal constructor is not needed.

This is as discussed in the April 20th GOOL code review meeting.

#2078 should be merged first. d708a33 is the first new commit.


bmaclach:
Builds are failing before they even run. I think the problem is on GitHub's end. I was getting Internal Server Errors earlier today when I was trying to push, too.

PR #2082, opened 2020-04-24
Rename RenderValueExp InternalValueExp

bmaclach:
In a previous PR I inadvertently renamed an `Internal` class to `Render` despite it not actually being a renderer-only class. This PR reverts that mistake.


bmaclach:
It is internal because the function it contains accepts parameters in the "wrong" order from a user's perspective, but in an order that makes implementations cleaner to write. Unlike the renderer classes, there are functions that are part of GOOL's public interface that call this class's method, so it must be implemented even for non-renderer `repr`s.

PR #2099, opened 2020-05-06
Fix small generator bugs

bmaclach:
Noticed some bugs in the generator that affect cases not covered by our current test suite:
- The parameters to the generated `InputParameters` constructor were wrong in some cases, and 
- the generated code should define constants before calling the `InputParameter` constructor because calculating derived inputs or checking constraints on the inputs might use those constants.


JacquesCarette:
I'll merge now, but won't delete the branch, you can then add more comments and do another PR on top of that?

PR #2103, opened 2020-05-07
Nathaniel-DataTable

Nathaniel-Hu:
Pull Request to merge Nathaniel-DataTable branch (with DataTable .xlsx file) to master branch

Note: sample.txt file can be deleted


JacquesCarette:
A 'README' is a good thing (which is roughly what you have in your `.txt` file) - we usually call it `README.md` and create it using makedown rather than plain text.

Also, you should have made 2 PRs, one for the data table (and the readme), and another for traceability graph(s). But I'll merge this one as-is.

JacquesCarette:
Further note that we create new branches for new work, so I will delete this branch. Just create new branches (with good names) for each new pieces of work.

PR #2107, opened 2020-05-08
muhammad_data_table and Muhammad Drasil/code/drasil-example/Drasil/SWHS/Body.hs implementation

muhammadaliog3:


JacquesCarette:
So does the other PR (that I've just pulled in) mean that this one should be closed?

PR #2113, opened 2020-05-13
Muhammad swhs edits pull request

muhammadaliog3:
I have made changes to 2 files, and hence 2 commits.


JacquesCarette:
Have you noticed that the automatic build failed?  [That's what the red X beside the commit id means.]  If you click through (the X is also a link), there is a report from Travis that will tell you what went wrong.

JacquesCarette:
Right, so I'm going to close this PR. Not because it's bad, but because while working on it, you stumbled upon a larger issue that needs a larger fix. That is **good**.

JacquesCarette:
I'm also deleting the branch, because the fixes to the bigger issue will be quite different / separate from these. It's best to start with a clean slate.

Issue #1499, opened 2019-06-05
Drasil should consider ODE solver parameters as required to get the output


bmaclach:
NoPCM and SWHS define quantities for absolute tolerance, relative tolerance, and time step. The purpose of these quantities is to pass them to an ODE solver. However, since they don't show up in any equations, Drasil would not recognize these as required inputs and thus would not throw an error if they were missing from the input list. Drasil should recognize that these ODE solver parameters are required inputs if an ODE solver is being used.

This is more something to keep in mind when teaching Drasil about ODEs (#473), not something that can really be fixed right now.

smiths:
@bmaclach, what about using the requirements to get the inputs?  There is a requirement in the SRS for SWHS that lists the parameters you are mentioning.  The requirement is human generated, but it is one place where all of the information is given.  Maybe we can use that as a source of information, while we are waiting for Drasil's automation to check up?

bmaclach:
@smiths Yes, that's how we are doing it now (or at least as of #1496). We have one `inputs` source, which is used for the requirement and everywhere else the input list is needed. This issue is more to address the possibility that the inputs were left out even in that one source.

bmaclach:
Actually, #1496 updates NoPCM to have a single input list source, but SWHS still has separate lists. I'll create a different issue for that.

JacquesCarette:
@bmaclach has your work on ODEs affected this in any way?

bmaclach:
As of #2114, users need to construct `ODEInfo` objects including all of the parameters discussed in this issue. Our function that checks whether there is a path from inputs to outputs will now consider all expressions in `ODEInfo` objects, and if those expressions involve anything that is not known or cannot be computed, an error is thrown saying there is not a path from knowns to outputs. This can be closed.

Issue #450, opened 2017-12-20
Code generation not implemented in the C++, C#, Java renderers


palmerst:
Re-opened milestone 0.3.1 as the code generation for ode solving is only working for Python right now.  Need to complete the other renderers.

bmaclach:
I can't find any working code generation for ode solving, for python or otherwise, even going back to commits from when this issue was opened. Do you know of something I might be missing @JacquesCarette @smiths?

smiths:
Until I read the text by @palmerst, I would have said we never had the code generation for the ODE solver working.  I don't know where the code would be, but I do know that we weren't aiming for generating code to directly solve an ODE.  We were aiming for code to interface with existing ODE solvers.  There are Python ODE solvers and we were just aiming to generate the code to interface with the existing solver.  We would have been targeting SWHS (and by association noPCM), since that is example that has an ODE that needs solving.

JacquesCarette:
If anywhere, I would have guess under `GOOLCodeGen`. I took a look to see if it might have been in a stale branch - but no, those are all gone.

JacquesCarette:
Does it make sense @bmaclach to keep this issue open, or is this essentially done?

bmaclach:
It is very close to done, but on the noPCMODE branch, which is still dependent on #2014 before a PR can be opened.

JacquesCarette:
Right, I'll try to get to that soon.  Review of writing first.

bmaclach:
As of #2114, code generation for ODE solving works in all languages. Closing.

Issue #2050, opened 2020-04-01
Create some kind of Contributing page


JacquesCarette:
where we can list the open projects to work on.  For example:

1. Incorporate Pandoc into DocLang.

2. Scientific knowledge ontology.

3. Generation of test cases based on data constraints and properties of a correct solution

4. Add 3D Model of the Aorta to Drasil

5. Generate jupyter notebooks - probably starting with generating simple physics experiments

smiths:
A potential projects page has been created at:

https://github.com/JacquesCarette/Drasil/wiki/Potential-Projects

PR #2114, opened 2020-05-14
Integrate ExternalLibrary in code generator

### LINKED ISSUE #394: Best way to implement external libraries for codegen?
@JacquesCarette I've been thinking about the best way to implement external libraries for code generation.  There seems to be 2 choices here:

1.  As an "External Library" data type that encodes a generic interface for a particular library type and given a language, converts it into `AbstractCode` that uses the existing GOOL AST.  This will mean generating a different language-specific `AbstractCode` for each language.  But we get the benefit of not having to modify the GOOL AST or the renderers every time we add a new library type or modify/extend existing libraries.

2.  Adding external library syntax directly into the GOOL AST, keeping the `AbstractCode` the same for all languages.  It seems like this approach will be much more tedious to modify and maintain.

I'm thinking option 1, but this will mean the GOOL `AbstractCode` is no longer "generic" and might add a lot of overhead later on when generating large programs in multiple languages.

bmaclach:
This PR does all of the work for incorporating the external library infrastructure in the generator. By the end of this PR, the generated code for NoPCM is complete and working. 

@JacquesCarette, I believe you reviewed the bulk of this a couple months ago when I first pushed. Since then, most, but not all, of the commits have been merge commits. You may want to look for the non-merge commits since April and review those (but they are mostly minor changes that I'm fairly confident aren't bad). 

Sorry again for the giant PR! 

This closes #394, #450, #1499, #1993, #2025.


JacquesCarette:
3500 files changed! But I reviewed the check-ins as they went, so this is all good.

smiths:
Twice now I have written a review of this merge, and both times I inadvertently closed the web-browser after I was distracted by something else.  :-(  You'll have to trust me that what I wrote previously was better than what I'm about to write.  I just don't have the energy to redo the work a third time.  :-(  It is a shame that GitHub doesn't issue any warning!  I know it is my fault, but it is still frustrating.

This is an impressive PR.  I find it very satisfying to see the fruition of a great deal of work!  I usually review the changes to stable, but there are just too many changes this time.  :-)  Instead, I focused on the generated code and running it myself.  I like the idea of impressing people by saying that they can just type `make run` and the code will run.  It didn't quite work like that, but pretty close.  :-)  I'll type up my experience with Python first, so that I can submit this comment and not risk losing it while I finish up with Java etc.

For Python `make run` did not work for two reasons:
1. For Python 3, I use `python3`, not `python`.  I don't think we want to generate the code for specific users by specifying the name of Python on their local system.  We could make it easier to modify `Makefile` though.  I suggest that we add the following to the top of `Makefile`: `PY = python3`, and then use `$(PY)` in place of `python`.  If we all agree, we can create an issue for this change.
2. `RUNARGS` was not defined.  I had to add the following to `Makefile`: `RUNARGS = input.txt`.  I feel like this was the original intention of `RUNARGS`, so I'm optimistic that this would be an easy fix.

I did not need to make any further changes to run the code.  Yay!  However, I do have one more question related to the Python version.  The code worked for me because I have `scipy` installed already.  Could we generate some simple documentation, like a `README` file, to tell the user what dependencies exist?  We know the dependency, so this seems like it wouldn't be hard to do.  Even better would be to generate a full installer, but I think that can wait.  If we agree, we could create an issue for generated a simple file listing the dependencies.  (If we could also list the version of the associated libraries, that would be even better.)

I'm going to write up my experience with Java next, but as a separate comment.  I'm not going to risk having to type up my Python experience a 4th time!  I already know though that the last point about `scipy` applies for Java as well, but in this case I don't already have the ODE solver installed.  Even if I did, I'm not sure I would have it installed where the Java compiler will look for it, but I'll give it a try now.

bmaclach:
Some quick early feedback:

If you look in code/datafiles/NoPCM/java you will find the library you need for Java (and similar for the other languages). You should be able to just copy and paste it into  build/NoPCM/src/java, and then make should work.

To run, I've been doing `make run RUNARGS=input.txt`. I had thought that was the intended use, but I'm actually not sure because @Mornix was the one who originally started the Makefile generation.

smiths:
Thanks for the note @bmaclach, you saved me some time.  :smile:  Yes, with your advice, it was easy to run the Java example.  I didn't realize that we had the files for the Java math library in our repo.  I do remember now our conversation about the boost library, which is I see is in the same area of our repo.  I like that having this library quickly available saved me some time, but it does seem strange to save local copies of things that are available on-line for download.  I guess the benefit is that we have the right versions for our code to work.

The immediate goal was to run the code.  This has mostly been accomplished (I haven't yet run the cpp or csharp examples.)  I do think we need to think more deeply about how we handle dependencies.  I do like the simple(?) idea of generating a file that accompanies the code and lists the dependencies and version numbers.

With respect to including `RUNARGS` in the call to `make`.  That makes sense to me.  The name of the input file is going to change, so "baking" it into `Makefile` doesn't make sense.  In the interest of making things easy for potential users of the generated code, I wonder if we could generate a `README` file that explains how to run the code?

bmaclach:
I like the idea to generate README files to explain dependencies and how to run the code. I agree that I expect that would be simple to do.

The reason we had to include the libraries in the repo was so that they'd be accessible during Travis builds, so we can build the NoPCM code during the build.

smiths:
I have completed my experiment.  Running the cpp code was easy.  I just had to copy over the boost library.  (It will be trickier for libraries that aren't header files, since we would need to compile and link the library.)

I wasn't able to run the csharp code, but I don't think that is unexpected, since the ODE library is from Microsoft.  Interestingly, I can get it to compile, but it fails when you try to run it.  I'm sure my Mac doesn't understand the dll file.  I did a quick google to see about compiling the ODE library for a Mac, but I don't think it is worth the effort.  Of course, if I'm missing something @bmaclach, please let me know.  If the csharp version is not support on Macs, that would be something we could say in the generated `README`.  If we generate a `README`.  :smile:

That makes sense that the libraries need to be around for the Travis builds.  I don't think I realized that part of the Travis build is to check that the generated code compiles.  Nice!  We don't actually do any test cases with the generated code, right?  That will be a nice thing to add in the future.

bmaclach:
Right, it makes sense the .dll file wouldn't work on other OS's. It might be possible to build the library in formats compatible with other OS's, but I agree probably not worth the effort (I had a hard enough time getting the .dll file, and I'm on Windows!). This could definitely be mentioned in the README.

And yes, we only compile the generated code but do not run or test it.

smiths:
I thought about creating an issue related to generating test cases for the generated code, and incorporating this into the Travis build, but I think that is too large an issue.  Issues work better when the path to closing them is reasonably clear, and not too time consuming.  Instead, I added a bit of wording to the potential projects document at:

https://gitlab.cas.mcmaster.ca/SEforSC/se4sc/-/tree/git-svn/Planning/PotGradStdProj

There is an issue assigned to me (#2050) to incorporate this kind of information into a "Contributing Page," but for now I've edited the potential project document.

smiths:
As a follow-up to the last comment, a potential projects page was created at:

https://github.com/JacquesCarette/Drasil/wiki/Potential-Projects

PR #2142, opened 2020-05-27
Stack Updates

Mornix:
This PR is two small stack related changes related to 340638bad523d7c7c5e2335fd66569a9f1985c25. 

1. Match the resolver used for the deploy site, uses its own YAML as its auxiliary and not regularly compiled with the rest of the project, with that of the project. The benefit of this is the dependencies will be the same version throughout and we're not using two different versions of GHC.

2. Update the minimum Stack version required in the Makefile. This change is for consistency. I though I may be able to remove the `check_stack.sh` script with the `required-stack-version` YAML flag, but if someone is on <1.9.1 then our `stack` command will fail with an "unknown argument" error. Of course this is bad and why the `check_stack.sh` script was added in the first place and is truly our minimum required Stack version. However, using `required-stack-version` in `stack.yaml` will cause stack invocations to fail if we're between 1.9.1 and 2.1.3. As such, I've updated our minimum version to 2.3.1 to prevent that case where someone with a pre-1.9.1 Stack tries to build, gets an error, updates to 1.9.1-2.1.3 (for some reason -- possibly dependency manager/OS related?) and gets another error message then requiring 2.3.1.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -76,7 +76,7 @@ ALL_EXPANDED_TARGETS = $(foreach P, $(PACKAGE_GEN_TARGET), $($(P)_PACKAGES)) $(f
 # make variables/configuration
 DIFF = diff --strip-trailing-cr -r -X ../.gitignore -x '*.txt'
 LOG_SUFFIX = _log.log
-MIN_STACK_VER = 1.9.1  # Version which adds --interleaved-output flag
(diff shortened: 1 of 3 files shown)


JacquesCarette:
Oddly, the builds didn't start automatically until I reviewed things. Not sure why they were stuck?

Issue #2116, opened 2020-05-14
Need a consistent purpDoc function among examples


muhammadaliog3:
### Context

While changing the purpose of the SWHS SRS example for [#1570](https://github.com/JacquesCarette/Drasil/issues/1570), I noted while every example intends to have a purpose, some examples have not implemented a purpose, and the examples that have a purpose each create there own purpose function.

### Problem:
There is no common function or methodology in creating the purpose section among the drasil examples. To promote software reuse we should have a common function/methodology for generating the purpose among Drasil examples. 

### The Facts

Note about Purpose section. All examples infact print out 2 paragraphs in the purpose section. However the second paragraph I found is always the same, and is hence probably automatically generated. Hence I will only be displaying code for the first paragraph which actually varies between examples. 

The following examples have not implemented a purpose yet: HGHC. 
The following examples share a purpDoc function already, hence output the same purpose: No_PCM and SWHS. 

**No_PCM and SWHS** 

Here is the function purpDoc:
```haskell
purpDoc :: Sentence -> CI -> Sentence
purpDoc spSent pro = foldlSent [S "The main", phrase purpose, S "of this",
  phrase document, S "is to describe the modelling of" +:+.
  spSent, S "The", plural goalStmt `sAnd` plural thModel,
  S "used in the", short pro, S "code are provided, with an emphasis",
  S "on explicitly identifying", plural assumption `sAnd` S "unambiguous" +:+.
  plural definition, S "This", phrase document,
  S "is intended to be used as a", phrase reference,
  S "to provide ad hoc access to all", phrase information,
  S "necessary to understand and verify the" +:+. phrase model, S "The",
  short Doc.srs, S "is abstract because the", plural content, S "say what",
  phrase problem, S "is being solved, but do not say how to solve it"]
```
The Latex text reads
"
The main purpose of this document is to describe the modelling of the solar water heating system incorporating PCM. The goal statements and theoretical models used in the SWHS code are provided, with an emphasis on explicitly identifying assumptions and unambiguous definitions. This document is intended to be used as a reference to provide ad hoc access to all information necessary to understand and verify the model. The SRS is abstract because the contents say what problem is being solved, but do not say how to solve it.
"
Note that as per issue #1570 that inspired this whole issue the function can be implemented with just the second parameter. 

The function ```purpDoc``` is called like ```purpDoc (phrase swhsPCM) progName```. 
```progName``` is according to the Concepts module 
```haskell 
[...]

PR #2155, opened 2020-05-29
Issue #2116 purp doc

muhammadaliog3:


muhammadaliog3:
I need some time to fix travis, so if you are reading this your time would be spent best elsewhere. Things may change!

The second push is for the stable version, so 5 of the examples that have an SRS have been updated, along with me finding a function that already titleizes commonIdeasWithDict in the Development/Sentence.hs module of doclang which simplifies code, ensuring that there is no redundancy in importing modules, and making ```Succinct```  a part of the DocLang module (I originally did this to test the purpDoc function but then I said that I might as leave it there cause it will likely be used in the future for user choices). 

(@oluowoj start reading here cause it concerns game physics)

The main issue is the following, (this can be found in the first sentence of the new purpose of SRS section in game physics that):

"The primary purpose of this document is to record the requirements of the game physics library."

"game physics library" is not capitalized. This can be explained by following codes
titleize is implemented as
``` titleize  n = NP.titleizeNP (n ^. term) ```
titleizeNP is implemented as 
```
titleizeNP  n = titleCase n phraseNP
```
titleCase is implemented as 
```
  titleCase n@ProperNoun {}         _ = phraseNP n
```
Hence as of right now there is no way to titleize a proper noun in drasil-lang. Note that the main difference between game physics and other examples that worked like SWHS is the fact that game physics is a proper noun while solar water heating system is common noun.  

```
progName    = commonIdeaWithDict "swhsName"   (nounPhrase "solar water heating system"
  "solar water heating systems") "SWHS" [materialEng]

gamePhysics = commonIdeaWithDict "gamePhysics" (pn "game physics library") "Game Physics" [physics]
```

Perhaps this is how the English language works? But it looks awkward in the only other place I could find "game physics library", where everything is capitalized except for "game physics library".

unnamed

We have 2 solutions: 
1. Change drasil-lang to capitalize proper nouns
2. Make "game physics library" as a common noun instead of proper noun
3. Make a separate variable where "game physics library" is a common noun, for cases where we need to use capitalize "game physics library". 

Probably the one of the 2 would be the best solution.

**Comment by smiths** [comment]
Great description of the problem @muhammadaliog3.  I especially like that you looked up the other instances where "game physics library" appeared.  Since it is used so rarely, and because in each instance it would make more sense if it were capitalized, my vote is that you follow your proposal 2 (make it a common noun).  @JacquesCarette may have some other thoughts with respect to unintended consequences.

File: code/drasil-docLang/Drasil/DocLang.hs [hunk 1]
```diff
@@ -12,13 +12,14 @@ module Drasil.DocLang (
   RefSec(..), RefTab(..), StkhldrSec(..), StkhldrSub(Client, Cstmr),
   TConvention(..), TraceabilitySec(TraceabilityProg), TSIntro(..), TUIntro(..),
   -- DocumentLanguage.Definitions
-  Field(..), Fields, InclUnits(IncludeUnits), Verbosity(Verbose), ddefn,
+  Field(..), Fields, InclUnits(IncludeUnits), Verbosity(..), ddefn,
```
JacquesCarette, line 15: Note that, in general, we prefer to be precise in our imports, `Verbose` is better than `..`. I won't block things just for that, but try to keep precise imports in the future.

code/drasil-docLang/Drasil/DocumentLanguage/Core.hs
```diff
@@ -96,7 +96,7 @@ data IntroSec = IntroProg Sentence Sentence [IntroSub]
 
 -- | Introduction subsections
 data IntroSub where
-  IPurpose :: Sentence -> IntroSub
+  IPurpose :: [Sentence] -> IntroSub
```
JacquesCarette, line 99: Note that you should consider using `NonEmpty` (which is a non-empty list), as we don't want to ever deal with length 0 lists. That can be a future improvement.

JacquesCarette:
And remember to close the issue.

Issue #2156, opened 2020-05-29
SSP: fixing scalar vector mismatch


smiths:
The discussion from #1115 motivates improving SSP by removing the scalar vector mismatch in the current definition of stress.  Removing the mismatch requires the following changes:

1. Change the current symbol tau (not bold) to tau^f. That is the symbol corresponding to "Shear Strength: The strength of a material against shear failure" should be changed from tau to tau^f.  This is the shear at failure, or shear strength.

2. Change DD:stress as follows:
    - change the Refname from DD:stress to DD:normStress
    - change the label to "Total normal stress"
    - change the Equation to sigma = F_n/A, where F_n is a new symbol, with the definition of "component of a force in the normal direction."  The normal force is a scalar, not a vector.
    - an associated change is that everywhere in the document where sigma is defined as total stress the corresponding change will be made to define it as "total normal stress."

3. Introduce a new data definition DD:tangStress, mostly the same as DD:normStress, except:
    - the label is "Tangential stress"
    - the equation is tau = F_t/A, where F_t is a new symbol, with the definition of "component of a force in the tangential direction."  (the symbol tau was previously used for shear strength, but by the change above, there is a new symbol (tau^f) for the shear strength, freeing up tau for the more conventional usage.)

4. GD:resShr currently references DD:stress, this should be changed to referencing DD:normStress and DD:tangStress.

@Nathaniel-Hu, I'm optimistic that these changes will be straightforward.  If you get stuck, I'm sure that @bmaclach can help us.  :smile:  Please do the items above in order, with a separate PR for each.  The more systematic we are, the more likely we'll notice a problem sooner, and we'll be able to back out our changes, if necessary.  Please reference this issue in your PR.

Nathaniel-Hu:
@smiths and @JacquesCarette, am I ok to change the variable names (e.g. totStress -> totNormStress or normStress) for the 3rd item of this issue? Or should I save that for the 4th item?

smiths:
@Nathaniel-Hu, I'm not sure what change you are referring to.  Probably best to stick to only making the changes in step 3.  If you have another change to propose, we should discuss it here.

Nathaniel-Hu:
@smiths, sorry, what I said was a bit confusing. This is what we have currently for normStress:
```haskell
--DD10

stressDD :: DataDefinition
stressDD = dd stressQD [makeCite huston2008] Nothing "normStress" []

stressQD :: QDefinition
[...]

PR #2157, opened 2020-06-01
Update SSP Shear Strength Symbol

Nathaniel-Hu:
- contains all changed files related to updating the Shear Strength Symbol from tau to tau^f for SSP
- i.e. Unitals.hs, SSP_SRS.tex, SSP_SRS.html
- to address Issue #2156 Item 1

### DIFF
--- a/code/drasil-example/Drasil/SSP/Unitals.hs
+++ b/code/drasil-example/Drasil/SSP/Unitals.hs
@@ -354,7 +354,7 @@ porePressure = uc' "u" (cn "pore pressure")
   "the pressure that comes from water within the soil" lU pascal
   
 shrStress = uc' "tau_i" (cn "shear strength")
-  "the strength of a material against shear failure" lTau pascal
+  "the strength of a material against shear failure" (sup lTau (Label"f")) pascal
 
 sliceHght = makeUCWDS "h_z,i" (cn "heights of interslice normal forces")
   ((plural height `inThe` phrase yDir) `ofThe` S "interslice normal forces on each slice")
--- a/code/stable/ssp/SRS/SSP_SRS.tex
+++ b/code/stable/ssp/SRS/SSP_SRS.tex
@@ -235,7 +235,7 @@ $σ'$ & Effective Stress: The stress in a soil mass that is effective in causing
 \\
 ${σ_{N}}'$ & Effective Normal Stress: The normal stress in a soil mass that is effective in causing volume changes; represents the average normal stress carried by the soil skeleton. & ${\text{Pa}}$
 \\
-$τ$ & Shear Strength: The strength of a material against shear failure. & ${\text{Pa}}$
+${τ^{\text{f}}}$ & Shear Strength: The strength of a material against shear failure. & ${\text{Pa}}$
 \\
 $\mathbf{τ}$ & Torque: A twisting force that tends to cause rotation. & $\text{N}\text{m}$
 \\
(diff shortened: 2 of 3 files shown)


Nathaniel-Hu:
@smiths and @JacquesCarette, I created branch Issue#2156 to store all my changes to address Issue #2156. It is possible for me to use this same branch multiple times for multiple pull requests for multiple changes linked to the same issue, correct?

bmaclach:
@Nathaniel-Hu You can, but the problem is that you can't have multiple PRs for the same branch open at the same time. If you push more commits to the same branch after opening a PR, those commits will get added to the open PR, which is usually not what we want.

What I usually did instead was create another new branch off of the first one (so in your case, branch off of the Issue#2156 branch). Then you have all of the changes from the first branch, but you can create a separate PR for the new branch. Then you should note in the description for your 2nd PR that it is dependent on the 1st one, so the 1st one should be merged first. 

I don't think the Contributor's guide is accurate with regards to this, so you probably want to update that. @JacquesCarette may have different thoughts about how best to do this too, so maybe wait for his input as well.

smiths:
@bmaclach, your advice sounds like the right approach to me.  Assuming @JacquesCarette agrees, this is something that we should explicitly state in the Contributor's Guide.

Nathaniel-Hu:
Ok then. Thanks for the insight @bmaclach. :)

JacquesCarette:
If this closes an issue, please say so specifically. From the discussion above, I won't delete this branch (yet), but ask if you want me to.

Nathaniel-Hu:
@JacquesCarette, this pull request was meant to be 1 of 4 that would be made in order to close Issue #2156 (as per @smiths requests). I will make sure to include whether the PR closes the issue or not in my next PRs. 

As per [Brooks' comment](https://github.com/JacquesCarette/Drasil/pull/2157#issuecomment-637079762), am I good to go and create a new branch off of branch Issue#2156 for my next PR? Also, @bmaclach for my 3rd PR, do I make it based off of the 2nd PR's branch, and then do the same for the 4th PR off of the 3rd PR Branch?

bmaclach:
@Nathaniel-Hu Assuming each PR requires the changes from the previous PR, yes. That may not be the case though (for example, looking at part 2 of #2156, it seems like it doesn't use the symbol from part 1, so that one may not actually be dependent). If you are not sure, a way that you can check whether two PRs are dependent is to assume that they are dependent (so branch off of the 1st PR's branch for the 2nd PR), then, after committing the changes (but before pushing), run:
```
git rebase --onto master old-branch-name new-branch-name
```
where `old-branch-name` should be replaced with the name of the previous branch, and `new-branch-name` should be replaced with the name of the current branch. This will attempt to re-structure the branches so that `new-branch-name` is branched directly off of `master`, instead of off of `old-branch-name`. If this succeeds, then the two branches aren't actually dependent, and `new-branch-name` will now be a branch off of `master`, which is what you would want in that case. If it fails saying there are conflicts, that means `new-branch-name` really was dependent on `old-branch-name`. That means the original structure of the branches was correct, so you can run `git rebase --abort` to revert back to that structure.

Also, since this PR has been merged now, you should just create a fresh branch from master for the future changes. If you already branched off of Issue#2156 and started to work on part 2, you can run the above `git rebase --onto ...` command to make it so your new branch is directly off of master (just remember to `git pull` on `master` first). Note that this only works if you haven't pushed the new branch to GitHub yet.

Let me know if any of this isn't making sense.

Nathaniel-Hu:
@bmaclach thanks for the insight. I have rebased a branch before, however I still have mixed feelings about using `git rebase` (not entirely comfortable with using it again quite yet). I'll just create a fresh branch for my next PR then. 

@JacquesCarette, you should be good to delete branch Issue#2156 now, I will just create a new branch for my next PR.

PR #2160, opened 2020-06-02
Update Stress to NormStress + Implement TangStress + Update Dependencies

Nathaniel-Hu:
- contains all changed files related to updating the Stress Data Definition to NormStress + Implementing the new TangStress Data Definition
- implementation of new TangStress Data Definition also necessitated creation of:
  - new symbol F_t for SSP
  - QDefinition TangStressQD
  - Expr TangStressEqn
  - Unital Chunk TangStress

- updated all references of totStress to totNormStress (changes related to Item 2)
- changed files: DataDefs.hs, TMods.hs, Unitals.hs, SSP_SRS.tex, SSP_SRS.html
- contributes to #2156 (addresses Issue Item 3, includes changes related to Item 2)

### DIFF
--- a/code/drasil-example/Drasil/SSP/DataDefs.hs
+++ b/code/drasil-example/Drasil/SSP/DataDefs.hs
@@ -30,7 +30,7 @@ dataDefs = [intersliceWtrF, angleA, angleB, lengthB, lengthLb, lengthLs,
   slcHeight, normStressDD, tangStressDD, torqueDD, ratioVariation, convertFunc1, 
   convertFunc2, nrmForceSumDD, watForceSumDD, sliceHghtRightDD, sliceHghtLeftDD]
 
---DD4
+--DD intersliceWtrF: interslice normal water forces
 
 intersliceWtrF :: DataDefinition
 intersliceWtrF = dd intersliceWtrFQD [makeCite fredlund1977] Nothing "intersliceWtrF"
@@ -51,7 +51,7 @@ intersliceWtrFEqn = completeCase [case1,case2,case3]
 
         case3 = (0, inxi waterHght $<= inxi slipHght)
 
---DD5
+--DD angleA: base angles
 
 angleA :: DataDefinition
 angleA = dd angleAQD [makeCite fredlund1977] Nothing "angleA" 
@@ -70,7 +70,7 @@ angleANotes = foldlSent [S "This", phrase equation, S "is based on the",
   phrase assumption, S "that the base of a", phrase slice,
   S "is a straight line", sParen (makeRef2S assumpSBSBISL)]
 
---DD6
+--DD angleB: surface angles
 
 angleB :: DataDefinition
 angleB = dd angleBQD [makeCite fredlund1977] Nothing "angleB"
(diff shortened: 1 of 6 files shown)


code/drasil-example/Drasil/SSP/DataDefs.hs
```diff
@@ -183,7 +194,7 @@ ratioVarEqn = completeCase [case1, case2]
         case2 = (sin (sy QM.pi_ * ((inxi slipDist - idx (sy slipDist) 0) /
                 (indxn slipDist - idx (sy slipDist) 0))), UnaryOp Not (sy constF))
 
---DD12
+--DD13
```
smiths, line 197: I'm not sure there is much value to comments with hard-coded numbers.  I can see you are making an effort to keep them in sync, but it is a challenge for maintenance.  The numbers are something that are automatically generated in the tex file.  We should either delete these comments, or change them to something meaningful, like the name of the DD that is being defined.

code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -237,6 +239,8 @@ \subsection{Table of Symbols}
 \\
 ${σ_{N}}'$ & Effective Normal Stress: The normal stress in a soil mass that is effective in causing volume changes; represents the average normal stress carried by the soil skeleton. & ${\text{Pa}}$
 \\
+$τ$ & Tangential Stress: The total force per area acting on the soil mass. & ${\text{Pa}}$
```
smiths, line 242: Please change this definition to "The shear force per unit area."  We don't have the concept of total force for shear.  We also don't need to mention the soil mass.  This definition is true whether we are talking about soil, or any other material.

bmaclach: @smiths If we are making an effort to keep these generic (i.e. by not mentioning soil mass, etc.), it may be worthwhile to move them (normal/tangential force/stress) to data-drasil instead of defining them directly in SSP, so they can be re-used by potential future examples. Thoughts?

smiths: Yes, @bmaclach, moving the generic concepts to data-drasil makes sense.  I don't think we need to make this a high priority at the moment, since ssp is the only example with the concept of stress, but this is the way things should go in the future.  I'd like to see us attack this problem for all of the examples, using two different approaches:

1. Top-down - we organize the information using an ontology for physics, math and computing knowledge.  Ideally we would be able to find an existing ontology, instead of having to invent something.  Unfortunately, my preliminary efforts to find existing ontologies haven't been that encouraging.
2. Example driven - we add more examples and let the ontology build up organically.  For instance, the concept of stress applies to all continuum mechanics problems, but the concept of effective stress is reserved for continuum mechanics problems where the material is a soil, or other granular material.
I'll edit the "potential projects" wiki page to mention refactoring the existing Drasil examples as the ontology evolves.

code/drasil-example/Drasil/SSP/DataDefs.hs
Nathaniel-Hu: Ok then. I've updated the comments by removing the numbers and adding in the DD name and label like so. 
```haskell
--DD ratioVariation: interslice normal to shear force ratio variation function

ratioVariation :: DataDefinition
ratioVariation = dd ratioVarQD [makeCite fredlund1977] Nothing 
  "ratioVariation" []

ratioVarQD :: QDefinition
ratioVarQD = mkQuantDef scalFunc ratioVarEqn

ratioVarEqn :: Expr
ratioVarEqn = completeCase [case1, case2]
  where case1 = (1, sy constF)

        case2 = (sin (sy QM.pi_ * ((inxi slipDist - idx (sy slipDist) 0) /
                (indxn slipDist - idx (sy slipDist) 0))), UnaryOp Not (sy constF))
```
Does that work @smiths?

Nathaniel-Hu:
@smiths, I have made the changes requested. Do any other changes need to be made?

PR #2166, opened 2020-06-04
Issue 2145

### LINKED ISSUE #2145: Don't print "private: " or "public: " in C++ classes if they are empty
In C++ header files, classes are structured like:
```
class Foo {
  public:
  // declarations of public variables and methods 
  ...
  private:
  // declarations of private variables and methods
  ...
};
```
Currently, when generating C++ classes, the `cpphClass` function in the C++ renderer generates the "public: " heading even if the class has no public members (and likewise for "private: "). We should add a check for if there are no public (or private) members, and not generate the corresponding heading if so.

muhammadaliog3:
Closes #2145

### DIFF
--- a/code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
+++ b/code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
@@ -714,10 +714,10 @@ instance (Pair p) => StateVarSym (p CppSrcCode CppHdrCode) where
   type StateVar (p CppSrcCode CppHdrCode) = StateVarData
   stateVar s p = pair1 (stateVar (pfst s) (pfst p)) (stateVar (psnd s) (psnd p))
     . zoom lensCStoVS
-  stateVarDef n s p vr vl = pair2
-    (stateVarDef n (pfst s) (pfst p)) 
-    (stateVarDef n (psnd s) (psnd p)) (zoom lensCStoVS vr) (zoom lensCStoVS vl)
-  constVar n s vr vl = pair2 (constVar n (pfst s)) (constVar n (psnd s))
+  stateVarDef s p vr vl = pair2
+    (stateVarDef (pfst s) (pfst p)) 
+    (stateVarDef (psnd s) (psnd p)) (zoom lensCStoVS vr) (zoom lensCStoVS vl)
+  constVar s vr vl = pair2 (constVar (pfst s)) (constVar (psnd s))
     (zoom lensCStoVS vr) (zoom lensCStoVS vl)
 
 instance (Pair p) => StateVarElim (p CppSrcCode CppHdrCode) where
@@ -1528,7 +1528,7 @@ instance StateVarSym CppSrcCode where
   stateVar s _ _ = onStateValue (on3CodeValues svd (onCodeValue snd s) (toCode 
     empty)) $ zoom lensCStoMS emptyStmt
   stateVarDef = cppsStateVarDef empty
-  constVar n s = cppsStateVarDef constDec' n s static
+  constVar s = cppsStateVarDef constDec' s static
   
 instance StateVarElim CppSrcCode where
   stateVar = stVar . unCPPSC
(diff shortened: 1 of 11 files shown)


code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
```diff
@@ -2606,17 +2606,19 @@ cpphClass :: Label -> CppHdrCode ParentSpec ->
...
-cpphClass n ps vars funcs pub priv = onCodeValue (\p -> vcat [
+cpphClass n ps vars funcs pub priv = let 
+  pubs = cpphVarsFuncsList Pub vars funcs
+  privs = cpphVarsFuncsList Priv vars funcs
+  indLi :: Doc -> Doc -> [Doc]
+  indLi pb pr
+    | isEmpty pb = [RC.scope priv <> colon, indent pr]
+    | isEmpty pr = [RC.scope pub <> colon, indent pb]
+    | otherwise   = [RC.scope pub <> colon, indent pb, blank,
+      RC.scope priv <> colon, indent privs]
```
bmaclach, line 2617: @muhammadaliog3 There's a helper function called `emptyIfEmpty` in Helpers.hs in the drasil-gool subpackage. I think you could use it to make this function cleaner/shorter.

muhammadaliog3: @bmaclach 
```haskell
emptyIfEmpty :: Doc -> Doc -> Doc
emptyIfEmpty ifDoc elseDoc = if isEmpty ifDoc then empty else elseDoc
```
I don't think I could use this function. Firstly because I need 3 if statements I would have to nest an ```emptyIfEmpty``` inside another ```emptyIfEmpty```, which may clutter things. The  other thing is that I need to return a list of [Doc] and not Doc. I can't just return ```pb``` if it is not empty, I need to return ```RC.scope pub <> colon``` and ```pb``` together in a list. Hence I think that using this function would complicate things.

bmaclach: I was thinking you could use `emptyIfEmpty` 3 times, once for "private:", once for "public:", and once for the blank line in between. Since you'd be using it on individual elements of the list of `Doc`, the types would work. 

Having to use it three times does add some clutter, but in return you would no longer need `indLi` and the duplicated code in its 3 cases. So I think it's worth thinking about.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
```diff
@@ -2610,15 +2610,14 @@ cpphClass :: Label -> CppHdrCode ParentSpec ->
...
-  indLi pb pr
-    | isEmpty pb = [RC.scope priv <> colon, indent pr]
-    | isEmpty pr = [RC.scope pub <> colon, indent pb]
-    | otherwise   = [RC.scope pub <> colon, indent pb, blank,
-      RC.scope priv <> colon, indent privs]
+  ifEmptyPubs = emptyIfEmpty pubs
+  ifEmptyPrivs = emptyIfEmpty privs
+  indLi = [ifEmptyPubs (RC.scope pub <> colon), ifEmptyPubs (indent pubs),
+          ifEmptyPubs (ifEmptyPrivs blank),
+          ifEmptyPrivs (RC.scope priv <> colon), ifEmptyPrivs (indent privs)]
```
muhammadaliog3, line 2617: Here was what the naive solution looked like 
```haskell
cpphClass n ps vars funcs pub priv = let 
  pubs = cpphVarsFuncsList Pub vars funcs
  privs = cpphVarsFuncsList Priv vars funcs
  indLi = [emptyIfEmpty pubs (RC.scope pub <> colon), 
          emptyIfEmpty  pubs (indent pubs),
          emptyIfEmpty  pubs (emptyIfEmpty privs blank),
          emptyIfEmpty privs (RC.scope priv <> colon), 
          emptyIfEmpty privs (indent privs)] ............
```
This takes the same number of lines of code, is easier to read, but perhaps looks more repetitive. If you would like this version I could use it.

You may be looking at the stable version and noticing that there are still blank lines. 

for example in Projectile.hpp
```C++
class Constants {
    public:
        static const double g_vect;
        static const double epsilon;
        
};
```
However I believe that this is because the public code probably has some kind of empty line. The reason I say this is 1. If you put your cursor on the empty line it is indented. 2. if you edit the code and remove everything except for the public code you still see this empty line.

bmaclach: Good catch about the blank line being indented. I bet it's the `cpphVarsFuncsList` function that is adding the blank line (which is not related to your changes). It's probably a quick fix. That might be something you are interested in doing (in a separate PR).

JacquesCarette: I'm fine with the repetition in this code. That can always be optimized later.

PR #2167, opened 2020-06-05
Preventing duplicate symbols

### LINKED ISSUE #1916: Detect and complain if the same symbol is used for two or more concepts within the same document

muhammadaliog3:
Closes #1916

### DIFF
--- a/code/drasil-docLang/Drasil/Sections/TableOfSymbols.hs
+++ b/code/drasil-docLang/Drasil/Sections/TableOfSymbols.hs
@@ -5,8 +5,7 @@ import Language.Drasil
 
 import Data.List (nub, (\\))
 import Control.Lens (view)
-import Data.Bifunctor (first)
-import Text.PrettyPrint.HughesPJ (render)
+import Text.PrettyPrint.HughesPJ (text, render, vcat, (<+>))
 
 import Drasil.DocumentLanguage.Units (toSentence)
 import Data.Drasil.Concepts.Documentation (symbol_, description, tOfSymb)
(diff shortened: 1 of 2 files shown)


code/drasil-docLang/Drasil/Sections/TableOfSymbols.hs
```diff
@@ -21,16 +20,22 @@ table st ls f
...
-        symUidPairDuplicates = [(first (render . symbolDoc) x, snd x) | x <- symUidPair, 
-          fst x `elem` symDuplicates]
+        -- If there are duplicates then the following will extract the UID's of duplicates symbols
+        extractPairs symb = filter (\x -> fst x == symb) symUidPair
+        extractUid  = map (\x -> snd x)
+        extractUidFromPairs = text . show . extractUid . extractPairs
+        errSymUidDuplicates = vcat $ map (\symb -> 
+          extractUidFromPairs symb<+>text "all have symbol"<+>symbolDoc symb) symDuplicates
+        errorMessage = "Same symbols for different quantities found: " ++ render errSymUidDuplicates
+
```
muhammadaliog3, line 38: This is as clean as I could get it.

The following is the new error message, say that we have 2 examples sharing the same symbol and 3 examples with another symbol, then the error will print different lines for each duplicated symbols. Like the following:
Same symbols for different quantities found: ["dimMax","dimMin"] all have symbol d_max
["cWeightMin","sdMax","sdMin"] all have symbol w_min

JacquesCarette: Good.

PR #2172, opened 2020-06-09
Generate Design Log

### LINKED ISSUE #2088: Generate design log with code
Drasil's code generator should generate a "design log" each time it generates an example. The design log would tell the user of any instances where the generator did not use their first choice, for the design choices where the user gives an preferentially-ordered list of options. Right now there are only three such choices: `conceptMatch`, `spaceMatch`, and `odeLibs` (which is in the `noPCMODE` branch).

If the generator can use the first choice, then it adds nothing to the log. But any time a choice can't be used and the generator moves on to the next choice, it should log that the first choice wasn't used and why.

bmaclach:
This PR updates the generator to build a design log (which required switching from `Reader` to `State`). The design log logs when there is a conflict between the users' choices, leading to one of their back-up choices being used instead. If there are no conflicts, no design log is generated.

I've added the generated design logs to stable.

Closes #2088.

### DIFF
--- a/code/drasil-code/Language/Drasil/Choices.hs
+++ b/code/drasil-code/Language/Drasil/Choices.hs
@@ -1,7 +1,7 @@
 module Language.Drasil.Choices (
   Choices(..), Modularity(..), InputModule(..), inputModule, Structure(..), 
   ConstantStructure(..), ConstantRepr(..), MatchedConceptMap, CodeConcept(..), 
-  matchConcepts, SpaceMatch, MatchedSpaces, matchSpaces, ImplementationType(..),
+  matchConcepts, SpaceMatch, matchSpaces, ImplementationType(..),
   ConstraintBehaviour(..), Comments(..), Verbosity(..), Visibility(..), 
   Logging(..), AuxFile(..), getSampleData, hasSampleInput, defaultChoices
 ) where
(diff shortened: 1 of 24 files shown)


bmaclach:
@smiths It can be done, but it's not straightforward. I think we could do a pass on `Choices` and render any choice that is fixed by that point (i.e. any choice that is not represented as a preferentially-ordered list) as a `Doc` that can be added to the log. Then for the choices that are not fixed until they are processed in `DrasilState`, (i.e. `conceptMatch`, `spaceMatch`, and `odeLibs`, the three choices logged by this PR), we add their values to the log as we currently do in this PR (except add a line to the log that specifies the value that was actually selected, instead of only logging the conflicts).

The trickiest part will still be the `SpaceMatch` choice, because this is technically an infinite mapping between `Space`s and `CodeType`s, and we don't want to generate an infinite log. That was one reason I held back from logging the actual values of the choices in this PR.

Now that you've asked the question, though, and I've thought about how I would do it if I had to, I think a potential solution to that problem would be to only log the space matches that are actually used in the generated program. That will always be a finite (and probably quite small) subset of all of the space matches, and it would not be hard to implement that.

So I guess, in answer to your actual question, it wouldn't be so hard after all.

smiths:
@bmaclach, thank you for the information.  @JacquesCarette will have to do his review, but I think we should merge in what you have, and leave a more detailed log as future work.  Nice to know that it is feasible to create one, but I think we can safely say that this is outside of the scope of your thesis.

bmaclach:
@smiths Sounds good, I'll create the new issue.

PR #2178, opened 2020-06-12
Issue 2076

### LINKED ISSUE #2076: Make GOOL's "FromData" functions stateful

muhammadaliog3:
Closes #2076 
To whomever it is that reviews the PR, it might be difficult going through the commits in order, as I re-did a lot of thingsto make it cleaner, especially in this commit:
c04a57e3d814bbfb79d576c1f8218b035690b5cd 
https://github.com/JacquesCarette/Drasil/commit/c04a57e3d814bbfb79d576c1f8218b035690b5cd


code/drasil-gool/GOOL/Drasil/LanguageRenderer/CLike.hs
```diff
@@ -54,20 +54,20 @@ charRender = "char"
...
 
 char :: (RenderSym r) => VSType r
-char = toState $ typeFromData Char charRender (text charRender)
+char = typeFromData Char charRender (text charRender)
 
 listType :: (RenderSym r) => String -> VSType r -> VSType r
-listType lst = onStateValue (\t -> typeFromData (List (getType t)) (lst 
-  `containing` getTypeString t) (text lst <> angles (RC.type' t)))
-  
+listType lst = on1StateWrapped (\t -> typeFromData (List (getType t)) (lst
```
JacquesCarette, line 66: I think I would prefer to use `do` notation instead of `on1StateWrapped`, everywhere.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CLike.hs
```diff
@@ -120,26 +121,30 @@ listSize v = v $. S.listSizeFunc
...
-  <+> RC.value vl)) (S.varDec vr) (zoom lensMStoVS vl')
+varDecDef vr vl' = on2StateWrapped (\vd vl -> mkStmt 
+  (RC.statement vd <+> equals <+> RC.value vl)) (S.varDec vr) $ 
+  zoom lensMStoVS vl'
 
 listDec :: (RenderSym r) => (r (Value r) -> Doc) -> SValue r -> SVariable r -> 
   MSStatement r
-listDec f vl v = on2StateValues (\sz vd -> mkStmt (RC.statement vd <> f 
-  sz)) (zoom lensMStoVS vl) (S.varDec v)
+listDec f vl v = on2StateWrapped (\sz vd -> mkStmt (RC.statement vd <> f sz))
```
JacquesCarette, line 149: I think a lot of `on2StateWrapped` would also be better using `do` notation. This one for sure.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/Constructors.hs
```diff
@@ -150,15 +151,13 @@ binExpr' = on3StateValues (\b v1 v2 ->
 -- either value passed to the operator is a float, this function preserves that 
 -- type by casting the result to a float.
 binExprNumDbl' :: (RenderSym r) => VSBinOp r -> SValue r -> SValue r -> SValue r
-binExprNumDbl' b' v1' v2' = do
-  b <- b'
-  v1 <- v1'
-  v2 <- v2'
+binExprNumDbl' = on3StateWrapped (\b v1 v2 -> do
```
JacquesCarette, line 154: I think I preferred the `do`-based version.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
```diff
@@ -286,8 +284,8 @@ instance (Pair p) => InternalVarElim (p CppSrcCode CppHdrCode) where
   variable v = RC.variable $ pfst v
 
 instance (Pair p) => RenderVariable (p CppSrcCode CppHdrCode) where
-  varFromData b n t d = pair (varFromData b n (pfst t) d) 
-    (varFromData b n (psnd t) d)
+  varFromData b n t' d = pair1 (\t ->varFromData b n t d)
```
JacquesCarette, line 286: This pattern (of having to introduce a lambda) often [but not always] means that the order of arguments in `varFromData` is sub-optimal, and it would be better if `t` was last. This should be investigated.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/LanguagePolymorphic.hs
```diff
@@ -201,10 +200,10 @@ litString :: (RenderSym r) => String -> SValue r
 litString s = mkStateVal S.string (doubleQuotedText s)
 
 valueOf :: (RenderSym r) => SVariable r -> SValue r
-valueOf = onStateValue (\v -> mkVal (variableType v) (RC.variable v))
+valueOf v' = v' >>= (\v -> mkVal (variableType v) (RC.variable v))
```
JacquesCarette, line 203: I prefer `do` over `>>=`. For human readability.

code/drasil-gool/GOOL/Drasil/RendererClasses.hs
```diff
@@ -60,7 +60,7 @@ class (BlockCommentSym r) => RenderFile r where
 
   commentedMod :: SFile r -> FS (r (BlockComment r)) -> SFile r
 
-  fileFromData :: FilePath -> r (Module r) -> SFile r
+  fileFromData :: FilePath -> FSModule r -> SFile r
```
JacquesCarette, line 63: I very much approve of making these stylistic changes. However, I would have preferred a few more PRs, each with its own set of smaller changes, rather than bundling these up with a bigger change.

code/drasil-gool/GOOL/Drasil/Helpers.hs
```diff
@@ -80,8 +80,23 @@ onStateList f as = f <$> sequence as
 on2StateLists :: ([a] -> [b] -> c) -> [State s a] -> [State s b] -> State s c
 on2StateLists f as bs = liftM2 f (sequence as) (sequence bs)
 
-on1StateValue1List :: (a -> [b] -> c) -> State s a -> [State s b] -> State s c
-on1StateValue1List f a as = liftM2 f a (sequence as)
+on1StateWrapped :: (Monad m) => (a -> m b) -> m a -> m b
+on1StateWrapped f a' = do
```
JacquesCarette, line 84: Ah, that's not what I meant!  I meant to use `do` notation in the code itself, **instead** of using `on1StateWrapped`, `on2StateWrapped`.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CLike.hs
JacquesCarette: Like here. There were reasons before to provide combinators, but because of various refactors, these have gone away, so that now the code would be clearer using `do` notation.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/Constructors.hs
muhammadaliog3: I have encountered a problem here. Due to the repetition unless I use on3StateWrapped, it won't pass the Hlint. Here is the repetition
```haskell
-- | Constructs binary expressions like v + w, for some operator + and values v 
-- and w, parenthesizing v and w if needed.
binExpr :: (RenderSym r) => VSBinOp r -> SValue r -> SValue r -> SValue r
binExpr b' v1' v2'= do 
  b <- b'
  v1 <- v1'
  v2 <- v2'
  let exprType = numType (valueType v1) (valueType v2)
      exprRender = binExprRender b v1 v2
  mkExpr (bOpPrec b) exprType exprRender

-- | Constructs binary expressions like pow(v,w), for some operator pow and
-- values v and w
binExpr' :: (RenderSym r) => VSBinOp r -> SValue r -> SValue r -> SValue r
binExpr' b' v1' v2'= do 
  b <- b'
  v1 <- v1'
  v2 <- v2'
  let exprType = numType (valueType v1) (valueType v2)
      exprRender = binOpDocD' (RC.bOp b) (RC.value v1) (RC.value v2)
  mkExpr 9 exprType exprRender

-- | To be used in languages where the binary operator returns a double. If 
-- either value passed to the operator is a float, this function preserves that 
-- type by casting the result to a float.
binExprNumDbl' :: (RenderSym r) => VSBinOp r -> SValue r -> SValue r -> SValue r
binExprNumDbl' b' v1' v2' = do 
  b <- b'
  v1 <- v1'
  v2 <- v2'
  let exprType = numType numType (valueType v1) (valueType v2)
      exprRender = binOpDocD' (RC.bOp b) (RC.value v1) (RC.value v2)
  e <- mkExpr 9 exprType exprRender
  binExprCastFloat t1 t2 e
```

JacquesCarette: You should pull out the common lines into a function.  One for applying `numType` and one for `binOpDocD'`. And then rewrite the above in terms of those.

PR #2186, opened 2020-06-16
Issue 2180 rm iter

### LINKED ISSUE #2180: Remove Iterator type from GOOL
GOOL currently has a type constructor called `iterator` and a variable constructor called `iterVar`. These only exist because, in C++, in for-each loops, the iterator variable needs to be de-referenced when used in the loop body. Thus, if someone writes GOOL and forgets to use `iterVar` when defining the body of a for-each loop, the generated code will not work in C++.

Now that we have state in GOOL, I think we can remove the `iterVar` and `iterator` functions from GOOL. We can instead have the `forEach` function update the state by storing the current iterator variable(s) --- each of which is passed as a parameter to `forEach` --- in the state. Then the C++ renderer can check (probably in the `valueOf` function) the state and add de-referencing syntax if the variable is a current iterator variable.

muhammadaliog3:
Closes #2180


muhammadaliog3:
The change: Iterators are no longer available outside of the foreach function within the CPPrenderer
Some key decisions:
1. HDR no longer has an iterator, this is because iterators are only available within forEach loops, and HDR had no forEach loop.
2. MethodState now carried another record/lens "iterators" which is reset to [] whenever the for loop ends.
3. Language Polymorphic no longer contains any function related to iterators
4. Cpp valueOf for srccode implements its own personal valueOf function that always checks if variable is an iterator (increases time complexity).
Potential Issue: Gool can no longer right

code/drasil-gool/GOOL/Drasil/ClassInterface.hs
```diff
@@ -594,8 +594,8 @@ convType Double = double
 convType Char = char
 convType String = string
 convType (List t) = listType (convType t)
+convType (Iterator _) = error "Iterators are only available within forEach loops in C++"
```
muhammadaliog3, line 597: I didn't really know much about convType, so this might be questionable.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
```diff
@@ -1165,7 +1166,14 @@ instance MathConstant CppSrcCode where
...
 instance VariableValue CppSrcCode where
-  valueOf = G.valueOf
+  valueOf vr' = do 
+    vr <- vr'
+    its <- zoom lensVStoMS getIter
+    let nvr = variableName vr 
+    if (nvr `elem` its) then (do
+        itvr <- iterator $ toState $ variableType vr
+        toState $ mkVal itvr (RC.variable vr)) 
+      else G.valueOf vr'
```
muhammadaliog3: This is what I was talking about with a do statement inside of a do statement, so make sure this is fine.

code/drasil-gool/GOOL/Drasil/State.hs
```diff
@@ -494,8 +494,8 @@ addIter st = over iterators ([st]++)
 getIter :: MS [String]
 getIter = gets (^. iterators)
 
-resetIter :: MethodState -> MethodState
-resetIter = set iterators []
+resetIter :: String -> MethodState -> MethodState
+resetIter st = over iterators (delete st)
```
muhammadaliog3: I had to change the resetIter function to this because I realized that my previous changes could lead to bugs with nested for each loops. If you want to know what bug is, then in a nutshell: Previously the code written within the outer for loop but after the inner for loop is closed would have produced errors because ```resetIter``` removed all Iter variables (including the outer one).

code/drasil-gool/GOOL/Drasil/ClassInterface.hs
bmaclach: You should be able to remove the `Iterator` node from the definition of `CodeType`. Then you can remove this whole line from `convType`.

I'm thinking the only place `Iterator` is currently used is in the `cppIterType` function, but you can just remove it there too.

JacquesCarette: I think all traces of `Iterator` should go, so that should make the above obsolete too.

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
bmaclach: You can use `mkStateVal` instead of `toState $ mkVal`. And `mkStateVal` takes a `VSType r`, so you won't need to use the do notation if you switch to that (i.e. you won't need to bind the iterator type to `itvr`).

code/drasil-gool/GOOL/Drasil/ClassInterface.hs
muhammadaliog3: Oh I can't do that, cause this function uses CodeType
```haskell
cppIterType :: (RenderSym r) => VSType r -> VSType r
cppIterType = onStateValue (\t -> typeFromData (Iterator (getType t)) 
  (getTypeString t `nmSpcAccess` cppIterator) (stdAccess' (RC.type' t) 
  `nmSpcAccess'` text cppIterator))```

code/drasil-gool/GOOL/Drasil/LanguageRenderer/CppRenderer.hs
JacquesCarette: Such a construct is fine.

JacquesCarette:
So this code all seems fine - but it also doesn't remove the `Iterator` construct (from `CodeType`?). Is that on purpose?

muhammadaliog3:
Its fixes now

PR #2187, opened 2020-06-19
Add Swift Renderer to GOOL

bmaclach:
I have implemented a Swift renderer for GOOL. Overall, Swift fit rather nicely into GOOL. Very few infrastructure changes were required. Unfortunately, synchronous file reading/writing does not seem straighforward in Swift. There's a `FileHandle` type, but the methods for reading and writing are deprecated. The only other way I could find was to use `InputStream` and `OutputStream`, but they are confusing and my attempts to get code using these types working locally were unsuccessful (it's difficult to test swift file I/O locally on Windows, since Swift is not meant to be run on Windows. Mostly I used online tools (like repl.it) to test generated swift code, but those also can't run file I/O code). Ultimately, I left the GOOL methods related to synchronous file I/O unimplemented.

I locally updated our GOOL tests (the ad-hoc ones I wrote, in drasil-code/Test, not the case studies) to generate Swift code, and tested them on repl.it to confirm the generated code works (except for the test involving file I/O, since I didn't implement that in Swift). I only did this locally because I don't see how we will be able to run Swift code in Travis. From my understanding, our Travis environment installs needed packages using `apt-get`, but swift cannot be installed that way. According to this page I found in the Travis docs, Swift code cannot be run in Linux Travis environments: https://docs.travis-ci.com/user/languages/objective-c/.

This PR will have many conflicts with the currently open PR (#2178). Probably it will be fastest to merge that first, assuming it is ready to be merged, and I'll then deal with the conflicts.

The following is a detailed discussion of this PR:

# Statistics
I spent about 3.5 work days implementing this. My estimate of 2 days during my seminar was definitely optimistic, but not that far off!

I organized every single GOOL method into buckets related to how they were implemented in Swift. Below is a count for each bucket. I also have more details (i.e. which methods are in which buckets), but I'll provide those if requested.

## Full re-use
These are methods that were implemented just by calling a generic function. 
Count: 133

## Full re-use with lifts
Similar to the above, but some lifting into the code/state monads was also required.
Count: 21

## Re-use with different arguments
These are methods that were implemented by calling a generic function, but one or more arguments to the generic function were specific to the Swift renderer. For example, they are of the form:
`method = G.method swiftSpecificArgument`
I further divided these into categories based on how complicated the swift-specific arguments were to write:

### Simple
These are simple, one-liner arguments.
Count: 17

### Simple, but multi-line
These are still straight-forward, but the argument's value required multiple lines to define.
Count: 4

## Partial re-use
These call generic functions in their implementation, but also require more. They generally have the form `method = ... (G.method ...) ...`
I further divided these into categories based on how complicated the non re-used part of the implementation was to write:

### Simple
These are implemented via one-liners, despite only part of the implementation re-using existing functions.
Count: 5

### Simple, but multi-line
These require multiple lines to implement, but are still straightforward.
Count: 11

### Complicated
These required some non-obvious logic to implement
Count: 2

## Note
The above counts include 35 methods where the re-used code first had to be pulled out into a generic function.

## Internal re-use
Sometimes Swift-specific code had to be written to implement one method, but then other methods could re-use that swift-specific code in their implementations. I categorized these other methods as having "internal re-use", and further divided them like below:

### Full internal re-use
These methods are fully implemented by re-using code for other swift methods.
Count: 34

### Partial internal re-use
These methods re-use swift-specific code, but also require even more specific code in their implementations. I did not further categorize these based on how complicated the non-reused part of the implementation was, because there is only one and it was simple to implement.
Count: 1

## No re-use
The implementations for these did not re-use any code, and were entirely swift-specific. I further categorized them based on how complicated they were to implement:

### Super obvious
These have very short implementations that are expected to be the same or analogous in any language renderer. Mostly, these are the "elimination" methods for extracting information from a GOOL entity, which are implemented by calling the `unRepr` function for the language.
Count: 34

### Obvious
These implementations were very Swift-specific, but still obvious in that they were just straightforward definitions of how something gets rendered in Swift.
Count: 23

### Complicated
These required some non-obvious logic or thinking to implement.
Count: 5

## Unimplemented
These are the functions related to synchronous file I/O that I left unimplemented because I couldn't figure them out:
- `getFileInput`
- `discardFileInput`
- `getFileInputLine`
- `discardFileInputLine`

# Infrastructure changes
I added a `mthdFromData` method so that I could write more generic functions for implementing some method constructors, but this was not related to anything about Swift specifically.

Swift uses named arguments with greater frequency than any of the other languages. We already supported named arguments, but I added an `objMethodCallNamedArgs` smart constructor for the case where all of a method's arguments are named.

The only really meaningful infrastructure change came from the fact that Swift doesn't have much in the way of built-in error types to throw. Instead, it provides an `Error` interface (or "protocol", as Swift calls it), and a Swift programmer must implement that interface to define error types to throw. Since GOOL's `throw` function is currently implemented to throw generic errors with an error message string, I implemented `throw` to render statements like `throw "error message string here"`, which only works if the `String` type (in Swift) implements the `Error` protocol. Thus, if `throw` is used, I store the fact that it was used in a new field in the state, and then generate `extension String: Error {}` at the top of a module that uses throw. This extension only needs to be generated once in the whole program, so I added another field to the state that says when it has already been generated, so we don't generate it extra times.

# Other Notes
If you look at Swift documentation, they refer to code like `object.var` and `object.var = x` as getters and setters, but this seems to me like just a difference in terminology, so I implemented `getMethod` and `setMethod` in GOOL the same way as I did for other languages (i.e. they define methods called "getX" or "setX").

Unlike our other target languages, Swift uses optional values frequently (like Haskell's `Maybe` monad), so there were a a few places where I had to render an `!` after a value to unwrap it from the optional context. 

Swift is not one of Doxygen's supported languages, so instead of generating Doxygen-style comments for GOOL's `doc` functions, the Swift renderer generates comments that are in a format that can be recognized by XCode (Apple's IDE for writing Swift) to show documentation when you hover over a function. There's a tool called Jazzy (https://github.com/realm/jazzy) that seems like it can compile these comments into documentation, similar to Doxygen, though I haven't looked deeply into it.


smiths:
Very impressive!  This is great to have another example, and the record of what was involved in adding it will really strengthen our eventual publication about this.  I'm impressed with the detailed notes you took and the quantifications of the number of methods that needed to be modified, not modified, etc.  Great!

Am I write that you haven't added any of the generated swift code to stable?  I'd like to see what the code looks like.  I'm running `make` now, so maybe I'll see the generated code in the build folder in a minute or two?  I know that we can't CI test the swift code, but we could still have it generated into stable, I think?  It is a shame that we won't be able to run the case study examples, because of the IO problems, but it would still be nice to see the code.

I suppose it makes sense that IO is problematic, since swift seems to target iOS devices.  You don't really have control of the file system on an iPhone.  I guess we could generate an interface for iOS devices to type in the inputs, but that is beyond the scope of what we currently do, and not really the use case that we envisioned.

My local build finished and I don't see any swift code?  I just used `make`.  Should I be doing something else to generate the swift code?

If there is something that I can test on my Mac, let me know.

Really good work @bmaclach.  If you need to learn any other programming languages for work, we should add them to Drasil too.  :smile:

bmaclach:
@smiths You aren't doing anything wrong, I just didn't commit anything that would actually generate Swift code. The only real candidates I could add to the build are my two GOOL tests that don't involve file I/O (called HelloWorld and PatternTest). All of the case studies, and the 3rd GOOL test, involve file I/O. The problem with turning Swift code generation on for these two tests is that the way our build scripts work, it checks every folder in the build/GOOLTest/ directory and goes into each and runs make to build the generated code. So I would need to add some logic to make it skip over the swift folder in the build. That's probably worth it, though, if it then means we can generate the code and store it in stable?

Mornix:
I believe there are builds of Swift on Linux and Windows. I don’t know how much they support, but I think they should have enough functionality to support the code we generate. I can do some tests later to confirm that. 

I also have some ideas on how to handle selective compilation depending on available tools (regardless of OS).

bmaclach:
I did download a "Swift for Windows" app, though it only supports up to Swift 4 (current Swift version is 5), and when I tried to test some file I/O code locally I either got a message like "not yet supported" or I got no errors yet also no files written to my system.

Mornix:
I think I saw the package of Swift 4 you mentioned. I also found instructions on how to build Swift from source on Windows. Looks a little annoying though: https://github.com/apple/swift/blob/master/docs/WindowsBuild.md#MSVC

smiths:
@bmaclach, if it is possible, I would like to see the generated swift code in stable, even though the code won't work.  I assume that modules like the calculate module for GlassBR would produce correct code, since there is no I/O for this module.  If we generated the code that can be generated, we would at least be able to show comparisons for all of the languages for a subset of the code.

How crazy an idea would it be to change the input related functions in Swift to instead of reading from a file to just return our default values for all of the parameters?  That is, try to replace the file reading with hard-coding the values of everything into code.  I'm sure that is outside of the scope of what we are currently discussing, but it is an interesting binding time discussion.  We have our inputs bound at run time, but with a code generator, we could bind our inputs at compile time.  To get new inputs, someone would have to re-run the generator.

bmaclach:
@smiths I just pushed a small change so that the "unimplemented" functions don't throw errors during generation. That way we can still generate the code, but the file I/O statements will just be missing. 

There is still the problem of preventing Travis from attempting to compile the Swift code. I think @Mornix's aforementioned ideas on selective compilation will be helpful for that.

With regards to your suggestion of adding the option of using hardcoded inputs instead of reading them from a file, I have a few thoughts. First, note that this would not solve our Travis problem because even with working Swift code, I don't think we can install a Swift compiler in our Travis environment. Second, I want to clarify that I don't think synchronous file I/O in Swift is impossible, I just wasn't able to find a way to do it that I had enough confidence in without being able to test it locally. With those two points being said, the idea of hardcoding inputs is definitely do-able and wouldn't require a crazy amount of work. One thing we would have to consider is which input values to use. In the past, when generating a sample input file, we had considered using the "reasonable values" stored with some chunks in Drasil, but @carette pointed out that those values are reasonable independently, but may not combine to form a reasonable data set (citing the fallacy of the average). Our solution for the sample input file was to have the user provide a sample data set in a file. The user only needs to provide a path to that file if they have generation of a sample input file turned on. Thus, if we wanted to re-use that file for the potential hardcoding-inputs design variability, we would need to figure out a new place, higher up, for the user to specify a path to a sample input file (and potentially that would mean forcing users to provide the path even if we did not end up reading the file due to the specific design choices they make).

With regards to running Swift on Windows, a hopefully easier way might be to install the Windows Subsystem for Linux and install Swift there. Swift does seem much easier to install on Linux than Windows.

JacquesCarette:
Because I just merged #2076 this now needs a merge.  I'll do a review after that (but I did look already, I don't really expect any big issues).

bmaclach:
@JacquesCarette I've resolved the conflicts.

Mornix:
@bmaclach I've added (initial), wacky support to Travis to build Swift. It's v5.2.4 if you wanted to test/play around with it, pull in the `swift_travis` branch. The branch installs Swift on Travis and builds a `Hello World` program before building Drasil, which could be modified to experiment with v5 IO stuff. Plus any added swift code to stable should be able to build with `swiftc`.

bmaclach:
@Mornix Awesome! Thanks! This will definitely be useful.

bmaclach:
Taking another look at File I/O in Swift, I noticed that the deprecated methods I needed on the `FileHandle` type were replaced with new, similar methods that I could use instead. I've just pushed some changes that I hope will work for writing to files.

Unfortunately, when it comes to reading from files, every way of doing this that I can find is either:
- Reading the entire file contents at once, or
- Requires specifying the number of bytes to read.

In GOOL, we have methods for reading files word by word or line by line, so reading the files all at once will not do. And we won't know ahead of time the number of bytes to read for each word or line, so those won't work either.

Thus, I think our best option right now for reading files word by word or line by line will be to fake it. We can read in the entire file contents, then _process_ it word by word or line by line. This is consistent with most of the Stack Overflow answers I've seen related to reading files line by line in Swift. To do this, we would generate code to read in the file contents, then split it on word breaks and line breaks, resulting in a list of lists of strings (this is easy to do in Swift). Then, in GOOL, I'd store a pair of indices in the state, starting at (0,0), and every time a GOOL function for reading a word or line is used, the indices will be updated, to ensure we are always reading the correct elements from the file contents list.

bmaclach:
I'm doing further experimenting with file I/O and adding Swift to stable in a branch off of this branch, so that this one can be merged in (assuming the lack of Swift code in stable isn't blocking this PR, and can be added in the next PR).

JacquesCarette:
Thanks for confirming that this is ready to merge. And yes, I'm fine with merging it in now, and upgrading later.

Issue #2195, opened 2020-06-24
Where does information belong?


JacquesCarette:
This issue has been central to Drasil for a long time, but now there is an issue (#2123) that is bringing parts of the problem back to the fore.

In part, `SystemInformation` is a big hack. We never defined what a 'System' is, so it's very hard to know if some information belongs there or elsewhere.

For example: authorship. A single example can contain multiple authors. One person might have written the (original) code, another the SRS, and yet a third might have written the description of the 'system' as a whole. We need to give proper attribution (this is indeed very important to science), so we need to define the entities within Drasil that are things whose authorship is possible.

muhammadaliog3:
I will be dividing this comment into the followiing sections
1. Code Investigation -- this time I explore systemInformation backwards, hopefully to get a different perspective. You also don't have to read it to understand the conceptual sections, it is more of a place to store my analysis.
2. Conceptual Problems
3. Conceptual (and high level code solutions)

**Code Investigation**


- _sysinfodb -------- This is a map that contains almost all of a drasil example, it contains all symbols (except input and output), all concepts, all units, tracebility map, reference map, all of the data  definitions in a single map, instance/theory/general definition map, conceptInstance map, section map, and labelled content map (which is figures but usually graphs).  

- _usedInfodb ------- I could not figure out what this was. What we do know is that this is a chunkDB with at max 2 fields filled out (atleast in the examples), a termMap and/or ConceptMap.

- _refbyMap ----------- self explanetory

- _constraints, _constants, _inputs, _outputs, _datadefs -------- self explanatory, they contain all the instances

- _defSequence -------------- this is a list of the following
```haskell
data Block QDefinition = Coupled QDefinition QDefinition [QDefinition] | Parallel QDefinition [QDefinition]
```
It is empty in some examples (such as Projectile) and filled in other examples (such as SSP)
However it is not used within drasil-code or drasil-docLang. It could, therefore, be removed without issue. Or it could be something that is not implemented yet.

- _concepts ------------- is always empty in all the examples and is not used within drasil-code or drasil-docLang

- _definitions ------------- are empty list in some examples and in other examples are a list of QDefinitions from various places (theory+instance+genDefinition models, data definitions, etc.). In CodeSpec.hs they are used to get derived inputs incase there are to data definitions and then ultimately to create an execution order of code definitions.

- _quants --------------- all symbols that are not input and not output symbols

- _authors -------------- list of persons
```haskell
data Person = Person { _given :: String
[...]

Issue #2123, opened 2020-05-15
Generate README with code


bmaclach:
The README should indicate:
- What the generated program's dependencies are (which external libraries are used, including version numbers)
- Are there any OS's the program is not supported on? (For example, the C# Oslo library comes from Microsoft and, while it is probably possible to run it on different OS's, the documentation is all Windows-focused, so it makes sense to say other OS's aren't supported).
- How to run the program (`make run RUNARGS=<path to input file, without the angle brackets>`)

Based on discussion in #2114 comments.

bmaclach:
We could also potentially make this README an auxiliary file that the user can turn on or off, where the default choice would be to have README generation turned on.

smiths:
Thank you for creating this issue @bmaclach.  I agree that this would be a helpful addition to the generated artifacts.  @JacquesCarette, is this a task that would be appropriate for one of the summer students?

smiths:
It is implicit in the original issue by @bmaclach, but just to be complete, I think the README should also mention the assumed version of the programming language.  This is particularly relevant for Python.

JacquesCarette:
Yes, I think this could be given to one of the summer students.

smiths:
When we generate the README file we should include a list of data files that the code depends on.  For instance, for GlassBR, we should say the txt files that are expected to exist in the same folder as the executable code.  This is discussed in #2138.

bmaclach:
I thought of a couple things to add to the generated README:
- The README should include instructions to build/compile the program (`make build`)
- The part of the README that says how to run the program should only be generated if the chosen `ImplementationType` is `Program`, since when `Library` is chosen the generated Makefile won't have a `run` rule.

Something to note with this issue is that it doesn't need to be done all at once. It will probably be easiest to start by generating a very simple README with only one small piece of information (the assumed version of the programming language, for example). Then in future PRs more information can be added to the generated README, and it can be updaed to be a toggle-able auxiliary file that the user can turn on or off.

smiths:
Great advice.  Thank you @bmaclach!

muhammadaliog3:
@smiths I had a quick question, Dr Carette asked me to put the markdown printers in the drasil-printers package. Here was his original message:
"the 'right' way to do this would be to create an AST for markdown and a printer for markdown (in drasil-printers), and then use that. Then you can separate the meaningful content of the ReadMe generation from the markup/display aspects."

[...]

PR #2196, opened 2020-06-25
Readme 2123

muhammadaliog3:
Contributes to #2123 
The purpose section is incomplete, due to number #2195


code/stable/glassbr/src/cpp/README.md
```diff
@@ -1,7 +1,22 @@
-###Summary of Key Information for Running GlassBR on C++
-C++ Version: gcc 10.1
-
+# GlassBR 
+> Authors:  Nikitha Krithnan, W. Spencer Smith 
+ >  __purpose__
```
JacquesCarette, line 3: Is this right?  Should `__purpose__` be in the document like that?

muhammadaliog3: No the __purpose__ is not supposed to be there, it is a place holder because this issue is incomplete.

According to Dr Smith we should have the purpose in the Readme's, however the purpose is stored as a sentence. In order to convert sentences to a doc I need several more parameters such as ChunkDB, Stage and Linearity (there is a function in drasil-printers sentenceToDoc that needs these). However due to the discussion of whether or not "System Information" should exist, I held out on such as large change.

JacquesCarette: Thanks. So I guess the question (@smiths ) is whether to merge this in now anyways, or wait.

PR #2197, opened 2020-06-27
Finish Swift Renderer, add Swift code to stable

bmaclach:
This PR first implements the previously-unimplemented file-reading GOOL methods using the strategy discussed in https://github.com/JacquesCarette/Drasil/pull/2187#issuecomment-648365868.

I then added Swift code to stable and merged @Mornix's swift_travis branch so I could test the generated code through Travis. This exposed a long series of bugs in the Swift renderer, which I fixed. Here is a summary of the interesting bugs and their fixes:

- Like Java, Swift requires a `throws` declarations on functions/methods that can throw an exception. Unlike Java, Swift requires the `try` keyword to be used when calling a function/method that `throws`. This didn't require any infrastructure change because we already built a "method exception map" from the static code analysis pass. We simply re-use that map to generate the `throws` and `try` for Swift. Since all GOOL methods for generating calls are based on the low-level `call` method, only that single method needed to be changed to generate the required `try` declarations. `try` is not generated for every call, only for calls that can throw.
- Every state variable in a class in Swift needs an explicit initial value, unlike the other languages. Since no values are passed to `stateVar`, GOOL's method for generating state variables, I added a function that would return a default value based on type. There are likely better solutions for this, but require some thinking/discussion. I'll create an issue about it.
- Swift's arithmetic and comparison operators (+, -, *, /, <,  <=, >, >=, ==, !=) do not work if they are used on operands of different numeric types. Thus, I had to generate explicit casts through the Swift renderer in cases like this.
- All parameters of Swift functions/methods are immutable by default. The `inout` keyword is required to make a parameter mutable, so I updated `pointerParam` for Swift to add that keyword, since our generator uses `pointerParam` for any parameter that is mutated in the function body.
- Unlike other languages, in Swift when calling a function with `inout` parameters, the `&` operator must precede the arguments for the `inout` parameters. For this purpose, I added a new method to GOOL: `pointerArg`, and updated our generator to use `pointerArg` in calls in the same cases where it uses `pointerParam` when defining functions (thus the use of `pointerArg` should always line up with `pointerParam`)
- Multiple output functions in Swift are possible by returning a tuple. Since the return type of a Swift function must appear in the function definition, I added `multiType` as a GOOL method in an internal type class, to be used as the return type for multiple-output functions.
- This is not related to GOOL, but to our Makefile generator, which assumed any `doc` rule would call `doxygen`. Since Swift isn't supported by Doxygen, I added a new `DocConfig` type that parameterizes our Makefile generator, to allow each target language to configure the `doc` rule to its needs.

I am going to update the Statistics part of the description of #2187 to reflect these changes.


smiths:
@bmaclach, it is great to see the swift code in stable.  I really like that another language has been added.   In evaluating the Swift code, what should my expectations be?  I know that it originally didn't work, but you made some changes to get it to compile.  Should it compile now?   (I don't recall all of the details of the discussion.)

I tested stable on my Mac and unfortunately I get error messages.  I haven't done anything to investigate this, since this might be a known problem.  Using `make` with `Drasil/code/stable/projectile/Projectile_C_P_NoL_B_U_V_D/src/swift`, the error message is as follows:

```
swiftc main.swift InputParameters.swift Calculations.swift OutputFormat.swift -o Projectile
OutputFormat.swift:19:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data("s = ".utf8))
                            ^~~~~~~~~~~~~
                             
OutputFormat.swift:24:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data(s.utf8))
                            ^~~~~~~~~~~~~
                             
OutputFormat.swift:25:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data("\n".utf8))
                            ^~~~~~~~~~~~~
                             
OutputFormat.swift:30:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data("d_offset = ".utf8))
                            ^~~~~~~~~~~~~
                             
OutputFormat.swift:35:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data(String(d_offset).utf8))
                            ^~~~~~~~~~~~~
                             
OutputFormat.swift:36:29: error: extraneous argument label 'contentsOf:' in call
        try outputfile.write(contentsOf: Data("\n".utf8))
                            ^~~~~~~~~~~~~
                             
make: *** [Projectile] Error 1
```

bmaclach:
@smiths It should compile now, so this is unexpected. Since the Travis builds are passing, my best guess is that the errors are caused by you using a different version of Swift than Travis. Travis is on v5.2.4. What version do you have installed?

smiths:
I meant to say the version in my previous post.  It is `5.1.3`.  I'll update the version, and let you know what happens.  :smile:

smiths:
I'm still working on testing with swift 5.2.4.  I've had to upgrade my OS, and now I have to upgrade XTools.  My home Internet is usually fine, but it struggles with large files.  :-(

smiths:
I've now updated to Swift 5.2.4 and the examples compile with no problems!  :-)

Should I be able to run the examples, or are we settling for compilation at this time?  When I run GlassBR it is looking for the sampleInput.txt file, but it has the wrong path to that file.  

```
naghdi:swift smiths$ GlassBR 
GlassBR: ../../datafiles/GlassBR/sampleInput.txt: openFile: does not exist (No such file or directory)
```
Is that what we expect at this time?

bmaclach:
@smiths Good to hear they are compiling now. For the purposes of this PR, successful compilation was what I was aiming for. I was unable to test running them, so your attempt should provide useful information. I am confused by that error, though. The generated code should make no reference to sampleInput.txt (it is used during the build to construct the input file, but is not the input file we actually pass to the generated code). What commanded resulted in that output? `make run...` or something different?

smiths:
I wasn't sure which way to run the code, so I tried two different approaches:

1. 
```
naghdi:swift smiths$ make run RUNARGS=input.txt
./GlassBR input.txt
Fatal error: Error raised at top level: "Error opening file.": file /AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1103.8.25.8/swift/stdlib/public/core/ErrorType.swift, line 200
make: *** [run] Illegal instruction: 4
naghdi:swift smiths$ 
```
2. Just using `make` and then running `GlassBr`
```
GlassBR: ../../datafiles/GlassBR/sampleInput.txt: openFile: does not exist (No such file or directory)
```

The second one isn't right though, I should have typed `./GlassBR input.txt`, which produces the same output as Case 1.

To be honest, Case 2 is very weird.  I have `.` on my path, so it should have run the same as the other versions.  We don't need to worry about this though.  Maybe, you have some ideas about Case 1?

bmaclach:
@smiths Yes, that error makes more sense, though it is still unexpected. It means we hit the catch block in the try-catch statement for opening a file in write or append mode. In our programs, we open the log.txt file in append mode and output.txt in write mode. I'm unsure why this would be failing... Maybe because the files do not yet exist? If you create empty versions of those two files before running `make run`, does anything change?

smiths:
@bmaclach, unfortunately providing empty versions of the files did not help.  The error message is unchanged.  :-(  I would manually change the swift code to get it to not have an exception, but I don't really know how to change it.  :-(

bmaclach:
@smiths Hmm... one more thing to try, when you have time, if you don't mind, is to try running a Projectile example with no logging (any of them with NoL in the folder name). I expect you will still get an error, but whether it is the same error as you are getting with GlassBR will shed some light on where exactly the error is happening.

smiths:
Interestingly, the error message changes to a read error (instead of an open error).  I ran this example:

```
naghdi:swift smiths$ pwd
/Users/smiths/Repos/Drasil/code/stable/projectile/Projectile_C_P_NoL_B_U_V_D/src/swift
naghdi:swift smiths$ ./Projectile 
Fatal error: Error raised at top level: "Error reading from file.": file /AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1103.8.25.8/swift/stdlib/public/core/ErrorType.swift, line 200
Illegal instruction: 4
naghdi:swift smiths$ 
```

bmaclach:
@smiths This was one of the possibilities I had in mind. When logging is turned on, the first file I/O that is attempted is related to writing to the log file. When logging is turned off, then the first file I/O is when we read the input file. The fact that both cases resulted in an error means it's likely that the common code between these two cases is throwing the error. This common code is generated by the `swiftOpenFile` function in the Swift renderer, so that would be the first place to look to troubleshoot.

Would it be okay to create an issue for this problem, rather than fixing it as part of this PR?

smiths:
Yes, a separate issue makes sense.  This PR has accomplished what you set out to do.  I'll approve the PR, but @JacquesCarette will have to approve the Drasil code.

PR #2204, opened 2020-07-13
Class diagram

muhammadaliog3:
Muhammad's class diagram. I started these class diagrams in may, so things may have changed since then, but I finished the gool package today (probably the most important) so I am confident that it is accurate. I don't think that the other packages changed. The more drasil evolves the more these class diagrams will become obsolete. But I hope that my work can provide a "code start", such that future users will need to simply tweak the diagrams rather than start a latex document from scratch.


smiths:
The diagrams look great @muhammadaliog3.  We'll need @JacquesCarette to review when he gets back from his vacation, but this looks like a nice visualization of the design.  @szymczdm,  you might find the diagrams from @muhammadaliog3 useful when looking into what has changed with Drasil.

The summary diagram for GOOL looks a bit odd for the right hand side of the diagram.  Could you reorder the classes attached to RenderSym so that the arrows don't have as far to go?  For instance, it seems that moving MethodTypeSym to the top of the block will simplify the dotted arrow connecting it to TypeSym.

JacquesCarette:
These are the kinds of diagrams that I would like to have automatically generated -- these are, in some ways, based on the same kinds of data that @Nathaniel-Hu is collecting.  Figuring out how to not hard-code the visualization is tricky though.

JacquesCarette:
I would very much like @Nathaniel-Hu 's explicit comments on this.

Nathaniel-Hu:
@JacquesCarette by modifying [SourceCodeReader.hs](https://github.com/JacquesCarette/Drasil/blob/Issue%23703_GenDataTable/code/scripts/SourceCodeReader.hs) (i.e. add a new function based upon existing functions), the data for these class diagrams can be obtained automatically. The code for the nodes also looks like it can mostly be generated automatically. My only concerns would be how to generate the code for the arrows connecting the nodes, and how to position the nodes relative to each other (mirroring @JacquesCarette's previous comments below). 

> These are the kinds of diagrams that I would like to have automatically generated -- these are, in some ways, based on the same kinds of data that @Nathaniel-Hu is collecting. Figuring out how to not hard-code the visualization is tricky though.

JacquesCarette:
I'm thinking we should be using `dot` (for now) instead of latex for these diagrams. Also, we need to know what the arrows mean for sure, so that information can be generated too.  @muhammadaliog3 the information is class dependency, right?

muhammadaliog3:
Yes it is. For example is we have
```haskell
class (FileSym r) => ProgramSym r .....
```
Then ProgramSym will point to FileSym.

PR #2213, opened 2020-07-20
Design log mess

### LINKED ISSUE #2173: Add to the design log the set of Choices used
As of #2172, the generated design log documents any cases where a user's first choice cannot be used, but it does not document the actual choices used by the generator, meaning a user would have to decipher that information themselves by looking at their choices specification and the default choices.

We should expand the design log by having it document which choice was used for every design variability. Ideas for how to do this are given in: https://github.com/JacquesCarette/Drasil/pull/2172#issuecomment-641506158

In more detail, we should implement a pass on `Choices` that defines a design log message for every field in the structure except for any fields specified with a preferentially-ordered list of options. This portion of the design log should be passed to `generator` and included in the initialization of the `_designLog` field.

For the three choices represented with preferentially-ordered lists, the `chooseConcept`, `chooseSpace`, and `chooseODELib` functions already have the infrastructure for defining design log messages. They currently define messages in the cases where a conflict is discovered. We just need to update them to also define a message in the "successful" case where there is no conflict. The message should say which `UID` was matched to which `CodeConcept` (for `chooseConcept`), which `Space` was matched with which `CodeType` (for `chooseSpace`), and which ODE library was used (for `chooseODELib`).

muhammadaliog3:
Closes #2173 

Some of the important logs. 

First match ode log
[designLog.txt](https://github.com/JacquesCarette/Drasil/files/4948640/designLog.txt)

Not first match ode log
[designLog.txt](https://github.com/JacquesCarette/Drasil/files/4948643/designLog.txt)

Code Concept match
[designLog.txt](https://github.com/JacquesCarette/Drasil/files/4948667/designLog.txt)


code/stable/projectile/Projectile_U_P_NoL_U_WI_V_D/src/swift/designLog.txt
```diff
@@ -0,0 +1,20 @@
+Languages selected as Python Cpp CSharp Java Swift..
```
smiths, line 1: There should only be one period at the end of the sentence.

code/stable/projectile/Projectile_U_P_NoL_U_WI_V_D/src/swift/designLog.txt
```diff
@@ -0,0 +1,20 @@
...
+Modularity selected as Unmodular.
+Input Structure selected as Unbundled.
+Constant Structure selected as WithInputs.
+Constant Representation selected as Var.
+Implementation Type selected as Program.
+Software Constraint Behaviour selected as Warning.
+Physical Constraint Behaviour selected as Warning.
+Comments selected as CommentFunc CommentClass CommentMod..
```
smiths, line 9: There should only be one period at the end of the sentence.

code/stable/projectile/Projectile_U_P_NoL_U_WI_V_D/src/swift/designLog.txt
```diff
@@ -0,0 +1,20 @@
...
+Dox Verbosity selected as Quiet.
+Dates selected as Hide.
+Log File Name selected as log.txt.
+Logging selected as.
+Auxiliary Files selected as SampleInput ../../../datafiles/Projectile/sampleInput.txt ReadME..
```
smiths, line 14: There should only be one period at the end of the sentence.

code/stable/glassbr/src/cpp/designLog.txt
```diff
@@ -0,0 +1,24 @@
+Languages selected as Python Cpp CSharp Java Swift..
```
smiths, line 1: Shouldn't the language say cpp?  The other languages have their own folders, where the selected language would correspond to the folder.

code/drasil-code/Language/Drasil/Choices.hs
```diff
@@ -67,10 +68,20 @@ data Choices = Choices {
   auxFiles :: [AuxFile]
 }
 
+class RenderChoices a where
```
JacquesCarette, line 71: So, in theory, the class `RenderChoices` should not really exist. All of the 'data' store in instances of `RenderChoices` should instead be stored in chunks. However, I will not ask you to change that right now - but please make an issue for a future refactor.

JacquesCarette:
@smiths the merge is blocked because you have to acknowledge whether the changes that you asked for have indeed been done.

muhammadaliog3:
Actually we should wait till the end of this week. Although stable has the design logs, the make rule has yet to be updated to compare the generated design logs.  So until the changes are made to scripts we should not merge the request.

smiths:
I agree with @muhammadaliog3 that we should wait until everything is done.  I just looked at the design logs in stable and they still have the extra periods.

Nathaniel-Hu:
I've been looking at how I would need to modify the `Makefile` rules to include the `design logs` as files to be tested/compared with the files in the `stable` version. However, I've noticed that the `Makefile` rule used to compare generated files with their `stable` counterparts is instructed to ignore .txt files (of which the `design logs` are). Is there a way to modify this rule to not ignore the `design log` files somehow? @JacquesCarette @smiths @Mornix

The following is an example of the `Makefile` commands used to compare generated and `stable` files and push differences to the `log` files.
```
cd "build/GlassBR" && stack exec -- "glassbr" 
diff --strip-trailing-cr -r -X ../.gitignore -x '*.txt' "stable/glassbr/" "build/GlassBR"/ > "logs/GlassBR_log.log" 2>&1
```

smiths (approved):

smiths:
The changes I requested were made, so I unblocked my review.  However, @muhammadaliog3 mentioned that a make rule still had to be updated.  I don't know if that update was made.

JacquesCarette (approved):
Even though there are some small tweaks that could still be done, this is better merged in than waiting forever.

PR #2222, opened 2020-07-23
Exactly equal constraint 1795

### LINKED ISSUE #1795: Express "exactly equal" constraints
Currently in Drasil, constraints can either be expressed as a `Range`, `EnumeratedReal`, or `EnumeratedStr`.

Another type of constraint we need to express are constraints that one thing is exactly equal to another. For example, in SSP we have the constraint that the size of the lists x<sub>slope</sub> and y<sub>slope</sub> are equal (and similar for other lists that represent x and y coordinates for different parts of the system). 

This could potentially be expressed as a `Range` where the upper and lower bounds are the same, though that seems like a hack.

muhammadaliog3:
Closes #1795


smiths:
@muhammadaliog3, the original issue (#1795) that inspired this PR didn't use the example of the exact equality of real numbers.  It used the example of the exact equality of integers (actually natural numbers).  The example was that the length of two lists are the same.  I'm fine with covering the case of exact equality or floats for completeness, but this case is not going to come up very often.  As you implied above tests for exact equality is a bad idea for floats.  We really do try to avoid it in our requirements.  Tests for equality of integers on the other hand are natural and will likely come up all the time.

muhammadaliog3:
> This is actually a non-trivial design decision, as it has all sorts of down-the-road ramifications.
> 
> Because of my knowledge of lots of higher-level math and CS, it is my belief that the 'proper' way to encapsulate this concept is that you define 1) a singleton space, and 2) express "x exactly equal to y" as `IsIn x (Singleton y)`. Even though 'exactly equal to' is a symmetric relation, the knowledge encoding here is asymmetric, basically because `x` here is not restricted to be a variable, but we will want to (eventually) have restrictions on `y`.

There are several problems with this design. 

1. The payload of a singleton cannot be a space. As it does not make sense. For example (sy numberOfEggs == Integer) , does not make sense, but (syNumberOfEggs == (SOME EXPR)) makes sense. So the payload must be an expr. But Expr.hs imports Space.hs, so Spaces.hs cannot use Expr's (Or an import cycle is generated). Contradiction. Therefore singleton cannot be a space.

2. We cannot express the exactly equal to constraint as an IsIn expr. This is because IsIn is not implemented in Gool yet. I know this  because if I convExpr, has the following

```haskell
convExpr :: (OOProg r) => Expr -> GenState (SValue r)
convExpr IsIn{}    = error "convExpr: IsIn"
```

The solution is to implement the exactly Singleton as a copy of RealInterval, which is a kind of middle ground between space and expr. Second is to try and use a binaryOp to check for equality or implement a new Expr Eql as before.

JacquesCarette:
Luckily, I'm very forgiving of this type of thing. Yes, I did not explain myself well enough, as I used concepts you didn't know (like that of a singleton), so you misunderstood. But you made a very serious mistake: assuming that the professor was wrong when you didn't understand the explanation. You might even have made a bigger error, and assumed you understood the explanation and then arrogantly thought the prof was wrong.

Yes, there is an import cycle issue. That is indeed something worth thinking about. But the concept of a "space that consists of a single point of another space" is very well defined, and is called a singleton. It even exists in Haskell (in a library) under that very name.

And of course using `==` between an element and a space doesn't make sense. But I never said that. I carefully used `IsIn`.

The fact that `IsIn` is not implemented in GOOL does not make it a bad design. It does make it a longer-term design. Whether that justifies a short-term hack to make it work depends on how important the feature is. If it was blocking us on implementing something crucial, then a workaround would be worth it. If it's not, then it might be best to just delay this feature.  [Some features were even taken out of Drasil because they were hacks that prevented more important features from being implemented.]

muhammadaliog3:
I think I am beginning to understand @JacquesCarette . I will be interpreting expressions as specific instances of spaces. For example the number 1.0 is an expression, whose space is "real numbers". 1.0 is a specific instance of all real numbers.

The problem is that singletons are a space of one point, hence singletons are both a space and an expression. 

A similar problem, that is already solved, is the space of discrete things. For example the set [1,2,3,4] can be represented in drasil with a discrete space of integers ```DiscreteI [Int]```. We can interpret each of the points in the discrete space of integers as expressions, so "the discrete space of integers = discrete space of integer expressions". 

In order to solve the import issue I think that we should define Expr's as a Space, and remove space from the Expr datatype. 

I propose changing
```haskell
data Space =....
  | DiscreteI [Int]  --ex. let A = {1, 2, 4, 7}
  | DiscreteD [Double]
  | DiscreteS [String] --ex. let Meal = {"breakfast", "lunch", "dinner"}
```
To 
```haskell
data Space =....
  | Discrete [Expr]
  | Singleton Expr
```
This solution is basically saying that we should classify spaces into a point (singleton) or a set of points (discrete spaces), and what those exact points are, whatever they are strings or vectors or rational, they are equivalent.

And change  IsIn from 
```haskell
data Expr where
  IsIn     :: Expr -> Space -> Expr 
```
to 
```haskell
data Expr where
  IsIn     :: Expr -> [Expr] -> Expr 
```

JacquesCarette:
(Sorry for taking so long to get back to all of this)

You do seem to understand now.  The problem with your proposal is the one you pointed to earlier: import cycle. And this one is 'deep', because you are, on purpose, defining spaces and expressions as being mutually recursive. That's not necessarily bad (and is part of the design of dependently-typed languages), but not something that I really want to jump to right now.

There is a second problem: the current `Space` definition has this very inelegant split into `DiscreteI`, `DiscreteD` and `DiscreteS` which you 'unify'; unfortunately, what that loses is _types_. The inelegant solution was at least mildly typed, but yours allows on to create a `Space` that could be `[Dbl 1.0, Str "1.0", Int 1, C boo]`. Any list of `Expr` could do.  We want to move things in the opposite direction (more types, not fewer).  [You suggestion to changing `IsIn` is then perfectly consistent with your change to `Space`.]

So you suggestion is **too** flexible. We want a much more constrained set of singletons.

Also: it's not even clear if singleton is even the best solution to the original problem! The original problem was about certain things being 'exactly equal'.  So the change to `Language.Drasil.Constraint` is good. It's all the other changes that are not as good.  I think it would be worthwhile to isolate the introduction of `Equality` in `Language.Drasil.Constraint` from the rest. We should discuss the design of that change too, just to make sure. And then see about the rest of the changes.

PR #2223, opened 2020-07-24
Blank project created for Double Pendulum example

oluowoj:
Blank project for double Pendulum example


JacquesCarette:
I think one really important next step would be to include comments for each of the fields that are currently empty lists (in `Body.hs`) describing what information each ought to contain. That would be very good information in a template.

JacquesCarette:
Should I be deleting the branch, or will you continue to use it to build on? [I guess we usually use new branches for new work, but sometimes keep one branch if the work really is very incremental]

oluowoj:
> Should I be deleting the branch, or will you continue to use it to build on? [I guess we usually use new branches for new work, but sometimes keep one branch if the work really is very incremental]

Please go ahead and delete the branch, I will start off with a new branch

oluowoj:
> I think one really important next step would be to include comments for each of the fields that are currently empty lists (in `Body.hs`) describing what information each ought to contain. That would be very good information in a template.

I will start to look into it, some of these information will be better understood when I start adding more information to the example.

oluowoj:
> @oluowoj, it is a good experiment to add another example to Drasil so that we can see the steps and record them for the 741 participants. However, when you are writing up the instructions, I don't think everyone in the class should be doing pull requests on Drasil. Instead you should write up their instructions so that they do a fork on Drasil. If an example turns out to be really great, we could do a pull request.

Yeah, you are absolutely right, I will add that as a to do

Issue #2171, opened 2020-06-08
Contributors Guide - User evaluation test


oluowoj:
Hello @muhammadaliog3 , @Nathaniel-Hu 

I'm aware that @Nathaniel-Hu is already working on the Contributor's guide and I read some of the information in the thread. We had a discussion with @smiths and @JacquesCarette and wanted to provide more info about on other information is required in your test,  if not already included.

Task:
1) Write down in point form what you wish you knew before you started working on Drasil
2) @Nathaniel-Hu , write a test that would sort of help new Drasil users evaluate whether they have knowledge of the stuff you and @muhammadaliog3 listed in point 1. You both can use this issue to provide feedback on point 1.
Let me know if you have any queries.

muhammadaliog3:
Things I wish I knew
- When I first started Drasil I thought that to use git all I needed to know was branching and adding/commiting/pushin/pulling. However you actually need to know a bit more, things like what is the difference between a remote branch and local branch, what is origin, how to reverse commits, what is head, which branch are you pulling from when you "git pull". You need to be confident in using the command line.
- The original book that me and Nathaniel read was [learn you a haskell](http://learnyouahaskell.com/), you really do need to learn a fair bit conceptually. Sections like the IO section are perhaps not important, but having a strong understanding of the concepts (especially monads and lensing, and a little bit of states and zoom) is important or you will get confused when reading Drasil code. Practice is not as important as you can always use those resources as reference. This is because in the issues that I have done, it is not the implementation that takes time but rather figuring out the code structure and then what to code, so nowadays I just use those resources as reference.
- Don't be afraid to ask questions
- You rarely have to implement at a low level, chances are many of the things you want to implement have already been implemented (i.e. check existing modules and packages there might be similar functions).

smiths:
@oluowoj,  please check that the information shared by @muhammadaliog3 is captured in the current contributor's guide.  If it isn't, please add the information.  This is also somewhat relevant for your Drasil user's guide.

Once you have completed this task, please close this issue.

JacquesCarette:
Re-assigning once more: @Awurama-N can you check the status of all of this?  Also, if there is additional information that you wish you knew when you started, do add it!  I'm going to ping @Ant13731 too.

Ant13731:
The above tips helped a lot, but I just want to add one more thing that might be helpful (although #2404 should hopefully relieve this):
[...]

PR #2226, opened 2020-07-24
Create new doc folder + Add Contributor's Test & Answers

### LINKED ISSUE #2161: Contributor Test
Now that you have almost completed #2136, the next task is to write a test for future contributors.  @JacquesCarette and I discussed approaches for training future developers and the idea we came up with was to start with a test.  @Nathaniel-Hu, please write a short test, consisting of multiple choice and short answer questions.  We'll use this test for future developers.

The test will be easier to write, and easier for @JacquesCarette and I to judge, if you include the answers.  You can create a branch for the test.  When we are done, we'll probably delete the branch, so that the answers aren't available for future developers.  :smile:

I'll attach LaTeX files that you can use as a starting point.  The files are from a previous undergrad test.  I haven't done anything to change the test to suit our current needs; I just deleted most of the questions.  You can remove most of the preamble.  We just need a title for the test.  (We could do the test with other software, but practicing with LaTeX would be useful (and fun!) for you.)

You will have to rename the first file to have the extension tex, and the second file to have the extension cls.  GitHub doesn't support those extensions, so I relabelled both files to have the extension txt.  :smile:

[PotentialContributorTest.txt](https://github.com/JacquesCarette/Drasil/files/4724969/PotentialContributorTest.txt)

[examtst.txt](https://github.com/JacquesCarette/Drasil/files/4724974/examtst.txt)

Nathaniel-Hu:
- contains a new `doc` folder in the main Drasil folder, with 4 new files
   - new files include ContributorTestAnswers.tex, examtst.cls, ContributorTest.pdf and ContributorTestAnswers.pdf
   - main file of interest is ContributorTestAnswers.tex; Contributor's Test intended to gauge new contributor's understanding of git and other Drasil related topics
   - ContributorTest.pdf and ContributorTestAnswers.pdf are the rendered PDF files of the test + answers
   - examtst.cls is a support file for rendering the PDF files from ContributorTestAnswers.tex

- also includes changes to README.md file
   - Summary of Folder Structure and File Contents section updated to include new entry for `doc` folder

- closes #2161; related to Issue #2171


smiths:
@Nathaniel-Hu, why do you have the old test in the PR?  It seems to me that we can just have the last version of the test in the repo.  I think it will be confusing for people in the future that look at the test and wonder why there are two versions.

Nathaniel-Hu:
Ok then, I will remove the old versions. Do I need to include the PDF rendered version of the test too, or no?

smiths:
I suggest you include two pdf versions of the test - one with answers and one without.  You can generate the different versions from your test tex file.

JacquesCarette:
I tend to not put generated things in repos. The main exceptions are: 1) ease of use, 2) snapshot of an official submission.

For this case, I think ease of use warrants putting them in.

Nathaniel-Hu:
Alright then. I will create two pdf versions; one with answers, and one without, and include them in the Contributor's Test folder.

Nathaniel-Hu:
@smiths and @JacquesCarette, should the examtst.cls file also be included here, or should that be excluded?

smiths:
I think you should include the `examtst.cls` file.  For the same reason that @JacquesCarette mentioned that you usually don't put pdf files under version control, but in this particular case we will make an exception.  The "right" way to do this would be to have a folder of tex resources that we reuse, or a pointer to the maintained repo where the tex resources are located.  This just isn't worth the effort though.  We want future students to be able to compile the tex file, and they'll need `examtst.cls` to do that.

Nathaniel-Hu:
Alright then. Is this pull request cleared to be merged into master then?

PR #2244, opened 2020-09-16
Fixes #2241 (2d is corrected to 2D)

### LINKED ISSUE #2241: Posible Two-Dimensional abreviation typo in GamePhysics IM:transMot table.
@smiths 
The label on IM:transMot table (page 33 of GamePhysics_SRS.pdf) reads "Force on the translational motion of a set of 2d rigid bodies".
Should it say "2D" instead?

smiths:
Closes #2241.  There is a typo in the original text that uses 2d instead of 2D.  At some point we should probably introduce a symbolic constant for 2D, so that 2D is always used, instead of relying on the person coding in the example to use the proper case.  (I suppose if someone wanted to, they could always make an "error" like this by using a string, instead of using the symbolic constant.

### DIFF
--- a/code/drasil-example/Drasil/GamePhysics/IMods.hs
+++ b/code/drasil-example/Drasil/GamePhysics/IMods.hs
@@ -43,7 +43,7 @@ transMotRC :: RelationConcept
 transMotRC = makeRC "transMotRC" transMotNP EmptyS transMotRel
 
 transMotNP :: NP
-transMotNP = nounPhraseSP "Force on the translational motion of a set of 2d rigid bodies"
+transMotNP = nounPhraseSP "Force on the translational motion of a set of 2D rigid bodies"
 
 transMotRel :: Relation -- FIXME: add proper equation
 transMotRel = sy accI $= deriv (apply1 velI time) time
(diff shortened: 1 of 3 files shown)


oluowoj:
> Closes #2241. There is a typo in the original text that uses 2d instead of 2D. At some point we should probably introduce a symbolic constant for 2D, so that 2D is always used, instead of relying on the person coding in the example to use the proper case. (I suppose if someone wanted to, they could always make an "error" like this by using a string, instead of using the symbolic constant.

@smiths, we  actually have a constructor for this in here:
image

smiths:
Thanks for the note about twoD @oluowoj.  We really should replace all occurrences of the string "2D" by twoD.  The harder task will be to train people writing Drasil code to use the proper constructor.

JacquesCarette:
Oops, ignore my comment above, it's a duplicate of what was already on here.  Q: @smiths why did you assign this to me? Did you mean to assign me as reviewer? I don't know what assigning a PR to someone means!!

oluowoj:
> In theory, that whole sentence should be rewritten using the lower level combinators - 'force', 'translational motion', '2D' and 'rigid bodies' are all concepts known to Drasil. So this bug should not have existed at all, if things had been encoded "as intended".

ha okay... combinators

smiths:
@JacquesCarette, to be honest, I didn't give much thought to Assign versus Review.  I figured that everything goes through you for approval, so I assigned it to you.  Next time I will use Review instead.  :smile:

Can you approve the pull request?  We can then create a new issue to use the combinators correctly.

smiths:
@JacquesCarette, I neglected to assign this issue to you.  Can you please approve so that we can merge?  A better fix should be done in the future, but this pull request does fix a minor typo.

JacquesCarette (approved):
Finally catching up...

PR #2280, opened 2020-11-25
Pendulum example- Added GDs and IMs and derivations

oluowoj:
Please merge

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Physics.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Physics.hs
@@ -13,7 +13,7 @@ import Data.Drasil.Citations (dampingSource)
 import Data.Drasil.Concepts.Education (mechanics)
 
 physicCon :: [ConceptChunk]
-physicCon = [acceleration, angAccel, angDisp, angVelo, angular, chgInVelocity,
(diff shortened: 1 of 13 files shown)


oluowoj:
Added IM and updated derivations
Added frequency and period DDs
Added angular frequency and period GD

JacquesCarette (approved):

JacquesCarette:
The interesting thing about this PR:
- it looks like it is done exactly as intended (i.e. just like all the other examples, I had 0 comments about things that need fixing before merging)
- it is horrible (see above though).

In other words, the Drasil infrastructure still has a lot of room for improvement if this is how things are supposed to be done.  Our means of 'knowledge encoding', 'recipes', etc, really are quite primitive. Sigh.

smiths:
I guess the silver lining is that this process of adding a new example to Drasil has highlighted where there is room for improvement.  The 741 students are presenting their Drasil examples in class later today.  Olu is also attending the class.  I'm sure we'll be able to get some additional feedback from the students.  I do have a worry that Olu's study on softifact generation isn't going to get as positive feedback on softifact generation because of the challenges of learning Drasil.  It is all feedback though, and we'll learn and improve.  :smile:

PR #2289, opened 2020-12-24
New example: PD Controller

muralidn:
Added a new example that demonstrates a PD Controller with a second-order system. Specifically, the example demonstrates solving a second-order differential equation.


muralidn:
A second-order ODE may be demonstrated in the DblPendulum example. This example is redundant and therefore closing the Pull Request.

JacquesCarette:
Actually, 'redundancy' of this kind is very good. I think @smiths and I would much prefer multiple demonstrations of the same idea. Having an example of a PD Controller is quite superior to not having any.  So are you ok if I reopen this PR?

muralidn:
Yes, I am ok with reopening this PR Dr. @JacquesCarette. Thank You!

JacquesCarette (approved):

Issue #2287, opened 2020-12-21
Switch from Travis to GitHub Actions


balacij:
As per Dr. Carette:

> Nice. Though we really ought to do a bigger move: to github actions. Travis is throttling its free accounts a lot.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/issues/2286#issuecomment-749144058_

balacij:
Related ticket: #2231

balacij:
For future reference, https://github.com/haskell/actions might be helpful for building this out.

balacij:
If it's alright, I can give this a go @JacquesCarette ?

JacquesCarette:
Yes please.

balacij:
For the rest of this ticket, we need to create a new Drasil workflow script that activates when a PR is accepted into the `master` branch. When it's activated, it will update the `gh-pages` branch to update the Drasil website.

balacij:
With PR #2461, we'll have deploying to `gh-pages`, but I'll still have left the the CI labels and `pr_label_update.py` script out of use. Looking at a few PRs, it seems like it adds `CI: *` labels to PRs. Do we still want to do this @JacquesCarette? With GitHub Actions, they don't appear directly on the PRs but they appear on PRs if you click on the "Details" of the various checks.

JacquesCarette:
Can you give me more details? I'm not really understanding that last comment of yours (the whole thing), making it hard to answer the question.

balacij:
Ah, yes.

I think the `pr_label_update.py` script attaches these
![image](https://user-images.githubusercontent.com/1627302/118319777-3ecd4c00-b4c9-11eb-8436-6efdc9ad9b9c.png)
on PRs.

I'm not quite sure I fully understand what the labels are for, nor why the labels only appear on some and not others. If they're only showing what steps did/didn't pass, then I think we can access it alternatively through the action run notes that appear on PRs
![image](https://user-images.githubusercontent.com/1627302/118320141-cf0b9100-b4c9-11eb-8263-aced1247868a.png)

JacquesCarette:
Indeed, that's what that script does. And yes, the labels only get attached when the build fails for those reasons, i.e. show what steps did not pass.  `drasil-bot` then removes those labels when the build passes those parts again. It's a fun feature, but not crucial.

balacij:
[...]

PR #2297, opened 2021-02-09
2295 : Remove unused imports

balacij:


JacquesCarette:
We're going to need to go over these one-by-one. I think some of them might be new warnings because things moved from a package to `Prelude` (like `elem` from `Data.List`), while others (like `dep`) has a 'FIXME' in `Language.Drasil` saying that it shouldn't be exported from there.

In other words, it would be better to redo this PR as several PRs, one for each 'reason' new warnings get triggered. The flip side, which we need to worry about, is that Drasil worked just fine in GHC 8.4 and 8.6, so it would be a shame to needlessly break those.  So we should put this on hold until we have #2287 done.

balacij:
Sounds good. We can follow up on this individually later as suggested. Sorry for the troubles!

PR #2325, opened 2021-04-30
2322 update website stack

balacij:


balacij:
@JacquesCarette Should we note that we based the code off MissingH's here? I really just rewrote their code (though the `endswith` was just an alias to Data.List's isSuffixOf, and the strip, lstrip, and rstrip are rather obvious implementations).

JacquesCarette:
May as well give credit where credit is due for the strip functions, even though they are obvious.

JacquesCarette (approved):

JacquesCarette (approved):

PR #2328, opened 2021-04-30
2321 : remove package minimums and parsec from dependencies from drasil-code

balacij:


code/drasil-code/drasil-code.cabal
```diff
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: e1ef900d66ab33dbfd54c8f4473845894f308f81b5aed1cf74b235a8feb0f420
+-- hash: a9a50a45f3a9e998f60776d2574553d0a145ad3e40e1f2062e6fec85b9bbc3ee
```
JacquesCarette, line 7: Does the hash need to keep changing? We did use to store hashes, but in a separate file that wasn't under revision control.

JacquesCarette (approved):

code/drasil-code/drasil-code.cabal
balacij: I believe so. The cabal files are regenerated each time the package.yaml file is altered, so I think we would want to commit them too. I think the cabal files in a stack project are just thought of as the intermediary between Stack, hpack/`package.yaml`, and Cabal.

JacquesCarette: Ok.

Issue #2323, opened 2021-04-30
Cleaning up DblPendulum/Concepts.hs


JacquesCarette:
See Projectile/Concepts.hs a 'template'.

1. remove all the commented out stuff
2. move `pendulumTitle` to Body.hs (see https://github.com/JacquesCarette/Drasil/commit/ade637d243bc4bc6641040f1b97684515ce5f31d)

Keep the file around, because there will be some stuff that will need to be moved to here from other places.

PR #2345, opened 2021-05-03
Fixed Issue #2323

Ant13731:
Closes issue #2323. Cleared comments and moved pendulumTitile function.


code/drasil-example/Drasil/DblPendulum/Body.hs
```diff
@@ -29,7 +29,7 @@ import Drasil.DblPendulum.Figures (figMotion)
 import Data.Drasil.Concepts.Math (mathcon, cartesian)
 import Data.Drasil.Quantities.Math (unitVect, unitVectj)
 import Drasil.DblPendulum.Assumptions (assumptions)
-import Drasil.DblPendulum.Concepts (pendulumTitle)
+import Drasil.DblPendulum.Concepts ()
```
JacquesCarette, line 32: Since no imports are needed from this file, this line should be entirely deleted.

JacquesCarette (changes requested):
`Assumptions` seems to import `pendulumTitle`, that should be investigated as to why and dealt with accordingly.

Did do a local 'make' that gave no warnings?  The build is failing, so I suspect not?

code/drasil-example/Drasil/DblPendulum/Body.hs
Ant13731: Ah okay, sorry about that. Next time I'll make sure the local make works first. I'll also remove the unneeded import line.

Ant13731:
Would I be able to move pendulumTitle to Assumptions.hs? I can't import pendulumTitle from Body.hs because Body.hs depends on Assumptions.hs.

smiths:
@Ant13731, my advice is to look at how projectile does it and following that example.

JacquesCarette:
So pendulum doesn't use `pendulumTitle` outside of body.

What we need to figure out is what `Assumption` really needs. The thing is, there are two concepts, that of a pendulum, and that of a document that describes a pendulum -- which might end up displaying the same thing "in English". So there might be a need for a `pendulum` concept (to go in to `Concept.hs`) that `Assumption` can use, and another one which signifies the title, that could end up being defined locally in `Body`.

This is hard/tricky, as the point is that there might be two different ideas that end up being said in the same way when rendered in a document. Yet, for Drasil, different ideas/concepts, need to be encoded differently.

Ant13731:
I'm not sure if I need to submit another PR as the guide stated that there shouldn't be any additional commits added to an open PR, but github automatically added my commits to this one.

JacquesCarette:
No, you shouldn't submit a new PR. It is fine that you add commits to an open PR as a means of fixing things upon feedback. This is the trigger for me to re-review.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -4,7 +4,8 @@ import Language.Drasil
 
 import Data.Drasil.Domains (physics)
 
-
+pendulum :: CI
+pendulum = commonIdeaWithDict "pendulumTitle" (pn "Pendulum") "Pendulum" [physics]
```
JacquesCarette, line 8: As this is now the concept a pendulum, the id "pendulumTitle" is not really apt anymore.  Can you make it just "pendulum"?

code/drasil-example/Drasil/DblPendulum/Main.hs
```diff
@@ -4,7 +4,8 @@ import GHC.IO.Encoding
 
 import Language.Drasil.Generate (gen, DocSpec(DocSpec), DocType(SRS, Website))
 
-import Drasil.DblPendulum.Body (srs, printSetting, pendulumTitle)
+import Drasil.DblPendulum.Body (srs, printSetting)
+
```
JacquesCarette, line 8: Try to add extra blank lines.

JacquesCarette (changes requested):

Ant13731:
Sorry. Should I remove the line in between the imports (and the one I added as well)?

Ant13731:
Also should I update the stable for dblpendulum? I didn't have to before but now it is generating an error.

JacquesCarette:
Yes to the two blank lines.  And yes to updating stable - it was wrong, as it hard-coded a capital letter when this was no appropriate. Guess no one caught that!

code/drasil-example/Drasil/DblPendulum/Main.hs
JacquesCarette: Oops, I see that I skipped the key word 'not' here!!! My apologies.

JacquesCarette (approved):

Issue #2341, opened 2021-05-03
Remove DblPendulum/TMods


JacquesCarette:
The model that is here should be in `Data.Drasil.Theories.Physics` instead. As a first step, what will be needed is that the 'note' attached to the model will need to be made generic - right now it is too specific to the double pendulum problem.

So this will need to be solved in two steps:
- [ ] change the note to be generic, and allow the new expected solution to be generic
- [ ] move the theory model to `drasil-theories`; move the definition `tMods = [accelerationTM, velocityTM, newtonSL, newtonSLR]` to (probably) DblPendulum.Body.

JacquesCarette:
Also @smiths mentioned "We can use the game physics example as given in Game Physics in the pendulum example.  We don’t need a different TM for pendulum."

PR #2348, opened 2021-05-04
Fix for Issue #2341

Ant13731:
closes issue #2341 
Removed TM from `DblPendulum/TMods.hs` into `drasil-theories`. Then modified other files' imports and `package.yaml` to work with the change.


Ant13731:
The second commit here is just resolving conflicts between the master branch and the generalSndRotLaw. The commit itself didn't mention it so I just put the message here instead.

code/drasil-data/Data/Drasil/Theories/Physics.hs
```diff
@@ -105,6 +106,27 @@ torqueDesc = foldlSent [S "The", phrase torque,
...
+newtonSLRRel :: Relation
+newtonSLRRel = sy QP.torque $= sy QP.momentOfInertia * sy QP.angularAccel
+
+newtonSLRNotes :: [Sentence]
+newtonSLRNotes = map foldlSent [
+  [S "The net", getTandS QP.torque, S "on a", phrase rigidBody `sIs`
+   S "proportional to its", getTandS QP.angularAccel `sC` S "where",
+   ch QP.momentOfInertia, S "denotes", phrase QP.momentOfInertia `ofThe`
+   phrase rigidBody, S "as the", phrase constant `sOf` S "proportionality"],
+  [S "We also assume that all", plural rigidBody, S "involved" `sAre`
```
JacquesCarette, line 127: Can you double-check this with @smiths ? I think that this 2D assumption was correct in the context of Pendulum, but that it is actually not embedded in this law as stated here, i.e. the law here is general.

JacquesCarette (changes requested):

code/drasil-data/Data/Drasil/Theories/Physics.hs
Ant13731: Okay I see what you mean. Should I take the 2D bit out? I think it still makes sense in the context of the pendulum without that specification. Otherwise, is there a way to just mention that more specifically in `DblPendulum/Body.hs` by appending that sentence to a generalized version of `newtonSLR`?

JacquesCarette: In some ways, the 2D note is redundant information, as it is already in the assumptions. And it will probably be in the instance model too. So since the theory model is supposed to be general, it makes sense to not have it at all.

If it should still be there (and hopefully @smiths will tell us), then the thing to do is to create a new theory model based on the generic one [using lenses to set the id field and the note field to be different].

Ant13731: Okay, sounds good.

smiths: Yes, we can take the 2D part out.  The equation looks exactly the same in 3D, so it is best if we keep the theory as general (and thus reusable) as possible.

JacquesCarette (approved):

PR #2354, opened 2021-05-05
Move "Bool Ops" out of BinOp, into BoolBinOp

balacij:
( Leaving as a draft until #2352 is merged in. )


code/drasil-printers/Language/Drasil/Printing/Import.hs
```diff
@@ -90,76 +92,80 @@ processExpo a
...
+expr (Str s)                  _ = P.Str s
+expr (Perc a b)              sm = P.Row [expr (Dbl val) sm, P.MO P.Perc]
   where
     val = fromIntegral a / (10 ** fromIntegral (b - 2))
-expr (AssocB And l)       sm = P.Row $ intersperse (P.MO P.And) $ map (expr' sm (precB And)) l
-expr (AssocB Or l)        sm = P.Row $ intersperse (P.MO P.Or ) $ map (expr' sm (precB Or)) l
-expr (AssocA Add l)       sm = P.Row $ intersperse (P.MO P.Add) $ map (expr' sm (precA Add)) l
-expr (AssocA Mul l)       sm = P.Row $ mulExpr l sm
-expr (Deriv Part a b)     sm =
+expr (AssocB And l)          sm =
```
JacquesCarette, line 105: In general, if you can keep these on a single line, that would be nice. The lines can go slightly over 80 characters if needed.

Another way would be to make a helper function for the pattern 
```haskell
P.Row $ intersperse (P.MO x) $ map (expr' sm (precB y)) l
```
used 3 times in the above.

JacquesCarette (approved):

code/drasil-printers/Language/Drasil/Printing/Import.hs
balacij: Thanks! I made a helper function along these lines to resolve this.

JacquesCarette (approved):

Issue #2347, opened 2021-05-04
Extra "The" in the DBLpendulum documentation


Ant13731:
In Assumptions.hs, the function:
```haskell
yAxisDirDesc :: Sentence
yAxisDirDesc = S "The direction" `ofThe'` phrase yAxis `sIs` S "directed opposite to" +:+. phrase gravity
```
is somehow producing an extra 'The' on the html and .tex files for dblpendulum. I'm not sure where the extra 'The' is coming from since the other functions in `DblPendulum/Assumptions.hs` follow the same format and appear properly.
![CodeFix1](https://user-images.githubusercontent.com/69334555/117029973-2d1bc580-accd-11eb-8465-208f20f0ade8.png)

smiths:
@Ant13731 you should assign your issues to someone, and you should label your issues when appropriate.  For questions related to the code, the most likely people to assign to are @JacquesCarette or @balacij.  If someone disagrees with the assignment, they can change it.  Another option is to self assign the issue.  This might be an issue that you can investigate yourself?

Ant13731:
Okay, sweet. I think I have an idea of how to fix it, so I'll give it a try.

PR #2360, opened 2021-05-05
Fixing duplicate occurrences of "the" in DblPendulum (Closes Issue #2347)

Ant13731:
closes issue #2347 
Edited some of the grammar and removed duplicate occurrences of "the".


code/drasil-example/Drasil/DblPendulum/Assumptions.hs
```diff
@@ -30,7 +30,7 @@ cartCoordRightDesc = S "The" +:+ phrase cartesian `sIs` S "right-handed where po
                          phrase xAxis `sAnd` phrase yAxis +:+ S "point right up"
 
 yAxisDirDesc :: Sentence
-yAxisDirDesc = S "The direction" `ofThe'` phrase yAxis `sIs` S "directed opposite to" +:+. phrase gravity
+yAxisDirDesc = S "direction" `ofThe'` phrase yAxis `sIs` S "directed opposite to" +:+. phrase gravity
```
JacquesCarette, line 33: This is very odd. The same kind of sentence in the `Projectile` example has the `The` in the sentence. So I wonder what is different here, i.e. where / why does it get added uniformly? Can you please track that down?

code/drasil-example/Drasil/DblPendulum/Body.hs
```diff
@@ -88,10 +88,10 @@ mkSRS = [RefSec $      --This creates the Reference section of the SRS
 
 justification :: Sentence
 justification = foldlSent [S "A", phrase pendulum, S "consists" `sOf` S "mass", 
-                            S "attached to the end of a rod" `sC` S "its moving curve" `sIs`
-                            S "highly sensitive to initial conditions" +:+ S "Therefore" `sC`
+                            S "attached to the end of a rod" `andIts` S "moving curve" `sIs`
```
JacquesCarette, line 91: improving grammar is a good thing. But I would actually prefer 2 PRs, one that focuses just on fixing the issue, and then another that improves such things. If some of the changes are controversial, then the good changes will still go in quickly, and only a few things can drag on for a while longer. Faster overall progress. 
Don't change anything for this PR, just for future ones.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -165,7 +165,7 @@ \subsection{Abbreviations and Acronyms}
 \end{longtable}
 \section{Introduction}
 \label{Sec:Intro}
-A pendulum consists of mass attached to the end of a rod, its moving curve is highly sensitive to initial conditions Therefore, it is useful to have a program the to simulate the motion of the pendulum to exhibit the chaotic characteristics of it The program documented here is called pendulum.
+A pendulum consists of mass attached to the end of a rod and its moving curve is highly sensitive to initial conditions. Therefore, it is useful to have a program the to simulate the motion of the pendulum to exhibit its chaotic characteristics. The program documented here is called pendulum.
```
JacquesCarette, line 168: This sentence now has a `the to` in it.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -179,7 +179,7 @@ \section{Specific System Description}
 
 \subsection{Problem Description}
 \label{Sec:ProbDesc}
-A system is needed to is needed to efficiently and correctly to predict the motion pendulum.
+A system is needed to is needed to efficiently and correctly to predict the motion of a pendulum.
```
JacquesCarette, line 182: `is needed to` is (still!) repeated twice. The `of a` was a good fix...

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -206,10 +206,10 @@ \subsubsection{Physical System Description}
 \end{figure}
 \subsubsection{Goal Statements}
 \label{Sec:GoalStmt}
-Given the the mass length of the rod, initial angle of the mass and the gravitational constant, the goal statements are:
+Given the mass and length of the rod, initial angle of the mass and the gravitational constant, the goal statements are:
 
 \begin{itemize}
-\item[Motion-of-the-mass:\phantomsection\label{motionMass}]{the Calculate the motion of the mass}
+\item[Motion-of-the-mass:\phantomsection\label{motionMass}]{the Calculate motion of the mass}
```
JacquesCarette, line 212: `the Calculuate`? This is definitely quite broken - something deeper is going on.  I'm guessing there's a hard-coded `the` being added to everything in an incorrect way. That's the likely source of the real problem.

JacquesCarette (changes requested):

code/drasil-example/Drasil/DblPendulum/Assumptions.hs
Ant13731: I think it might be because of the `ofThe` function since it adds an extra "the" before joining its arguments with the words "of the".
```haskell
ofThe  a b = S "the" +:+ sentHelper "of the" a b
```
Most of the projectile sentences don't use `ofThe` at the beginning except for this one:
```haskell
yAxisGravityDesc :: Sentence
yAxisGravityDesc = S "direction" `ofThe'` phrase yAxis `sIs` S "directed opposite to" +:+. phrase gravity
```
which also does not include a literal "the" at the beginning.

code/drasil-example/Drasil/DblPendulum/Body.hs
Ant13731: Ah okay, sounds good.

code/drasil-example/Drasil/DblPendulum/Assumptions.hs
JacquesCarette: Oh, that's weird. And not the expected behaviour. I would changed `ofThe` to stop doing that, and create a differently named combinator for that.  So the correct fix would be 
1. to create that new combinator that does what `ofThe` currently does
2. change the current correct uses of `ofThe` to use that new one
3. have `ofThe` do what's expected
4. use `ofThe` appropriately
5. put back those initial "the" (or simply revert this PR and restart).

Note that we want to minimize explicit capitalization of words, and let the framework deal with it. Take a further look at Projectile as an example.

Ant13731: I agree many of the use cases for `ofThe` currently make sense, it's just some few sentences that read weirdly because of the naming of the combinator. There's also similar functions in `Utils.Drasil.Sentence` such as `ofThe'` (capital variant),`ofGiv`, `ofGiv'`, and `isExpctToHv`. So would making those combinators into something slightly different such as `theOfThe` and `theOfGiv` work well? After that, then should I away the extra "the" from `ofThe` and manually change the ones that will make more sense?

With that being said, is there an easy way to rename a function in the whole repo? I've used VSCode to rename all functions in an individual file before, but I was wondering if there was a better way to go about it for a larger project like this. I think it might be easier to merge the PR and then I'll search and replace all `ofThe` badly acting functions on a new PR.

Also, for minimizing explicit capitalization, I think that `ofThe'` currently handles that better than putting a literal "The" at the beginning of the sentence. The projectile definition still makes enough sense without the explicit "The" at the beginning. Changing this might make it less readable if we don't include the explicit "The" when using the 'new' definition of `ofThe`.

JacquesCarette: Perhaps `the-ofThe` might be even better, if `-` is allowed in a name. `the_ofThe` otherwise. I forget the exact lexical conventions of Haskell.  And then yes, get right of the extra "the" in `ofThe` and adjust accordingly.

Renaming all functions in a repo is a function of the editor / IDE you use. The one I use wouldn't help me... but maybe @balacij uses something cleverer?  There is a Haskell plugin for VSCode.

JacquesCarette:
Just to be clear: you think it would be best to merge this PR in as-is, and you'll make further changes in subsequent ones?

code/drasil-example/Drasil/DblPendulum/Assumptions.hs
Ant13731: Okay, good to know. Thanks!

Ant13731:
Yes, I think that would be the easiest as some of the changes are still proper even with the renamed functions.

JacquesCarette (approved):

PR #2363, opened 2021-05-05
Removed explicit imports of applicative and monad. Closes Issue #2355

### LINKED ISSUE #2355: Explicit imports of Applicative, <$>, & Monad can be removed
`Applicative`, <$>, & `Monad` have been in `base/Prelude` for all versions of GHC 8.* so I think we can comfortably remove their explicit imports without causing any incompatibility issues for 8.*+.

[As early as GHC 8.0.1's `base/Prelude` had Applicative and Monad](https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Applicative)
[Directory of GHC versions to base versions](https://wiki.haskell.org/Base_package)

Ant13731:
Closes #2355 
Removed some imports of Control.Applicative as they are included in the Prelude now.


JacquesCarette (approved):
My favourite kind of PR: only code removed!

Issue #2343, opened 2021-05-03
line up CIs in PDController/Concepts


JacquesCarette:
https://github.com/JacquesCarette/Drasil/blob/236594a4a00940b0168a7699c75d3a3465fd0ad4/code/drasil-example/Drasil/PDController/Concepts.hs#L19-L32 should be "lined up" - see https://github.com/JacquesCarette/Drasil/blob/236594a4a00940b0168a7699c75d3a3465fd0ad4/code/drasil-example/Drasil/Projectile/Concepts.hs#L18-L22 for a good template.

Awurama-N:
```
pidControllerSystem = commonIdeaWithDict "pdControllerApp" (pn "PD Controller") "PD Controller" []
pdControllerCI = commonIdeaWithDict "pdControllerCI" (pn "Proportional Derivative") "PD" []
proportionalCI = commonIdeaWithDict "proportionalCI" (pn "Proportional") "P" []
derivativeCI = commonIdeaWithDict "derivativeCI" (pn "Derivative") "D" []
integralCI = commonIdeaWithDict "integralCI" (pn "Integral") "I" []
pidCI = commonIdeaWithDict "pidCI" (pn "Proportional Integral Derivative") "PID" []
```

PR #2368, opened 2021-05-06
lining up Cls in PDcontroller  (Closes Issue #2343)

Awurama-N:
closes issue #2343
lined up the code in PDcontroller. what was previously lines 19-32.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "editor.renderWhitespace": "all",
-    "diffEditor.ignoreTrimWhitespace": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "editor.renderWhitespace": "none",
+    "diffEditor.ignoreTrimWhitespace": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,4 @@
 {
-    "editor.renderWhitespace": "none",
+    "editor.renderWhitespace": "all",
     "diffEditor.ignoreTrimWhitespace": false
 }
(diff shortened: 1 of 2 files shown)


JacquesCarette (changes requested):
By "lined up", I meant that all the `=` signs should be in the same column, and that all the `(pn`... arguments too, then the strings, then the last `[]` argument.  You'll need to use spaces to do that, like the example that I linked to.

The code should end up looking like a table.

Awurama-N:
Okay thank you, i moved over the = and the (pn but i wasn't sure about the strings & bracket, because i thought it'd get too long. I'll add that now

Awurama-N:
This is what the code looks like to me now, i tried to create a commit to push but it says no changes were made.

Screenshot 2021-05-05 at 8 34 03 PM

Awurama-N:
Screenshot 2021-05-05 at 8 36 45 PM

JacquesCarette:
Right, that code looks perfect. I'll investigate a bit.

Awurama-N:
okay thank you

JacquesCarette:
So the file that is in the repo, in your branch, is not lined up. My guess is that your editor is "too smart" and removes all those spaces before saving!  Whatever editor you're using, you'll need to investigate your settings, and turn off this option.

Awurama-N:
Oh okay okay, I will sort that out

smiths:
@Awurama-N, are you sure that you changed the right branch?  It looks like you checked out the branch, but didn't make any changes to it.   Maybe you edited another branch, maybe master?, by mistake?

Awurama-N:
hmm, how can i be sure i'm working in the right branch?
 is there code i should input after `git checkout`?

JacquesCarette:
You did commit some code to branch `issueSolution`, and pushed it, as I was able to pull it down. It was changed, in that the extra returns and blank lines were gone.  So you did those edits in the right branch, in the right way.

What seems to have disappeared are the extra spaces on those lines.

Awurama-N:
Oh okay then it is probably the settings of the editor ignoring the spaces. I'm going to try to turn that off.

JacquesCarette:
You should not be checking in your private vscode settings into the repo itself. Only the changes to the Haskell files please.

JacquesCarette:
Yes, that worked!  Now you just have to remove `settings.json` from this commit, and it will be good to go.

JacquesCarette (approved):

PR #2372, opened 2021-05-06
Modified ofThe function (addresses follow-up for issue #2360)

Ant13731:
Changed all occurences of `ofThe` to `the_ofThe` for easier readability. Added in new `ofThe` function to work as expected.
closes issue #2360

### DIFF
--- a/code/.hlint.yaml
+++ b/code/.hlint.yaml
@@ -54,7 +54,7 @@
 # Ignore some builtin hints
 # - ignore: {name: Use let}
 # - ignore: {name: Use const, within: SpecialModule} # Only within certain modules
-  - ignore: {name: "Use camelCase"}
+- ignore: {name: "Use camelCase"}
 
 
 # Define some custom infix operators
--- a/code/.hlint.yaml
+++ b/code/.hlint.yaml
@@ -54,6 +54,7 @@
 # Ignore some builtin hints
 # - ignore: {name: Use let}
 # - ignore: {name: Use const, within: SpecialModule} # Only within certain modules
+  - ignore: {name: "Use camelCase"}
 
 
 # Define some custom infix operators
--- a/code/drasil-data/Data/Drasil/Concepts/Physics.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Physics.hs
@@ -118,7 +118,7 @@ pendulum = dccWDS "pendulum" (cn "pendulum")
 position = dcc "position" (cn' "position")
   "an object's location relative to a reference point"
 positionVec = dccWDS " positionVec" (cn' "position vector")
-   (S "a vector from the origin" `the_ofThe` phrase cartesian +:+ S "defined"
+   (S "a vector from the origin" `ofThe` phrase cartesian +:+ S "defined"
     `toThe` phrase point +:+ S "where the" +:+ phrase force +:+ S "is applied")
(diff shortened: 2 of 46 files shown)


Ant13731:
Is the underscore the reason for the error here? Should I just use pure camel case?

smiths:
Yes, @Ant13731, HLint failed on the camelCase test.  It seems that we could change `the_ofThe` to `theOfThe`.  I'm sure it is also possible to override the camelCase warning, but you would want permission from @JacquesCarette before going down that road.

JacquesCarette:
If there's a way to make `the_ofThe` an exception to the camel case warning, that would be the best route. I like the underscore in that name. Otherwise, I think we'll turn off that warning.

smiths:
There might be a hint on how to turn off the warning here:

https://github.com/haskell/haskell-ide-engine/issues/955

code/drasil-data/Data/Drasil/Concepts/Physics.hs
```diff
@@ -118,10 +118,10 @@ pendulum = dccWDS "pendulum" (cn "pendulum")
 position = dcc "position" (cn' "position")
   "an object's location relative to a reference point"
 positionVec = dccWDS " positionVec" (cn' "position vector")
-   (S "a vector from the origin" `ofThe` phrase cartesian +:+ S "defined"
+   (S "a vector from the origin" `the_ofThe` phrase cartesian +:+ S "defined"
```
JacquesCarette, line 121: This one seems odd, i.e. wouln't it generate "the a vector..." ?

code/drasil-example/Drasil/GamePhysics/GenDefs.hs
```diff
@@ -107,7 +107,7 @@ accelGravityDerivSentence2 = [S "The above equation governs the gravitational at
 
 accelGravityDerivSentence3 :: [Sentence]
 accelGravityDerivSentence3 =  [S "Given the above assumptions" `sC` S "let", ch mLarger `sAnd` ch QPP.mass, 
-        S "be the", phrase QPP.mass `ofThe` S "massive and light body respectively.",
+        S "be the", phrase QPP.mass `the_ofThe` S "massive and light body respectively.",
```
JacquesCarette, line 110: won't this generate "the the" ?

JacquesCarette (changes requested):

JacquesCarette (approved):

JacquesCarette:
Except that your change to the hlint yaml seems to have a syntax error, which cased the build to fail. Otherwise, looks good.

code/.hlint.yaml
```diff
@@ -54,6 +54,7 @@
 # Ignore some builtin hints
 # - ignore: {name: Use let}
 # - ignore: {name: Use const, within: SpecialModule} # Only within certain modules
+  - ignore: {name: "Use camelCase"}
```
balacij, line 57: I believe you just need to remove the leading whitespace in front of the `-` so that it can be parsed for hlint.

Ant13731: Ah okay I'll try that. Thanks!

smiths: @Ant13731, you might want to run HLint before doing a pull request.  I think there is a make rule for hlint?

Ant13731: make hlint didn't work but I think the program passed the build test for hlint at least.

Issue #2370, opened 2021-05-06
Inconsistent naming in GamePhysics example.


JacquesCarette:
In words (and in UIDs) things are subscripted with `j`, but in the Haskell, `I` (capital i) is used instead. This is confusing and inconsistent. An old fix was in https://github.com/JacquesCarette/Drasil/commit/266a4805dacdbcba1f05cdbf014336952838d2fb but that is too old and can't be merged easily, it will have to be hand-patched.

Ant13731:
@JacquesCarette just for clarification, should they all have the `j` subscript or the `I`?

JacquesCarette:
I think the words are correct, i.e. it should be `j`.

PR #2373, opened 2021-05-06
Change Subscripts in GamePhys

Ant13731:
Changed all subscript `I` into `j` for GamePhys.
Closes issue #2370


code/drasil-example/Drasil/GamePhysics/DataDefs.hs
```diff
@@ -43,7 +43,7 @@ ctrOfMass = mkQuantDef posCM ctrOfMassEqn
 
 -- FIXME (Variable "i") is a horrible hack
 ctrOfMassEqn :: Expr
-ctrOfMassEqn = sumAll (Variable "i") (sy massI * sy posI) / sy mTot
+ctrOfMassEqn = sumAll (Variable "i") (sy massj * sy posj) / sy mTot
```
JacquesCarette, line 46: Hmm, something here seems amiss. The hack here is likely quite wrong. It does look like `mass` and `pos` depend on `j` (as defined in `Unitals`), but this sum over `i` is then wrong. Could you check in the generated PDF and HTML if that's indeed the case?

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
```diff
@@ -319,7 +319,7 @@ momentOfInertia :: QDefinition
 momentOfInertia = mkQuantDef QP.momentOfInertia momentOfInertiaEqn
 
 momentOfInertiaEqn :: Expr
-momentOfInertiaEqn = sumAll (Variable "i") $ sy massI * (sy rRot $^ 2)
+momentOfInertiaEqn = sumAll (Variable "i") $ sy massj * (sy rRot $^ 2)
```
JacquesCarette, line 322: Here as well, of course.

code/drasil-example/Drasil/GamePhysics/Unitals.hs
```diff
@@ -171,23 +171,23 @@ posCM = ucs "p_CM" (nounPhraseSP "Center of Mass")
...
                 (cn "of the j-th particle")) (phrase QPP.mass)) 
                 (sub (eqSymb QPP.mass) lJ) Real kilogram
 
-posI = ucs' (dccWDS "p_j" (compoundPhrase' (QP.position ^. term) 
+posj = ucs' (dccWDS "p_j" (compoundPhrase' (QP.position ^. term) 
                (cn "vector of the j-th particle")) (phrase QP.position))
                (sub (eqSymb QP.position) lJ) Real metre
 
-accI = ucs' (dccWDS "accI" (compoundPhrase' (cn "the i-th body's")
+accj = ucs' (dccWDS "accj" (compoundPhrase' (cn "the i-th body's")
```
JacquesCarette, line 182: Here `accj` has `i` in its English form!

code/drasil-example/Drasil/GamePhysics/Unitals.hs
```diff
@@ -171,23 +171,23 @@ posCM = ucs "p_CM" (nounPhraseSP "Center of Mass")
...
   (sub (eqSymb QP.position) lCMass) Real metre
 
-massI = ucs' (dccWDS "m_j" (compoundPhrase' (QPP.mass ^. term)
+massj = ucs' (dccWDS "m_j" (compoundPhrase' (QPP.mass ^. term)
                 (cn "of the j-th particle")) (phrase QPP.mass)) 
                 (sub (eqSymb QPP.mass) lJ) Real kilogram
 
-posI = ucs' (dccWDS "p_j" (compoundPhrase' (QP.position ^. term) 
+posj = ucs' (dccWDS "p_j" (compoundPhrase' (QP.position ^. term) 
                (cn "vector of the j-th particle")) (phrase QP.position))
```
JacquesCarette, line 179: and here `pos` has `j` in its English form - this is inconsistent.

JacquesCarette (changes requested):

code/drasil-example/Drasil/GamePhysics/Unitals.hs
Ant13731: Should I change the English description to "j-th body" in order to match the other variable changes. And does the same go for "F_i"?

Ant13731: I also noticed that torque and velocity use the "i-th" body in their description too. Should I change those to "j-th"?

JacquesCarette: They should all consistently use the same variable, for now. Putting a variable name in there is a hack, but one we have to live with right now.  Let's do `j`, as `i` is over-used and could clash with other things.

JacquesCarette (approved):

JacquesCarette (approved):

Issue #2376, opened 2021-05-06
Main README file missing information on pdController example


balacij:
In the [build specific examples area](https://github.com/JacquesCarette/Drasil#building-specific-examples), we are missing a row entry for `pdController_diff`.

In the [running specific examples area](https://github.com/JacquesCarette/Drasil#running-the-examples), we are missing a row entry for `pdController`.

balacij:
@Awurama-N Would you mind having a look at this please?

balacij:
With #2380 merged in, we can close this ticket. Thank you @Awurama-N

PR #2380, opened 2021-05-06
Adding pdController missing rows

Awurama-N:
adding pdController missing rows 
closes Issue #2376


Awurama-N:
added "deleting unwanted references" commit to this accidently, will attempt to delete that commit

JacquesCarette (approved):

PR #2382, opened 2021-05-06
regenerate drasil-example cabal file

balacij:
I'm not too sure what happened, but it was giving errors in the `make` command!

A simple regeneration of the `drasil-example` cabal file is sufficient to fix it luckily. Maybe we'll need to write some documentation about using this build system as well later?


JacquesCarette (approved):
It might have been me - I had to do a hand-merge, and I might have chosen the wrong hash to copy in.

PR #2384, opened 2021-05-06
#2369 : convert TMods' lrIsSafe into a QDefinition

### LINKED ISSUE #2369: In GlassBR, the `lrIsSafe` theory model should use a QDefinition
instead of a `RelationConcept`.  See https://github.com/JacquesCarette/Drasil/commit/560d77699d707d2aa2f59baa9f37e25a11ad0c3f for the details.

balacij:
Closes #2369

### DIFF
--- a/code/drasil-example/Drasil/GlassBR/TMods.hs
+++ b/code/drasil-example/Drasil/GlassBR/TMods.hs
@@ -27,10 +27,8 @@ tMods = [pbIsSafe, lrIsSafe]
 lrIsSafe :: TheoryModel
 lrIsSafe = tm (cw lrIsSafeQD)
    [qw isSafeLoad, qw tmLRe, qw tmDemand] ([] :: [ConceptChunk])
-   [relToQD locSymbMap lrIsSafeQD] [sy isSafeLoad $= lrIsSafeExpr] [] [makeCite astm2009] 
+   [lrIsSafeQD] [sy isSafeLoad $= lrIsSafeExpr] [] [makeCite astm2009] 
    "isSafeLoad" [lrIsSafeDesc]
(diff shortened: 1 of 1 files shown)


code/drasil-example/Drasil/GlassBR/TMods.hs
```diff
@@ -25,16 +25,18 @@ tMods = [pbIsSafe, lrIsSafe]
 
 
 lrIsSafe :: TheoryModel
-lrIsSafe = tm (cw lrIsSafeRC)
+lrIsSafe = tm (cw lrIsSafeQD)
    [qw isSafeLoad, qw tmLRe, qw tmDemand] ([] :: [ConceptChunk])
-   [relToQD locSymbMap lrIsSafeRC] [sy isSafeLoad $= sy tmLRe $> sy tmDemand] [] [makeCite astm2009] 
+   [relToQD locSymbMap lrIsSafeQD] [sy isSafeLoad $= lrIsSafeExpr] [] [makeCite astm2009]
```
JacquesCarette, line 30: So it should be possible to not use `relToQD` here. There was a different combinator in that old branch that was more sensible for use here.  Through this pass, we need to reduce the number of calls to `relToQD` to as close to 0 as possible.

JacquesCarette (changes requested):

JacquesCarette (approved):

code/drasil-example/Drasil/GlassBR/TMods.hs
balacij: My apologies, I overlooked that bit. I just fixed it. The related area in IMods will be using the getEqMods function to grab the QDefinitions there later. I'll do that one in a separate PR (if I can get the logs to not break -- I might need to submit that as an issue instead).

Issue #2342, opened 2021-05-03
Move 2 references out of PDController/References


JacquesCarette:
`laplaceWiki` and `pidWiki` should be moved into Data.Drasil (see #2324) for a similar issue).

Awurama-N:
after taking out the two references, there are warnings that "The type signature for ‘pidWiki’ lacks an accompanying binding", same for laplacewiki, would you like me to take it out of the citation as well.

JacquesCarette:
The declarations of those two signatures should be moved along with the code. They belong together.  Did you just move ½ ?

Awurama-N:
oh okay... 
Yes at first i just took out the references themselves and nothing else

JacquesCarette:
If they are for something quite generic, yes. But not necessarily for everything - there's still a judgement call to make.

PR #2385, opened 2021-05-07
moving wiki references

Awurama-N:
moving wiki references
closes Issue #2342


Awurama-N:
should i have changed something in the "TModel.hs" file as well

code/drasil-example/Drasil/PDController/References.hs
```diff
@@ -1,27 +1,12 @@
 module Drasil.PDController.References where
 import Data.Drasil.Citations (smithLai2005)
-import Data.Drasil.People (wikiAuthors)
 
 import Language.Drasil
 
 citations :: BibRef
-citations = [pidWiki, laplaceWiki, johnson2008, abbasi2015, smithLai2005]
+citations = [johnson2008, abbasi2015, smithLai2005]
```
JacquesCarette, line 7: So those two citations should still be in the `citations` list here. This is a list of all the citations used in this example, regardless of where they are defined.

JacquesCarette (changes requested):
Yes, you need to make changes to `TModel` as well, i.e. import those references.

In general, you need to do `make` locally, and make sure that everything compiles successfully, and the tests pass too.

JacquesCarette (approved):

PR #2381, opened 2021-05-06
Fixing .gitattributes file

balacij:
Closes #2184 

Example of what the code stats will end up looking like: https://github.com/balacij/Drasil/

Breakdown of what will be counted towards the code statistics with this configuration file: ~~SNIP: See below comments~~

PR #2386, opened 2021-05-07
Remove Projectile's src folder

balacij:
As per [Dr. Smith's](https://github.com/JacquesCarette/Drasil/pull/2381#issuecomment-833813074) note, these files are no longer needed.


balacij:
Should we also remove the MG and MIS folder that [Dr. Carette](https://github.com/JacquesCarette/Drasil/pull/2381#issuecomment-833837399) mentioned as well?

JacquesCarette (approved):

smiths:
Yes @balacij, please remove the MG and MIS folders from the Projectile example.  I will move those files to the caseStudies repo (https://github.com/smiths/caseStudies).  There is a chance that some of this content will be useful in the future, but it makes more sense with the other manual created case studies.

Issue #2394, opened 2021-05-08
split content out of Concepts.hs into Unitals.hs


JacquesCarette:
The latter will have to be created.  Everything from line 131 onwards should be moved, and then everything else fixed up accordingly.

PR #2402, opened 2021-05-09
Creating Unitals file and moving neccessary data

Awurama-N:
moving content from 131 onwards to new file "Unitals"
closes issue #2394

### DIFF
--- a/code/drasil-example/drasil-example.cabal
+++ b/code/drasil-example/drasil-example.cabal
@@ -215,6 +215,7 @@ executable pdController
       Drasil.PDController.Requirements
       Drasil.PDController.SpSysDesc
       Drasil.PDController.TModel
+      Drasil.PDcontroller.Unitals
   hs-source-dirs:
       ./
   ghc-options: -Wall -Wredundant-constraints -threaded -O2 -main-is Drasil.PDController.Main
--- a/code/drasil-example/package.yaml
+++ b/code/drasil-example/package.yaml
@@ -155,6 +155,7 @@ executables:
     - Drasil.PDController.Requirements
     - Drasil.PDController.SpSysDesc
     - Drasil.PDController.TModel
+    - Drasil.PDcontroller.Unitals
      
   ssp:
     main: Drasil.SSP.Main


Awurama-N:
when i ran make i didn't come across issues related to the Unitals file. But there's a warning here that the module cannot be found. please How can I fix that?

code/drasil-example/Drasil/PDController/DataDefs.hs
```diff
@@ -4,6 +4,8 @@ import Drasil.PDController.Assumptions
 
 import Drasil.PDController.Concepts
 
+import Drasil.PDcontroller.Unitals
```
JacquesCarette, line 7: while you're at it, can you delete all those extra blank lines in the import list please?

code/drasil-example/drasil-example.cabal
```diff
@@ -215,6 +215,7 @@ executable pdController
       Drasil.PDController.Requirements
       Drasil.PDController.SpSysDesc
       Drasil.PDController.TModel
+      Drasil.PDcontroller.Unitals
```
JacquesCarette, line 218: This is going to get automatically generated - it is better not to add it by hand (unlike for `package.yaml`)

JacquesCarette (changes requested):

Awurama-N:
okay will do. is there anything else i have to do to stop the build check from failing?

code/drasil-example/Drasil/PDController/Body.hs
```diff
@@ -71,6 +71,7 @@ import Language.Drasil.Code
 import Language.Drasil.Printers (PrintingInformation(..), defaultConfiguration)
 import Theory.Drasil (DataDefinition, GenDefn, InstanceModel, TheoryModel)
 import Utils.Drasil
+import Drasil.PDcontroller.Unitals
```
JacquesCarette, line 74: This should be `PDController`, not `PDcontroller` (capital C, not lower-case c). This occurs multiple times - that's the reason it cannot be found.

JacquesCarette (changes requested):

code/drasil-example/Drasil/PDController/Body.hs
Awurama-N: oh okay

PR #2405, opened 2021-05-10
Move order related binary ops into OrdBinOp

balacij:
Separate order related binary operations and the general binary operations


code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
```diff
@@ -15,7 +16,7 @@ liftS = fmap (: [])
 
 -- For an expression using a less than operator, gets the upper bound
 getUpperBound :: Expr -> Expr
-getUpperBound (BinaryOp Lt _ b) = b
+getUpperBound (OrdBinaryOp Lt _ b) = b
```
JacquesCarette, line 19: Please file an issue about getting rid of `getUpperBound` - it's a horrible hack. We should not have "get"-style functions that work on `Expr` at all. The knowledge should be encoded in other ways so that there is no `error` case. We'll first need to analyse all the uses of `getUpperBound` to design a proper replacement.

JacquesCarette (approved):

PR #2410, opened 2021-05-10
Moving concepts into data-drasil

Awurama-N:
Is this what you had in mind for moving concepts into data-drasil concepts files?
If so, i can look for the others to move int o their respective files as well.
And if i delete them from the file they were originally in, should i go ahead and import the files they are now located in into any files that were importing Data.PDContoller.Concepts for those specific concepts?

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Math.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Math.hs
@@ -25,6 +25,15 @@ angle, area, axis, calculation, cartesian, centre, change, constraint, diameter,
   rightHand, shape, surArea, surface, unitV, unit_, vector, xAxis, xCoord, xComp, xDir,
   yAxis, yCoord,  yComp, yDir, zAxis, zCoord, zComp, zDir, iAngle :: ConceptChunk
 
+symbols :: [QuantityDict]
+symbols = [qdNegInf, qdPosInf, qdInvLaplaceTransform] 
+qdNegInf, qdPosInf, qdInvLaplaceTransform :: QuantityDict
+
+symnegInf, symposInf, syminvLaplace :: Symbol
+symnegInf = Variable "-∞"
+symposInf = Variable "∞"
+syminvLaplace = Variable "L⁻¹[F(s)]"
+
 
 angle       = dcc "angle"        (cn' "angle")                   "the amount of rotation needed to bring one line or plane into coincidence with another"
 area        = dcc "area"         (cn' "area")                    "a part of an object or surface"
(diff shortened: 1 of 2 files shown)


code/drasil-data/Data/Drasil/Concepts/Math.hs
```diff
@@ -25,6 +25,15 @@ angle, area, axis, calculation, cartesian, centre, change, constraint, diameter,
...
   yAxis, yCoord,  yComp, yDir, zAxis, zCoord, zComp, zDir, iAngle :: ConceptChunk
 
+symbols :: [QuantityDict]
+symbols = [qdNegInf, qdPosInf, qdInvLaplaceTransform] 
+qdNegInf, qdPosInf, qdInvLaplaceTransform :: QuantityDict
+
+symnegInf, symposInf, syminvLaplace :: Symbol
+symnegInf = Variable "-∞"
+symposInf = Variable "∞"
+syminvLaplace = Variable "L⁻¹[F(s)]"
```
JacquesCarette, line 35: Hmm, this is a hack as a 'symbol', so let's not put that in `Concepts.Math` right now.

code/drasil-data/Data/Drasil/Concepts/Math.hs
```diff
@@ -100,3 +109,8 @@ perpV   = dcc "perp_vect"     (compoundPhrase' (perp ^. term) (vector ^. term))
 rOfChng = dcc "rOfChng"       (rate `of_` change) "ratio between a change in one variable relative to a corresponding change in another"
 surArea = dcc "surArea"       (compoundPhrase' (surface ^. term) (area ^. term)) "a measure of the total area that the surface of the object occupies"
 unitV   = dcc "unit_vect"     (compoundPhrase' (unit_ ^. term) (vector ^. term)) "a vector that has a magnitude of one"
+
+
+qdPosInf              = vc "qdPosInf"             (nounPhraseSent (S "Infinity"))                                symposInf     Real
+qdInvLaplaceTransform = vc "qInvLaplaceTransform" (nounPhraseSent (S "Inverse Laplace Transform of a function")) syminvLaplace Real
```
JacquesCarette, line 115: and so this will, unfortunately, have to go too.

code/drasil-example/Drasil/PDController/TModel.hs
```diff
@@ -17,6 +17,7 @@ import Utils.Drasil
 
 import Data.Drasil.Citations(laplaceWiki)
 
+
```
JacquesCarette, line 20: why the extra blank line?

Also, this PR should delete `symnegInf` and `sumposInf` from PDController at the same time.

A separate PR can probably be made for re-using these in other examples.

JacquesCarette (changes requested):

code/drasil-example/Drasil/PDController/TModel.hs
JacquesCarette: What do you mean by 'format' here?  Take a look at what is already in `drasil-data` as a guide for how it is currently done. I think `Pi` already exists, so do the same for the infinities, for example.

Awurama-N: oh okay okay, i see Pi, i'll write them in as such then.
by "format" i was asking what the line should look like. initially i just wrote them in the same way they were put in the PDController/concepts file

Issue #2411, opened 2021-05-10
Error from using xPos and yPos in DblPendulum


Ant13731:
```cd "build/DblPendulum" && stack exec -- "dblpendulum" 
dblpendulum.EXE: Symbol: yPos not found in SymbolMap
CallStack (from HasCallStack):
  error, called at ./Database\Drasil\ChunkDB.hs:81:24 in drasil-database-0.1.1.0-1H2SDTZem9VDdavf3C6Z5k:Database.Drasil.ChunkDB
make: *** [Makefile:142: dblpendulum_gen] Error 1
```
When trying to use the `xPos` or `yPos` symbols in `dblpendulum`, I received the above error. However, switching those out with `ixPos` and `iyPos` seems to make the error go away.

balacij:
Would adding `xPos` and `yPos` to the Unitals/unitalChunks list resolve this?

Disclaimer: I'm not entirely sure if this what _should_ be done. We'd likely need Dr. Carette or Dr. Smith to discuss if these should be added or not.

Ant13731:
That worked! Thanks

Issue #2391, opened 2021-05-07
DblPendulum Derivation Missing


Ant13731:
Looking through the `DblPendulum` file, there are clear derivations for angular frequency, period of the pendulum, etc., but the derivations for the components of linear acceleration and velocity are just repeats of the equation given in `GenDefs`. I don't think the derivation is terribly complicated, so should I try to fill it in for the sake of completeness? On that note, is there a definition written out for velocity = angular velocity * radius?
![image](https://user-images.githubusercontent.com/69334555/117495344-7bd29500-af43-11eb-974c-ba95f3cc732d.png)
![image](https://user-images.githubusercontent.com/69334555/117495227-5ba2d600-af43-11eb-8a2a-31373775f361.png)

smiths:
@Ant13731, improving the double pendulum example is a great idea.  However, you don't want to be fixing it and refactoring it at the same time.  It can be difficult if you are trying to make changes to what it presents at the same time as you are changing the Drasil code on how it generates what it presents.  :smile:

If you and @JacquesCarette are done with the fixes to the double pendulum example, we can look at improving the generated documentation.  Are there any open issues for fixes that bring DblPendulum in-line with the previous fixes @JacquesCarette?  If not, adding a derivation would be good.

The relationship between velocity and angular velocity is given by Chasles's theorom.  This theorem can be found in the game physics example:

https://jacquescarette.github.io/Drasil/examples/GamePhysics/srs/GamePhysics_SRS.html#DD:chaslesThm

Ant13731:
Okay sounds good! I just finished making a rough draft for the velocity derivations, but it shouldn't break anything too much, even if we do end up changing other files. I figured that doing this would also help me get more familiar with Drasil language, so it's good practice nevertheless.

I also found that using the definition of arc length worked out decently enough for the velocity derivation, but I can always change it.

JacquesCarette:
There will be more refactoring issues coming, I had not had time to put them all in. I will try to keep them all small, so that this can be done in between as well.

One of the things to keep in mind while doing this: what is generic physics, and what is specific to this example? We want to take as much generic physics out of the examples and into our code database of 'facts'.

Ant13731:
Hi @smiths, I'm a little lost in deriving the equation for acceleration (the one in the picture above). I tried looking online but I can't seem to piece it together. Is there somewhere I can find the derivation or some help starting out?

Ant13731:
Alright, I'll check it out. Thanks!

PR #2413, opened 2021-05-10
Adding velocity and acceleration derivations in pendulum

Ant13731:
I apologize in advance for the large PR. After discussing issue #2391 some more, Dr. Smith gave me the link to the derivations of the velocity and acceleration equations, so most of the changes are just implementing those. `xPos` and `yPos` were also added to the list of symbols (which closes issue #2411)


Ant13731:
Yes! I enjoyed how the equations were automatically formatted and easily changeable

PR #2426, opened 2021-05-11
Convert velocityIX RC into QD

balacij:


balacij:
@JacquesCarette I'm converting this one back into a draft as well. It still uses ExprRelat's `relat` as in #2409. I _think_ you wanted to keep it, but I just want to make sure before we merge. It's fairly easy to remove usage of the `relat` if we want to do so.

balacij:
Sounds good. Would that be more in tune with when we make more use of EquationalModels vs ODEModels vs OthModels then? If I'm understanding correctly, then the EquationalModels would be more `:=` vs the others which would be more of an `=`.

JacquesCarette:
Correct. We'll also likely want other kinds of models, like MultiEquationalModel (or a better name) for when several things are defined simultaneously. There will be more. We basically want to minimize `OthModels` as much as we can.

PR #2433, opened 2021-05-11
Convert DblPendulum's accelerationIY GD into an EquationalModel

balacij:
Relies on #2432


balacij:
I'm going to try to avoid having PRs that rely on each other unless I absolutely need it. I was assuming that GitHub (like BitBucket) would recognize that once the convDblPenAccIXRC branch was merged into master, it would use a diff between branches to get the changes, but it seems it didn't (since it still shows changes that were done in the IX branch).

JacquesCarette:
You can tell me that one PR should be merged first, that works too.

JacquesCarette (approved):

Issue #2424, opened 2021-05-11
Potential Abstraction of Words in Pendulum


Ant13731:
Referring to second part of #2398. I'm not sure if these all are good candidates for use in functions, but here is a general list I found @JacquesCarette. I didn't find very many nouns like `rod` that could be replaced with functions.

Because these words have some conceptual/mathematical meaning and are often seen in physics problems, these could even be generally defined as a `CI` or a `ConceptChunk`:
horizontal
vertical
component
arc length
motion

Zooming in on the pendulum example, these could be defined:
amplitude
simple harmonic motion

Other phrases that I think already exist but are still not being used fully in `dblpendulum`:
newton's second law of rotation
length
mass
angle

Some other words that I felt were not as important but could still be used:
origin
direction
positive
point

For common English phrases I found:
denotes the
with respect to
defined as

There were also quite a few linking words, but it'll probably do more harm than good to change those. Edit: I found a whole list of where those words are already defined, so as a note for me it is under `drasil-utils/Utils/Drasil/Misc.hs`. I might try incorporating these as well.

Ant13731:
@JacquesCarette should the `denotes`, `wrt`, and `defnAs `combinators be Sentence-level or Concepts-level? I think that they are more English-based words rather than concept combinators but there's a case to be made about their conceptual and relational side. On this topic, should they have a type of `Sentence -> Sentence -> Sentence` or `(NamedIdea a, NamedIdea b) => a -> b -> someOtherType`? I feel like the first one would be more useful in `drasil-examples` but I'm not completely sure.

JacquesCarette:
Let's start with Sentence level. I do think that, in time, these will acquire meaning, but we're not there yet.

PR #2438, opened 2021-05-12
Abstraction of words in pendulum part 1

Ant13731:
Contributes to #2424 
Added in words for the first two lists and submitting a PR just so it can be added incrementally. Once this PR is done, should I just make a new branch to address the other parts of #2424?


JacquesCarette (approved):
Approved.  I would like @smiths to comment on the concepts added to `Concepts`, for both wording and whether any should be moved to `drasil-data` (and where).

JacquesCarette:
And yes to a new branch for the rest.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -9,10 +9,27 @@ pendulum :: CI
...
 rod = nc "rod" (cn' "rod")
+horizontal = nc "horizontal" (cn "horizontal") 
+vertical = nc "vertical" (cn "vertical") 
+---
+
+defs:: [ConceptChunk]
+defs = [shm, amplitude, component, arcLen, motion]
+
+shm, amplitude, component, arcLen, motion :: ConceptChunk
+shm = dcc "SHM" (nounPhraseSP "simple harmonic motion") ("repetitive movement back and forth through" ++
```
smiths, line 24: I think we  can improve a bit on this definition.  SHM still applies when there is damping and in this case the maximum displacements on each side won't be identical over time.  On of the displacements on each oscillation will be less than the other.  I've looked at some on-line definitions, and I like the ones that mention period motion and the notion of an equilibrium position.  How about defining SHM as:

"Periodic motion through an equilibrium position.  The motion is sinusoidal in time and demonstrates a single resonant frequency."

The definition is based on Wikipedia.  Is there a way to cite this source, at least as a comment in our code?

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -9,10 +9,27 @@ pendulum :: CI
...
+                                                         "an equilibrium position such that the maximum" ++
+                                                         "displacement on one side is equal to the maximum" ++
+                                                         "displacement on the other side")
+amplitude = dcc "amplitude" (nounPhraseSP "amplitude") ("the maximum displacement of a body measured from" ++
```
smiths, line 28: We might want to use amplitude for something other than displacements.  We talk about amplitude for any sine wave, whether the dependent variable is displacement, or voltage, or current, or whatever.  The Wikipedia page for the  Sine wave (https://en.wikipedia.org/wiki/Sine_wave) has a definition that I like:

amplitude: the peak deviation for the function from zero

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -9,10 +9,27 @@ pendulum :: CI
...
+                                                        "its equilibrium position")
+component = dcc "component" (nounPhrase "component" "components") "one part or direction of a vector"
```
smiths, line 30: I don't like this definition of vector components, since a component isn't a direction, but the projection of the vector onto one of the basis vectors of the coordinate system.  How about something like:

The scalar quantity defining the contribution of the vector in one of the coordinate directions

I actually found this one a bit tricky to say in a single sentence.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -9,10 +9,27 @@ pendulum :: CI
...
+arcLen = dcc "arc length" (nounPhraseSP "arc length") "the distance between two points on a curve"
+motion = dcc "motion" (nounPhraseSP "motion") "the act of changing position from one place to another"
```
smiths, line 32: I think we want to include time in the definition of motion: "change in position over time"

code/drasil-example/Drasil/DblPendulum/Concepts.hs
smiths: I think the home for this knowledge would be in `Physics`.  We probably will want to subdivide physics at some point, but this is a good home for now.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
smiths: I'm tempted to classify the definition of amplitude as belonging to `Math`.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
smiths: I think the definition of vector component belongs to `Math`.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
smiths: It seems to me that motion would be defined in `Physics`.

JacquesCarette (approved):

JacquesCarette:
I assumed this was complete. If that was incorrect, new things can still be added, which is why I did not delete the branch. Please delete if it was indeed complete.

Ant13731:
Yes I think its complete for now - any new stuff should probably go on a new branch anyways

Issue #1944, opened 2019-10-22
Refactor UnaryOp, BinaryOp


JacquesCarette:
Those classes were also created because of syntax. They should be split into classes according to type (without typing information).  This will cause the 'inheritance graph' of the classes to change a fair bit.

The next step will be to actually merge the new classes into more semantics-driven classes (around arithmetic, boolean logic, etc).

balacij:
I think this is more or less complete now @JacquesCarette. Should we be continuing in #2212 or do we want to branch into additional issues related to this one?

JacquesCarette:
Yes, I think this one is complete. We should open new issues for tasks that move us ever closer to having typed `Expr`.

The next one is going to be to split some things out of `Expr` entirely. I would guess that the programming-oriented pieces are the ones that seem easiest to achieve.

PR #2442, opened 2021-05-12
Move Vec -> Vec -> Vec binary operations

balacij:
Contributes to #1944

Maybe we should we rename `VVVBinOp` to `VecVecVecBinOp` or anything of the sorts?


JacquesCarette:
This is going to change again, so there's no need to come up with good names now.

JacquesCarette (approved):
So the additional thing would be for the the smart constructor `cross` to only operate on vectors and complain otherwise. Right now, that's probably still too hard. Nevertheless, while doing this, you should check that in all example, it typechecks. And if it doesn't, create an issue, as it's going to be a blocker later.

balacij:
Thank you @JacquesCarette, I was able to find one area (mentioned in #2451 now). I'll try to look for the same in other operations as well.

Issue #2396, opened 2021-05-08
Move some concepts from PDController in to drasil-data


JacquesCarette:
Or re-use them from drasil-data if they already exist. Coming to mind are:
- positive infinity
- negative infinity
- Laplace transform

There are likely others, which should be discussed in this issue.

PR #2447, opened 2021-05-13
Adding Negative & Positive Infinities to math concepts

Awurama-N:
closes Issue #2396

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Math.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Math.hs
@@ -55,19 +55,19 @@ number      = dcc "number"       (cn' "number")                  "a mathematical
 orient      = dcc "orientation"  (cn' "orientation")             "the relative physical position or direction of something"
 parameter   = dcc "parameter"    (cn' "parameter")               "a quantity whose value is selected depending on particular circumstances"
 --FIXME: Should "parameter" be in math?
-perp         = dcc "perp"         (cn' "perpendicular")           "At right angles"
+perp         = dcc "perp"         (cn' "perpendicular")          "At right angles"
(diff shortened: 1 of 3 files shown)


code/drasil-data/Data/Drasil/Concepts/Math.hs
```diff
@@ -12,7 +12,7 @@ mathcon :: [ConceptChunk]
 
 mathcon = [angle, area, axis, calculation, cartesian, centre, change, constraint, diameter,
   equation, euclidN, euclidSpace, gradient, graph, law, line, matrix, norm, normal,
-  normalV, number, orient, parameter, perp, perpV, pi_, point, probability, rOfChng,
+  normalV, number, orient, parameter, perp, perpV, pi_, qdNegInf, qdPosInf, point, probability, rOfChng,
```
JacquesCarette, line 15: please omit the `qd` prefix.  Just `negInf` and `posInf`.

JacquesCarette (changes requested):

Awurama-N:
is this more like what you needed done?
if so, then i'll create pull requests for the other concepts as well. 
I wasn't sure whether you needed it to be "NegInf" or "qdNegInf" (since this is what is already in the system).

code/drasil-data/Data/Drasil/Concepts/Math.hs
Awurama-N: Screenshot 2021-05-13 at 7 38 51 AM
there's a warning that there's a parse error with NegInf, after i take out the "qd"?

JacquesCarette: `negInf`, not `NegInf`. Look at the capitalization. Identifiers in Haskell cannot start with a capital letter, those are reserved for constructors.

JacquesCarette:
You should add to `Quantity` too, at the same time.

JacquesCarette (approved):

Issue #2404, opened 2021-05-10
CompoundPhrase Use in Concepts.Physics


Ant13731:
As I was working through the DblPendulum example, I found this:
![image](https://user-images.githubusercontent.com/69334555/117669859-59c05900-b175-11eb-847e-34135a2a6e66.png)
In the grand scheme of things, this is not that important, but I think the use of `compoundPhrase'` instead of `compoundPhrase` in the definition of `angAccel` in `Data.Drasil.Concepts.Physics` is causing the second word to be capitalized in the `dblpendulum` file. However, I am not sure if other files are relying on the capitalization of `compoundPhrase'` to work correctly. Here is the definition:
```Haskell
angAccel = dcc "angularAcceleration" (compoundPhrase' (angular ^. term) (acceleration ^. term))
  "the rate of change of angular velocity"
```
Would this be good for me to look into (once the other issues for dblpendulum are done)?

smiths:
Yes, @Ant13731 this sounds like a good thing to look into.  I wonder if it would be helpful to do a brief write-up to distinguish all of the combinators for phrases.  @JacquesCarette might know of an existing summary, maybe through the haddock documentation, but I know that I still find it confusing every time it comes up.  :smile:  I don't really have confidence that we've used the combinators consistently in the different examples.  That seems like a good avenue for exploration to me.

Ant13731:
Alright, thanks! I feel that even having all the combinator definitions in one spot could be helpful too. The Haddock documentation works well enough most of the time, but I have a bit of trouble trying to jump into the different constructor types (for concepts and sentences). For example, I really only found the `ofThe` constructor because of the duplicate "the"s in `dblpendulum`. Maybe even having a Drasil generated documentation of its language could help us incorporate it more reliably and improve generalization. This could be a decent continuation of issue #2399 as well.

smiths:
Yes, having Drasil information in Drasil is our eventual goal.  We would like Drasil to generate its own documentation.  :smile:

Ant13731:
For now, would it be a good idea to start a manual Latex document/glossary for the different combinators? Or would I be able to use the template files to make the documents within Drasil?

JacquesCarette:
I think it would be worthwhile to try to following:
1. augment the current Haskell sources with good Haddock comments, so that Drasil's own documentation has good descriptions of its own combinators
2. make some kind of hand-written HTML meta-document that points to the generated docs as a kind of roadmap.

[...]

PR #2448, opened 2021-05-13
Review Haddock Documentation for Utils

Ant13731:
Also separated some changes into smaller pieces (this originally started out as one idea and got messy from there, so I'm separating them here).
Tried to make the Haddock Documentation more informative (especially to newer users).
Contributes to #2404


JacquesCarette:
The build failed for hlint reasons - can you fix please?

code/drasil-example/Drasil/DblPendulum/Unitals.hs
```diff
@@ -49,7 +49,7 @@ lenRod = makeUCWDS "l_rod" (cn "length of rod")
         (sub cL lRod) metre
 
 pendDisplacementAngle = makeUCWDS "pendDisplacementAngle" (cn "displacement angle of pendulum")
-        (phrase angle `the_ofThe` phrase pendulum)
+        ( S "the" +:+ phraseNP ((angle `ofThe'` pendulum)))
```
JacquesCarette: Perhaps this change shouldn't be with the Haddock-oriented changes?  Also, perhaps there should be a `the_ofThe'` combinator?  I hate to see a `S "the"` re-appear...

JacquesCarette (changes requested):

JacquesCarette (approved):

Issue #2441, opened 2021-05-12
Some Concept Combinator Function Definitions Weird


Ant13731:
While I was looking through `Utils.Drasil`, I found that `andRT` uses a plural case for the second `NamedIdea` parameter when it should be singular. Also, according to the definition of `nounPhrase''`, the plural case for this is only taking singular phrases:
```Haskell
-- | Customizable `and_` combinator (takes two title case capitalization rules)
andRT :: (NamedIdea c, NamedIdea d) => 
  (c -> Sentence) -> (d -> Sentence) -> c -> d -> NP
andRT f1 f2 t1 t2 = nounPhrase''
  (phrase t1 `sAnd` plural t2) --here should be singular but uses plural
  (phrase t1 `sAnd` phrase t2) --and here should be plural but uses singular
  (Replace (atStart t1 `sAnd` phrase t2))
  (Replace (f1 t1 `sAnd` f2 t2))
```
I investigated a little more and found that switching those two around did not seem to break anything:
```Haskell
-- | Customizable `and_` combinator (takes two title case capitalization rules)
andRT :: (NamedIdea c, NamedIdea d) => 
  (c -> Sentence) -> (d -> Sentence) -> c -> d -> NP
andRT f1 f2 t1 t2 = nounPhrase''
  (phrase t1 `sAnd` phrase t2) --swapped
  (phrase t1 `sAnd` plural t2) --swapped
  (Replace (atStart t1 `sAnd` phrase t2))
  (Replace (f1 t1 `sAnd` f2 t2))
```
Should I leave it as the first way or change it to the second. I doubt `andRT` is using those cases of `NP` anyway, but it might've broken something not related to the documentation.

Ant13731:
In addition, the type signature of `for` is:
```Haskell
for :: (NamedIdea c, NamedIdea d) => c -> d -> Sentence
```
but it may be better if it is changed to:
```Haskell
for :: (NamedIdea c, NamedIdea d) => c -> d -> NP
```
in order to match the rest of the functions listed in `Phrase.hs`

Ant13731:
`compoundNC'` and `compountNCPlPl` are pretty much the same thing, but with a slightly more constrained type:
```Haskell
compoundNC' :: (NamedIdea a, NamedIdea b) => a -> b -> NamedChunk
compoundNC' t1 t2 = nc
[...]

PR #2453, opened 2021-05-13
Changed some combinators and their uses

Ant13731:
contributes to issue #2441 
Changed `theCustom` to `the''` to match similar constructors
modified `andRT`, removed `compoundNCPlPl`, moved and changed `for` family of functions so that they are with other `Sentence` level combinators


Ant13731:
Referring to the last list in #2424, should the `denotes`, `wrt`, and `defnAs` combinators be Sentence-level or Concepts-level?  I think that they are more English syntactic sugar than concept combinators but there's a case to be made about their conceptual and relational side. On this topic, should they have a type of `Sentence -> Sentence -> Sentence` or `(NamedIdea a, NamedIdea b) => a -> b -> someOtherType`? I feel like the first one would be more useful in `drasil-examples`.

JacquesCarette (changes requested):

JacquesCarette (approved):

Issue #2294, opened 2021-02-09
Haddock docs missing for some files


balacij:
For example, `Language.Drasil.URI.AST` does appear not have haddock documentation built.

Related: https://github.com/JacquesCarette/Drasil/pull/2293#discussion_r573059414

balacij:
@JacquesCarette It appears that Haddock ignores files that are not explicitly in the "exposed-modules" of the Cabal build files. But since all of the "exports" of a library are generally inside of the main "Drasil.hs" file of each library where all of the exports get threaded through, the documentation is still there but not divided into the individual files from which they originated. Were you wanting it to be divided in the haddock documents? For example, "InstanceModel" appears in Theory.Drasil https://jacquescarette.github.io/Drasil/docs/drasil-theory-0.1.0/Theory-Drasil.html#t:InstanceModel but actually comes from Theory.Drasil.InstanceModel.

JacquesCarette:
Hmm, that's a bit of a pain.  We kind of need 2 sets of documentation then, one external and one internal. Or something like that. Can you investigate what the recommended way to deal with such a situation is?  I don't mean exactly the solution I outline, but the general problem of documenting things not in `exposed-modules`.

balacij:
I'll take a look!

PR #2454, opened 2021-05-13
Small change to ChunkDB

### LINKED ISSUE #2411: Error from using xPos and yPos in DblPendulum

Ant13731:
Small PR, documented the `ChunkDB` type with a little bit more detail. Closes #2411 for now. I will probably take a look at the entire folder once I am more familiar with the functions/chunk system.

### DIFF
--- a/code/drasil-database/Database/Drasil/ChunkDB.hs
+++ b/code/drasil-database/Database/Drasil/ChunkDB.hs
@@ -15,7 +15,7 @@ import Data.List (sortOn)
 import Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Map as Map
 
--- The misnomers below are not actually a bad thing, we want to ensure data can't
+-- | The misnomers below are not actually a bad thing, we want to ensure data can't
 -- be added to a map if it's not coming from a chunk, and there's no point confusing
 -- what the map is for. One is for symbols + their units, and the others are for
 -- what they state.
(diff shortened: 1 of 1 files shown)


Ant13731:
Some of the Mapping data types are not included in the haddock documentation (even though they have the proper comment syntax) as they do not import to `Database.Drasil`. Should they be imported or is there a way to get the documentation out of the nested files?

JacquesCarette (approved):

JacquesCarette:
I think you mean that some don't show up in the Haddock because they are not exported from this current file?  Imports don't change anything wrt visibility, but export does. And what do you mean here by 'nested file'?

Ant13731:
Yes, that's what I meant. By nested files I mean that the haddock documentation really only documents the main file (in this case, `Database.Drasil`) instead of all the smaller files that will eventually be imported (and exported) by `Database.Drasil`. In better wording, is there a way for the haddock documentation to "see" some of the types that are not exported?

balacij:
@Ant13731 Yep, we're going to work on the Haddock docs in #2294 once the gh-pages automation is done (soon!). We previously discussed having external vs internal Haddock docs where the external one is fairly minimal in comparison to the internal one (which would show everything [excluding external dependencies]).

Right now the docs are deployed using https://github.com/JacquesCarette/Drasil/blob/5fbcb79c17aa8c7820abac644ddb23a38ea1ee64/.github/workflows/Drasil.yaml#L69 which causes a bit of problems as you already noticed.

The main issue lies in which modules are exposed in our package.yaml files, we just need to override them with the haddock args.

Ant13731:
Okay I see. Sounds good!

PR #2456, opened 2021-05-13
adding physics concepts to drasil-data

Awurama-N:
contributes to Issue #2396


JacquesCarette (approved):
Do remember to go back and use these new concepts in (all) the examples too. But as a follow-up PR.

Issue #2450, opened 2021-05-13
Double Spacing in PDController


Ant13731:
Some of the definitions have double spacing (having manual spaces in addition to the automatic spaces generated by `+:+` in the PDController files)

PR #2459, opened 2021-05-14
Removed Whitespaces in PDController

Ant13731:
Removed unneeded whitespaces and used commas over `+:+` and `<>` operators.
Closes issue #2450


JacquesCarette (approved):
Excellent!

PR #2462, opened 2021-05-14
moving software concepts into drasil-data

Awurama-N:
closes Issue #2396


JacquesCarette:
Actually, none of these 3 concepts are accompanied by good definitions. The text you use here is way too example-specific. Let's leave those as is.

Issue #2399, opened 2021-05-08
Sentence vs concept-level combinators


JacquesCarette:
This issue should eventually apply to all examples, but it can start at DblPendulum.  It should be done after the first part of #2398 is done.

https://github.com/JacquesCarette/Drasil/blob/dc6da83d0ddf6c04c98bc1fdb0f43f8a8f9c1a57/code/drasil-example/Drasil/DblPendulum/Unitals.hs#L47-L57

Note the ` (phrase angle `the_ofThe` phrase pendulum)` (and same with `rod`). What's going on is that there is a sentence-level combinator `the_ofThe` being used. That's a fine thing, but in common use, it's applied on concepts, so explicit extraction needs to happen.

Better would be a combinator, in a separate package that can be imported qualified (say as `C`) so that this could instead be `angle C.the_ofThe pendulum`.  In fact, even better would be if the sentence-level combinator was imported qualified (as `S`) and the concept-level ones not qualified, so that this could be `angle the_ofThe pendulum` or even `the angle ofThe pendulum` (there's a `the` combinator too!).

The point is that this piece of English actually is very semantic. It embodies the pattern "the unique attribute X associated to the unique object Y" where here X = angle and Y = pendulum.  Recognition of the semantics of what the English is actually saying will let us generalize things.

This is another issue that requires some local fixes, as well as the creation of further issues that detail the future work that needs to be done to improve the examples overall.

Ant13731:
We could also split up Utils.Drasil into two separate files (the concepts go with the general file while the sentence combinators get their own separate imports (this way they can be qualified). I got something to work but it's a really hacky solution that's probably worse than before.

Ant13731:
As a note to myself, #2423 should also add a sentence ending operator

Ant13731:
I'm having a little bit of trouble incorporating `:.` as an operator. I looked it up and I think the `:` is reserved for data/type constructors, so implementing it as a function would be impossible. I also tried to use the postfixoperator language but I keep getting parse errors when trying to implement it. Is there an example of other postfix operators I could take a look at?
```
Parse error in pattern: (+.)
|
| ((+.) a) = a :+: S "."
|
```
(this was also posted under the PR but that has been merged and closed so I put it here too)

balacij:
[...]

Issue #2397, opened 2021-05-08
Re-use of phrases in PDController quantities.


JacquesCarette:
Take for example
```haskell
qdCtrlVarTD
  = vc "qdCtrlVarTD" (nounPhraseSent (S "Control Variable in the time domain"))
      symCT
      Real
```
The phrase is of the form "X in the Y", for X and Y both concepts that pre-exist. So that's how it should be described, i.e. this should be essentially
```haskell
qdCtrlVarTD = vc "qdCtrlVarTD" (controlVar `inThe` timeDomain) symCT Real
```
Note that I am not sure of the names `controlVar` or `timeDomain` (something like them should exist, or be created). Similarly, there might already be a `inThe` combinator that may or may not do the right thing. This needs to be investigated and dealt with first.

Also, I think there's a combinator that defaults to `Real` so this could be shortened further.

Awurama-N:
` inThe` combinator does indeed exist and is used (for example) in the following situation:
```
oIDQConstruct x = cic "outputInputDerivVals" (foldlSentCol [
  titleize output_, S "the", plural inValue `sAnd`
  S "derived", plural value  `inThe` S "following list"] +:+.
  foldlList Comma List x) "Output-Input-Derived-Values" funcReqDom
```

JacquesCarette:
Right, this was discussed at the meeting today, and also arose in things that @Ant13731 was doing. The "problem" is that some of the combinators are for sentences (like `inThe`), while we'd also want a variant for concepts.

I think that @Ant13731 is working on changing `drasil-utils` in this direction. Once that's done, this can be changed.

Awurama-N:
@JacquesCarette I'm going to break the question into bits. So firstly:
How can I know what these "nounPhraseSent" strings should look like on either side of the combinator?
```
qdSetPointFD = vc "qdSetPointFD" (nounPhraseSent (S "Set-Point" `inThe` S "frequency domain")) symYrS Real
```
I was originally thinking frequency domain might be ccFrequencyDomain, but that still raises warnings.

JacquesCarette:
`nounPhraseSent` is a function that creates a noun-phrase from a Sentence. [It's actually a bit of a hack, things should go in that direction!].

What would be optimal would be if there were concepts for "Set-Point" and "frequency domain", so that the phrase used to describe this new quantity could be derived from these.

[...]

PR #2464, opened 2021-05-14
Separated Concept and Sentence from Utils.Drasil

Ant13731:
Contributes to #2399 and #2397 
Starting the process of separating and separately importing Sentence combinators and Concept combinators. Also made new Concept-level combinators (under `Concepts.hs`) to help chain words together with their relations. I haven't quite separated everything from `Utils.Drasil` into `Utils.Drasil.Concepts` but this at least starts separating `Utils.Drasil.Sentence` out. Most of this PR is importing the separated `Sentence` file. The first two commits are adding the concept combinators and the third one is adding all the import statements.

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Computation.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Computation.hs
@@ -35,9 +35,9 @@ dataStruct, dataStruct', dataType, dataType',
   computerLiteracy, computerApp :: NamedChunk
 
 dataStruct       = compoundNCPlPh datum structure
-dataStruct'      = compoundNCPlPl datum structure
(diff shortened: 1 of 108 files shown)


JacquesCarette (approved):

JacquesCarette:
Good work, but the build failed, that will need fixed first.

Ant13731:
Yep, I think the reason it keeps breaking is that my local branch is behind the remote master branch so it sometimes calls a function that uses an import that wasn't needed before, so I'll have it fixed soon.

JacquesCarette (approved):

JacquesCarette:
Can you merge master, there's now a merge conflict.

JacquesCarette (approved):

PR #2465, opened 2021-05-14
More Abstracted Words

Ant13731:
Incorporates 3rd list of words for abstraction which were previously defined but unused in dblpendulum.
Contributes #2424

### DIFF
--- a/code/drasil-example/Drasil/DblPendulum/Concepts.hs
+++ b/code/drasil-example/Drasil/DblPendulum/Concepts.hs
@@ -5,7 +5,7 @@ import Data.Drasil.Theories.Physics (newtonSLRRC)
 
 
 concepts :: [IdeaDict]
-concepts = map nw [rod, horizontal, vertical] ++ map nw defs ++ [nw newtonSLRRC] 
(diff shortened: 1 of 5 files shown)


code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -1,10 +1,11 @@
 module Drasil.DblPendulum.Concepts where
 
 import Language.Drasil
+import Data.Drasil.Theories.Physics (newtonSLRRC)
 
 
 concepts :: [IdeaDict]
-concepts = map nw [rod, horizontal, vertical] ++ map nw defs 
+concepts = map nw [rod, horizontal, vertical] ++ map nw defs ++ [nw newtonSLRRC]
```
JacquesCarette, line 8: When adding a single word at the end, use `: nw newtonSLRRC` instead. That's much more efficient.

JacquesCarette (changes requested):

JacquesCarette (approved):
that works too.

PR #2471, opened 2021-05-17
More abstract words for dblPendulum

Ant13731:
Adds list four of abstract words contributing to issue #2424. I'm not too sure if the definitions are good (notably the direction definition).


JacquesCarette (approved):
Not the best definitions ever, but that can be dealt with later.

Issue #2476, opened 2021-05-18
Adding Higher Level Operators to DBLPendulum


Ant13731:
From the last comment of #2399. Now that `Utils.Drasil.Sentence` imports as qualified, I think trying to implement as many level 3 (or level 2) combinators into `dblPendulum` would be a good first step to completing #2399.

In addition, I think it would be good to have a separate `the_ofThe` combinator for each level, but I'm not sure how a good naming scheme would work. So far, the most common among combinators are:
Level 1 `Sentence -> Sentence` level combinators: prepends an `s` to most combinators
Level 2 `NP or NounPhrase` level combinators: appends `NP` to most combinators
Level 3 `NamedIdea` or higher level conceptual combinators: are the cleanest ones (only appended by `'` and `_`, although they convert `NamedIdeas` into `NP`s)

Edit: For clarification, this issue is meant to incorporate more conceptual-level combinators for ideas and phrases, rather than using `Sentence`-level combinators. This is to allow Drasil to gain more information about the semantics of what we mean and be more self-sufficient, rather than just telling it the exact sentences we want it to generate.
From #2399:
> The point is that this piece of English actually is very semantic [referring to combinators such as `the_ofThe`]. It embodies the pattern "the unique attribute X associated to the unique object Y" where here X = angle and Y = pendulum. Recognition of the semantics of what the English is actually saying will let us generalize things.

JacquesCarette:
I think that all 3 levels should be called `the_ofThe`, and that users of these should be forced to disambiguate which one they mean at the point of use.

We should undo the naming conventions listed above (prepend `s`, append `NP`); these conventions pre-date our better understanding of what we want.

Note that this issue is not as precise as it should be. While it is perfectly understandable to us now, it may not be so to someone next year, if somehow this issue were not closed over this summer. [There are many such issues already, we want to minimize them, not let them multiply!]

Ant13731:
@JacquesCarette in order to have `the_ofThe` named the same at all 3 levels, should I separate the level 2 (`NounPhrase`) combinators from `Utils.Drasil` so they can be qualified and imported separately just as `Utils.Drasil.Sentence` is separated? 

As for the prepending `s`, some of the combinators would clash with Haskell's syntax (namely `of` and `and`). Should an underscore just be appended to the end of those ones? That's how the conceptual combinators are currently named too.

Ant13731:
[...]

PR #2479, opened 2021-05-18
Implementing new combinators into dblPendulum

Ant13731:
Contributes #2399
Contributes #2476 
Implemented some higher level combinators in the pendulum example. I think `compoundNC` allows for a level 3 combination between `NamedIdeas`, but I had to place it into the database. Although I'm not sure if `Assumptions.hs` is the best place for it.


code/drasil-example/Drasil/DblPendulum/Assumptions.hs
```diff
@@ -20,18 +22,18 @@ yAxisDir        = cic "yAxisDir"          yAxisDirDesc        "yAxisDir"
 startOrigin     = cic "startOrigin"       startOriginDesc     "startOrigin"     assumpDom
 
 pend2DMotionDesc :: Sentence
-pend2DMotionDesc = S "The" +:+ phrase pendulum +:+ phrase motion `S.sIs` phrase twoD +:+. sParen (getAcc twoD)
+pend2DMotionDesc = atStartNP (the (compoundNC pendulum motion)) `S.sIs` phrase twoD +:+. sParen (getAcc twoD)
```
JacquesCarette, line 25: it's probably not a good idea to use `compoundNC` inline like this. You should give this a Haskell-level name.

code/drasil-example/Drasil/DblPendulum/Body.hs
```diff
@@ -94,12 +94,12 @@ justification :: Sentence
 justification = foldlSent [S "A", phrase pendulum, S "consists" `S.sOf` phrase mass, 
                             S "attached to the end of a", phrase rod `S.andIts` S "moving curve" `S.sIs`
                             (S "highly sensitive to initial conditions" !.), S "Therefore" `sC`
-                            S "it is useful to have a", phrase program, S "to simulate", phrase motion
-                            `S.the_ofThe` phrase pendulum, (S "to exhibit its chaotic characteristics" !.),
+                            S "it is useful to have a", phrase program, S "to simulate", phraseNP (motion
```
JacquesCarette, line 97: here too, this should be named. Same below.

code/drasil-example/Drasil/DblPendulum/Concepts.hs
```diff
@@ -3,9 +3,18 @@ module Drasil.DblPendulum.Concepts where
...
 
+--below imports needed compoundNC part to work
+import Utils.Drasil
+import Data.Drasil.Concepts.Physics (pendulum, motion, position, velocity, force)
+
 
 concepts :: [IdeaDict]
-concepts = nw newtonSLRRC : map nw [rod, horizontal, vertical] ++ map nw defs
+concepts = nw newtonSLRRC : map nw [rod, horizontal, vertical] 
+  ++ map nw defs ++ map nw [compoundNC pendulum motion,
```
JacquesCarette, line 13: aha, with all the previous things named, this list would then shorten!

JacquesCarette (changes requested):
Overall, good stuff. But as this is being edited, may as well do it 'right'.  Nice side-effect of fixing bad periods on the output.

Ant13731:
This should be okay to merge now (and resolve the requested changes). I know we already discussed changing the naming scheme and moving some combinators in other issues, but I think it would be easier to just start on a fresh branch.

JacquesCarette (approved):

PR #2484, opened 2021-05-19
Moving fterms to Utils

### LINKED ISSUE #2483: Moving fterms function out of drasil-data
@JacquesCarette @smiths I found this function while looking through `Data.Drasil.Concepts.Documentation`:
```Haskell
-- FIXME: fterms is here instead of Utils because of cyclic import
-- | Apply a binary function to the terms of two named ideas, instead of to the named
-- ideas themselves. Ex. @fterms compoundPhrase t1 t2@ instead of
-- @compoundPhrase (t1 ^. term) (t2 ^. term)@
fterms :: (NamedIdea c, NamedIdea d) => (NP -> NP -> t) -> c -> d -> t
fterms f a b = f (a ^. term) (b ^. term)
```
I think this can now be moved to `Utils` because `Data.Drasil.Concepts.Documentation` already imports `Utils.Drasil` in the first place. This also would allow `fterms` to be used in `Data.Drasil.Concepts.Math` and `Data.Drasil.Concepts.Physics`

Ant13731:
Moved `fterms` to `Utils.Drasil.Misc`, used `fterms` more in `Data.Drasil.Concepts.Math`.
Closes #2483.

### DIFF
--- a/code/drasil-data/Data/Drasil/Concepts/Thermodynamics.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Thermodynamics.hs
@@ -6,7 +6,6 @@ import Utils.Drasil
 import Data.Drasil.Concepts.Documentation (source, theory)
 import Data.Drasil.Concepts.Physics (energy)
 
---import Control.Lens((^.))
 
 thermocon :: [ConceptChunk]
 thermocon = [boilPt, boiling, degree_', enerSrc, heat, heatCapSpec, heatTrans,
--- a/code/drasil-data/Data/Drasil/Concepts/Thermodynamics.hs
+++ b/code/drasil-data/Data/Drasil/Concepts/Thermodynamics.hs
@@ -1,11 +1,12 @@
 module Data.Drasil.Concepts.Thermodynamics where
 
 import Language.Drasil
+import Utils.Drasil
 
 import Data.Drasil.Concepts.Documentation (source, theory)
 import Data.Drasil.Concepts.Physics (energy)
 
-import Control.Lens((^.))
+--import Control.Lens((^.))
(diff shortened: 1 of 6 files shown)


code/drasil-data/Data/Drasil/Concepts/Thermodynamics.hs
```diff
@@ -1,11 +1,12 @@
 module Data.Drasil.Concepts.Thermodynamics where
 
 import Language.Drasil
+import Utils.Drasil
 
 import Data.Drasil.Concepts.Documentation (source, theory)
 import Data.Drasil.Concepts.Physics (energy)
 
-import Control.Lens((^.))
+--import Control.Lens((^.))
```
JacquesCarette, line 9: go ahead and delete such imports rather than just commenting them out.

code/drasil-utils/Utils/Drasil/Misc.hs
```diff
@@ -227,3 +227,10 @@ checkValidStr s [] = Right s
 checkValidStr s (x:xs)
   | x `elem` s = Left $ "Invalid character: \'" ++ [x] ++ "\' in string \"" ++ s ++ ['\"']
   | otherwise  = checkValidStr s xs
+
+-- FIXME: fterms is here instead of Utils because of cyclic import
```
JacquesCarette, line 231: You should remove the `FIXME` comment!

JacquesCarette (changes requested):
Two very minor things to fix first.

Really nice to see how these combinators really shorten many definitions. Nicely spotted.

JacquesCarette (approved):

Issue #2477, opened 2021-05-18
More Detailed Haddock Documentation


Ant13731:
Building off of the discussion from #2404, this is just to keep track of the progress of making more detailed haddock documentation.
> 1. augment the current Haskell sources with good Haddock comments, so that Drasil's own documentation has good descriptions of its own combinators
> 2. make some kind of hand-written HTML meta-document that points to the generated docs as a kind of roadmap.
> 
> Doing it this way "by hand" may give us some insights as to how we'd want such things documented, as a prelude to seeing if it can be done in Drasil itself.

Progress list (I'm not sure if `example` and `data` need to be checked, but I'll put them here anyway):
- [x] `drasil-build` documentation
- [x] `drasil-code-base` documentation
- [x] `drasil-code` documentation
- [x] `drasil-data` documentation
- [x] `drasil-database` documentation
- [x] `drasil-docLang` documentation
- [ ] `drasil-example` documentation
- [x] `drasil-gen` documentation
- [ ] `drasil-gool` documentation (skip this one for now)
- [x] `drasil-lang` documentation
- [x] `drasil-metadata` documentation
- [x] `drasil-printers` documentation
- [x] `drasil-theory` documentation
- [x] `drasil-utils` documentation
- [x] `drasil-website` documentation

Some of the above folders are too large to complete in one go, so I'll list more details of them below.
For `drasil-code`:
- [x] Data
- [x] Language
    - [x] Chunk
    - [x] Code
        - [x] Imperative
            - [x] Build
            - [x] Doxygen
            - [x] GOOL
            - [x] Parsers
            - [x] Rest of Imperative files
        - [x] Rest of Code files
    - [x] Data
- [x] Test

For `drasil-data`:
- [x] Concepts
- [x] Equations
- [x] Quantities
- [x] Software
- [x] Theories
- [x] Units
- [x] rest of Drasil folder

For `drasil-docLang`:
- [x] DocLang
- [x] DocumentLanguage
- [x] Sections
- [x] rest of Drasil folder

For `drasil-gool`:
- [ ] LanguageRenderer
- [ ] rest of Drasil folder

For `drasil-lang`:
- [x] Chunk
- [x] Classes
- [x] Data
- [x] Development
- [x] Document
- [x] Expr
- [x] Label
- [x] NounPhrase
- [x] Sentence
- [x] Symbol
- [x] Uncertainty
- [x] URI
- [x] rest of files before README
- [x] rest of files after README 

[...]

PR #2485, opened 2021-05-19
Review Utils Documentation

Ant13731:
While making a summary of the combinators, I found some things I didn't like from my last changes to the Utils documentation, so these should remedy that. Tried to add more detail and keep comments to one style.
Contributes #2477


JacquesCarette (approved):
Just waiting for the build to complete.

PR #2486, opened 2021-05-19
Addition & Multiplication separation for different types

### LINKED ISSUE #2212: Drasil's expression language should have different kinds of multiplication
A mutliplication between numbers of the same type should be represented in Drasil differently than a multiplication between numbers of different types (i.e. an integer times a real number). This distinction should be added to `Expr`, probably to the `ArithOper` type.

See related discussion starting with this comment: https://github.com/JacquesCarette/Drasil/commit/a40c9a32df0af722bc9a0b07cda9bdb584633856#commitcomment-40209185

balacij:
Closes #2212 

* Disables `Num Expr`
* Disables `Fractional Expr`
* Separates addition and multiplication

~~I just have 1 question @JacquesCarette:~~

~~We mentioned that pi/2 was an exact number (as opposed to pi/2.0), but it seems that we don't do this constant folding ourselves and leave it to be done by the exported language's compiler/runtime. Should we be folding it ourselves?~~

~~I leave this as a draft because I have not yet re-implemented the functionality of Fractional, but it's getting very close.~~


code/stable/nopcm/src/cpp/input.txt
```diff
@@ -22,7 +22,5 @@
 1.0e-10
 # relative tolerance
 1.0e-10
-# temperature of the water (degreeC)
-40.0
```
balacij, line 26: I just wanted to explain this, it happened automatically as a result of building the software artifacts.

However, I think it makes sense, this file originally had 14 variables, but checking the [reader file](https://github.com/JacquesCarette/Drasil/blob/ef6297da096fa03dc27609c59db0c219dd97751f/code/stable/nopcm/src/cpp/InputParameters.cpp#L15), it only reads 13, and the related [sampleInput file](https://github.com/JacquesCarette/Drasil/blob/master/code/datafiles/NoPCM/sampleInput.txt) also only carries these leftover 13.

JacquesCarette: Good that you explained, as I would have asked! It's a bit odd still. @smiths do you know why the temperature is not actually used here?

code/drasil-data/Data/Drasil/Constraints.hs
```diff
@@ -3,5 +3,5 @@ module Data.Drasil.Constraints where
 import Language.Drasil
 
 gtZeroConstr, probConstr :: Constraint
-gtZeroConstr = physc $ UpFrom  (Exc, 0)
-probConstr   = physc $ Bounded (Inc, 0) (Inc, 1)
+gtZeroConstr = physc $ UpFrom  (Exc, Int 0)
```
JacquesCarette, line 6: As this is a "physical constraint", it is really about reals, so this is the wrong type.

code/drasil-data/Data/Drasil/Constraints.hs
```diff
@@ -3,5 +3,5 @@ module Data.Drasil.Constraints where
...
+probConstr   = physc $ Bounded (Inc, Int 0) (Inc, Int 1)
```
JacquesCarette, line 7: Ditto: these are probabilities.

code/drasil-data/Data/Drasil/Equations/Defining/Physics.hs
```diff
@@ -20,16 +20,16 @@ newtonSLRel, weightEqn, weightDerivAccelEqn, weightDerivNewtonEqn, weightDerivRe
...
-weightDerivNewtonEqn      = sy QP.weight $= sy QPP.mass * sy QP.gravitationalAccel
-weightDerivReplaceMassEqn = sy QP.weight $= sy QPP.density * sy QPP.vol * sy QP.gravitationalAccel
-weightDerivSpecWeightEqn  = sy QP.weight $= sy QPP.vol * sy QPP.specWeight
+weightEqn                 = sy QP.weight $= mulRe (sy QPP.vol) (sy QPP.specWeight)
+weightDerivNewtonEqn      = sy QP.weight $= mulRe (sy QPP.mass) (sy QP.gravitationalAccel)
+weightDerivReplaceMassEqn = sy QP.weight $= mulRe (sy QPP.density) (mulRe (sy QPP.vol) (sy QP.gravitationalAccel))
+weightDerivSpecWeightEqn  = sy QP.weight $= mulRe (sy QPP.vol) (sy QPP.specWeight)
 
-weightDerivAccelEqn       = sy QP.acceleration $= vec2D 0 (sy QP.gravitationalAccel * sy QM.unitVectj)
+weightDerivAccelEqn       = sy QP.acceleration $= vec2D (int 0) (mulRe (sy QP.gravitationalAccel) (sy QM.unitVectj))
```
JacquesCarette, line 30: This wouldn't type... the first part is an integer, the 2nd part a real.

code/drasil-example/Drasil/DblPendulum/DataDefs.hs
```diff
@@ -60,7 +60,7 @@ frequencyDDQD :: QDefinition
 frequencyDDQD = mkQuantDef QP.frequency frequencyDDEqn
 
 frequencyDDEqn :: Expr
-frequencyDDEqn = 1 / sy QP.period
+frequencyDDEqn = dbl 1 $/ sy QP.period
```
JacquesCarette, line 63: The more I see things like this, the more I think we need an `Exact` constructor that maps integers to reals.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -729,7 +729,7 @@ \subsubsection{General Definitions}
 Substituting for $\mathbf{I}$
 
 \begin{displaymath}
-m {L_{\text{rod}}}^{2} \frac{\,d\frac{\,d{θ_{p}}}{\,dt}}{\,dt}=-{L_{\text{rod}}} m \mathbf{g} \sin\left({θ_{p}}\right)
+m {L_{\text{rod}}}^{2.0} \frac{\,d\frac{\,d{θ_{p}}}{\,dt}}{\,dt}=-{L_{\text{rod}}} m \mathbf{g} \sin\left({θ_{p}}\right)
```
JacquesCarette, line 732: That this is a `2.0` is odd. Exponents are usually / often integers. There should be a real-to-a-natural 'power' function.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -761,7 +761,7 @@ \subsubsection{General Definitions}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           T=2 π \sqrt{\frac{{L_{\text{rod}}}}{\mathbf{g}}}
+           T=2.0 π \sqrt{\frac{{L_{\text{rod}}}}{\mathbf{g}}}
```
JacquesCarette, line 764: And this is where we need `Exact`. Even though in code, we'd want `2.0`, in the display math, we'll still want `2`. So we do need to embed the integers into the reals in our representation.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -895,7 +895,7 @@ \subsubsection{Data Definitions}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           f=\frac{1}{T}
+           f=\frac{1.0}{T}
```
JacquesCarette, line 898: And here too, we'd still want a `1`, at least in the document.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -1014,17 +1014,17 @@ \subsubsection{Instance Models}
          
 \\ \midrule \\
 Input Constraints & \begin{displaymath}
-                    {L_{\text{rod}}}\gt{}0
+                    {L_{\text{rod}}}\gt{}0.0
```
JacquesCarette, line 1017: This, on the other hand, is more justified. Here, I think this is an improvement.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -437,7 +437,7 @@ \subsubsection{Theoretical Models}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           \mathbf{F}=G \frac{{m_{1}} {m_{2}}}{\text{||}\mathbf{d}\text{||}^{2}} \mathbf{\hat{d}}=G \frac{{m_{1}} {m_{2}}}{\text{||}\mathbf{d}\text{||}^{2}} \frac{\mathbf{d}}{\text{||}\mathbf{d}\text{||}}
+           \mathbf{F}=G \frac{{m_{1}} {m_{2}}}{\text{||}\mathbf{d}\text{||}^{2.0}} \mathbf{\hat{d}}=G \frac{{m_{1}} {m_{2}}}{\text{||}\mathbf{d}\text{||}^{2.0}} \frac{\mathbf{d}}{\text{||}\mathbf{d}\text{||}}
```
JacquesCarette, line 440: This `2.0` really is a `2`, mathematically.

code/stable/gamephysics/SRS/GamePhysics_SRS.tex
```diff
@@ -986,7 +986,7 @@ \subsubsection{Data Definitions}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           KE=\frac{m \mathbf{v}^{2}}{2}
+           KE=\frac{m \mathbf{v}^{2.0}}{2.0}
```
JacquesCarette, line 989: That `2.0` in the denominator might also be exact, but I'm not sure. The meaning might be "cut in half" rather than "divide by 2.0".

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -481,7 +481,7 @@ \subsubsection{Data Definitions}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           B=\frac{k}{\left(a b\right)^{m-1}} \left(E h^{2}\right)^{m} LDF e^{J}
+           B=\frac{k}{\left(a b\right)^{m-1.0}} \left(E h^{2}\right)^{m} LDF e^{J}
```
JacquesCarette, line 484: this is likely another case where the exponent might be exact - not sure.

code/stable/glassbr/SRS/GlassBR_SRS.tex
```diff
@@ -668,7 +668,7 @@ \subsubsection{Data Definitions}
         
 \\ \midrule \\
 Equation & \begin{displaymath}
-           NFL=\frac{{\hat{q}_{\text{tol}}} E h^{4}}{\left(a b\right)^{2}}
+           NFL=\frac{{\hat{q}_{\text{tol}}} E h^{4.0}}{\left(a b\right)^{2}}
```
JacquesCarette, line 671: whereas this one's odd.  fourth powers are rare - but it might be exact too.

code/stable/glassbr/src/cpp/DerivedValues.cpp
```diff
@@ -26,7 +26,7 @@ void derived_values(InputParameters &inParams) {
     outfile << " in module DerivedValues" << std::endl;
     outfile.close();
     
-    inParams.LDF = pow(3.0 / 60, 7.0 / 16);
+    inParams.LDF = pow(3.0 / 60.0, 7.0 / 16.0);
```
JacquesCarette, line 29: This is definitely an improvement.

code/stable/glassbr/src/swift/Calculations.swift
```diff
@@ -45,7 +45,7 @@ func func_J_tol(_ inParams: inout InputParameters) throws -> Double {
         throw "Error closing file."
     }
     
-    return log(log(Double(1) / (Double(1) - inParams.P_btol)) * (pow(inParams.a * inParams.b, 7.0 - Double(1)) / (2.86e-53 * pow(7.17e10 * pow(inParams.h, 2), 7.0) * inParams.LDF)))
+    return log(log(Double(1) / (Double(1) - inParams.P_btol)) * (pow(inParams.a * inParams.b, 7.0 - 1.0) / (2.86e-53 * pow(7.17e10 * pow(inParams.h, 2), 7.0) * inParams.LDF)))
```
JacquesCarette, line 48: One less cast! That is definitely an improvement.

JacquesCarette (changes requested):
Herculean work! In an excellent direction.

code/drasil-example/Drasil/DblPendulum/DataDefs.hs
balacij: I like the idea of `Exact`! I'll make it! Thank you.

code/drasil-data/Data/Drasil/Constraints.hs
balacij: This should be fixed now, thank you

code/drasil-data/Data/Drasil/Constraints.hs
balacij: Thank you, this should be fixed now

code/drasil-data/Data/Drasil/Equations/Defining/Physics.hs
balacij: Thank you, this should be fixed now

code/drasil-example/Drasil/DblPendulum/DataDefs.hs
balacij: Now, it and similar equations use `exactDbl` to construct them, leaving us with prettier looking equations when printed, but the more appropriate variants using `1.0` for the code generation. Thank you!

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
balacij: I was thinking that one of the next tasks might be to add a whole series of functions like this. Though, it seems that Java, C#, and Swift at least typically have their respective "Math.Pow" as Double X Double -> Double

balacij: For now, I switched it to using `square` which uses the new `exactDbl`, so it will be visually pleasing at least

code/stable/glassbr/SRS/GlassBR_SRS.tex
balacij: For the moment, I changed this one back to the exact 4

code/stable/glassbr/src/swift/Calculations.swift
balacij: Now all of the casts in this one are removed :smile: , thanks to exactDbl!

JacquesCarette (approved):
This batch of changes is good.

code/drasil-example/Drasil/GlassBR/DataDefs.hs
```diff
@@ -79,8 +78,8 @@ hFromt = dd hFromtQD [makeCite astm2009] Nothing "minThick" [hMin]
 
 --DD3-- (#749)
 
-loadDFEq :: Expr 
-loadDFEq = (sy loadDur / 60) $^ (sy sflawParamM / 16)
+loadDFEq :: Expr
+loadDFEq = (sy loadDur $/ dbl 60) $^ (sy sflawParamM $/ dbl 16)
```
JacquesCarette, line 82: I'm pretty sure this 60 comes from a minute or second computation, so is actually exact. No idea about the 16,

code/drasil-example/Drasil/PDController/IModel.hs
```diff
@@ -21,10 +21,10 @@ instanceModels = [imPD]
 imPD :: InstanceModel
 imPD
   = im (OthModel imPDRC)
-      [qwC qdSetPointTD $ UpFrom (Exc, 0), qwC qdPropGain $ UpFrom (Exc, 0),
-       qwC qdDerivGain $ UpFrom (Exc, 0)]
+      [qwC qdSetPointTD $ UpFrom (Exc, dbl 0), qwC qdPropGain $ UpFrom (Exc, dbl 0),
```
JacquesCarette, line 24: These all should be exact as well.

code/drasil-example/Drasil/Projectile/Expressions.hs
```diff
@@ -16,60 +16,60 @@ import Data.Drasil.Quantities.Physics (gravitationalAccelConst, ixVel, iyVel, xP
 import Drasil.Projectile.Unitals (launAngle, launSpeed, targPos, tol, landPos, flightDur, offset)
 
 flightDur', iyPos, yConstAccel, iSpeed :: Expr
-flightDur' = 2 * sy launSpeed * sin (sy launAngle) / sy gravitationalAccelConst
-iyPos = 0                                   --  launchOrigin
-yConstAccel = - sy gravitationalAccelConst  -- accelYGravity
+flightDur' = exactDbl 2 `mulRe` sy launSpeed `mulRe` sin (sy launAngle) $/ sy gravitationalAccelConst
+iyPos = dbl 0                                   -- launchOrigin
```
JacquesCarette, line 20: exact

code/stable/glassbr/SRS/GlassBR_SRS.tex
balacij: I _think_ you're correct in that it should be an exact as well. I changed it for now, we can alter it later if we would like.

code/drasil-example/Drasil/Projectile/Expressions.hs
```diff
@@ -16,60 +16,60 @@ import Data.Drasil.Quantities.Physics (gravitationalAccelConst, ixVel, iyVel, xP
...
 --
 speed' :: Expr
-speed' = sy QP.iSpeed + sy QP.constAccel * sy time
+speed' = sy QP.iSpeed `addRe` (sy QP.constAccel `mulRe` sy time)
 
 rectVelDerivEqn1, rectVelDerivEqn2 :: Expr
 rectVelDerivEqn1 = sy QP.constAccel $= deriv (sy speed) time
-rectVelDerivEqn2 = defint (eqSymb speed) (sy QP.iSpeed) (sy speed) 1 $=
-                   defint (eqSymb time) 0 (sy time) (sy QP.constAccel)
+rectVelDerivEqn2 = defint (eqSymb speed) (sy QP.iSpeed) (sy speed) (dbl 1) $=
```
JacquesCarette, line 51: That `1` should be exact

code/drasil-example/Drasil/SSP/DataDefs.hs
```diff
@@ -44,28 +44,28 @@ intersliceWtrFQD = mkQuantDef watrForce intersliceWtrFEqn
 
 intersliceWtrFEqn :: Expr
 intersliceWtrFEqn = completeCase [case1,case2,case3]
-  where case1 = ((inxi slopeHght - inxi slipHght) $^ 2 / 2  *
-          sy waterWeight + (inxi waterHght - inxi slopeHght) $^ 2 *
-          sy waterWeight, inxi waterHght $>= inxi slopeHght)
+  where case1 = ((square (inxi slopeHght $- inxi slipHght) $/ dbl 2) `mulRe`
```
JacquesCarette, line 47: `half` ?

code/drasil-example/Drasil/SSP/DataDefs.hs
```diff
@@ -44,28 +44,28 @@ intersliceWtrFQD = mkQuantDef watrForce intersliceWtrFEqn
...
+          sy waterWeight `addRe` (square (inxi waterHght $- inxi slopeHght) `mulRe`
+          sy waterWeight), inxi waterHght $>= inxi slopeHght)
 
-        case2 = ((inxi waterHght - inxi slipHght) $^ 2 / 2  * sy waterWeight,
+        case2 = ((square (inxi waterHght $- inxi slipHght) $/ dbl 2)  `mulRe` sy waterWeight,
```
JacquesCarette, line 51: `half` ?

code/drasil-example/Drasil/SSP/DataDefs.hs
```diff
@@ -178,23 +178,23 @@ tangStressQD :: QDefinition
...
+ratioVariation = dd ratioVarQD [makeCite fredlund1977] Nothing
   "ratioVariation" []
 
 ratioVarQD :: QDefinition
 ratioVarQD = mkQuantDef scalFunc ratioVarEqn
 
 ratioVarEqn :: Expr
 ratioVarEqn = completeCase [case1, case2]
-  where case1 = (1, sy constF)
+  where case1 = (dbl 1, sy constF)
```
JacquesCarette, line 194: I think this is exact

code/drasil-example/Drasil/SSP/GenDefs.hs
```diff
@@ -257,126 +257,126 @@ momentEql = makeRC "momentEql" (nounPhraseSP "moment equilibrium")
   momEqlDesc momEqlRel -- genDef6Label
 
 momEqlRel :: Relation
-momEqlRel = 0 $= momExpr (\ x y -> x +
-  (inxi baseWthX / 2 * (inxi intShrForce + inxiM1 intShrForce)) + y)
+momEqlRel = dbl 0 $= momExpr (\ x y -> x `addRe`
+  ((inxi baseWthX $/ dbl 2) `mulRe` (inxi intShrForce `addRe` inxiM1 intShrForce)) `addRe` y)
```
JacquesCarette, line 261: `half`?

code/drasil-example/Drasil/SWHS/Body.hs
```diff
@@ -541,10 +541,10 @@ propCorSolDeriv1 lce ewat en co pcmat g1hfc g2hfp su ht  =
 
 propCorSolDeriv2 :: Contents
 propCorSolDeriv2 = eqUnR'
-  (sy watE $= defint (eqSymb time) 0 (sy time)
-  (sy coilHTC * sy coilSA * (sy tempC - apply1 tempW time))
-  - defint (eqSymb time) 0 (sy time)
-  (sy pcmHTC * sy pcmSA * (apply1 tempW time -
+  (sy watE $= defint (eqSymb time) (dbl 0) (sy time)
```
JacquesCarette, line 544: both of these are exact 0

JacquesCarette (changes requested):
- I did not catch all the `dbl 0` and `dbl 1` that should be exact. Many more should be. Looks like you caught most of the `dbl 2`. Most other numbers shouldn't be.
- a few needed calls to `half` should be added
- add a `recip` (for 1/) smart constructor.

Quite a few languages have a `^` operator for when the base is real and the power is a natural number.  Right now, GOOL doesn't as far as I could tell, so that's a deeper problem (make an issue please).

balacij:
Thank you! I've converted it back to a draft for the time being. I'll perform some updates shortly and re-mark it as ready for review when it's appropriate.

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
JacquesCarette: `^` in a number of these languages is what's wanted.

code/drasil-example/Drasil/SWHS/Body.hs
balacij: Thank you, fixed it now

code/drasil-example/Drasil/SSP/GenDefs.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/SSP/DataDefs.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/SSP/DataDefs.hs
balacij: Fixed

code/drasil-example/Drasil/SSP/DataDefs.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/Projectile/Expressions.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/Projectile/Expressions.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/PDController/IModel.hs
balacij: Thanks! Fixed

code/drasil-example/Drasil/GlassBR/DataDefs.hs
balacij: Thanks, fixed the 60. I also opted to change the 16 to an exact as well since it's an equation. I think it just makes the output look nicer anyways. It's a reasonably low number too.

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
```diff
@@ -285,18 +285,19 @@ coeffRestitutionDD = ddNoRefs coeffRestitution Nothing "coeffRestitution"
...
   sy normalVect
 
 coeffRestitutionDesc :: Sentence
 coeffRestitutionDesc = foldlSent [S "The", getTandS QP.restitutionCoef,
   S "determines the elasticity of a collision between two" +:+. plural rigidBody,
   foldlList Comma List [
-  E (sy QP.restitutionCoef $= 1) +:+ S "results in an elastic collision",
-  E (sy QP.restitutionCoef $< 1) +:+ S "results in an inelastic collision",
-  E (sy QP.restitutionCoef $= 0) +:+ S "results in a totally inelastic collision"]]
+  E (sy QP.restitutionCoef $= dbl 1) +:+ S "results in an elastic collision",
```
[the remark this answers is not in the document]
balacij: Thanks! Fixed

code/drasil-example/Drasil/GamePhysics/DataDefs.hs
```diff
@@ -241,18 +241,18 @@ impulseVDerivSentence3 :: [Sentence]
...
 impulseVDerivEqn1 :: Expr
-impulseVDerivEqn1 =  sy QP.force $= sy QPP.mass * sy QP.acceleration
-                     $= sy QPP.mass * deriv (sy QP.velocity) QP.time
+impulseVDerivEqn1 =  sy QP.force $= sy QPP.mass `mulRe` sy QP.acceleration
+                     $= sy QPP.mass `mulRe` deriv (sy QP.velocity) QP.time
 
 impulseVDerivEqn2 :: Expr
 impulseVDerivEqn2 = defint (eqSymb timeT) (sy time_1) (sy time_2) (sy QP.force) $=
-                    sy QPP.mass * defint (eqSymb QP.velocity) (sy velo_1) (sy velo_2) 1
+                    sy QPP.mass `mulRe` defint (eqSymb QP.velocity) (sy velo_1) (sy velo_2) (dbl 1)
```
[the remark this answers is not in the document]
balacij: Thanks! Fixed

code/drasil-example/Drasil/DblPendulum/IMods.hs
```diff
@@ -25,20 +25,20 @@ iMods = [angularDisplacementIM]
 
 ---
 angularDisplacementIM :: InstanceModel
-angularDisplacementIM = imNoRefs (OthModel angularDisplacementRC) 
-  [qwC lenRod $ UpFrom (Exc, 0)
-  ,qwC initialPendAngle $ UpFrom (Exc, 0)
-  , qwC gravitationalAccel $ UpFrom (Exc, 0)]
-  (qw pendDisplacementAngle) [UpFrom (Exc, 0)]
+angularDisplacementIM = imNoRefs (OthModel angularDisplacementRC)
+  [qwC lenRod $ UpFrom (Exc, dbl 0)
```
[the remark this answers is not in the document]
balacij: Thanks! Fixed them all now

balacij:
Thank you for your review @JacquesCarette . I changed most, if not all, the whole numbered doubles into exacts. I think the majority of them were in spots that made sense for them to be, but I'm not too sure what the standard is with displaying constants in physics works, so I might be wrong. 

Do we want to have any specific guideline for restricting the `exactDbl`s? For example, I think it might be good to have a min/max value an exactDbl could take on. I think that some larger numbers might be better off displayed with engineering/scientific notation instead of as a whole integer.

JacquesCarette (approved):

JacquesCarette:
Having guidelines for what real numbers can be entered as exact is a good idea. Right now, I think I only have rough ideas which I'm not sure I could translate into guidelines!

code/drasil-example/Drasil/GlassBR/DataDefs.hs
smiths: Yes, the 60 is 60 seconds in a minute.  I also don't know the meaning of the 16, but I agree that it is exact.  The source of these GlassBR equations is:

https://github.com/smiths/caseStudies/blob/master/CaseStudies/glass/refs/1BasisForASTME1300AnnealedGlassThicknessSelectionCharts.pdf

There is another paper cited there (Beason and Morgan) that likely explains the 16 better, but I don't think it is worth our time to fret over this, especially when an exact 16 is so likely to be our conclusion.

code/drasil-example/Drasil/SSP/DataDefs.hs
smiths: @balacij what is the difference between `half` and an exact 2?  In the slope stability problem most of the 2s come from needing from taking moments of forces about the centre of a "slice".  They are exactly 2, because they come from mathematical definitions.  (The 3s are exact too.)

code/drasil-example/Drasil/SSP/DataDefs.hs
smiths: Agreed.

code/stable/glassbr/SRS/GlassBR_SRS.tex
smiths: Yes, I agree that is should be exact.  I believe that the (m-1) is assumed and then fitting of experimental data is done to find the values for m.  I could look into this in more detail, but the GlassBR example isn't one that will have its knowledge spread to many future examples.  I don't think it is something we need to worry about.

Issue #2487, opened 2021-05-20
Adding Concepts to PDController concepts


Awurama-N:
Adding "proportional control" and "Derivative Control"

PR #2488, opened 2021-05-20
adding derivative and proportional control to PDController concepts.hs

Awurama-N:
closes Issue #2487


JacquesCarette (changes requested):
This looks good, but there is an inconsistency: one definition ends in a period, the other not. I'm not sure what the convention is, but definitely one of the two is incorrect.

JacquesCarette (approved):

PR #2496, opened 2021-05-20
Alternative style for printing vector magnitude?

balacij:
Builds on #2495 

Just a suggestion, I think the standard way to have vector norms in LaTeX is `\|` wrapped around a vector, so this changes ours from `||`:
image

to:
image


JacquesCarette:
Agreed. And it looks better too.

JacquesCarette (approved):

smiths:
I agree that this looks better, and we should use it for now, but in the future, there may be cases where people wish to use the | v | notation (single bars) to represent a norm.  In some cases this would be a style preference, or the authors may wish to distinguish the vector norm (single bars) and the matrix norm (double bars).

PR #2498, opened 2021-05-21
reuse of phrases with "in the frequency domain"

Awurama-N:
for the phrases with "in the time domain" it seems like what should be the equivalent of "ccFrequencyDomain" doesn't exist yet. I can create a "ccTimeDomain". and then get started on those.
I can also investigate and work on phrases with "of the" and "of a". I could also look at phrases in the other examples afterwards as well?
is that okay?
@JacquesCarette 

contributes to Issue #2397.

### DIFF
--- a/code/drasil-example/Drasil/PDController/Concepts.hs
+++ b/code/drasil-example/Drasil/PDController/Concepts.hs
@@ -131,9 +131,9 @@ concepts = map nw defs
 defs :: [ConceptChunk]
 defs
   = [pidCL, pidC, summingPt, powerPlant, secondOrderSystem, processError,
-     simulationTime, processVariable, setPoint, propGain, derGain,
-     ccFrequencyDomain, ccLaplaceTransform, controlVariable, stepTime,
-     ccAbsTolerance, ccRelTolerance, ccTransferFxn, ccDampingCoeff,
-     ccStiffCoeff]
+     simulationTime, processVariable, setPoint, propGain, derGain, propControl, 
+     derControl, ccFrequencyDomain, ccLaplaceTransform, controlVariable, 
+     stepTime, ccAbsTolerance, ccRelTolerance, ccTransferFxn, 
+     ccDampingCoeff, ccStiffCoeff]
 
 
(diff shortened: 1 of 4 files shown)


JacquesCarette:
Yes please on "in the time domain".  And go ahead with "of the" and "of a"; the urgency of that is not so high, **but** it will give you a chance to read the other examples. While you're doing that, try to spot patterns that should be abstracted. And ask questions about anything that's not clear.

Urgency not so high = it's fine if you only do some of these, but are not systematic about it.

JacquesCarette (approved):

Awurama-N:
Okay great
Thank you

Awurama-N:
hmmm seems like i missed an error:
```
error, called at ./Database/Drasil/ChunkDB.hs:81:24 in drasil-database-0.1.1.0-1H2SDTZem9VDdavf3C6Z5k:Database.Drasil.ChunkDB
```
But when I try to follow the links in my local repo the files can't be found
@JacquesCarette i started this after i updated master but i'm not sure if i missed something?

JacquesCarette:
You didn't give enough of the error message for me to help - there's probably stuff after this that's important.

My guess is that one of the new definitions you have made needs to be added to the right spot in `Body.hs`. The full error will tell you which symbol isn't found.

Ant13731:
@Awurama-N I can help if you're having some trouble here. I went through the same thing when trying to make new ConceptChunks (and NamedIdeas) with the DblPendulum example. You have to map it into a `ChunkDB` function (most likely `symbMap` in the `Body.hs` file of the example). If you have more than one new definition, I suggest first making the new definitions into a list and then putting that list into the `symbMap` function (with the rest of the lists). It'll probably be easier to look at what I did in the `Concepts.hs` file of the `dblpendulum` example than my explanation here (just follow where the `concepts` definition goes and how it is created). Let me know if you need any clarification.

smiths:
Thank you @Ant13731 for offering your help to @Awurama-N.

@JacquesCarette - it looks like we have a really good team working with us!  :smile:

Awurama-N:
Thank you! @Ant13731

JacquesCarette:
Why did you close and delete this branch?  There are ways to fix conflicts without restarting from scratch!

Awurama-N:
Yes sorry I found i was spending a bit too much time trying to figure it out (issues kept popping up after i fix one) so i thought it would be faster to restart for now and ask for a more effecient way for next time

Issue #2481, opened 2021-05-19
New Combinator Documentation


Ant13731:
Continuing #2476. In the Wiki, there is now a small summary for [Combinator Documentation](https://github.com/JacquesCarette/Drasil/wiki/Combinator-Documentation). For level 1 and 2 combinators, I specified some of the differences between similar functions. However, for level 3, I think the naming scheme with `'` and `_` needs to be changed first. I also added a Notes section at the bottom for some possible changes. Please let me know what you think @smiths @JacquesCarette 

Here are some ideas to make the naming scheme for combinators a little more cohesive.
To change the naming scheme, I think that the base function (no extra `'` or `_`) should have the plural case default to `phrase t1 combine plural t2`. This would also match the level 2 combinators.
A single `'` could denote that the plural case is `plural t1 combine phrase t2`.
A double `''` could denote that the combinator takes two functions for the plural case.
A triple `'''` could allow for a combinator that takes four functions for both singular and plural cases.
An appended `T` (or `TT`) could denote that default case will be titleized.
Maybe an `_` could be used for a `plural t1 combine plural t2` case?
I'm not too sure from here, there are a couple of cases that would be missing, even with this naming method.

smiths:
I really like the summary @Ant13731.  It is great to see all of the information in one spot.  I don't have an opinion on the best naming scheme; I defer to the judgement of @JacquesCarette on that.  However, I think you have made his job easier by your summary.  It is also good that you have a specific proposal for the naming.

I think we can likely accept it if there are a few cases that aren't captured by your naming conventions.

JacquesCarette:
Excellent work indeed!

On to the naming convention:
- I like the idea of having the default being singular+plural
- I don't like using `'` for the variants.  What about instead using S and P?  So `fooPS` would be `foo` with `plural` for the first, and `phrase` for the 2nd. We could also have `PP` instead of `''`. For `'''`, I'd prefer going for something like `fooGen` (for 'general'). 
- that would fit with using `T` postfix

The above is my suggestion. Feel free to improve on it!

smiths:
I agree with the move away from the `'` variants.  The `'` doesn't mean anything from a mnemonic point of view.  S and P (for instance) carry information that will let the programmer remember the meaning of the combinator.

Ant13731:
[...]

PR #2500, opened 2021-05-21
Separating and renaming combinators

### LINKED ISSUE #2399: Sentence vs concept-level combinators

Ant13731:
First commit separates `NounPhrase.hs` from `Utils.Drasil`. Second commit separates `Concepts.hs` (level 3 combinators) from `Utils.Drasil`. Every import of Concepts is unqualified but I wasn't sure if I should merge it with the general `Utils.Drasil` module.
Contributes #2476, #2481.
Closes #2399


Ant13731:
Converted to a draft for now because I noticed a few mistakes that should be fixed.

Ant13731:
Okay, it should be good now.

JacquesCarette (approved):
Nicely done!

Issue #2371, opened 2021-05-06
Remove as many `RelationConcept`s from the instance models as possible, by using `ModelKind` more


JacquesCarette:
See https://github.com/JacquesCarette/Drasil/commit/f23bf7b5757b6686d4fa1a6aec371520300ee77b for example. This is part of adding `ModelKind` to `InstanceModel`. 

Basically, every instance model in all examples should be looked at, and either
1. its model should be changed from `OthModel` to something else (one per PR!)
2. an issue should be created regarding why it can't be done.

balacij:
Trying to better keep track of them and their similarities:

Things not ticked off are still `OthModel`s.

- [ ] DblPendulum
    - [ ] GenDefs
        - [x] velocityIXGD - EquationalModel
        - [x] velocityIYGD - EquationalModel
        - [x] accelerationIXGD - EquationalModel
        - [x] accelerationIYGD - EquationalModel
        - [ ] hForceOnPendulumGD
        - [ ] vForceOnPendulumGD
        - [ ] angFrequencyGD
        - [ ] periodPend
   - [ ] IMods
       - [ ] angularDisplacementIM 
- [ ] GamePhysics
    - [x] GenDefs
        - [x] accelGravityGD - EquationalModel
        - [x] impulseGD  - EquationalModel
    - [ ] IMods
        - [ ] transMot
        - [ ] rotMot
        - [ ] col2D
    - [ ] TMods
        - [ ] newtonSL
        - [x] newtonTL - EquationalModel
        - [ ] newtonLUG
        - [ ] newtonSLR 
- [ ] GlassBR
    - [ ] IMods
        - [ ] pbIsSafe
        - [ ] lrIsSafe
    - [x] TMods
        - [x] pbIsSafe - EquationalModel
        - [x] lrIsSafe - EquationalModel
- [x] HGHC
- [ ] NoPCM
    - [ ] GenDefs
        - [ ] rocTempSimp
        - [ ] htFluxWaterFromCoil
    - [ ] IMods
        - [ ] eBalanceOnWtr
        - [ ] heatEInWtr
- [ ] PDController
    - [ ] GenDefs
        - [ ] gdPowerPlant
    - [ ] IMods
        - [ ] imPD
    - [ ] TMods
        - [ ] tmLaplace
        - [ ] tmInvLaplace
        - [ ] tmSOSystem
- [ ] Projectile
    - [ ] GenDefs
        - [ ] rectVelGD
[...]

PR #2510, opened 2021-05-24
Convert rectVel GD into an EquationalModel

### LINKED ISSUE #2493: Projectile's rectVelGD
Presently, we have a data definition and a general definition that both define the same thing, `speed` through their relations (e.g., in the left-hand side of their equations, they both define `speed` at the moment). The general definition, `rectVelGD`, currently relies on OthModel/RelationConcepts. While trying to convert `rectVelGD` into something other than an OthModel (for #2371), I wanted to make it an EquationalModel using a QDefinition, but then it causes a UID conflict between itself and the related data definition, resulting in references that once referred to `rectVelGD` changing into references to the related data definition of speed.

We need to either merge the 2 models into 1, or change the UIDs so they don't conflict. Do we have any preferred way for this? If we want the latter case, I imagine we would want to just change the quantity that  `rectVelGD` defines into something new. If so, how should we define the new `rectVel` quantity that it defines?

Related data definition:
image

Related general definition:
image

balacij:
Closes #2493 
Contributes to #2371 

* Creates 2 new variants of speed as per dsicussion in #2493
* I left the symbol of the vector magnitude variant as just "v", but changed the symbol of the 1D constant acceleration variant as "v^f".

I'm not quite sure if the names/symbols/descriptions are appropriate, so please let me know if they are okay.

Thank you!


code/stable/projectile/SRS/Projectile_SRS.tex
```diff
@@ -243,10 +245,10 @@ \subsubsection{Assumptions}
...
 \item[freeFlight:\phantomsection\label{freeFlight}]{The flight is free; there are no collisions during the trajectory of the projectile. (RefBy: \hyperref[constAccel]{A: constAccel}.)}
 \item[neglectCurv:\phantomsection\label{neglectCurv}]{The distance is small enough that the curvature of the Earth can be neglected. (RefBy: \hyperref[targetXAxis]{A: targetXAxis} and \hyperref[cartSyst]{A: cartSyst}.)}
-\item[timeStartZero:\phantomsection\label{timeStartZero}]{Time starts at zero. (RefBy: \hyperref[GD:velVec]{GD: velVec}, \hyperref[IM:calOfLandingTime]{IM: calOfLandingTime}, \hyperref[GD:rectPos]{GD: rectPos}, \hyperref[GD:rectVel]{GD: rectVel}, and \hyperref[GD:posVec]{GD: posVec}.)}
+\item[timeStartZero:\phantomsection\label{timeStartZero}]{Time starts at zero. (RefBy: \hyperref[GD:velVec]{GD: velVec}, \hyperref[IM:calOfLandingTime]{IM: calOfLandingTime}, \hyperref[GD:rectVel]{GD: rectVel}, \hyperref[GD:rectPos]{GD: rectPos}, and \hyperref[GD:posVec]{GD: posVec}.)}
```
balacij, line 251: Changes in ordering of references is normal, caused by ordering by UIDs instead of labels

code/stable/projectile/SRS/Projectile_SRS.tex
```diff
@@ -350,20 +352,20 @@ \subsubsection{General Definitions}
...
 \paragraph{Detailed derivation of rectilinear velocity:}
 \label{GD:rectVelDeriv}
-Assume we have rectilinear motion of a particle (of negligible size and shape, from \hyperref[pointMass]{A: pointMass}); that is, motion in a straight line. The velocity is $v$ and the acceleration is $a$. The motion in \hyperref[TM:acceleration]{TM: acceleration} is now one-dimensional with a constant acceleration, represented by ${a^{c}}$. The initial velocity (at $t=0$, from \hyperref[timeStartZero]{A: timeStartZero}) is represented by ${v^{\text{i}}}$. From \hyperref[TM:acceleration]{TM: acceleration}, using the above symbols we have:
+Assume we have rectilinear motion of a particle (of negligible size and shape, from \hyperref[pointMass]{A: pointMass}); that is, motion in a straight line. The velocity is ${v^{\text{f}}}$ and the acceleration is $a$. The motion in \hyperref[TM:acceleration]{TM: acceleration} is now one-dimensional with a constant acceleration, represented by ${a^{c}}$. The initial velocity (at $t=0$, from \hyperref[timeStartZero]{A: timeStartZero}) is represented by ${v^{\text{i}}}$. From \hyperref[TM:acceleration]{TM: acceleration}, using the above symbols we have:
```
JacquesCarette, line 355: Here, it should definitely by $v$, which is the velocity at some time $t$, which is not necessarily the final velocity. Somewhere the wrong object is being used.

code/stable/projectile/SRS/Projectile_SRS.tex
```diff
@@ -350,20 +352,20 @@ \subsubsection{General Definitions}
...
 
 \begin{displaymath}
-{a^{c}}=\frac{\,dv}{\,dt}
+{a^{c}}=\frac{\,d{v^{\text{f}}}}{\,dt}
```
JacquesCarette, line 358: Here as well.

code/stable/projectile/SRS/Projectile_SRS.tex
```diff
@@ -350,20 +352,20 @@ \subsubsection{General Definitions}
...
 \end{displaymath}
 Rearranging and integrating, we have:
 
 \begin{displaymath}
-\int_{{v^{\text{i}}}}^{v}{1}\,dv=\int_{0}^{t}{{a^{c}}}\,dt
+\int_{{v^{\text{i}}}}^{{v^{\text{f}}}}{1}\,d{v^{\text{f}}}=\int_{0}^{t}{{a^{c}}}\,dt
```
JacquesCarette, line 363: Here, it's even worse. I think the `t` is free, i.e. not the final time, so both the upper `v` and the one in the `dv` should be just `v`. Even if `t` was `t^f`, the `dv` part should still not have a superscript.

JacquesCarette (changes requested):

code/stable/projectile/SRS/Projectile_SRS.tex
balacij: To fix this, I had to replace the "v" symbol left for the vector speed variant. I just changed it to a bold faced v with a superscript "speed". Is there any preference for this?

code/drasil-example/Drasil/Projectile/Concepts.hs
```diff
@@ -31,14 +32,17 @@ projMotion = compoundNC projectile motion
...
 
-launcher, projectile, target :: ConceptChunk
+launcher, projectile, target, velVecSpeed, speed1DAcc :: ConceptChunk
 launcher   = dcc "launcher"   (nounPhraseSP "launcher")  ("where the projectile is launched from " ++
                                                           "and the device that does the launching")
 projectile = dcc "projectile" (nounPhraseSP "projectile") "the object to be launched at the target"
 target     = dcc "target"     (nounPhraseSP "target")     "where the projectile should be launched to"
 
+velVecSpeed = dccWDS "velVecSpeed" (cn' "vector speed") (S "magnitude" `S.the_ofThe` phrase velocity +:+ phrase vector)
+speed1DAcc  = dccWDS "speed1DAcc"  (cn' "final speed") (S "1D speed under" +:+ phrase constant +:+ phrase acceleration)
```
JacquesCarette, line 44: This here is confusing. Why is it called `speed1DAcc`, termed "1D speed under constant acceleration" but labelled "final speed"?

code/drasil-example/Drasil/Projectile/Unitals.hs
```diff
@@ -1,16 +1,19 @@
...
 import Data.Drasil.SI_Units (radian, metre, second)
 import Data.Drasil.Units.Physics (velU)
 
-import qualified Drasil.Projectile.Concepts as C (flightDur, offset,
-  flightDur, landPos, launAngle, launSpeed, offset, targPos)
+import qualified Drasil.Projectile.Concepts as C
+
+velVecSpeed, speed1DAcc :: UnitalChunk
+velVecSpeed = uc C.velVecSpeed (sup (vec lV) $ Label "speed") velU
+speed1DAcc  = uc C.speed1DAcc  (sup lV $ Label "f") velU
```
JacquesCarette, line 16: The f gets attached here - but why? I'm no longer sure what this quantity even refers to! (See previous comments).

code/drasil-example/Drasil/Projectile/Unitals.hs
```diff
@@ -1,16 +1,19 @@
...
 import Data.Drasil.Constraints (gtZeroConstr)
 import Data.Drasil.SI_Units (radian, metre, second)
 import Data.Drasil.Units.Physics (velU)
 
-import qualified Drasil.Projectile.Concepts as C (flightDur, offset,
-  flightDur, landPos, launAngle, launSpeed, offset, targPos)
+import qualified Drasil.Projectile.Concepts as C
+
+velVecSpeed, speed1DAcc :: UnitalChunk
+velVecSpeed = uc C.velVecSpeed (sup (vec lV) $ Label "speed") velU
```
JacquesCarette, line 15: And `v^speed` ??

JacquesCarette (changes requested):
This is still not right. There is some conceptual problem here.  The fix really has to come from capturing the correct idea(s).

balacij:
@JacquesCarette Thank you for your review. I'm not quite sure what we should name these ideas, or if we have any preference in the symbols.

For the purpose of the ticket #2493 (and this PR), we need to give them different UIDs (and therefore symbols as well). We have the option of using 2 or 3 different symbols. I chose to go down the route of 3 different symbols so as to completely differentiate between the 3 ideas.

The related unital chunks are as follows:
| variable | symbol | label | description | related GD/DD | related QDef's equation |  
|------------|-----------|--------|----------------|----------------------|-------------|
| speed | `v` | speed | magnitude of the velocity vector | <n/a> | <n/a> |
| speed1DAcc | `v^f` | final speed | 1D velocity under constant acceleration | [GD: rectVel](https://jacquescarette.github.io/Drasil/examples/Projectile/srs/Projectile_SRS.html#GD:rectVel), replacing the current `v` | `v^f = v^i + a^c * t` |
| velVecSpeed | `bold(v)^speed` | vector speed | magnitude of the velocity vector | [DD: vecMag](https://jacquescarette.github.io/Drasil/examples/Projectile/srs/Projectile_SRS.html#DD:vecMag), replacing the current `v` | `\|\| bold(v) \|\|` |

The `bold(v)` is the related velocity vector.

The core library's speed (`v`) is used when deriving `speed1DAcc`, and then I had it note that it was replaced by the symbol for `speed1DAcc`. This is why I differentiated between `speed` and `velVecSpeed` -- so as not to cause any accidental/unwanted relations between `DD: vecMag` and `GD: rectVel`.

I'm now realizing that what I had wrote as `speed1DAcc` would probably be more appropriate as `constAcc1DVel`, `finalVel`, or maybe even `finalVelConstAcc`.

Do we have any sort of preference in what we should symbols and names we should be giving these concepts?

Alternatively, we can use only 2 unital chunks (e.g., speed and `speed1DAcc` [or under a new name]), perhaps leaving `DD: vecMag` as defining the core library's speed (`v`), and then just creating a new symbol for `speed1DAcc`? My only concern with this way is that we would be deriving the equation for `GD: rectVel` using the symbol for `DD: vecMag`'s defined variable until the last step.

smiths:
@balacij I'm not 100% sure of what is being proposed, but I know there are things I don't like about it.  :smile:  

1. I really don't like `bold(v)^speed`.  This notation is going to confuse any reader of the document.  I think you might have introduced an artificial difference between speed and vector speed?  We call the magnitude of the velocity vector the speed.  I don't think there is any thing more to it.  Why do we need a UID for speed and one for velVecSpeed?  Can we just use v for speed?  (If we need to change the document itself in some way, I'm open to making changes.  I think these discussions are often going to highlight problems with the documentation, since human beings wrote the documents, which means some ambiguity very likely slipped in.)
2. v^f is problematic in `speed1DAcc`.  We might have to think about another UID, since I think there are two concepts that are being mixed here.  In the original equation v = v^i + a^c * t the type of v is a function.  v: R -> R.  You give v a time and it returns a velocity.  As time changes the velocity changes.  When we solve a specific problem though we may know the final time, let us call it t^f.  In that case v^f = v^i + a^c*t^f.  In this case the type of v^f is R.  You can see an example of this in the projectile lesson Jupyter lab book: https://github.com/smiths/caseStudies/blob/master/CaseStudies/projectile/projectileLesson/ProjectileLesson.ipynb

It is natural in writing to use the same symbol v for the magnitude of the velocity vector and for a 1D equation for velocity under constant acceleration.  In 1D the direction of the velocity is known (except for the sign).

I thought about whether we should replace the DD defining speed with a generic (math knowledge) DD explaining how to decompose a vector into its components in a 2D Cartesian coordinate system?  That is what we use the magnitude of the vector for.  We then wouldn't have to mention the concept of speed, or need to use v to mean speed.  You could then use the symbol v for `speed1DAcc`.

I don't like this solution though because speed is one of our inputs.  We don't enter the magnitude of the velocity vector, we enter speed.  Speed is the concept that will be most easily accessible to the reader/user.

I thought about another solution of using the symbol $s$ to mean speed.  If our goal is to have different symbols for different concepts, this makes sense.  However, we rely on the implicit connection between the symbols.  We use v for speed and v for speed in 1D under constant acceleration and then we relate the speed to the "speed in 1D under constant acceleration."  Therefore, another option is to do what we do now and use the symbol v for both.

I have a suggestion that I don't think will be much work with Drasil.  @balacij, can you change the symbol for speed to $s$.  We should also change the symbols v^i in DD:speedIY and DD:speedIX.  They should just have an $s$ in place of it.  I'd like to look at the documentation with the new symbol ($s$) and see how it fits together.  My current thinking is that this is the best solution, but I'd like to see how it would change stable before proceeding.

@JacquesCarette might have another idea on what we should do.

balacij:
> @balacij I'm not 100% sure of what is being proposed, but I know there are things I don't like about it. 

That's perfect. I didn't quite know how we wanted to capture this knowledge, but I think your messaged really helped me understand it. Thank you!

> 1. I really don't like `bold(v)^speed`.  This notation is going to confuse any reader of the document.  I think you might have introduced an artificial difference between speed and vector speed?  We call the magnitude of the velocity vector the speed.  I don't think there is any thing more to it.  Why do we need a UID for speed and one for velVecSpeed?  Can we just use v for speed?  (If we need to change the document itself in some way, I'm open to making changes.  I think these discussions are often going to highlight problems with the documentation, since human beings wrote the documents, which means some ambiguity very likely slipped in.)

Indeed, I did. If we remove the "velVecSpeed" and just use speed (and it's symbol) there, it's fine.

> 2. v^f is problematic in `speed1DAcc`.  We might have to think about another UID, since I think there are two concepts that are being mixed here.  In the original equation v = v^i + a^c * t the type of v is a function.  v: R -> R.  You give v a time and it returns a velocity.  As time changes the velocity changes.  When we solve a specific problem though we may know the final time, let us call it t^f.  In that case v^f = v^i + a^c*t^f.  In this case the type of v^f is R.  You can see an example of this in the projectile lesson Jupyter lab book: https://github.com/smiths/caseStudies/blob/master/CaseStudies/projectile/projectileLesson/ProjectileLesson.ipynb

This might be my point of confusion. If this is the case, would we want to change it's left-hand side to "v(t)" as a "functional model"? In which case, we would close this PR and wait for the functional models to be built first.

> I thought about another solution of using the symbol $s$ to mean speed. If our goal is to have different symbols for different concepts, this makes sense. However, we rely on the implicit connection between the symbols. We use v for speed and v for speed in 1D under constant acceleration and then we relate the speed to the "speed in 1D under constant acceleration." Therefore, another option is to do what we do now and use the symbol v for both.
> 
> I have a suggestion that I don't think will be much work with Drasil. @balacij, can you change the symbol for speed to $s$. We should also change the symbols v^i in DD:speedIY and DD:speedIX. They should just have an $s$ in place of it. I'd like to look at the documentation with the new symbol ($s$) and see how it fits together. My current thinking is that this is the best solution, but I'd like to see how it would change stable before proceeding.
> 
> @JacquesCarette might have another idea on what we should do.

We could try this too, but I don't think we're able to define things using the same symbol but different UIDs. There's currently code that errors out when it finds these instances. Should we try this with $s_y$ and $s_x$ for speedIY and speedIX?

Thank you.

balacij:
Thank you, I just implemented them with 1 exception -- for `projSpeed`, we have to use a symbol other than the one already registered for speed (otherwise the symbol table generator throws an error about different UIDs sharing the same symbol). Since it seems we want this to be dependent on time, I changed it's symbol to be `v(t)`. Is this change okay?

JacquesCarette (approved):

JacquesCarette:
Aha, so we do have a check that in a single document, symbols must be unique! I wasn't sure, and may have said the opposite in a conversation with @smiths . 

The `v(t)` is a minor hack, that we'll need to fix later. Otherwise, the rest is looking good.

balacij:
Perfect, thank you!

PR #2512, opened 2021-05-25
 Reuse of phrases with "in the frequency domain"

Awurama-N:
reuse of phrases & error fixes

contributes to Issue #2397.

### DIFF
--- a/code/stable/pdController/SRS/PDController_SRS.tex
+++ b/code/stable/pdController/SRS/PDController_SRS.tex
@@ -65,15 +65,15 @@ $-∞$ & Negative Infinity & --
 \\
 $AbsTol$ & Absolute Tolerance & --
 \\
-${C_{\text{s}}}$ & Control Variable in the frequency domain & --
+${C_{\text{s}}}$ & Control Variable in the Frequency Domain & --
 \\
 $c$ & Damping coefficient of the spring & --
 \\
 ${c_{\text{t}}}$ & Control Variable in the time domain & --
 \\
-${D_{\text{s}}}$ & Derivative control in the frequency domain & --
+${D_{\text{s}}}$ & Derivative Control in the Frequency Domain & --
 \\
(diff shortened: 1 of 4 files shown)


JacquesCarette (commented):
Hmm, something's odd - too many words seem capitalized now. Probably due to `inThe` use - @Ant13731 since you're now the expert on those combinators, could you take a look please?

Ant13731:
Converting to a draft just to compare with stable from master branch. I think the capitalization is from the definitions, and there's a list of terminology that doesn't automatically titleize the definitions.

code/stable/pdController/SRS/PDController_SRS.tex
```diff
@@ -247,13 +247,15 @@ \subsubsection{Terminology and Definitions}
 \item{Set-Point: The desired value that the control system must reach. This also knows as the reference variable.}
 \item{Proportional Gain: Gain constant of the proportional controller.}
 \item{Derivative Gain: Gain constant of the derivative controller.}
-\item{Frequency Domain: The analysis of mathematical functions in terms of frequency, instead of time.}
+\item{Proportional control: A linear feedback control system where correction is applied to the controlled variable which is proportional to the difference between desired and measured values.}
+\item{Derivative control: Monitors the rate of change of the error signal and contributes a component of the output signal (proportional to a derivative of the error signal).}
```
Ant13731, line 251: Making the other section lowercase means making these definitions look out of place in this list. I'll see if there's a way to just titleize them here.

Ant13731: I can't seem to find an explicit place where the capitalization rules are defined for this list, but I think Drasil is only capitalizing the first word of each definition on its own. The rest of the capitalization comes from the functions that hold the terminology definitions (which have been hard-coded to be capitalized).

JacquesCarette (approved):

Issue #2499, opened 2021-05-21
Using More Concept Combinators


Ant13731:
As a follow up for #2399, this issue will be a list to keep track of which examples use higher level combinators (and compound ideas) and which ones still need to be changed (see the [Wiki](https://github.com/JacquesCarette/Drasil/wiki/Combinator-Documentation) for more info about combinators).

- [x] DblPendulum
- [x] GamePhysics 
     - [x] Update Combinator documentation
- [x] GlassBR 
- [x] HGHC (doesn't seem to be any for now)
- [x] NoPCM 
- [x] PDController 
- [x] Projectile
- [x] SSP
- [x] SWHS

PR #2514, opened 2021-05-25
Adding concept-level combinators to GlassBR example

Ant13731:
Adding concept-level combinators to GlassBR example, introduced some new combinators to Utils.Drasil.Concepts used in the GlassBR example (this will likely have some merge conflicts with the GamePhys PR, but it should quick to fix when it's ready).
Contributes #2499


JacquesCarette (approved):

JacquesCarette:
As expected, some conflicts. But otherwise nice.

PR #2516, opened 2021-05-25
Splitting up Haddock documentation

### LINKED ISSUE #2294: Haddock docs missing for some files

balacij:
Closes #2294 

As discussed previously, we would have 2 variants of Haddock documents. One with all modules fully exposed, and one variant as it is right now.

These changes edit the scripts just slightly, but it adds a bit of overhead. Unfortunately, it seems that Haddock requires a whole recompilation (!) in order to change the settings of the haddock documents last built. So, this adds ~3-5 min of overhead to our build times. Luckily, I think we can remove a bit of time elsewhere by using pre-compiled binaries of graphmod/dotgen.

Demo: https://balacij.github.io/Drasil/

Change to textual links:
image

Normal docs: https://balacij.github.io/Drasil/docs/index.html
Full docs (fully exposed modules): https://balacij.github.io/Drasil/docs/full/index.html

### DIFF
--- a/code/website/index.html
+++ b/code/website/index.html
@@ -215,7 +215,7 @@ $endfor$
 $if(docsUrl)-$
   <h3 id="haskDoc">Haddock Documentation</h3>
   <p>
-    The current <a href="$docsUrl$">Haddock documentation</a> for the Drasil framework. $if(fullDocsUrl)-$A variant with fully exposed modules is also available <a href="$fullDocsUrl$">here</a>.$-endif$
+    The current <a href="$docsUrl$">Haddock documentation</a> for the Drasil framework. $if(fullDocsUrl)-$A variant with <a href="$fullDocsUrl$">fully exposed modules</a> is also available.$-endif$
   </p>
 $-endif$
 
--- a/code/website/index.html
+++ b/code/website/index.html
@@ -215,7 +215,7 @@ $endfor$
 $if(docsUrl)-$
   <h3 id="haskDoc">Haddock Documentation</h3>
   <p>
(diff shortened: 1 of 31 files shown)


code/website/index.html
```diff
@@ -215,7 +215,7 @@ <h3 id="caseStudies">Case Studies</h3>
 $if(docsUrl)-$
   <h3 id="haskDoc">Haddock Documentation</h3>
   <p>
-    The current <a href="$docsUrl$">Haddock documentation</a> for the Drasil framework.
+    The current <a href="$docsUrl$">Haddock documentation</a> for the Drasil framework. $if(fullDocsUrl)-$A variant with fully exposed modules is also available <a href="$fullDocsUrl$">here</a>.$-endif$
```
JacquesCarette, line 218: Please make "fully exposed modules" the link words. "here" links are an anti-pattern.

JacquesCarette (changes requested):
Also, the docs have a bunch of top-level `Paths_*` files - could these somehow not be indexed?

balacij:
For the `Paths_*` modules, it's a pretty weird issue. The hpack docs just mention that it's "not harmful most of the time and it's helpful in some cases", but doesn't really seem to discuss where it's helpful/harmful. It also seems pretty often that people just omit it -- quite a few discussions about removing it from their GitHub issue tracker. So I just removed it as well, and it didn't cause any visible issues for me (and it also fixed the issue of haddock building information for the `Paths_*` modules).

JacquesCarette (approved):

PR #2518, opened 2021-05-26
Add debugging swift step to CI

balacij:
As per https://github.com/JacquesCarette/Drasil/commit/58728e973538388fd98bf8b0e3b79b4cff7aa4f5#commitcomment-51335621

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -21,6 +21,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: "Debug: Swift on CI"
+        run: |
+          echo "print(\"Hello World\")" > ~/test.swift
+          swiftc ~/test.swift -o test_swift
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Ah, I didn't see that the original thing was just a stub for debugging, rather than full-fledged testing of our generated swift code. The latter is really the intent. We should be doing test runs of the code that we generate.  [Though perhaps in a weekly CI run, rather than for all? We don't want to blow up the time to something too crazy]

balacij:
Ah! I agree, I would like to see that as well.

I don't think any of our current examples have long execution times, and we already compile them via a `make code` step in CI, so we may not actually slow down our CI builds too much if we curate smaller input sets for them. Should we give it a go and see what the impact is? Then we can make a decision on whether to run with each or weekly I think.

JacquesCarette:
Ah indeed. I was afraid of the compilation time - but if we already grab binary distribs for the compilers, indeed, this should have very little impact.

PR #2520, opened 2021-05-26
Adding PDController to Website

### LINKED ISSUE #2515: Website case studies table missing PDController example
image

Ant13731:
Added PDController to website under the index.html. @balacij does this look right?
Closes #2515

### DIFF
--- a/code/website/index.html
+++ b/code/website/index.html
@@ -69,6 +69,16 @@ td {
     <td>D</td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
I actually don't know if this is going to work! It's not bad change per se, but I'm not sure if it's this simple. Let's try it and see, I guess.  I'll remove the "closes" from the PR comment though, as we'll need for someone to check by hand that it does.

Issue #2525, opened 2021-05-27
How to see full documentation locally


Ant13731:
Hello, I'm having trouble finding the full docs when running `make docs` locally. Is it in a different file path from the normal documentation?

balacij:
Ah, I had the `make docs` changed slightly. It builds the `docs` folder, but embeds the "full" version in a `full` folder within the `docs` folder.

balacij:
Might be worth noting this somewhere now that I think about it, any ideas where we can?

Thank you!

smiths:
I like the idea of adding the note to the end of the script.  Would there also be value in updating the Contributor's Guide?  There is a section on make (https://github.com/JacquesCarette/Drasil/wiki/Contributor's-Guide#make-and-drasil-packages), although the make section doesn't currently discuss `make doc`.

Ant13731:
Mine doesn't seem to generate a full folder. Also I think mentioning in the Quick Start and Readme files would be good too.

balacij:
@smiths I think that's a great idea! I can write a note about it there too.

@Ant13731 Did you make sure to `git pull origin master` to get upstream changes? Were there any errors in the logs?

Ant13731:
Oh sorry, I thought my master was up to date. Let me see if it works now.

Ant13731:
Okay, I found it! I think I was looking in the wrong place (its under `code/docs/full` but I was looking under `code/.stack-work/install/48ba28e6/doc`). Thanks for your help @balacij!

balacij:
I added a little blurb about the Haddock documentation, a note at the end of the `make docs`, and updated the related README area.

Let me know if this is good and we can close this ticket!

balacij:
@Ant13731 Are we okay to close this now?

Ant13731:
Yep!

PR #2527, opened 2021-05-27
Pass GHC options to make_docs.sh, move flags to stack.yaml, and outpu…

balacij:
…t location of generated docs

Note: `--interleaved-output` can be omitted since it's enabled by default since Stack 1.8

Contributes to: #2525

### DIFF
--- a/code/stack.yaml
+++ b/code/stack.yaml
@@ -58,6 +58,11 @@ extra-deps:
 - unicode-names-3.2.0.0@sha256:fef7241d93170e26265e84553090253a5e8ee207645d47cf271816f5834d07d2
 - unicode-properties-3.2.0.0@sha256:239766a6ac4322329353f6b9cd546024fd8c5f0235c8e32b3cba2d6bd245a699
 
+# Used only for a basic set of flags for haddock
+build:
+  haddock-deps: false
+  haddock-hyperlink-source: false
+
 # Override default flag values for local packages and extra-deps
 flags: {}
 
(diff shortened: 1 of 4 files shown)


balacij:
If we can figure out an optimal thread count for github actions, we can likely speed up builds _a bit_. Passing the GHC flags also passes our current GHCTHREADS option, which will help with this (assuming we get gains from a >2 thread count from the Actions host server).

code/stack.yaml
```diff
@@ -58,6 +58,11 @@ extra-deps:
 - unicode-names-3.2.0.0@sha256:fef7241d93170e26265e84553090253a5e8ee207645d47cf271816f5834d07d2
 - unicode-properties-3.2.0.0@sha256:239766a6ac4322329353f6b9cd546024fd8c5f0235c8e32b3cba2d6bd245a699
 
+# Used only for a basic set of flags for haddock
+build:
+  haddock-deps: false
+  haddock-hyperlink-source: false
```
JacquesCarette, line 64: Why?

JacquesCarette (approved):
I think most people have found that -j2 speeds things up, and that anything higher doesn't really. It might be worth experiment with.

code/stack.yaml
balacij: These were the original flags used for generating the Haddock docs. `haddock-deps` is a switch for building haddocks for dependencies, and `haddock-hyperlink-source` is for creating the links to the source code. I think we can re-enable that, we'd just need to tweak the settings a little bit for it to go to github.

JacquesCarette: Right, I was only asking about the hyperlinks. It would be nice to have that.

balacij: I'm sure we can figure it out. I'll make a ticket.

Issue #2440, opened 2021-05-12
New ModelKind: `EquationalRealm`


balacij:
> I agree with the reasoning.
>
> We could indeed use an `NEList` of expressions (for the right hand sides) in `EquationalModels` to collapse the two together. But, let's not do that now. Using something from a list requires a choice, but in the 1 case, there is none - and from a design perspective, that's quite different.
>
>I'm thinking `EquationalRealm`. We use `Realm` in a paper to mean a theory with multiple interfaces.  Let's start with that?

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/issues/2436#issuecomment-839853035_

balacij:
I started some preliminary work in the [equationalRealmMK](https://github.com/JacquesCarette/Drasil/compare/equationalRealmMK) branch, but very quickly running into some troubles/big questions.

Assuming we wanted an `EquationalRealm [QDefinition]` data constructor for ModelKinds, ...

1. Would we want to allow it to be instantiated for anything other than TheoryModels? If not, should we potentially have separate ModelKinds ADTs for just TheoryModels?
2. Should each QDefinition in the EquationalRealm be forced to carry the same data? In particular, should each QDefinition in the EquationalRealm have the same UID?
4. If we don't necessarily want all the of QDefinitions to be equivalent (excluding Exprs and definitional Sentences), then when we [grab some information from something that carries a ModelKind](https://github.com/JacquesCarette/Drasil/blob/2b43623ad864ab051db278fd8421629713d22866/code/drasil-theory/Theory/Drasil/ModelKinds.hs#L35), which QDefinition should we grab information from when it's an EquationalRealm? Right now, I set it to be the first element (working on the assumption that each internal QDefinition carries the same information [even though I know they don't, they each, at least, carry different definition Sentences and `relat`s]

However, these questions lead me to think we would actually want some slightly different structure: `EquationalRealm' QuantityDict [(Sentence, Expr)]` where we can use an EquationalRealm to generate QDefinitions (since a QDefinition is currently made up of a QuantityDict, defn :: Sentence, and an Expr [though, technically, it also has a ConceptDomain that isn't used at the moment too])

JacquesCarette:
`NEList`, not `[]`.  We should have it for both `InstanceModel` and `TheoryModel`. At least for now.

[...]

PR #2532, opened 2021-05-27
Basic EquationalRealms

### LINKED ISSUE #2436: Converting "Newton's law of universal gravitation" Theory Model
Related to #2371 

While converting the following TheoryModel into "not an OthModel":

https://github.com/JacquesCarette/Drasil/blob/acf6d8dbd9632e14954e8e5027768edfd8ceb7d8/code/drasil-example/Drasil/GamePhysics/TMods.hs#L45-L60

and in the context of https://github.com/JacquesCarette/Drasil/pull/2426#issuecomment-838992397 , it seems inappropriate to naively convert the following TheoryModel into an EquationalModel because it contains 2 expressions for which it can use to calculate the `force`. However, if I'm not mistaken, if we want an EquationalModel to mean "something is calculated by this 1 formula", then I think we need to create a new kind of ModelKind to accommodate this.

Assuming I'm correct, would we potentially want some sort of "AnyEquationModel" (weird name, but we can probably figure something out that's better) that contains a list of QDefinitions, for which any can be used to calculate the thing in question?

**EDIT**: Alternatively, we could even just make "EquationalModels" just allow for this list of QDefinitions describe above, and create a smart constructor for all the existing models that currently have just 1 single related QDefinition.

balacij:
Contributes to #2440 
Closes #2436


JacquesCarette:
I'm going to review this "later", as it will need quite a bit of effort.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
```diff
@@ -1,42 +1,63 @@
...
+import Control.Lens ((^.), to, lens, set, makeLenses, makeLensesFor,
+  Getter, Lens', Setter')
 import Data.Maybe (mapMaybe)
+import qualified Data.List.NonEmpty as NE
+
+import Language.Drasil (($=), sy, Expr, RelationConcept,
+  NamedIdea(..), HasUID(..), ExprRelat(..), ConceptDomain(..), Definition(..),
+  Idea(..), DefiningExpr(..), UID, Sentence, QDefinition, QuantityDict)
+
+data RealmVariant = RV {
```
JacquesCarette, line 14: Can you add some documentation as to what is meant by this? I'm not even sure I know what the `Expr` here is supposed to represent!

code/drasil-theory/Theory/Drasil/ModelKinds.hs
```diff
@@ -1,42 +1,63 @@
...
+  _desc :: Sentence,
+  _expr :: Expr
+}
+
+makeLensesFor [("_expr", "expr")] ''RealmVariant
 
 data ModelKinds = EquationalModel QDefinition
+                | EquationalRealm QuantityDict [UID] (NE.NonEmpty RealmVariant)
```
JacquesCarette, line 22: Are the lengths of the list and the nonempty supposed to be related?

code/drasil-theory/Theory/Drasil/ModelKinds.hs
```diff
@@ -1,42 +1,63 @@
...
-setMk (EquationalModel q) f _ x = EquationalModel $ set f x q
-setMk (DEModel q)         _ g x = DEModel $ set g x q
-setMk (OthModel q)        _ g x = OthModel $ set g x q
-
-lensMk :: forall a. Lens' QDefinition a -> Lens' RelationConcept a -> Lens' ModelKinds a
-lensMk lq lr = lens g s
+instance HasUID        ModelKinds where uid      = lensMk uid uid uid
+instance NamedIdea     ModelKinds where term     = lensMk term term term
+instance Idea          ModelKinds where getA     = elimMk (to getA) (to getA) (to getA)
+instance Definition    ModelKinds where defn     = lensMk defn (error "ambiguous definition (defn) in EquationalRealm") defn
```
JacquesCarette, line 31: These errors are just as bad as `undefined`. If the information is not available, or not uniquely available, in all model kinds, then it should be considered available at all.

If it's crucial that the information is available (which could indeed be the case), then this means that the current design is flawed and needs to be revised.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
```diff
@@ -1,42 +1,63 @@
...
+instance ExprRelat     ModelKinds where 
+  relat (EquationalRealm q _ rs) = sy q $= foldr1 ($=) (NE.map (^. expr) rs)
+  relat (EquationalModel q)      = relat q
+  relat (DEModel q)              = relat q
+  relat (OthModel q)             = relat q
+-- TODO: implement MayHaveUnit for ModelKinds once we've sufficiently removed OthModels (else we'd be breaking too much of `stable`)
+
+-- | Retrieve internal data from QDefinitions/RelationConcepts
+elimMk :: Getter QDefinition a -> Getter QuantityDict a -> Getter RelationConcept a -> ModelKinds -> a
+elimMk l _ _ (EquationalModel q)     = q ^. l
```
JacquesCarette, line 43: So what this `elimMk` seems to say is that a "better" representation for a `ModelKind` would be a product of a `QDefinition` and "the rest of the information", labelled with the kind of model.  So `EquationalModel` would have no payload, only `EquationalRealm` would.

JacquesCarette (changes requested):
This is much more a "design review" than 'request changes'.

code/drasil-example/Drasil/GamePhysics/TMods.hs
```diff
@@ -44,22 +46,21 @@ newtonTLNote = foldlSent [(S "Every action has an equal and opposite reaction" !
 
 -- T3 : Newton's law of universal gravitation --
 
+-- FIXME: Missing ConceptDomain!
+newtonLUGModel :: ModelKinds
```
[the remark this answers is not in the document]
balacij: I wanted this one to be the prototype of sorts for the EquationalRealm since it was the original model that started this work. I think it makes sense to work with at least one model while building the EquationalRealms so that we know what to expect them to look like.

code/drasil-theory/Theory/Drasil/GenDefn.hs
```diff
@@ -18,26 +18,27 @@ data GenDefn = GD { _mk :: ModelKinds
...
-instance Definition         GenDefn where defn = lensMk defn defn
-instance ConceptDomain      GenDefn where cdom = elimMk (to cdom) (to cdom) . view mk
-instance ExprRelat          GenDefn where relat = elimMk (to relat) (to relat) . view mk
-instance HasDerivation      GenDefn where derivations = deri
+          s gd_ x = set mk (setMk (gd_ ^. mk) lq lqd lr x) gd_
+
+instance HasUID             GenDefn where uid           = lensMk uid uid uid
+instance NamedIdea          GenDefn where term          = lensMk term term term
+instance Idea               GenDefn where getA          = getA . (^. mk)
+instance Definition         GenDefn where defn          = lensMk defn undefined defn
```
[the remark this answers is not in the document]
balacij: I agree! The `undefined`s are definitely very odd.

Additionally, while working on it, I _really_ wanted to take the `lensMk`, `elimMk`, and `setMk`'s 3 function parameters (acting on RelationConcepts, QuantityDicts, and QDefinitions) and merge them into 1 param using typeclasses, but I couldn't get a sufficient "common denominator" between the various models.

The "definition" of an EquationalRealm is interesting. I think we want it be undefined until the EquationalRealm is observed and converted into an EquationalModel using a specific RealmVariant with the attached QuantityDict.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: Yes, I should definitely do that. I'll merge in master and add some documentation for this. I'll let you know when I've done so.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: No, the `[UID]` is meant to be the concept domain of the variants. It's related, I think, in that it would be the union of the concept domains of the various RealmVariants. I think we discussed that we wanted it to be for the whole model instead of the individual RealmVariant, but thinking about it again from your question... it might be the case that we want both, with the top-level CD of the Realm being calculated using the CDs of the individual RealmVariants.

It also looks like the current instance of the ConceptDomain isn't aligned with this. I will fix this.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: Would we be able to form a QDefinition for an EquationalRealm if we want EquationalRealms to represent "more than one way to define a quantity"? I think the "right-hand side" of the QDefinition would be ambiguous for an EquationalRealm, unless we wanted to have the Realm's QDefinition be of the form `x $= y $= z $= ...`.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: I agree. I'm not much of a fan of the errors that I had written for it to be honest.

If we had an InstanceModel/GenDefn/TheoryModel with an EquationalRealm, what would we want to say the;
* "Definition" is? Perhaps we just need a custom "Definition" sentence for an EquationalRealm? I think this sounds about right actually -- I don't know why I didn't think about this before.
* "DefiningExpr" is? Perhaps we need DefiningExpr to be returning a "Maybe Expr" instead of a `Lens c Expr` -- this would be following suit in the current ExprRelat (in that ExprRelat's `relat` function just returns a plain Relation), but against the redesign proposed of `ExprRelat`

Perhaps we should be changing the definition of EquationalRealms (`EquationalRealm QuantityDict (NE.NonEmpty RealmVariant)`) into `EquationalRealm QuantRealm` where a `QuantRealm` contains a `QuantityDict` and `(NE.NonEmpty RealmVariant)`. Then we'd be probably be able to avoid these errors and clean up the code a bit to avoid `undefined`. If the above 2 "Definition" and "DefiningExpr" are actually things we should define instead of creating errors, then I think this would be a decent solution.

balacij: I actually really liked the idea of having a separate Realm carrier, I think it makes a lot of sense in this context. So, I made a prototype in e45bc854231efe71a8a7e93f45f9376ace92a8b4 (though, it will be easier to review in this branch, since it contains this commit).

It allowed us to define the "Definition" instance, and I think that definition would be good to have as a relational statement between each equation, which may not be something that we want to procedurally generate via the definitions of the underlying variants.

I imagine that if we have only 1 RealmVariant in a Realm, then the DefiningExpr would be that 1 RealmVariant's Expr. However, if we have more than 1, it will be ambiguous. If we change the type of `defnExpr`/DefiningExpr to return a `Maybe (Lens c Expr)`/`Maybe Expr`, then I think we can define something for Realms.

Also, I named the bundle "Realm" but it's probably not the best name for it. Maybe QDefRealm?

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: I added docs with the proto-Realms as described below, but the meaning of things will still be the same.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
JacquesCarette: See the discussion on that commit.  This is getting a lot closer.

However, I definitely don't want to use `Maybe` here. Basically one should not be asking for 'the' definition of something that doesn't have it. So we need to understand why we're even asking the question. So we'll need to see who is using those features of IM/TM/etc and in what way.

I think that in most cases, we'll be able to make sense of what it's trying to do, which will feed back into the design of this feature.

code/drasil-theory/Theory/Drasil/ModelKinds.hs
JacquesCarette: Correct, the 'right hand side' will not make sense. 'a' right hand side will. I think a `QDefinition` needs to be precise. So indeed, there would be no such `Getter`. I consider that to be a good thing, as it will force us to be precise about what we mean. It would be interesting to see who uses this getter.  Maybe it should return an `NonEmpty` ?

code/drasil-theory/Theory/Drasil/ModelKinds.hs
balacij: I see. I will try to check usage of defnExpr (of DefiningExpr) and relat (of ExprRelat), I think this also relates to the below discussion about `elimMk` too. I'll post when I've understood a bit more about the existing usage.

balacij: As of right now, `defnExpr` is almost exclusively used on QDefinitions and DataDefinitions. So, it looks like I didn't need (nor should I have) added the extra instance for `DefiningExpr` for IM/TM/GD. `relat` is primarily used for cosmetics and quickly grabbing "x $= ..." Exprs. However, if we intend to replace fully replace `relat` with `defnExpr`, it will require a bit more work since `defnExpr` is not intended for this (and incompatible, because it currently is geared towards uniquely identifying exprs).

If we return a NonEmpty list of Exprs instead of a single Expr, for all the instances that use `relat`, we can just intercalate `$=`, prepending a symbol as well, but then it might not work nicely with the existing usage for DataDefinitions -- for example, defining unique expressions for constants. I think we'll need to have a both in some capacity.

JacquesCarette: Right - this is all very valuable design information that shouldn't be lost as discussion on the PR. Probably it should go in the `Classes` file that defines the various classes (and from there, `relat` and `defnExpr`) as the reasoning behind them.

Once we design what each of them means, then it will be clear what to do. And what to do might include deleting some, creating some new ones (with a different meaning), etc.

balacij: That's a good idea. I just moved it now. Hopefully it can be helpful in the future. Thank you!

JacquesCarette (approved):

PR #2536, opened 2021-05-28
Adding `ccTimeDomain` to PDController concepts

Awurama-N:
contributes to Issue #2397


JacquesCarette (approved):
As-is, this is good - but a new PR should start to use this in the example.

PR #2539, opened 2021-05-28
Convert Newton's SLRM TM into an EquationalModel

balacij:
Depends on #2538 and #2537

### DIFF
--- a/code/drasil-lang/Language/Drasil/Development/Sentence.hs
+++ b/code/drasil-lang/Language/Drasil/Development/Sentence.hs
@@ -2,7 +2,7 @@
 -- Various helpers for building Sentences from other bits.
 -- Really ought to be moved out to (likely) docLang, but is here for now.
 module Language.Drasil.Development.Sentence (short, introduceAbb, atStart,
-  atStart', titleize, titleize', phrase, plural, phrasePoss, pluralPoss) where
+  atStart', titleize, titleize', phrase, phrase', plural, phrasePoss, pluralPoss) where
 
 import Control.Lens ((^.))
 
@@ -38,9 +38,13 @@ titleize  n = NP.titleizeNP (n ^. term)
 -- | Plural title case.
 titleize' n = NP.titleizeNP' (n ^. term)
 
--- | Helper for getting the phrase from a 'NamedIdea'.
+-- | Helper for getting the phrase from a 'NamedIdea' using it's UID.
 phrase :: (HasUID n, NamedIdea n) => n -> Sentence
-phrase n = sentenceTerm (n ^. uid) --NP.phrase (n ^. term)
+phrase n = sentenceTerm (n ^. uid)
+
+-- | Helper for getting the phrase from a 'NamedIdea'.
+phrase' :: NamedIdea n => n -> Sentence
+phrase' n = NP.phraseNP (n ^. term)
 
 -- | Helper for getting the plural of a phrase from a 'NamedIdea'.
 plural :: (HasUID n, NamedIdea n) => n -> Sentence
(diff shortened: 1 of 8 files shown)


code/drasil-lang/Language/Drasil/Development/Sentence.hs
```diff
@@ -38,9 +38,13 @@ titleize  n = NP.titleizeNP (n ^. term)
...
 
--- | Helper for getting the phrase from a 'NamedIdea'.
+-- | Helper for getting the phrase from a 'NamedIdea' using it's UID.
 phrase :: (HasUID n, NamedIdea n) => n -> Sentence
-phrase n = sentenceTerm (n ^. uid) --NP.phrase (n ^. term)
+phrase n = sentenceTerm (n ^. uid)
+
+-- | Helper for getting the phrase from a 'NamedIdea'.
+phrase' :: NamedIdea n => n -> Sentence
+phrase' n = NP.phraseNP (n ^. term)
```
balacij, line 47: This is the only thing I'm a _bit_ concerned about. We use this for grabbing the "Newton's second law of rotational motion" phrase, but I'm wondering if we should be using this directly or using `titleize` instead here. If we use titleize, we'd change stable just to capitalize each word I think.

JacquesCarette: Why this instead of `phrase` ?  They both end up grabbing the same thing, one directly (the new one) and one indirectly (`phrase`).

JacquesCarette (changes requested):

code/drasil-lang/Language/Drasil/Development/Sentence.hs
balacij: `phrase` grabs the `uid`, not the name/defining term, which is why I needed to use this (the code for it was already there, it was just commented out). For this particular example, it grabs `torque` instead of `Newton's second law of rotational motion` (which is what stable wants + makes sense). `titleize` appears to grab the "correct" text but it also is intended to alter the  text.

balacij: (and `torque` is what newtonSLR TM defines)

JacquesCarette: The issue seems to be that `phrase` is grabbing the wrong thing, i.e. `torque`.  Maybe the `term` instance is odd? I get the feeling that there's something much deeper going wrong here, and that `phrase'` is a symptom.

balacij: Revisiting this, it looks like `term`'s instance is fine. It looks like the issue actually lies in the UID used for newtonSLR (torque) and registering the newtonSLR in the symbol map. Registering it in the symbolMap before torque "overrides" the `torque` UID/term, resulting in changing all `torque`s in text to `Newton's second law ...`. We'd need to change the UID of the newtonSLR TM so that references between them don't collide.

Should we add UID to the smart constructors of TM so that we manually set them instead of inheriting them from underlying RelationConcepts (which have customizable UIDs)? The new ModelKinds inherit their UIDs from their underlying structures (in particular, the UID of a QDefinition is inherited from the thing it defines -- maybe this is wrong and should be manually set as well?).

JacquesCarette: Aha, that is indeed the real bug!  They should not have the same UID. And adding duplicate UIDs to a symbolMap should be an error!

Yes, TMs should have their own UIDs. For others, it's less clear.

balacij: Sounds good, I just changed it so that we have 2 variants of the constructors; 1 with the manually set UID, and 1 that derives it from the underlying ModelKind. I imagine we would want to slowly convert the variants with the "derived UID" into the one with it manually set until we can remove the derived variant.

JacquesCarette: Yes - please create an issue for that.

JacquesCarette (approved):

JacquesCarette:
So this is now ready?

balacij:
Yes, it's ready now.

PR #2540, opened 2021-05-28
Started adding comments to docLang files

Ant13731:
Contributes #2477


code/analysis/DataTable.csv
```diff
@@ -0,0 +1,675 @@
+Package,	,	,	,	,	,Class Instances
```
JacquesCarette, line 1: I don't think this `DataTable.csv` should be part of this PR.

JacquesCarette (changes requested):

code/analysis/DataTable.csv
Ant13731: Oh sorry, I didn't know it added that. I was looking at the data table from the other issues so I might've done something to it there.

Issue #2231, opened 2020-08-20
Including the auto-generated Analysis in the Nightly Deployments/Travis Builds


Nathaniel-Hu:
Based upon discussions with @smiths, a good addition to the Drasil framework nightly deployments would be the addition of the automatically-generated Data Table (merged in pull request #2208). By making this addition, the scripts to generate the Data Table would also automatically be run in the Travis Builds in future pull requests. This issue is related to/compliments Issue #2221.

Note for @smiths and @JacquesCarette: Would there ever be a case where the Data Table would be treated like a stable folder file, or will it be allowed to just change without needing to be compared to a stable version of the Data Table?

smiths:
Thank you for creating the issue @Nathaniel-Hu.  No, I do not think we would consider any design stable, just like we don't consider the Haddock documentation stable.  We assume that when we make changes to the design and the documentation, we are making each better.  Your tool is just giving us a snapshot of the current design.

JacquesCarette:
In theory 3aea095 does the deploy. Once it looks like it worked, I'll hook it in to the web page too.

Issue #2221, opened 2020-07-22
Posting the auto-generated Analysis to Website


Nathaniel-Hu:
Now that pull request #2208 has been merged, the next step would be to post the auto-generated analysis (the DataTable.csv file) to the [Drasil Nightly Generated Case Studies and Documentation](https://jacquescarette.github.io/Drasil/) website.

The `deploy` target in the Makefile (in /code/) has already been updated to trigger the `analysis` target (which runs the scripts that auto-generate the analysis). Changes will likely need to be made to the `deploy_lite` target (as discussed in this [comment](https://github.com/JacquesCarette/Drasil/pull/2208#discussion_r458941627)) as well as the script deploy_stage.sh (in /scripts/) at the very least.

JacquesCarette:
Changes to the appropriate scripts now done (but not necessarily debugged). The website itself still need to have the results hooked in.

Ant13731:
@JacquesCarette I'm not really sure how to go about doing this. Is there another example I can follow or something I can read up on?

balacij:
@Ant13731 I don't believe the analysis artifact(s) are currently being [uploaded](https://github.com/JacquesCarette/Drasil/tree/gh-pages/) either so that'll probably also be an issue, the related scripts are the [deploy_stage.sh](https://github.com/JacquesCarette/Drasil/blob/master/code/scripts/deploy_stage.sh) and the main [Makefile](https://github.com/JacquesCarette/Drasil/blob/master/code/Makefile#L95-L96). You should check the [website Hakyll build setup](https://github.com/JacquesCarette/Drasil/blob/master/code/website/Main.hs#L171-L180) for how the links are generated/passed along as environment variables for generation in the index.html file (for how you should add in the link too).

Ant13731:
Okay, I'll check it out. Thanks @balacij!

JacquesCarette:
And do continue to ask for help. I don't know how much you know about this kind of shell scripting and web infrastructure - it's a pretty deep rabbit hole if you haven't delved into it before.

Ant13731:
Will do. I think I can get by with my current bash knowledge but the web stuff is new to me.  I'm going to try and imitate how the other files are generated on the website (like the documentation and examples) and see if that works.

Ant13731:
@JacquesCarette Is this issue and #2231 okay to close now?

JacquesCarette:
[...]

Issue #2524, opened 2021-05-27
Swift; missing doxygen generation


balacij:
Right now, if you click on any of the "Generated Code Documentation" links on the main website page, it will 404.

caused by:
![image](https://user-images.githubusercontent.com/1627302/119843148-429f9c00-bed5-11eb-90b6-502a4f701702.png)

![image](https://user-images.githubusercontent.com/1627302/119843355-7c70a280-bed5-11eb-98b6-26dc4eea94bc.png)

balacij:
Well, just changing the ticket after a bit of discovery. It looks like we're not generating doxygen docs for Swift, and it also appears that doxygen doesn't support Swift at the moment.

JacquesCarette:
Right - so we should remove the Swift links entirely. We can revisit later, if the need arises.

balacij:
@Ant13731 Would you mind taking a look at this since you're familiar with the template language for Hakyll?

Ant13731:
Sure! Should I just remove the links from the webpage (and anywhere the links are generated)?

balacij:
Yep. If possible as well, if you could eliminate those swift-related `stat` errors (in the original post) that just signal that the target directory in question doesn't exist, that would be great too, but I don't think that's too important really.

Ant13731:
I'm having trouble finding a way to not insert the missing doxygen stuff. I tried doing some things on the haskell file but it's not really working. I tried changing the `mkExampleCtx` function and the `mkExamples` function but even if I replace the file path with an empty string it still shows up on the website. Is there documentation available for the index.html file syntax (for example, could there be an if statement that checks if a file really exist or if the variable name is "Swift")?

balacij:
Hmm, that's weird. Looking at the [Hakyll docs](https://jaspervdj.be/hakyll/reference/Hakyll-Web-Template.html), it looks like the `if` is only for checking existence of keys and whether or not boolean keys are true.

[...]

Issue #2300, opened 2021-03-01
Responsive CSS for templates


szymczdm:
Our current CSS is not responsive nor accessible on mobile views, it would be worthwhile to update it to be legible on small screens as a stopgap until it can be fully customized.

This should be pretty easy to achieve and would likely reduce the overall complexity of our current CSS.

PR #2545, opened 2021-05-31
Test for adding datatable to website

Ant13731:
I got this error when I tried to run this locally (using `make` in `.\code\website`):
```
 website.EXE: DEPLOY_FOLDER: getEnv: does not exist (no environment variable)
make: *** [Makefile:7: run] Error 1
```
So I wanted to see if the automated tests would work. I'm not really how to generate the website to see if it worked locally.

Contributes (maybe closes?) #2221 #2231


Ant13731:
After trying to run make deploy, I found that I had to install graphviz first.

balacij:
The CI for the website only runs in the full `Build.yaml` script (not the mini `Test.yaml` variant) which only runs on the `master` branch. So, it shouldn't run here.

I think the `deploy_lite` target is what you'd be interested in with this? https://github.com/JacquesCarette/Drasil/blob/b36114f5355ff8b0a290622c6097d9068f13a571/code/Makefile#L212-L216

I previously had a small script to help me with mock deployments locally. I ran through the full steps in the Test.yaml CI, and then used it:

https://github.com/JacquesCarette/Drasil/blob/151e6638fea05bf25d57cb84a06c3e55c302fe9a/code/rn.sh

You'd just need to run `sh rn.sh` in the `code` folder, with that file placed in the `code` folder.

Ant13731:
Okay thanks! I'll give it a try.

Ant13731:
I'm currently going through the process listed in the Test.yaml section, but before I did, I tried the script and it gave this: 
```
Only perform deploys for master (attempted on: master).
Skipping.
```
Does it only run on master? Do I need to fork the repo and try it on the master of the forked one?
I'll try it again once make deploy finishes (its taking some time to compile) but I'm hoping the website will work after its done anyways.

balacij:
Ah, if you're using the `rn.sh` script I made, you'll need to update the `GITHUB_REF=master` to `refs/heads/master`. It might be good to also update that error message.

Sorry about that! I forgot that I updated it since that script's original version.

Ant13731:
Ah okay thanks!

Ant13731:
I tried the new script and it works but now I'm getting a lot of different errors (mostly from the deploy_wrapper.sh file.

My git bash says that `rev` command not found - I think the git bash shell that I'm using on Windows doesn't have it so I might have to try something else. 

Next, I've gotten a bunch of `cannot stat: No such file or directory` for all of the object oriented language html files. 

For some doxygen files, I get the following:
```
doxygen doxConfig
process_begin: CreateProcess(NULL, doxygen doxConfig, ...) failed.
make (e=2): The system cannot find the file specified.
make[1]: *** [Makefile:22: doc] Error 2
```

It did generate an index.html file in the deploy folder, but my additions aren't generating there.

balacij:
Yep, for the `cannot stat` errors, we have a ticket going for it: https://github.com/JacquesCarette/Drasil/issues/2524 

Hmm, `rev` is just a tool that reverses the characters in inputted text.

Perhaps you can install BusyBox for some minimal tools that are on Linux? I'm not too sure what the norm is for development on Windows. https://busybox.net/

Ant13731:
Thanks! I did some looking around and I found that the cygwin shell has make so I can try it there.

Ant13731:
@balacij I just finished installing the windows subsystem for linux so I'm getting all the tests ready to go. I was looking at the `rn.sh` script again and I was just wondering what the GITHUB_SHA represents and if the GITHUB_REPOSITORY variable should be JacquesCarette/Drasil instead? Thanks for all your help!

balacij:
@Ant13731 The `GITHUB_SHA` is meant to be the latest commit hash on the branch and the `GITHUB_REPOSITORY` is meant to be the related GitHub repository location. They're dynamically set as environment variables by the Github Actions runner, check [their docs](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) if you'd like to know more about it. So, you don't need to worry about what values you give them, because they'll always be the "correct" data when running on GitHub Actions.

Note that the script is only meant to help with replicating the environment variables that GitHub Actions automatically creates for each CI run.

Ant13731:
I'm really close to getting it to work fully! Running the build steps in order almost works perfectly, except the analysis folder doesn't generate automatically (I have to run `make analysis` before it works). Also, when I click the link for the data table, it just downloads the csv file. Is this okay?

balacij:
You'd need to ask about whether just a download or a link to the GitHub tree is okay to @JacquesCarette. I'm not quite sure what's desired.

Yep, you'll probably need to add that as a step in the CI. Would you need to add it to the Test.yaml as well? I imagine we would want it there too, but I'm not sure what the analysis is for.

Hmm, that's an interesting issue. Let's try this:
1. In your Windows version of Stack, run `stack --numeric-version` and make note of the version
2. In your WSL, run `stack upgrade --binary-version <version>` where `<version>` is the version from ^.

Ant13731:
Okay thanks. My linux one was using 2.7.1 while my windows is on 2.5.1, so I changed my linux one back to 2.5.1 and it seems to work now.

Ant13731:
I looked into making the data table visible on the website and I found this [link that uses php to convert a csv file into an html table](https://stackoverflow.com/questions/518795/dynamically-display-a-csv-file-as-an-html-table-on-a-web-page). I don't really know any php (or how it fits in with html files), but I think this might be able to help display the file on the website.

balacij:
Nice! You could probably do this with Python too if that's easier (Python comes with really really nice tooling for CSVs).

Though, if you'd prefer to do it via Haskell, there is [`Cassava`](https://www.stackage.org/lts-16.31/package/cassava-0.5.2.0) which we can use. I'm not too sure how many dependencies it would be adding to the website package, but it's probably worth investigating. It can then be used in the website package in the template for `index.html`.

Ant13731:
Do you think it would be good if there was a python file in the scripts folder that could create a new `database.html` file, and then we could link that to the main website? I'm guessing there's a way to call python files to run within a bash script. Python seems to look a lot easier than haskell in terms of parsing csv files (especially since this one is pretty big). I think it might also be worth looking into colouring some of the cells of an html table to easily view the data types and places they are defined.

balacij:
There's a couple of ways we could do this.

If we're okay with using a [DataTables](https://datatables.net/) table to display the data, the sizing wont be an issue, and the scripting will be rather straightforward. With this way, we'd only need to upload the csv file to some location we know. Then, with Javascript in the web browser, we can load that csv file into the DataTable. With this way, we'd have very little programming to do really. If you want to add colouring too, Javascript + DataTables will make this near trivial -- there's a "loading data event" that we can hook into to set colour of cell based on value of cell. I think adding DataTables would be good regardless of where we gather the data from. It allows us to have cool features like sorting, searching, paging, etc. It also follows the theme of #2300.

The Python way or the Haskell way will have about the same difficulty. The Haskell way is certainly easier if you want to have our Home/`index.html` page contain the actual table itself.

There is also another way with Haskell. We could edit the actual `DataTableGen.hs` file to output an html file alongside the csv. This would likely be a good solution too. I think you'd just want to play around with this function:
https://github.com/JacquesCarette/Drasil/blob/b470ff3dac52dc433a027570793cdef94de7d888/code/scripts/DataTableGen.hs#L116-L126

I think this way would also result in little programming.

Ant13731:
Okay, sounds good. I tried to make something with python and it's not really working so I'll try modifying the existing haskell file now

JacquesCarette:
Excellent discussion.  Modifying the Haskell generator to output html as well as csv seems like the best solution.  The point of the csv is that Excel really is a decent tool for trying to look more deeply into the data. For more cursory investigation, some kind of html-based table seems like a good idea. Strictly speaking, we don't need the csv on the web site; but I like being fully open, and it doesn't hurt to have it there too, so why not?

Ant13731:
I think this should be just about ready to implement, so I'll fix any mistakes from the auto tests and mark it as ready for review. Thanks for all your help @balacij!

JacquesCarette (approved):

PR #2567, opened 2021-06-07
Rename stable folders to align with those generated in the builds

balacij:
This just simplifies adding a `make stabilize` target a lot and makes the folder names in `stable/` and `build/` align better


balacij:
Yep, it's a bit more difficult than it appears. The website would break with just this. It needs to be fixed too.

Issue #2582, opened 2021-06-08
Convert SSP's GenDefn normShrR into EquationalModel


balacij:
Similar to #2580 

https://jacquescarette.github.io/Drasil/examples/SSP/srs/SSP_SRS.html#GD:normShrR

Awurama-N:
@balacij this one seemse a bit different from the others in the sense that it's "on its own"

```
normShrR :: RelationConcept
normShrR = makeRC "normShrR"
  (nounPhraseSP "interslice normal and shear force proportionality")
  nmShrRDesc nmShrRRel
```

with the others i usually had to edit both things to make it work

```
offsetIM :: InstanceModeloffsetIM = imNoDerivNoRefs (OthModel offsetRC)
  [qwC landPos $ UpFrom (Exc, exactDbl 0)  
  ,qwC targPos $ UpFrom (Exc, exactDbl 0)]  
  (qw offset) [] "offsetIM" [landPosNote, landAndTargPosConsNote]

offsetRC :: RelationConcept
offsetRC = makeRC "offsetRC" (nounPhraseSP "offset") EmptyS $ sy offset $= E.offset
```

I initially thought it would be:

```
normShrR :: QDefinition 
normShrR = mkQuantDef nmShrRDesc nmShrRRel
```
But there's an error reading that there are "no instances of (Quantity Sentence)"

Awurama-N:
also after this is figured out, i see other Relationship concepts in this file similar to this one, would you like me to create issues and change them to equational models as well?

balacij:
```
normShrR :: QDefinition 
normShrR = mkQuantDef nmShrRDesc nmShrRRel
```
The issue here is that `nmShrRDesc` is a Sentence. You should be converting the RelationConcept (`sy offset $= E.offset`) into a QDefinition using mkQuantDef/any of the other QDefinition constructors (if you CMD+click on mkQuantDef, it should bring you to the file containing mkQuantDef, showing other constructors -- namely, mkQuantDef', fromEqnSt', fromEqnSet, etc). Specifically, here we are defining the `offset` using the expression `E.offset`.

Thank you, the other RelationConcepts in the file are slightly different unfortunately however, they need a new kind of model to assist in defining them. We wouldn't be able to jig them into any of the existing ModelKinds, I think it will need to be some sort of forall/discrete/ranged model. I'm not quite sure yet how we can define them.

Awurama-N:
[...]

PR #2584, opened 2021-06-09
Converting GenDef's normShrR to an equational model

Awurama-N:
closes Issue #2582


Awurama-N:
There were no errors when i did it this way but i'm still not certain this is the right way though

balacij:
Nice!

With this one, we shouldn't be breaking stable in this way. In the `λ=\mathbf{X}=λ \mathbf{f} \mathbf{G}` change to stable, we have a redundant `λ=\mathbf{X}=λ \mathbf{f}`. You'll need to reduce nmShrRRel's `... $= X` into just `X` so that it no longer contains the left-hand side nor the `$=`. With QDefinitions, we're saying "this quantity is defined using this expression", and the `ExprRelat` instance of a QDefinition is set to recreate this `nmShrRRel`-stylized Relation/Expr (e.g., `sy quantity $= expr`). After reducing the `nmShrRRel` relation, it should work nicely.

The other change of relabeling is caused by it no longer using `(nounPhraseSP "interslice normal and shear force proportionality")`, so you might need to use another constructor to override the term of QDefinition -- `mkQuantDef'` maybe?

Awurama-N:
@balacij Oh okay, for the changes i made to `normShrR` should i keep them the same and just work on `normShrRRel` (except changing mkQuantDef) or it should be how it was before i made any changes and then just work on `normShrRRel`?

balacij:
Yep, start with undoing the changes to `stable/`, and then work on fixing the QDefinition made using `mkQuantDef` and the `normShrRRel` Relation.

code/drasil-example/Drasil/SSP/GenDefs.hs
```diff
@@ -192,14 +192,14 @@ effNormFDeriv = mkDerivNoHeader [foldlSent [
...
 
---
-normShrR :: RelationConcept
-normShrR = makeRC "normShrR"
-  (nounPhraseSP "interslice normal and shear force proportionality")
-  nmShrRDesc nmShrRRel -- genDef5Label
+-- 
+
+normShrR :: QDefinition 
+normShrR = mkQuantDef' normToShear  (nounPhraseSP "interslice normal and shear force proportionality") nmShrRRel
```
balacij, line 198: I think this should be defining `intShrForce`, not `normToShear`. Does it still break stable if you try this instead?

Awurama-N: okay i just made the change, i think it should work, the first time i tried it was without the noun phrase and it caused errors, so i didn't think to go back to it, but it should be better now.

balacij: Hmm, it looks like it still causes breakage of `stable/`. Where is the breakage located if you don't give the `mkQuantDef` a custom term? We might just need a custom term for the gd instead.

JacquesCarette (approved):
The code changes are fine. But it is very odd that these changes cause `normal and` to get added to the text, even though that's what it looks like was there before!

My guess: `nmShrRDesc` was used before, and now it's the `nirmShrR`'s description that is used. @smiths should also look at this before this gets merged.

code/stable/ssp/Website/SSP_SRS.html
```diff
@@ -1055,7 +1055,7 @@ <h3>Goal Statements</h3>
                   </p>
                   <p>
                     <div id="determineShearF">
-                      Determine-Shear-Forces: Determine the interslice shear forces between each pair of vertical slices of the slope.
+                      Determine-Shear-Forces: Determine the interslice normal and shear force proportionality between each pair of vertical slices of the slope.
```
smiths, line 1058: This change is not correct.  The goals for determining the shear force and the normal force are separate goals.

code/stable/ssp/Website/SSP_SRS.html
```diff
@@ -1523,7 +1523,7 @@ <h4></h4>
                             <em><b>W</b></em> is the weights (<em>\(\frac{\text{N}}{\text{m}}\)</em>)
                           </li>
                           <li>
-                            <em><b>X</b></em> is the interslice shear forces (<em>\(\frac{\text{N}}{\text{m}}\)</em>)
+                            <em><b>X</b></em> is the interslice normal and shear force proportionality (<em>\(\frac{\text{N}}{\text{m}}\)</em>)
```
smiths, line 1526: This change is incorrect.

smiths (changes requested):
I'm afraid this change has modified the meaning of the original document.  I've edited in some comments in the html version to highlight points where the change makes the SRS incorrect.  I stopped looking at the differences, because they all seem to be incorrect.  In the analysis of the forces and stresses, the shear and normal forces are generally tracked separately.  The new wording incorrectly combines them.

We cannot accept this pull request as given.

balacij:
It looks like we need to pull in changes from `master` into this branch as there are conflicts with GenDefs.hs now.

The way to fix this looks like we need to add a custom term to GenDefs, like we did for InstanceModels. Check out PR #2564's [changed files](https://github.com/JacquesCarette/Drasil/pull/2564/files) for the way to do this.

Though, since we've added custom fields to InstanceModels, TheoryModels, and GenDefs now, it sounds like we need to have a  better revisit to it -- we'd likely want to add an overridable packet of information above ModelKinds entirely, for each of the various models.

JacquesCarette:
I think that indeed the issue is more conceptual. We don't seem to have quite the right information in the right places. It might be that the examples also are using the 'wrong' information, but were built in a way that made them correct anyways.

JacquesCarette:
@Awurama-N can you merge master please? @smiths can you approve this PR?

smiths (approved):

JacquesCarette:
This looks ready to go now, right @Awurama-N ?

PR #2585, opened 2021-06-09
Create EquationalConstraints Model + convert SSP's TM `equilibrium` into EquationalConstraints model

### LINKED ISSUE #2581: New ModelKind: `EquilibriumModel`?
Related to #2371 

Considering SSP's [equilibrium TM](https://jacquescarette.github.io/Drasil/examples/SSP/srs/SSP_SRS.html#TM:equilibrium) and [momentEql GD](https://jacquescarette.github.io/Drasil/examples/SSP/srs/SSP_SRS.html#GD:momentEql), I don't these are models that we can model as `EquationalRealm`s because an `EquationalRealm` says that some quantity can be defined using one of any number (> 1) expressions. I think we need a new kind of model that says:
* for some pool of expressions, they are equivalent under certain conditions (conditions may be trivial/empty)
* we can calculate settings where the above equivalence holds using a set of QDefinitions that define the variables

Related models:
image
image

balacij:
Closes #2581 

One notable thing is that it divides the constraints for the `equilibrium` TM into 3 separate equations:
So, from ->
image

we get ->
image

### DIFF
--- a/code/drasil-example/Drasil/SSP/TMods.hs
+++ b/code/drasil-example/Drasil/SSP/TMods.hs
@@ -4,9 +4,10 @@ module Drasil.SSP.TMods (tMods, factOfSafety, equilibrium, mcShrStrgth, effStres
 
 import Control.Lens ((^.))
 import Prelude hiding (tan)
+import qualified Data.List.NonEmpty as NE
 
 import Language.Drasil
-import Theory.Drasil (TheoryModel, tm, tm', ModelKinds(EquationalModel, OthModel))
+import Theory.Drasil
 import Utils.Drasil
 import Utils.Drasil.Concepts
 import qualified Utils.Drasil.Sentence as S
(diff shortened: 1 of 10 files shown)


JacquesCarette (approved):

smiths:
I agree with @JacquesCarette that this is better knowledge encoding.  I do think that we'll want to format the information differently in the future.  A scientist or engineer that reads the SRS might not know how to interpret the and symbol.  From a display point of view, the more traditional way to show it would be to have each condition on a separate line, without any and symbol.

JacquesCarette:
Right, that's what I was thinking too. It's probably relatively easy to do. An issue making that as a feature request?

balacij:
I see. I agree as well. To be honest, I've still got a note about @smiths last comment from PR #2496 as well. I wonder if we can create a few kinds of "language internationalization & description variation" (actual English language, expression language, etc) since we have a few "languages". Specifically if we can create a renderer that would be able to infer these various alternative display methods -- that would be nice I think.

I'll create an issue from this.

JacquesCarette:
I don't think we want to 'infer', but we do want to be able to specify. There's a strong CSS analogy here. CSS helps with modularization of web content, by helping you separate the content (XML/HTML with no style) from display options.

Though I guess, as with web sites, we could 'infer' which set of CSS to use depending on context.

balacij:
Ah, that would be better to specify. I was thinking in terms of inferring from ProgExprs, but that's hacky. It should definitely be specified and then outputted differently depending on configuration.

PR #2596, opened 2021-06-11
Port QDef-In-InstMod's 39f2c22

balacij:
QDef-In-InstMod's 39f2c222cc1a96fe5085e247a7b3dc80574e9bd9 commit took out the `_definitions` list of QDefinitions from the SystemInformation. The related discussion noted that it was used for code generation, so I think a good half-way until fully automating is to convert it into an InstanceModels list because we can at least generate more things from the InstanceModels (with richer models) than we can with just QDefinitions. For example, the DEModels will be something that we can later move away from the Choices and have derived like our current `getEqModQdsFromIm` get the QDs from IMs.

(Also, I'm leaving the related commit hash in the title, I hope GitHub fixes it and makes it the short-style too, but we'll see -- I'll change it if it doesn't).


JacquesCarette (approved):

JacquesCarette:
We're seriously going to need to **design** `SystemInformation`. But, for now, this fixes things nicely.

Issue #2598, opened 2021-06-11
Formally displaying EquationalConstraints


balacij:
> I agree with @JacquesCarette that this is better knowledge encoding.  I do think that we'll want to format the information differently in the future.  A scientist or engineer that reads the SRS might not know how to interpret the and symbol.  From a display point of view, the more traditional way to show it would be to have each condition on a separate line, without any and symbol.

_Originally posted by @smiths in https://github.com/JacquesCarette/Drasil/issues/2585#issuecomment-859704303_

There's a bit of discussion on this PR that explains what we would want from this ticket.  For starters, the EquationalConstraints should be separated by new lines instead of logical `and` symbols.

It could however be extended to adding full on i18n of language outputs so that the way we output some ideas varies based on the configuration.

For example, we could have a renderer that allows either styles and changes like the below quote:

> I agree that this looks better, and we should use it for now, but in the future, there may be cases where people wish to use the | v | notation (single bars) to represent a norm.  In some cases this would be a style preference, or the authors may wish to distinguish the vector norm (single bars) and the matrix norm (double bars).

_Originally posted by @smiths in https://github.com/JacquesCarette/Drasil/issues/2496#issuecomment-846143933_

balacij:
The EquationalConstraints component of this ticket was completed with #2600, however, the greater aspect of having full i18n of different forms isn't -- that would require non-trivial design and would require the sign-off of @JacquesCarette + @smiths before really starting that (e.g., if we even want it). So, I think we can close this ticket.

PR #2600, opened 2021-06-14
Creating a basic DisplayExpr language

### LINKED ISSUE #387: Redesign ExprRelat (Theoretical Models and Instance Models)
The underlying infrastructure of `ExprRelat` needs revision with respect to the needs of code generation.  Currently relations are just unrestricted `Expr`s which requires pattern matching hacks which may fail in order to get the information needed to generate code.

balacij:
Closes #387 

Related to:

>Agreed. The only uses of `relat` right now should all be for display purposes.

> So it would be interesting to start a new `Display` language that contains only two things: `Expr` and a new thing for `:=`. Then `relat` could be changed to output `Display` instead of `Expr` -- and we could see what breaks. Possibly not too much.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/issues/387#issuecomment-858115513_

I added 1 extra component to the DisplayExpr; multiple exprs to partially address #2598. The related stable-breaking change puts the "sum of the Fxs = 0", "sum of the Fys = 0", and "sum of the Ms = 0" onto new lines.

Creating a draft PR so I can comment on code to ask a few questions.


code/drasil-docLang/Drasil/ExtractDocDesc.hs
```diff
@@ -43,26 +43,26 @@ secConPlate mCon mSec = preorderFold $ purePlate {
...
   (concatMap egetSec) `appendPlate` (preorderFold $ purePlate {
   scsSub = Constant <$> \case
-    (TMs _ _ t) -> let r = concatMap (\x -> x ^. invariants ++
-                           defExp (x ^. defined_quant ++ x ^. defined_fun) ++
-                           r (x ^. valid_context)) in r t
-    (DDs _ _ d _) -> map sy d ++ defExp d
-    (GDs _ _ g _) -> expRel g
-    (IMs _ _ i _) -> expRel i
+    (TMs _ _ t)   -> goTM t
+    (DDs _ _ d _) -> go d -- ++ map (toDispExpr . sy) d ++ map (toDispExpr . (^. defnExpr)) d -- TODO: this *was* a very different definition in comparison to those around it...
```
balacij, line 51: The commented out method is a 1-1 way of translating the existing code into this new DisplayExpr language, but it seems to make no difference between this way or the way the current code works either.

code/drasil-example/Drasil/SWHS/Body.hs
```diff
@@ -116,7 +116,7 @@ symbMap = cdb (qw heatEInPCM : symbolsAll) -- heatEInPCM ?
   ++ map nw prodtcon ++ map nw physicCon ++ map nw mathcon ++ map nw mathcon' ++ map nw specParamValList
   ++ map nw fundamentals ++ map nw educon ++ map nw derived ++ map nw physicalcon ++ map nw unitalChuncks
   ++ [nw swhsPCM, nw algorithm] ++ map nw compcon ++ [nw materialProprty])
-  (cw heatEInPCM : map cw symbols ++ srsDomains) -- FIXME: heatEInPCM?
+  (cw heatEInPCM : map cw symbols ++ srsDomains ++ map cw specParamValList) -- FIXME: heatEInPCM?
```
balacij, line 119: Without these concepts added to the map, we get an error for not having many reachable concepts (all auxiliary constants that weren't in the table of symbols). I'm not entirely sure which change in the code results in this change of concept enforcement, but I think it looks like a positive one since it now seems to enforce that the symbols are there.

(Also, regarding the note that I made 2 comments up, this change is required regardless of the translation.)

code/drasil-lang/Language/Drasil/Classes.hs
```diff
@@ -154,16 +156,18 @@ class UnitEq u where
...
 
 -----------------------------------------------------
--- TODO: It is ok to be able to view a (defining?) 'Relation', but not necessarily
--- to 'set' it,  as it might just not be settable. So Getter it is.
+
 -- | Has a function that turns into a 'Relation'.
 class ExprRelat c where
   -- | Holds a relation.
   relat :: c -> Relation
+  -- TODO: rename to `toRelat`/`asRelat`? essentially saying "this is an idea that we can convert into a logical equivalence checking relation"
```
balacij, line 164: My comment is outdated here, but it's still true that I'm changing the meaning of ExprRelat.

It's no longer much of a "Relation", it's more so just a "is this a thing that we can encode as an expression/relation?".

For example, things that just contain UIDs to variables will be just `sy UID`, while things containing defining expressions will be `sy UID $= definingExpr`.

This ExprRelat should likely be renamed, but I couldn't think of a better name.

code/drasil-lang/Language/Drasil/Expr/Display.hs
```diff
@@ -0,0 +1,32 @@
...
+  defines, multiExpr, multiExprNE) where
+
+import qualified Data.List.NonEmpty as NE
+
+import Language.Drasil.Expr (Expr)
+
+data DisplayExpr where
+    AlgebraicExpr :: Expr -> DisplayExpr
+    Defines       :: DisplayExpr -> DisplayExpr -> DisplayExpr
+    MultiExpr     :: NE.NonEmpty DisplayExpr -> DisplayExpr
```
balacij, line 13: AlgebraicExpr is just the basic expressions. I leave it without UIDs (yet?) because I'm not yet sure of the impacts. This is also why the `Defines`'s LHS is a DisplayExpr as well (as opposed to a UID).

code/drasil-lang/Language/Drasil/Expr/Display.hs
```diff
@@ -0,0 +1,32 @@
...
+
+
+class Display c where
+  toDispExpr :: c -> DisplayExpr
+
+instance Display Expr where
+  toDispExpr = AlgebraicExpr
+
+instance Display DisplayExpr where
+  toDispExpr = id
```
balacij, line 23: I really wanted to place these in the `Classes` file, but I seem to run into circular dependency issues fairly quickly when I add in the 2 instances. I'll probably try to move it again though.

code/drasil-printers/Language/Drasil/Printing/Import.hs
```diff
@@ -97,6 +99,11 @@ processExpo a
 assocExpr :: P.Ops -> Int -> [Expr] -> PrintingInformation -> P.Expr
 assocExpr op prec exprs sm = P.Row $ intersperse (P.MO op) $ map (expr' sm prec) exprs
 
+dispExpr :: DisplayExpr -> PrintingInformation -> P.Expr
+dispExpr (AlgebraicExpr e) sm = expr e sm
+dispExpr (Defines l r)     sm = P.Row [dispExpr l sm, P.MO P.Eq, dispExpr r sm] -- TODO: Use new symbol?
```
balacij, line 104: Do we want to replace the Equals symbol with a new symbol for `Defines`?

code/drasil-theory/Theory/Drasil/MultiDefn.hs
```diff
@@ -50,6 +50,10 @@ instance Definition    MultiDefn where defn     = rDesc
 -- | The complete Relation of a MultiDefn is defined as the quantity and the related expressions being equal
 --   e.g., `q $= a $= b $= ... $= z`
 instance ExprRelat     MultiDefn where relat q  = sy q $= foldr1 ($=) (NE.map (^. expr) (q ^. rvs))
+-- TODO: How do we want to display MultiDefns?
+instance Display       MultiDefn where
+  -- toDispExpr md = multiExpr $ map (^. expr) $ NE.toList $ md ^. rvs
+  toDispExpr = toDispExpr . relat  -- this one at least doesn't break stable (yet?)
```
balacij, line 56: I think MultiDefns are the hackiest thing I left in before PR.

I'm not quite sure how we'd want to display, so I leave it as it is.

For example, if we had `X` which can be defined by `Y` or `Z`.

Do we want to display it as `X = Y = Z` or as:

```
(i) X = Y
(ii) X = Z
```

or anything else similar?

code/drasil-utils/Utils/Drasil/Fold.hs
```diff
@@ -25,11 +25,12 @@ foldle1 f g (x:y:xs) = foldle f g (f x y) xs
 -- | Helper for formatting constraints.
 foldConstraints :: (Quantity c) => c -> [Constraint] -> Sentence
 foldConstraints _ [] = EmptyS
-foldConstraints c e  = E $ foldl1 ($&&) $ map constraintToExpr e
+foldConstraints c e  = E $ multiExpr $ map constraintToExpr e
   where
     constraintToExpr (Range _ ri)         = realInterval c ri
     constraintToExpr (EnumeratedReal _ l) = isin (sy c) (DiscreteD l)
     constraintToExpr (EnumeratedStr _ l)  = isin (sy c) (DiscreteS l)
+      -- TODO: Should these 2 above be moved into DisplayExpr?
```
balacij, line 33: This relates to your discussion in https://github.com/JacquesCarette/Drasil/commit/b585428f1626a9a6aebd93bb7309869c7ed720d7#diff-1c876a4c1a461cc7dbbf3a2edfccc788d3979c4c1f436cf377a8ce323aababc5

Should we being trying something like this next?

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
```diff
@@ -1104,9 +1104,9 @@ \subsubsection{Properties of a Correct Solution}
 \\
 \midrule
 \endhead
-$α$ & $α\gt{}0$
+$α$ & ${α\gt{}0}$
 \\
-${θ_{p}}$ & ${θ_{p}}\gt{}0$
+${θ_{p}}$ & ${{θ_{p}}\gt{}0}$
```
balacij, line 1109: I'm not yet sure what's causing this double {}, but it's only appearing in the TeX version as of right now. I think that current changes wrap things in a PrintingAST Row twice somehow, but I haven't figured it out yet.

code/stable/glassbr/Website/GlassBR_SRS.html
```diff
@@ -2058,9 +2058,7 @@ <h3>Data Constraints</h3>
                     </tr>
                     <tr>
                       <td><em>a</em></td>
-                      <td>
-                        <em>a&thinsp;&gt;&thinsp;0 &and; a&thinsp;&ge;&thinsp;b</em>
```
balacij, line 2062: It appears to display on the next line, but I'm not quite sure if this is the "right" way to change things. I didn't actually touch much of the display/formatting code, so I haven't dug deep in this yet.

code/stable/ssp/SRS/SSP_SRS.tex
```diff
@@ -513,12 +513,22 @@ \subsubsection{Theoretical Models}
...
 \\ \midrule \\
 Equation & \begin{displaymath}
-           \displaystyle\sum{{F_{\text{x}}}}=0\land{}\displaystyle\sum{{F_{\text{y}}}}=0\land{}\displaystyle\sum{M}=0
+           \displaystyle\sum{{F_{\text{x}}}}=0
+           \end{displaymath}
+           \begin{displaymath}
+           \displaystyle\sum{{F_{\text{y}}}}=0
+           \end{displaymath}
+           \begin{displaymath}
+           \displaystyle\sum{M}=0
```
balacij, line 522: All are displayed on new lines, but I don't know if this is a standard or if I should be adding a HARDNL between them.

code/stable/swhs/SRS/SWHS_SRS.tex
```diff
@@ -79,6 +81,10 @@ \subsection{Table of Symbols}
 \\
 $AR$ & Aspect ratio & --
 \\
+${AR_{\text{max}}}$ & Maximum aspect ratio & --
```
balacij, line 84: The auxiliary constants are now added to the table of symbols.

code/drasil-example/Drasil/SWHS/Body.hs
JacquesCarette: It would be interesting to know a bit more as to why this is needed. I agree that it is an improvement!

code/drasil-lang/Language/Drasil/Classes.hs
JacquesCarette: we need to look at all uses to see what is going on. From your description, it's almost like it's being *abused* to get two different features!

code/drasil-lang/Language/Drasil/Expr/Display.hs
JacquesCarette: I think this definition is good as is. It really ought to be an embedding of `Expr` (your first constructor) along with some extras that are there just for display purposes (the rest).

code/drasil-lang/Language/Drasil/Expr/Display.hs
JacquesCarette: Add a `DisplayClasses` file, i.e. `Language.Drasil.DisplayClasses`, and put the class there.

code/drasil-printers/Language/Drasil/Printing/Import.hs
JacquesCarette: Later PR. Let's minimize the changes in this one.

code/drasil-theory/Theory/Drasil/MultiDefn.hs
JacquesCarette: Can we in fact leave it as using 'and' in this PR?

code/drasil-utils/Utils/Drasil/Fold.hs
JacquesCarette: Yes. In particular, the target to `isin` should be a `DisplayExpr`.

code/stable/ssp/SRS/SSP_SRS.tex
JacquesCarette: `displaymath` are automatically displayed on new lines, so it's good as is.

code/drasil-lang/Language/Drasil/Classes.hs
balacij: Sounds good.

We use it as follows right now:
* For TM definitions, since we have manually set equations and relations (not using ModelKinds enough here really), we use `relat <QD>` to get the equations related to EquationalModels. I think this one can be nicely minimized/removed by having a tighter integration with the ModelKinds.
* MultiDefn: for the `toDispExpr`, we intentionally don't break stable by using `relat` to display the whole relation
* In the last equation of a few "equation derivation lists".
* 
```
drasil-docLang/Drasil/DocumentLanguage/Definitions.hs:  buildDescription v u (relat g) m []) : fs
drasil-docLang/Drasil/DocumentLanguage/Definitions.hs:  foldr (\x -> buildDescription v u x m) [] [relat i]) : fs
```
This one was actually in a TODO note for me to look into today.

code/drasil-lang/Language/Drasil/Expr/Display.hs
balacij: Thank you!

code/drasil-lang/Language/Drasil/Classes.hs
JacquesCarette: I think we should strive to remove `relat` and `ExprRelat` entirely. We would need to create functions that implement similar functionality, on top of the existing classes.

code/drasil-lang/Language/Drasil/Expr/Display.hs
balacij: Sounds good, just did this in 83dbb923653f43e7bf4212646c6e65d845b47bab

code/drasil-printers/Language/Drasil/Printing/Import.hs
balacij: Sounds good.

code/drasil-theory/Theory/Drasil/MultiDefn.hs
balacij: Ah, yes, this one currently displays in the `a $= b $= ... $= z` format to be clear. But, of course, we can leave it as it is.

code/stable/ssp/SRS/SSP_SRS.tex
balacij: That's good to know. Thank you!

code/stable/dblpendulum/SRS/DblPendulum_SRS.tex
balacij: Fixed

code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -375,9 +374,9 @@ convCall c x ns f libf = do
   
 -- Converts a Constraint to an Expr
 renderC :: (HasUID c, HasSymbol c) => c -> Constraint -> Expr
-renderC s (Range _ rr)          = renderRealInt s rr
-renderC s (EnumeratedReal _ rr) = IsIn (sy s) (DiscreteD rr)
-renderC s (EnumeratedStr _ rr)  = IsIn (sy s) (DiscreteS rr)
+renderC s (Range _ rr)         = renderRealInt s rr
+renderC _ (EnumeratedReal _ _) = error "EnumeratedReal IsIn not supported yet" -- IsIn (sy s) (DiscreteD rr)
+renderC _ (EnumeratedStr  _ _) = error "EnumeratedStr IsIn not supported yet" -- IsIn (sy s) (DiscreteS rr)
```
balacij, line 379: It doesn't seem like this is ever reached. Same as `convExpr (IsIn ..)`.

code/drasil-utils/Utils/Drasil/Fold.hs
balacij: I gave this a go in ff34fe093b8898383232e2822f478a394243aef9. I'm not too sure if I'm happy with the result. If `IsIn` is one of many future operators that are meant to be something that's only for display and we never generate code for, then it may be the case that DisplayExpr should just fully recreate Expr within DisplayExpr and purely work in terms of DisplayExpr for all things related to printing (of course, excluding code generation).

code/stable/glassbr/Website/GlassBR_SRS.html
balacij: This is fixed now

code/drasil-utils/Utils/Drasil/Fold.hs
JacquesCarette: There won't be that many more things in `Expr` that will migrate to `DisplayExpr`. This will all need to be revisited once we go typed, so it's not worth doing anything complicated now.

code/drasil-lang/Language/Drasil/Classes.hs
balacij: I think we can do that. The TMs, I think we can minimize much of it with a tighter integration with the ModelKinds. It seems that much of the current usage in the "last equation of a few 'equation derivation lists'" should really be converted into DisplayExprs. For the MultiDefn, we can just move the logic of relat into the DisplayExpr. I'll just need to inspect the Definitions.hs usage I mentioned but I think it will be doable.

JacquesCarette: Sounds good.

code/drasil-code/Language/Drasil/Code/Imperative/Import.hs
balacij: Update: this will later be moved to a new CodeExpr in #2602

code/drasil-example/Drasil/GlassBR/TMods.hs
```diff
@@ -23,7 +23,7 @@ tMods = [pbIsSafe, lrIsSafe]
 lrIsSafe :: TheoryModel
 lrIsSafe = tm (EquationalModel lrIsSafeQD)
    [qw isSafeLoad, qw tmLRe, qw tmDemand] ([] :: [ConceptChunk])
-   [lrIsSafeQD] [sy isSafeLoad $= lrIsSafeExpr] [] [makeCite astm2009] 
+   [lrIsSafeQD] [] [] [makeCite astm2009]
```
balacij, line 26: This was duplicate information.

code/drasil-theory/Theory/Drasil/Theory.hs
```diff
@@ -16,8 +16,9 @@ class Theory t where
   quantities    :: Lens' t [QuantityDict]
   operations    :: Lens' t [ConceptChunk] -- FIXME: Should not be Concept
   defined_quant :: Lens' t [QDefinition]
-  invariants    :: Lens' t [Relation]
+  invariants    :: Lens' t [DisplayExpr]  -- TODO: temporary hack until designed, previously `Lens' t [Relation]`
   defined_fun   :: Lens' t [QDefinition]
+  display_exprs :: t -> [DisplayExpr]     -- TODO: temporary hack until designed
```
balacij, line 21: This area needs a little bit of design work.

Should the `invariants` be a list of Relations? If we only intend to display them, then this change might be "okay".

The `display_exprs` is meant to be a list of all the related defined quantities, functions, and invariants.

Would it be okay to merge the defined_fun (functions) and defined_quants (quantities)? It doesn't appear that we use the defined_fun list yet, and we can later filter based on the Space if need be (once we have function spaces).

code/drasil-theory/Theory/Drasil/Theory.hs
```diff
@@ -70,6 +71,11 @@ instance HasReference       TheoryModel where getReferences = ref
 instance ConceptDomain      TheoryModel where cdom = cdom . view con
 -- | Finds any additional notes for the 'TheoryModel'.
 instance HasAdditionalNotes TheoryModel where getNotes = notes
+
+-- TODO: I think we should be gathering these from the ModelKinds of the TheoryModel.
+--       If we need "more than 1 ModelKind" in the TheoryModel, we may need to create 
+--       a "stacked model" that allows for composing them.
```
balacij, line 77: If we choose to do something like this with a "MultiModel" of sorts, then the _defq, _invs, _dfun lists should likely be completely derived from the the models held within the "MultiModel". However, it may prove a bit difficult to work with for the above Theory typeclass as I don't see a clear path to creating a nice Lens for a "MultiModel".

balacij: One nice thing about a "MultiModel" however, is that it would make TM definitions a bit shorter.

code/drasil-theory/Theory/Drasil/Theory.hs
JacquesCarette: The various pieces of a theory were defined from semantic considerations. We did not yet use all its features.

Once we do have types, yes, we can merge `defined_fun` and `defined_quants`.

code/drasil-theory/Theory/Drasil/Theory.hs
JacquesCarette: Let's wait and see if/when this actually arises in practice.

code/drasil-example/Drasil/GlassBR/TMods.hs
balacij: Before one of my changes in this last commit, the first list of QDefs were unused in showing the equations, it completely relied on the list of invariants. So I have to remove the invariants from this since it, as an "invariant", was already given by lrIsSafeQD.

code/drasil-lang/Language/Drasil/DisplayExpr.hs
```diff
@@ -7,7 +7,14 @@ import Language.Drasil.Expr (Expr)
...
 data DisplayBinOp = Defines | IsIn
-data DisplayAssocBinOp = And
+  deriving (Eq, Show)
+
+-- | Associative Binary Display Operations
+--
+--   Note: 'Equal' makes sense in the Display language, but not in the
+--         general computation language.
+data DisplayAssocBinOp = And | Equal
+  deriving (Eq, Show)
```
balacij, line 17: Well, on second thought, 'Equal' may not be good for the associative list.

balacij: In a sense, it can be, but in another, it's definitely not. I'm starting to confuse myself with it. Any thoughts on this @JacquesCarette?

JacquesCarette: It's indeed quite tricky. Let's call it `Equivalence` instead, that will be less confusing.

balacij: Thanks! I changed the name now.

JacquesCarette (approved):

PR #2601, opened 2021-06-14
`drasil-code` documentation 1

Ant13731:
Contributes #2477 
Also changed a type name spelling error. It was only used in the one file where it was defined, so it shouldn't change anything.

### DIFF
--- a/code/drasil-build/Build/Drasil/Make/MakeString.hs
+++ b/code/drasil-build/Build/Drasil/Make/MakeString.hs
@@ -39,15 +39,15 @@ renderVar f (Os nm _ _ _) = f nm
 renderVar f (Implicit nm) = "\"" ++ f nm ++ "\""
 renderVar f (Free nm) = f nm
 
--- | Constructor for converting a String into a MakeString.
+-- | Constructor for converting a 'String' into a 'MakeString'.
 makeS :: String -> MakeString
 makeS = Mr
 
--- | Constructor for Windows OS variables.
+-- | Constructor for Windows 'Os' variables.
 mkWindowsVar :: VarName -> VarVal -> VarVal -> MakeString
 mkWindowsVar n w e = Mv $ Os n w e e
 
--- | Constructor for 'OS' variables.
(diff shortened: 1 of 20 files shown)


code/drasil-build/Build/Drasil/Make/MakeString.hs
```diff
@@ -39,15 +39,15 @@ renderVar f (Os nm _ _ _) = f nm
 renderVar f (Implicit nm) = "\"" ++ f nm ++ "\""
 renderVar f (Free nm) = f nm
 
--- | Constructor for converting a String into a MakeString.
+-- | Constructor for converting a 'String' into a 'MakeString'.
 makeS :: String -> MakeString
 makeS = Mr
 
--- | Constructor for Windows OS variables.
+-- | Constructor for Windows 'Os' variables.
```
JacquesCarette, line 46: No, here OS is short for Operating System, so having both capitalized was correct (here and elsewhere).

code/drasil-code/Language/Drasil/Chunk/Code.hs
```diff
@@ -20,126 +20,161 @@ import qualified Data.Map as Map
 import Text.PrettyPrint.HughesPJ (render)
 
 -- not using lenses for now
+-- | Information that is needed to translate an idea into code.
```
JacquesCarette, line 23: This class is the other way around: "code as an idea". So it says that a `CodeIdea` must include both some code and its name.

JacquesCarette (changes requested):

JacquesCarette:
Note, some of the GOOL files are huge, and not worth fully documenting with Haddock-style comments right now. Other issues are more pressing.

Ant13731:
By GOOL files, do you mean the ones in `drasil-gool`? Should I continue with the `drasil-code` haddock documentation?

JacquesCarette:
Yes and yes.

JacquesCarette (approved):

Issue #2489, opened 2021-05-20
`section'` Function Behaviour


Ant13731:
@JacquesCarette, @smiths  This issue builds on top of a bunch of functions and I'm not completely sure what is going wrong, so I apologize in advance if there is any confusion here. I initially found this while trying to rename all the concept-level functions.
Under `Drasil.DocLang.SRS`, the helper function `section'` is defined as:
```Haskell
--function that sets the shortname of each section to be the reference address
section' :: Sentence -> [Contents] -> [Section] -> String -> Section
section' a b c d = section a b c (makeSecRef d (toString a))
  where
    toString :: Sentence -> String
    toString (S x) = x
    toString ((:+:) s1 s2) = toString s1 ++ toString s2
    toString _ = error "Term is not a string"
```
This is used in defining a reference such as this one:
```Haskell
propCorSol cs ss = section' (titleize' Doc.propOfCorSol) cs ss "CorSolProps"
```
This works well for the most part, but `toString` always returns the error case when I try to change `ofA`'s capitalization rules from using a `Replace Sentence` constructor to using `CapWords`
For reference, here is the current definition of `ofA`:
```Haskell
-- | Same as 'of__', except combining Sentence piece is "of a"
ofA :: (NamedIdea c, NamedIdea d) => c -> d -> NP
ofA t1 t2 = nounPhrase'' 
  (plural t1 +:+ S "of a" +:+ phrase t2)
  (plural t1 +:+ S "of a" +:+ phrase t2)
  (Replace (atStart' t1 +:+ S "of a" +:+ phrase t2))
  (Replace (titleize' t1 +:+ S "of a" +:+ titleize t2)) --Replacing this line with CapWords breaks the section' function above
```
I want to be able to change the last line to just `CapWords` in order to reduce any confusion in naming (and reuse combinators instead of making new ones), but I'm not sure if/how to go about solving it. The `titleize'` function used in `propCorSol` should be taking the plural case and then making every word capital in `Sentence` form, but right now it's only using what is given by the `Replace` constructor in `ofA`. Using `CapWords` with the singular `titleize` also calls the same error.

JacquesCarette:
The problem is that `toString` is a horrible, horrible hack. It should be removed. The real bug is, why does `makeSecRef` need a String for its 2nd input?  We should investigate that first.  Then we can see about potential solutions.

Ant13731:
[...]

Issue #1261, opened 2019-05-03
Merge Reference.hs and RefProg.hs


samm82:
Unless there's a specific reason for having separate modules (@smiths, @JacquesCarette?), I think we can merge the two modules into one Reference.hs. 

RefProg:
https://github.com/JacquesCarette/Drasil/blob/1184b84c48d0d3cdf336f2f3d83cce87b172266f/code/drasil-lang/Language/Drasil/RefProg.hs#L1-L20

Reference:
https://github.com/JacquesCarette/Drasil/blob/1184b84c48d0d3cdf336f2f3d83cce87b172266f/code/drasil-lang/Language/Drasil/Reference.hs#L1-L23

smiths:
@samm82 I don't have an opinion on the proposed consolidation.  However, we shouldn't just be looking at the current code, but ways the files could change in the future.  Although `RefProg.hs` is small now, maybe it won't always be small?  I think we are better off thinking about the reason we have two separate files in the first place.  It could just be how the design grew, or there might be a rationale behind the division.  From a software engineering perspective, we should ask: What is the secret of each module?  If the secrets are different, we should keep the files separate.  If not, we can combine them.

samm82:
I would be in favour of merging them, since `RefProg.hs` defines the `Reference` type, and `Reference.hs` defines its constructors. I'm not sure if these two secrets are discrete enough to separate, and we keep these together in similar modules throughout Drasil (eg. [NounPhrase](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-lang/Language/Drasil/NounPhrase.hs)). @smiths

smiths:
That sounds like a good reason for merging @samm82.  If @JacquesCarette doesn't like it, he'll stop it with the pull request.  :smile:

JacquesCarette:
A lot of this is a product of evolution.  Before we merge, we should take a look at the 'uses' graph (I should check those utilities in) to see if there is a structural reason these were kept separate.

smiths:
Great point @JacquesCarette.  The last thing we want to do is go in circles with design decisions.  Splitting files, remerging them and later splitting them again would be a mistake.  We want to be sure that we are making decisions that we are reasonably sure will "stick" going forward.

samm82:
[...]

Issue #1142, opened 2019-02-28
Using custom text as an internal reference link


bmaclach:
Some internal references throughout our example SRSs duplicate information. For example, here's a line from the SRS:
![image](https://user-images.githubusercontent.com/18634568/53579183-1c98bf00-3b47-11e9-9898-2670febd5485.png)
The phrases "goal statements", "theoretical models", and "instance models" are repeated. Ideally, we would want this to say "The goal statements are refined to..." where the phrase "goal statements" itself is an internal reference to the corresponding section. This means we need a way to have custom text get rendered as a link.

The text that gets rendered as a link is passed as a `ShortName` to a `Reference` in Drasil. Thus, if we want to use custom text for the reference text, we need to convert from `String` to `ShortName` or even from `Sentence` to `String` to `ShortName` (though the need for a `Sentence` to `String` conversion is a sign that the `Sentence` may have been 'premature'). We need to think about the design and how best to achieve this functionality.

Ant13731:
I think this is also relevant to #2489, so I'm going to assign myself as well.

Ant13731:
I've been thinking about this for a bit, and with the new Ref constructor and shortname formation, I think this should be doable. However, I think the way that shortnames are handled in drasil may contribute to this problem. 
Some ideas I have are:
- Since we are looking up references by `UID` now, there is no real reason for the `Reference` itself to hold a `shortname` (or even `RefInfo` for that matter) because we want to be able to change the display information on a per-use basis rather than at the central database. In other words, I think the `Ref` constructor in `Sentence` should hold the display name (as a `ShortName` or a `Sentence` and additional display info (as `RefInfo`).
- We could keep `Reference` the same, and modify `RefInfo` so that it could take a display name (by adding something like a `RefDisplay` constructor). The default format would be the `ShortName` within the `Reference`, but if `RefInfo` contains something other than `None`, we could override the `ShortName`. 

The second idea could probably be implemented in smaller parts, but the first could be more concise. @JacquesCarette, what do you think?

JacquesCarette:
Two thoughts: 1. this seems excellent, and 2. there's too much implicit information not spelled out for me to go from 'seems' to 'is'.

[...]

PR #2610, opened 2021-06-17
Sentence short name

Ant13731:
Converted shortname to take in a Sentence.
Contributes (maybe closes?) #1261.  `IRefIntro` still needs `RefProg.hs` in order to avoid import cycles between `Reference.hs` and `Sentence.hs`, though I suppose it could be moved to `Sentence.hs` and then we could remove `RefProg.hs`.
Contributes #1142. I am hoping that this change will allow us to use Sentences as display words.
Contributes (maybe closes?) #2489. I think the original goal is complete here, but there was some discussion about using `Sentence` for other parts of `Reference` constructors.


JacquesCarette (approved):

JacquesCarette:
This does seem to deal with #1261. I might want to rename the file (`RefPRog`), as it is now quite obscure. It does seem to deal with #2489 too.

Issue #1808, opened 2019-08-07
Reasonable value improvements


bmaclach:
Currently reasonable values are introduced at the same level as constraints in the chunk hierarchy (`ConstrainedChunk` and `ConstrConcept`). Reasonable values should instead be introduced independently from constraints. New chunk types should be created to lie between `QuantityDict` and `ConstrainedChunk` and between `DefinedQuantityDict` and `ConstrConcept`. These new chunk types would add the field for reasonable values, and then `ConstrainedChunk` and `ConstrConcept` should be updated to remove their reasonable value fields. 

Also, the current `HasReasVal` class provides function `reasVal` which is a lens to a `Maybe Expr`. This should instead be just a lens to `Expr`, and the new chunks described above should represent the reasonable value as `Expr`, not `Maybe Expr`. Then a new class, `MayHaveReasVal`, should be created and provide a lens to a `Maybe Expr`, since some things will not have reasonable values.

JacquesCarette:
Have any updates on this issue been made?

bmaclach:
No updates, the original description is still accurate.

Awurama-N:
I started working on this ticket by looking for where everything was defined to gain some more understanding on the "Chunk hierarchy"  and "chunk types".

- ConstrConcepts are conceptual symbolic quantities ('DefinedQuantityDict') with 'Constraint's and maybe a reasonable value (no units!).
- DefinedQuantityDict is the combination of a 'Concept' and a 'Quantity'. Contains a 'ConceptChunk', a 'Symbol' dependent on 'Stage', a 'Space', and maybe a 'UnitDefn'.
- The ConceptChunk datatype is a Concept that contains an idea ('IdeaDict'), a definition ('Sentence'), and a domain (['UID']).
- ConstrainedChunks are symbolic quantities ('QuantityDict') with 'Constraint's and maybe a typical value ('Maybe' 'Expr')
- QuantityDict is a combination of an 'IdeaDict' with a quantity. Contains an 'IdeaDict', 'Space', a function from 'Stage' -> 'Symbol', and 'Maybe' a 'UnitDefn'.

I still have a few questions though:

- What are "Reasonable values"?
- What are examples of the "new chunk types" that need to be created?
[...]

PR #2614, opened 2021-06-21
reasonable value improvements 

Awurama-N:
contributes to #1808


Awurama-N:
After resetting the branch and introducting one at a time (`ConstrReasQDef`) first, i realised the main issue is with the type for reasV changing from `Maybe Expr` to `Expr`.
Please Is there something else i'm missing?

Awurama-N:
When i went through the code and changed the class `HasReasVal` to the type `Maybe Expr`, it fixed the issue. But this time i started with `ConstrainedQDef`chunk type which doesn't have the `reasV` access fucntion.

JacquesCarette:
Sorry for not responding earlier, many other tasks on my plate.

Part of the point is to get rid of those `Maybe`s. `HasReasVal` should return `Expr`. The point is that `ConstrainedQDef` should not have an instance of `HasReasVal` at all.  Now, it might make sense to create a new class, `MayHaveReasVal`, which would have `Maybe`.

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
```diff
@@ -50,6 +50,33 @@ instance Eq            ConstrainedChunk where c1 == c2 = (c1 ^. qd . uid) == (c2
 instance MayHaveUnit   ConstrainedChunk where getUnit = getUnit . view qd
 -- ^ Finds units contained in the 'QuantityDict' used to make the 'ConstrainedChunk'.
 
+data ConstrainedQDef = ConstrainedQDef { _qd' :: QuantityDict
```
JacquesCarette, line 53: This is good.

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
```diff
@@ -50,6 +50,33 @@ instance Eq            ConstrainedChunk where c1 == c2 = (c1 ^. qd . uid) == (c2
...
+                                       , _constr' :: [Constraint] -- should really be NEList
+                                       }
+
+makeLenses ''ConstrainedQDef                          
+
+instance HasUID        ConstrainedQDef   where uid = qd' . uid
+-- ^ Finds 'UID' of the 'QuantityDict' used to make the 'ConstrainedQDef  '.
```
JacquesCarette, line 60: Note that a recent PR from Anthony has moved all these comments from below to above the `instance`, as I think Haddock (the documentation generator) prefers that. Please coordinate.

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
```diff
@@ -60,13 +87,16 @@ cvc :: String -> NP -> Symbol -> Space -> [Constraint] -> Maybe Expr -> Constrai
...
+cnstrw c = ConstrainedChunk   (qw c) (c ^. constraints) (c ^. reasVal)
+
+
+
 
 -- | ConstrConcepts are conceptual symbolic quantities ('DefinedQuantityDict')
 -- with 'Constraint's and maybe a reasonable value (no units!).
 data ConstrConcept = ConstrConcept { _defq :: DefinedQuantityDict
-                                   , _constr' :: [Constraint]
+                                   , _constr'' :: [Constraint]
```
JacquesCarette, line 99: why the extra \' ?

JacquesCarette (commented):
This change on its own seems fine. Might be better to fix the comments now.

Do you want this merged now, before this is 'plugged in', or wait until it sees some use in an example?

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
Awurama-N: > This change on its own seems fine. Might be better to fix the comments now.
> 
> Do you want this merged now, before this is 'plugged in', or wait until it sees some use in an example?

I'd prefer to wait please

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
Awurama-N: Okay

code/drasil-lang/Language/Drasil/Chunk/Constrained.hs
Awurama-N: Since constrained chunk was still present i had to change the `_constr` in `ConstrainedQDef` to `_constr'` and this to `_constr''`

After all the new chunk types are in all the ' should be uniform then

JacquesCarette: Ah, indeed, makes sense.

Awurama-N:
I'll uncomment the new definition after i have checked all the places with the existing constructors, and do the same for `ConstrReasQDef`

Awurama-N:
I have now put two of the new chunk types in. The major error i'm trying to deal with is the Expr/ Maybe Expr issue. 
Another question i had is with the constructors, I understand that i'm suppose to search for their use in the examples.
What i want to be sure of is: for examnple, if i'm working with `ConstrainedQDef` which doesn't have a reasonable value, am i going through the use of the constructors to see which constructor fits that description and then switch the definition of that constructor from `ConstrainedChunk` to `ConstrainedQDef` or making the changes and then going through to debug.

I guess the major question is what should my objective be whilst looking at the use of the constructors in the examples?

JacquesCarette:
The code for the new chunk types looks good to me.

To deal with the `Maybe` issue, you will have no choice but to
1. change `HasReasVal` to return an `Expr` (as a lens)
2. add a `MayHaveReasVale` to return a `Maybe Expr` (as a getter only)
3. adjust some of the downstream code accordingly
This will be slightly tricky in some cases.

About changing the use of constructors: your description is correct. Hmm, I see that the simplest examples are actually for `ConstrConcept` rather than for `ConstrainedChunk`.  (for example
```haskell
constrainedNRV' q cs = ConstrConcept (dqdWr q) cs Nothing
```
would be a prime candidate to change, because of that explicit `Nothing`).

In this case, one to consider is
```haskell
cuc i t s u space cs rv = ConstrainedChunk (qw (unitary i t s u space)) cs (Just rv)
```
which has an explicit `Just`, and thus should be building some that **has** a reasonable value.

JacquesCarette:
So I think your 'next step' would be to change `ConstrainedChunk` to have an `Expr` instead of `Maybe Expr` for `_reasV`. Then adjust the classes `HasReasVal` and `MayHaveReasVal`. Then adjust the constructors to build the most appropriate structure. Then we'll see what breaks downstream.

Awurama-N:
Thank you.
I'm having a bit of trouble trying to figure this error out though:
`instance HasReasVal    ConstrReasQDef where reasVal     = reasV''`   
` --couldn't match type Expr with ConstrReasQDef -> f ConstrReasQDef`

JacquesCarette:
Seems like one is expecting a `Lens` while the other is just a function. In one direction, you can use `view`. The other is impossible, so might be a deeper mismatch.

Awurama-N:
Since i changed the constructors to match the new chunk types i was going through the examples to change those as well. in my most recent commit i changes `nomThick` from `cuc` to `cnstrw`. And i just wanted to make sure that was right before proceeding and pushing other changes

JacquesCarette:
It certainly is not wrong. Out of curiosity, why is the change needed? Why is `cuc` no longer the right constructor? [I want you to spell out the reasons, I am not questioning the decision itself.]

Awurama-N:
when i was editing the constructor definitions from ConstrainedChunk to the new chunk types:

```
-- | Creates a constrained unitary chunk from a 'UID', term ('NP'), 'Symbol', unit, 'Space', 'Constraint's, and an 'Expr'.
cuc :: (IsUnit u) => String -> NP -> Symbol -> u
  -> Space -> [Constraint] -> ConstrainedQDef
cuc i t s u space = ConstrainedQDef (qw (unitary i t s u space))

-- | Creates a constrained unitary chunk from a 'UID', term ('NP'), 'Symbol', 'Space', 'Constraint's, and a 'Maybe' 'Expr' (Similar to 'cuc' but no units).
cvc :: String -> NP -> Symbol -> Space -> Expr -> ReasonableValueQDef
cvc i des sym space = RVQD (qw (vc i des sym space))

-- | Creates a new ConstrReasQDef from either a 'ConstrainedChunk', 'ConstrConcept', 'UncertainChunk', or an 'UncertQ'.
cnstrw :: (Quantity c, Constrained c, HasReasVal c, MayHaveUnit c) => c -> ConstrReasQDef  
cnstrw c = CRQD   (qw c) (c ^. constraints) (c ^. reasVal)
```

for cuc i took out the reasonable value so it fit `ConstrainedQDef`, for cvc I took out the constraint so it fit `ReasonableValueQDef` so now `cnstrw` is the only constructor which is used when theres a constraint and a reasonable value.

And in the unitals file `nomThick` has both a constraint and a reasonable value. It hadn't raised any problems yet, but i thought it made sense to make the change.

JacquesCarette:
Great explanation - and yes, it makes perfect sense.

JacquesCarette (approved):
The changes as they are now are good. Is this ready to go?

Awurama-N:
Yes it can be merged, I am still going through the constructors to make changes but i can just push them after though

Awurama-N:
Should i go ahead and change wherever `HasReasVal` is used? since now its lens is `Expr` and not `Maybe Expr`.
I could change it to `MayHaveReasVal` instead so there are no errorrs of "couldn't match Expr to MAybe Expr"

JacquesCarette:
Change them to `MayHaveResVal`, since that will be simplest. We can do a second pass on that later.

There is a conflict, so can you merge master first?

Awurama-N:
Okay

JacquesCarette:
And now the build is failing...

Awurama-N:
yeah i was trying to figure out why, and I'm thinking it's from the inconsistencies with `Expr`.

```
/home/runner/work/Drasil/Drasil/code/drasil-lang/Language/Drasil/Chunk/Constrained.hs:125:59: error:
drasil-lang    >     • Couldn't match type ‘Expr’
drasil-lang    >                      with ‘ConstrReasQDef -> f ConstrReasQDef’
drasil-lang    >       Expected type: (Expr -> f Expr)
drasil-lang    >                      -> ConstrReasQDef -> f ConstrReasQDef
drasil-lang    >         Actual type: ConstrReasQDef -> Expr
drasil-lang    >     • In the expression: reasV''
drasil-lang    >       In an equation for ‘reasVal’: reasVal = reasV''
drasil-lang    >       In the instance declaration for ‘HasReasVal ConstrReasQDef’
drasil-lang    >     • Relevant bindings include
drasil-lang    >         reasVal :: (Expr -> f Expr) -> ConstrReasQDef -> f ConstrReasQDef
drasil-lang    >           (bound at Language/Drasil/Chunk/Constrained.hs:125:45)
```

balacij:
```
drasil-lang    >     • Couldn't match type ‘Expr’
drasil-lang    >                      with ‘ConstrReasQDef -> f ConstrReasQDef’
drasil-lang    >       Expected type: (Expr -> f Expr)
drasil-lang    >                      -> ConstrReasQDef -> f ConstrReasQDef
```
I think that means that it's looking for a `Lens' ConstrReasQDef Expr` but you gave it a plain Expr without a Lens. Note that `makeLenses` only makes lenses for things prefixed with a `_`, unless you specify with `makeLensesFor`.

Awurama-N:
oh okay thank you

JacquesCarette:
Let me know when the build goes through (it still isn't).

Awurama-N:
Okay I will let you know when the errors are fixed and the build goes through

Awurama-N:
is it okay if I make `MayHaveReasVal` a getter and then make changes to wherever in the code it's present to suit that?

JacquesCarette:
Yes.

Awurama-N:
UncertainChunk used ConstrainedChunk originally but now that ConstrainedChunk has been split into 3 different chunk types do we want to split UncertainChunk into 3 as well? or do we want to make 3 new constructors.
This issue is what is currently breaking the build.

JacquesCarette:
Take a look at what kinds of `UncertainChunk` are used in practice. We'll need to support what is currently in use.

Awurama-N:
### UncertainChunk:


### Reasonable value only
Drasil.GlassBr.Unitals:
- plateWidth

Drasil.SWHS.Unitals:
- absTol
- relTol

### Both constraint and reasonable value
Drasil.GlassBr.Unitals:
- pbTol
- tNT
- plateLen
- CharWeight


I am a bit unsure about where to place absTol & relTol,
```
unconstrained :: [UncertainChunk]
unconstrained = [absTol, relTol]
```

I  see they're labeled as unconstrained even though their definitions are in this format:

```
absTol = uvc "absTol" (nounPhraseSP "absolute tolerance") 
  (sub cA lTol) Real
  [physc $ Bounded (Exc, exactDbl 0) (Exc, exactDbl 1)] 
   (dbl (10.0**(-10))) (uncty 0.01 Nothing)
```

and physc is `RealInterval Expr Expr -> Constraint` and it's a smart constructor for a range of physical constraints between 2 expressions. But there's not a constraint list in it's definition like how `pbTol` has `[probConstr]` for example.

If they count as being constrained then everywhere UncertainChunk is used in the examples has both reasonable value and constraint.

Awurama-N:
Okay great. Should i create a ticket so calling them `unconstrained` can be changed?

Also this is the current definition for `UncertainChunk`:

data UncertainChunk  = UCh { _conc :: ConstrReasQDef , _unc' :: Uncertainty }
makeLenses ''UncertainChunk

JacquesCarette:
Yes on the ticket (but you might want to just do it here).  So it looks like the current definition of `UncertainChunk` doesn't have to change, if I'm reading things correctly.

Awurama-N:
yes okay, it seems like it already does what it's suppose to do so instead would we need to make more constructors? i'm trying to figure out what the next step should be to fix the error occuring right now

JacquesCarette:
Put the details of the error (i.e. where it occurs and what's the error) in the issue here.  Hopefully @balacij can help, as I'll be offline for the next hour.

Awurama-N:
Okay.

```
-- | Smart constructor that can project to an 'UncertainChunk' (also given an 'Uncertainty').
uncrtnChunk :: (Quantity c, Constrained c, MayHaveReasVal c, MayHaveUnit c) => 
  c -> Uncertainty -> UncertainChunk
uncrtnChunk q = UCh (cnstrw q)
```

Screenshot 2021-06-29 at 11 38 45 AM (2)

```
-- | UncertQs are conceptual symbolic quantities with constraints and an 'Uncertainty'.
-- Contains a 'ConstrConcept' and an 'Uncertainty'.
data UncertQ = UQ { _coco :: ConstrConcept , _unc'' :: Uncertainty }
makeLenses ''UncertQ
  
<img width="1092" alt="Screenshot 2021-06-29 at 11 38 55 AM (2)" src="https://user-images.githubusercontent.com/72299644/123827736-e2828800-d8ce-11eb-92c6-74d18a716d82.png">
```

JacquesCarette:
The error message is quite good: you have `MayHaveReasVal` as a class constraint, but need `HasReasVal` instead.

Awurama-N:
`HasReasVal` was what was originally there, but because I changed `HasReasVal` to be just `Expr` instead of `Maybe Expr` I thought it was right to change this as well from `HasReasVal` to `MayHaveReasVal` since it now does what `HasReasVal` previosuly did.

So should i just change that back then?

JacquesCarette:
Let the types tell you what to do. I think you require a reasonable value, so `Expr` is needed.

Awurama-N:
I reverted some changes i made since the last push and changed the `MayHaveReasVal` back to `HasReasVal`, but it brought back the old error i was originally trying to fix with these changes.

Screenshot 2021-06-29 at 4 15 21 PM

Screenshot 2021-06-29 at 4 13 22 PM

JacquesCarette:
Perhaps you're missing an instance? That seems to be what the error indicates.

JacquesCarette:
Perhaps if you could push what you think is something reasonable, I can take a deeper look.

JacquesCarette:
Thanks, looking now.

JacquesCarette:
In digging deeper, it seemed that this exact design was destined to fail. A new design will have to be drawn up.

Issue #2612, opened 2021-06-18
Best way to display a dependency graph/how to use graphviz


Ant13731:
Referring to the emails about creating new introduction documents, I think something that would be good to have is a graphical and a visual set (venn diagram-like) way of showing type dependencies. I was originally planning to use tikz (as I am somewhat familiar with it), but after looking at Nathaniel's work and the [drasil website dependency graphs](https://jacquescarette.github.io/Drasil/graphs/drasil-build.pdf), I was wondering if using latex is a good way to go about doing something like this. I think graphviz would be nicer as it has more colours/options for displaying, but I'm not really sure how to start something. Is there a good source that I could read to learn more, or should I stick with Tikz?

JacquesCarette:
Definitely `dot` is going to give you the most flexibility.

There are 3 big pieces of information:
- the relations between the data-structures (the ones under `Chunk/`
- the relations between the classes
- the relations between the classes and the data-structures.

The latter is what I asked Nathaniel to generate in those giant spreadsheets. The most important ones to focus on are the ones in `drasil-lang`. These are the core of Drasil.  All 3 of the above, when drawn, would yield some information about oddities in what we have. [There are known oddities, so we know that we should be able to see at least those; @Awurama-N is actively working on fixing one at the data-structure level.]

Ant13731:
Okay, I see. I just made a new branch that should start to cover the first piece, but it's still a WIP.

JacquesCarette:
I saw all the `.dot` files. Do you have scripts to automatically render them?

balacij:
@Ant13731 Regarding `DTG_Config.txt`, does `code-base` and `metadata` need to be added to the DTG_Config.txt file under the scripts folder? If so, do you know what the comment "(lowercase, space separated, in order of display in output file)" means from the file? I'm not quite sure I understand what "output file" it's referring to.

Ant13731:
I think the comment means that the order in which the folder names are given is the order the folders will appear on the type/class dependency .csv file that Nathaniel generated (its a giant spreadsheet that records types, classes, and instances). `DTG_Config.txt` is used when `make analysis` is called. I think those two new folders should be added, but I don't think the order really matters to the program, its just so we can change which folder appears first on the spreadsheet.

balacij:
Thanks! I'll make a PR for it then.

Ant13731:
Just a quick update on this issue:

[...]

PR #2621, opened 2021-06-23
Adding type dependency graph

### LINKED ISSUE #2620: PDFs for drasil-lang type dependencies
I have pushed the pdf files in the branch `addingTypeDependencyGraph` under the `People/Anthony/Type_Dependency_Graphs` folder. In that folder, there are two different folders with the pdfs inside (along with all the `.dot` files).
Here is the pdf including low-level type dependencies:
https://github.com/JacquesCarette/Drasil/blob/a3cf83b149242f8b281a2418ada0522d0e9f23c8/People/Anthony/Type_Dependency_Graphs/JoinedTypes/June_18_2021/drasil-lang.pdf
And here is another one with separated low-level dependencies (a little bit cleaner):
https://github.com/JacquesCarette/Drasil/blob/a3cf83b149242f8b281a2418ada0522d0e9f23c8/People/Anthony/Type_Dependency_Graphs/SeparatedLowLvlTypes/June_18_2021/drasil-lang.pdf
The autogen ones will probably look like the first one, but I figured I could separate it for now so it is easier to read. Some of the nodes are still separate from the entire graph, but that is probably because I didn't count types defined with `newtype` or `type`. The autogen graphs will though.

For the first graph:
The dotted lines signify that the tip of the arrow could contain the tail type, but may not have to (it is either a `Maybe type` or it is using constructors). Coloured nodes are for those that are defined. 

For the second graph:
The separated low-level graph should have separate nodes for `Maybe types` and will use a solid line instead. The dotted lines are just for type constructor values.

Please let me know anything I could improve/any changes I should make.

Ant13731:
Closes #2620 
Contributes #2612 
The autogen type graphs will now appear in the analysis folder after using `make analysis`. Each `drasil-` package has its own graph and it also generates a full graph of all types in drasil. 

One thing that isn't quite right:
- some types aren't read properly by the source code reader (especially with types that accept other types in their definition), though this functionality should be able to fit nicely with what I have right now. At the moment, these types are just defined with no spaces. They still show up on the graph, but their name might not be quite exact. For example (as defined in `Language.Drasil.Code.Imperative.GOOL.LanguageRenderer.SwiftRenderer`):
    ```Haskell
    type Auxiliary SwiftProject = AuxData
    type AuxHelper SwiftProject = Doc
    ```
    will show up as the types `AuxiliarySwiftProject`  and `AuxHelperSwiftProject` instead of a generalized `Auxiliary` and `AuxHelper` types.

The nodes of the graph have different colours (which can be changed easily) depending on the way the type is defined:
- record types using the `data` syntax will appear as `cyan3`
- non-record types using the `data` syntax will appear as `darkviolet`
- types using the `newtype` syntax will appear as `darkgreen`
- types using the `type` syntax will appear as `red2`

I think that most of my helper functions for the `SourceCodeReaderTypes.hs` file can be made much more concise by using `map` and `filter` instead of pattern matching, but I just wanted to get this PR out for now.


JacquesCarette (changes requested):
I don't think we really want all these `.dot` files under `People/Anthony` in the repo. I'm fine with them being generated by the analysis, but those ones will go stale really quickly, and I don't want anyone using them "by accident."

JacquesCarette (commented):
This is an 'approval' for the added commit. I still have to review the code.

code/Makefile
```diff
@@ -191,8 +194,16 @@ graphs: $(GRAPH_PACKAGES)
 ###
 
 analysis: check_stack
+	- rm -rf $(ANALYSIS_FOLDER)
 	@mkdir -p $(ANALYSIS_FOLDER)
 	cd $(SCRIPT_FOLDER) && stack exec -- runghc DataTableGen.hs
+	cd $(SCRIPT_FOLDER) && stack exec -- runghc TypeGraphGen.hs
+	stack install dotgen graphmod
```
JacquesCarette, line 201: typically, it's best to do the `stack install` first because if it fails, there's no point doing any of the rest.

JacquesCarette (changes requested):
The code could be cleaned up a lot (there is a lot of duplication, on top of the things that could be done with map/filter/fold), but it's fine for a first go, modulo the one comment.

JacquesCarette (approved):

Issue #2602, opened 2021-06-15
Move Actor/Messages+EnumeratedThings out of Expr, into new "CodeExpr"


balacij:
Related to discussion on https://github.com/JacquesCarette/Drasil/commit/ff34fe093b8898383232e2822f478a394243aef9#r52161989

balacij:
This task is proving a little bit more complicated than separating the DisplayExpr-related components. With this one, we frequently embed CodeExprs in Exprs, unlike with the DisplayExpr-related components.

JacquesCarette:
Concrete examples?

For `CodeExpr`, we might have to bite the bullet and make a parallel hierarchy.

balacij:
In drasil-code, we need to convert `FuncStmt` (since most `new` calls are within `FuncStmt`s) and `Step` (because of below `appendCurrSol`) to be using CodeExprs instead of Exprs, but when we do this, the ExternalLibrary related components are the first to breakdown.

For example, 
https://github.com/JacquesCarette/Drasil/blob/268bbdd5b552e2a4c51eebffcdf7303d2689d0a4/code/drasil-code/Language/Drasil/Code/ExternalLibrary.hs#L276-L283

^ we would be comparing a CodeExpr (field ..) to an Expr (ub)

https://github.com/JacquesCarette/Drasil/blob/268bbdd5b552e2a4c51eebffcdf7303d2689d0a4/code/drasil-code/Language/Drasil/Code/ExternalLibrary.hs#L270-L274

^ here the `v` is a CodeExpr, so we would need to instantiate a Matrix containing a single value; a CodeExpr -- we might need to have a Matrix idea in CodeExpr too.

https://github.com/JacquesCarette/Drasil/blob/268bbdd5b552e2a4c51eebffcdf7303d2689d0a4/code/drasil-code/Language/Drasil/Code/ExternalLibrary.hs#L292-L294

^ here the first parameter would need to become a CodeExpr, but we use that CodeExpr in the `idx` call, which means that we do in fact need a copy/idea of a function in the CodeExpr as well.

https://github.com/JacquesCarette/Drasil/blob/268bbdd5b552e2a4c51eebffcdf7303d2689d0a4/code/drasil-code/Language/Drasil/Code/ExternalLibrary.hs#L233-L237

I'm going to try to spend a bit more time seeing if I can the wrapped version going well. If it continues to be difficult, we'll try out the 1-1 way.

JacquesCarette:
[...]

Issue #2414, opened 2021-05-10
Expr `getUpperBound` causes errors we can potentially avoid


balacij:
> Please file an issue about getting rid of `getUpperBound` - it's a horrible hack. We should not have "get"-style functions that work on `Expr` at all. The knowledge should be encoded in other ways so that there is no `error` case. We'll first need to analyse all the uses of `getUpperBound` to design a proper replacement.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/2405#discussion_r629619744_

balacij:
There's a few more functions like `getUpperBound` -- `getInnerSpace`, `getActorName`, etc

balacij:
`getUpperBound`:
https://github.com/JacquesCarette/Drasil/blob/542f6aa85f23cde493536be924211fb3827b8a5d/code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs#L17-L20

Current usage of `getUpperBound` is limited to 1 instance:
```
-- | Converts a 'FuncStmt' to a GOOL Statement.
convStmt :: (OOProg r) => FuncStmt -> GenState (MSStatement r)
```
https://github.com/JacquesCarette/Drasil/blob/542f6aa85f23cde493536be924211fb3827b8a5d/code/drasil-code/Language/Drasil/Code/Imperative/Import.hs#L538-L542

We use `getUpperBound` to grab the right-hand side of a `a < b` expression.

`FFor` is only constructed via `ffor`, and `ffor` is only used in GlassBR's ModuleDefs, in a fashion similar to:
https://github.com/JacquesCarette/Drasil/blob/542f6aa85f23cde493536be924211fb3827b8a5d/code/drasil-example/Drasil/GlassBR/ModuleDefs.hs#L155-L156

It looks like we can create extra For-loop constructors to resolve this and get rid of `getUpperBound`.

PR #2622, opened 2021-06-23
Proto-CodeExpr

balacij:
I think we're _almost_ there with CodeExpr. I have  few notes in the code that I will review in this PR shortly.

However, notably, in this PR, we split up some files, create a new proxy package `drasil-code-base`, and create CodeExpr as a one-to-one copy of Expr, removing New/Message/Field from the original Expr.


code/drasil-code-base/Language/Drasil/Code/Expr.hs
```diff
@@ -0,0 +1,135 @@
+{-# LANGUAGE GADTs #-}
+
+module Language.Drasil.Code.Expr where
+
+import Language.Drasil (UID, DomainDesc, Completeness, DerivType, RealInterval)
+
+-- Binary functions
+
+-- | Arithmetic operators (fractional, power, and subtraction).
+data ArithBinOp = Frac | Pow | Subt
```
JacquesCarette, line 10: Maybe it would make sense to share all these `Op` types? They are not recursive, so they are independent.

code/drasil-code-base/Language/Drasil/Code/Expr.hs
```diff
@@ -0,0 +1,135 @@
...
+-- | Negation operator (not).
+data UFuncB = Not
+  deriving Eq
+
+-- | Data for vectors (normal, dimensions).
+data UFuncVec = Norm | Dim
+  deriving Eq
+
+-- | A one-to-one clone of Expr, with extra OO/code-related functionality.
+data CodeExpr where
```
JacquesCarette, line 60: This is the one that can't be shared.

code/drasil-code-base/Language/Drasil/Code/Expr/Precedence.hs
```diff
@@ -0,0 +1,95 @@
...
+import Language.Drasil.Code.Expr (CodeExpr(..), UFuncVec, UFuncB(..),
+    UFunc(..), AssocBoolOper(..), AssocArithOper(..), VVNBinOp,
+    VVVBinOp, OrdBinOp, LABinOp, BoolBinOp, EqBinOp, ArithBinOp(..))
+
+-- These precedences are inspired from Haskell/F# 
+-- as documented at http://kevincantu.org/code/operators.html
+-- They are all multiplied by 10, to leave room to weave things in between
+
+-- | prec2Arith - precedence for arithmetic-related binary operations.
+prec2Arith :: ArithBinOp -> Int
```
JacquesCarette, line 12: And then a lot of these could be shared too.

code/drasil-code-base/Language/Drasil/Code/Expr/Render.hs
```diff
@@ -0,0 +1,116 @@
...
+
+import qualified Language.Drasil as L
+import qualified Language.Drasil.Development as LD
+
+import Data.Bifunctor (Bifunctor(bimap, second))
+
+import Language.Drasil.Code.Expr
+
+-- | Render an algebraic expression into our code expression language
+renderExpr :: L.Expr -> CodeExpr
```
JacquesCarette, line 13: `render` made sense when it was more of a printing operation, and this propagated.  A simple `toCode` might be better, but I'm open to suggestions that are less printing-oriented.

JacquesCarette: Also, because this whole file is about 'rendering', there is no point embedding that in the name too.  So `expr`, etc, might end up being best?

code/drasil-code/Language/Drasil/Chunk/CodeDefinition.hs
```diff
@@ -15,53 +15,52 @@ import Control.Lens ((^.), makeLenses, view)
...
 qtov :: QDefinition -> CodeDefinition
-qtov q = CD (codeChunk $ quantvar q) (q ^. defnExpr) [] Definition
+qtov q = CD (codeChunk $ quantvar q) (renderExpr $ q ^. defnExpr) [] Definition
 
 -- | Constructs a 'CodeDefinition' for an ODE.
 odeDef :: ODEInfo -> CodeDefinition
-odeDef info = CD (codeChunk $ quantfunc $ depVar info) (Matrix [odeSyst info]) 
-  (map ($ info) [tInit, tFinal, initVal, absTol . odeOpts, relTol . odeOpts, 
+odeDef info = CD (codeChunk $ quantfunc $ depVar info) (matrix [map renderExpr $ odeSyst info])
+  (map (renderExpr . ($ info)) [tInit, tFinal, initVal, absTol . odeOpts, relTol . odeOpts,
```
balacij, line 65: I'm not yet sure if we want to change `ODEInfo`s to use `CodeExpr`s instead of `Expr`s.
 https://github.com/JacquesCarette/Drasil/blob/5232d494995ca2d558dc60c2c81a8d4682a3c2e9/code/drasil-code/Language/Drasil/Data/ODEInfo.hs#L14-L23

Any thoughts on this?

code/drasil-code/Language/Drasil/Code/ExtLibImport.hs
```diff
@@ -4,12 +4,13 @@
 module Language.Drasil.Code.ExtLibImport (ExtLibState(..), auxMods, defs, 
   imports, modExports, steps, genExternalLibraryCall) where
 
-import Language.Drasil
+import Language.Drasil (HasSpace(typ), getActorName, NamedArgument)
```
JacquesCarette, line 7: It surprises me that `getActorName` is exported by `Language.Drasil`. Wouldn't this be move to `drasil-code`?

code/drasil-code/Language/Drasil/Chunk/CodeDefinition.hs
balacij: I would like to change it to use CodeExprs, but I'm not quite sure how the ODEs would use any objects as of right now.

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
```diff
@@ -15,7 +15,7 @@ liftS :: State a b -> State a [b]
 liftS = fmap (: [])
 
 -- | For an expression using a less than operator, gets the upper bound.
-getUpperBound :: Expr -> Expr
+getUpperBound :: CodeExpr -> CodeExpr
```
JacquesCarette, line 18: Can you open an issue to get rid of this function? We shouldn't doing 'introspection' like this.

code/drasil-example/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -86,41 +88,40 @@ filename = var "filename" "name of file with x y and z data"
 -- Some semantic functions
 
 -- Given two points (x1,y1) and (x2,y2), return the slope of the line going through them
-slope :: (Expr, Expr) -> (Expr, Expr) -> Expr
+slope :: (CodeExpr, CodeExpr) -> (CodeExpr, CodeExpr) -> CodeExpr
```
JacquesCarette, line 91: This is very interesting. `slope` is not necessarily a `CodeExpr` generator (though that is indeed how it is used). It kind of wants to be polymorphic over both [though I don't want to go there right now.]

Other routines below really are "pure code".

I don't want you to change anything here, this is just a comment.

code/drasil-lang/Language/Drasil/Constraint.hs
```diff
@@ -9,26 +9,30 @@ import Language.Drasil.Space (RealInterval(..))
 
 -- | The reason behind the constraint's existence.
 data ConstraintReason = Physical | Software
+
+-- | Type synonym for 'ConstraintE'
+type ConstraintE = Constraint Expr
+
 -- | Holds constraints. May occur between an interval of 'Expr', a list of 'Double's, or a list of 'String's.
-data Constraint where
+data Constraint a where
```
JacquesCarette, line 17: This is a nice generalization.

JacquesCarette (changes requested):

code/drasil-code/Language/Drasil/Chunk/CodeDefinition.hs
JacquesCarette: I think all of this can (and should) be changed to `CodeExpr`. If we need an ODE representation as an `Expr`, it will have to have its own definition there.

balacij: Sounds good. I'll get this done and let you know when it's completed.

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
balacij: Ah, yes, we had this issue before actually. #2414 is the related ticket, we only have 1 usage of it -- and it's very small and in `drasil-code`, I think we can add it to the scope of this PR too.

code/drasil-example/Drasil/GlassBR/ModuleDefs.hs
balacij: Thanks, yes, I agree. I believe this particular piece of code originally (pre-me starting the Expr split up) had used the Num typeclass to constrain the points so this code was more generalized then.

balacij: I'm not yet sure how this piece will look like when we have typed Expr, but hopefully it will revert to the original Num constraint (maybe it was Num, Fractional actually).

code/drasil-lang/Language/Drasil/Constraint.hs
balacij: Thanks!

code/drasil-code-base/Language/Drasil/Code/Expr/Render.hs
balacij: Yep, I agree. I changed the name of the function to just 'expr', and I also renamed the file to Convert.hs, I think it's a bit more suitable.

code/drasil-code/Language/Drasil/Chunk/CodeDefinition.hs
balacij: Fixed -- I think it looks a lot better now. Thanks!

code/drasil-code/Language/Drasil/Code/ExtLibImport.hs
balacij: `getActorName` is another function like `getInnerSpace` -- exact same functionality but instead of it being for the "Vect" Space, it's for the "Actor" Space. I guess, if anything, it would be another function marked for removal, like `getInnerSpace`.

JacquesCarette (approved):

code/drasil-code-base/Language/Drasil/Code/Expr.hs
balacij: Would we ever want to add some sort of typing to `CodeExpr`? If we do, we might want to keep it as is and continue to have it closely follow `Expr` in development, otherwise, I think we should definitely share the `Op` types, it would definitely tidy things up!

JacquesCarette: We will want to have both `Expr` and `CodeExpr` be typed. So if sharing might interfere with doing that, let's wait.

balacij: Sounds good, let's wait a bit to see how it ends up. Thanks!

code/drasil-code/Language/Drasil/Code/Imperative/Helpers.hs
balacij: There's a few too many to add on, I actually confused `getUpperBound` above with another, `getInnerSpace` which I used in an example. Let's leave it to another PR -- I'll try to get it done today however.

code/drasil-example/Drasil/NoPCM/ODEs.hs
```diff
@@ -0,0 +1,19 @@
...
+  absTol, relTol)
+
+
+noPCMODEOpts :: ODEOptions
+noPCMODEOpts = odeOptions RK45 (sy absTol) (sy relTol) (sy timeStep) (exactDbl 0)
+
+noPCMODEInfo :: ODEInfo
+noPCMODEInfo = odeInfo (quantvar time) (quantvar tempW)
+  [quantvar tauW, quantvar tempC] (exactDbl 0) (sy timeFinal) (sy tempInit)
+  [recip_ (sy tauW) `mulRe` (sy tempC $- idx (sy tempW) (int 0))] noPCMODEOpts
```
balacij, line 19: I forgot to mention this -- I moved the ODEs to a new file so that we could create `CodeExpr`s instead of `Expr`s by importing `Language.Drasil.CodeExpr` instead of `Language.Drasil` -- it has the same function names and operators, so it looks the same but creates different things (CodeExpr vs Expr). If we wanted to keep it in the original Body.hs file, we'd probably need to import the CodeExpr with a qualified name, which would also make the equations a bit more difficult to read.

balacij:
Contributes to #2602

code/drasil-example/Drasil/NoPCM/ODEs.hs
JacquesCarette: Saw that - good move.

JacquesCarette (approved):

JacquesCarette:
I'm actually going to hold off on merging, and let the one from @Awurama-N come first. There will likely be non-trivial merge conflicts, and I'll want you to deal with them.

balacij:
Sounds good. Thank you! :smile:

PR #2624, opened 2021-06-24
Display name reference

Ant13731:
Contributes #1142 

I also changed the original example referenced in  #1142 (the introduction section) to use the new reference names.

Would the next step be to create an issue outlining some places where we could implement these changes?


JacquesCarette (approved):
Given our discussions, the complexity of referencing, and the sheer size of this PR, I was concerned that it would take me forever to review - and that it would take a while before it could be merged. So I am quite pleasantly surprised that this was not so! It's ready to go.

Issue #2152, opened 2020-05-28
Makefile should require and run `hlint` locally


JacquesCarette:
That would help people not check in things that will break the build.

Mornix:
I touched on this in #2151, but not including HLint in the build was intentional to avoid friction from intermediate/working changes failing a build, or if the return code is to be ignored in the Makefile, that may mean the terminal messages output by HLint may be ignored or missed before code is pushed. I suggested that some depending on individual developer preference either `hlint . && make` or `hlint .; make` can be run to achieve the desired result of an additional Makefile target.

JacquesCarette:
Or have an `hlint` target in the Makefile. I'm ok with not having it on by default. The usefulness of having it in the Makefile is that we can automate dealing with installing + dependencies.

JacquesCarette:
Do you still want this assigned to yourself?

balacij:
We now have `hlint` + `hot_hlint` targets. HLint seems to take ~1min to run, so it may be best to leave it as a target and not have it run with each build, or else it would cause a bit of delay in debugging one's own work. I think that we should run `make hlint` or `make hot_hlint` (preferred) before creating PRs from branches we've been working on.

JacquesCarette:
We should lint when we also run tests but not when we just build. Having a separate target definitely makes sense.

balacij:
That sounds good. If I understand you correctly, then I think we can close this ticket. The expected workflow would be to manually run `make` followed by a `make hot_hlint` to see if code is "PR Ready". We could also add a new target "pr_ready" which just has a dependency on both "all" and "hot_hlint". What do you think?

JacquesCarette:
`make pr_ready` is an excellent suggestion.

balacij:
I just need to remember to add this to the wiki in a place or two, and then we should be able to close this.

balacij:
I added the `make pr_ready` in a few places across the wiki already, but I also added a few more to the `Workflow` page for good measure. I'm going to close this ticket, but, of course, we can re-open it if desired.

PR #2625, opened 2021-06-24
Add a `hlint` + `hot_hlint` Makefile target

balacij:
* Added `.hlint*` to the .gitignore
* Wrapped string fields in double quotes in the makefile
* Add 2 new make targets -- `hlint` and `hot_hlint` for using hlint, respectively, via a local installation or via hot download


balacij:
This ticket does also contribute to #2152. The only issue with adding it as a hard requirement is that the hlint version on Stack (which would be installable on user machines via `stack`) gives an error on our current code, but the latest hlint version does not -- if we were to add hlint as a requirement, we would just need to decide on which version to use.

JacquesCarette (approved):
Having the targets is good. But yes, now seems premature to make it required.

PR #2634, opened 2021-06-25
capitalized pendulum in the introduction

### LINKED ISSUE #2615: Update Introduction for Double Pendulum
The introduction for Double Pendulum does not follow the pattern of the other examples.  Please make the following changes:

- Please modify the wording so that the Pendulum is capitalized in the last word of the first paragraph.
- The introduction says that the purpose will be presented, along with other parts that are not included.  We should add those other parts to the document.  If adding the other parts is a significant amount of work @cd155, please create issues for each subsection.

cd155:
- The Pendulum is capitalized in the last word of the first paragraph.

- I created additional three tickets to add three new sections.

close #2615

### DIFF
--- a/code/drasil-example/Drasil/DblPendulum/Body.hs
+++ b/code/drasil-example/Drasil/DblPendulum/Body.hs
@@ -102,7 +102,7 @@ justification = foldlSent [ atStartNP (a_ pendulum), S "consists" `S.of_` phrase
                             (S "highly sensitive to initial conditions" !.), S "Therefore" `sC`
                             S "it is useful to have a", phrase program, S "to simulate", phraseNP (motion
                             `the_ofThe` pendulum), (S "to exhibit its chaotic characteristics" !.),
-                            atStartNP (the program), S "documented here is called", phrase pendulum]
+                            atStartNP (the program), S "documented here is called", phrase pendulumTitle]
 scope :: Sentence
 scope = foldlSent [phraseNP (NP.the (analysis `ofA` twoD)), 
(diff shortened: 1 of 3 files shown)


smiths (changes requested):
@cd155, when you make changes that cause changes in the generated code, you need to also commit the corresponding changing in the generated code.  We put the correct generated code in the stable folder.  The build process checks that any changes generate the "correct" code.  If we change our opinion on what is correct, then we need to update stable to reflect this change.

JacquesCarette (approved):

JacquesCarette:
@smiths could you approve too? The changes you requested have been done.

smiths (approved):

JacquesCarette:
There's something quite satisfying about pushing that 'merge' button, isn't there?

smiths:
Yes!

cd155:
Is it a routine to delete the branch once a pull request is completed?

smiths:
Yes, we usually delete branches after the feature is merged because otherwise we end up with multiple branches in the future and we don't know which ones have useful code in them.  I restored the branch, in case you still need it for something.  If you don't need it, we should delete it again.

cd155:
No problem, thanks for the explanation. I will delete the brach.

For the next time, I think I should name the branch in a more meaningful name.

smiths:
Yes, meaningful names for branches are helpful.  You also want to keep your branch to one feature to resolve one issue, whenever possible.  I'll make an effort to ask next time before deleting the branch.  :smile:

PR #2636, opened 2021-06-28
Make RealI use Expr instead of a UID

balacij:
Builds on #2622 + #2626

Just adds a bit of extra flexibility, we can stop exporting the constructor to not allow the extra flexibility however.


code/drasil-lang/Language/Drasil/Expr.hs
```diff
@@ -136,10 +136,8 @@ data Expr where
   --   of an 'Expr'.  Could be called BigOp.
   --   ex: Summation is represented via 'Add' over a discrete domain.
   Operator :: AssocArithOper -> DomainDesc Expr Expr -> Expr -> Expr
-  -- | The expression is an element of a space.
-  -- IsIn     :: Expr -> Space -> Expr
-  -- | A different kind of 'IsIn'. A 'UID' is an element of an interval.
-  RealI    :: UID -> RealInterval Expr Expr -> Expr
+  -- | A different kind of 'IsIn'. An 'Expr' is an element of an interval.
+  RealI    :: Expr -> RealInterval Expr Expr -> Expr
```
JacquesCarette, line 140: The reason for UID is that we really only want 'names' as the first component. It's almost ok if the data-structure allows more, as long as the smart constructor does not.

code/drasil-lang/Language/Drasil/Expr/Math.hs
```diff
@@ -130,8 +130,8 @@ prodAll v = Operator MulRe (AllDD v Discrete)
 -- TODO: Above only does for Reals
 
 -- | Smart constructor for 'real interval' membership.
-realInterval :: HasUID c => c -> RealInterval Expr Expr -> Expr
-realInterval c = RealI (c ^. uid)
+realInterval :: (HasSymbol c, HasUID c) => c -> RealInterval Expr Expr -> Expr
+realInterval c = RealI (sy c)
```
JacquesCarette, line 134: So this is a bit odd for the user: you can put in anything, as long as it has a symbol, and that's what's used, everything else is forgotten. It's not wrong, but I'm not 100% sure that's what we want?

code/drasil-lang/Language/Drasil/Expr.hs
balacij: I was thinking that this could add in extra flexibility to have those non-name cases, e.g., `1 < (x / 2pi) < 5`. However, I now also think there is also benefit to changing it to Expr for when we add typing. It seems a bit difficult to "add type information" from UIDs. Having Exprs instead, where the typing is will already be done by the `sy` (though, we might need to later replace `sy` with variants for each type or somehow allow it to gain the type information from UIDs), will allow for later processing to be done a bit smoother.

I started writing some general thoughts on a new [jasonWriting](https://github.com/JacquesCarette/Drasil/tree/jasonWriting) branch and coming across `C`, I was thinking that we would need to do something like this:
https://github.com/JacquesCarette/Drasil/blob/846c0308c87b675ccee16cca4c54f2da5de51e14/People/Jason/Typing.lhs#L83-L105
So I think this will, at least, allow us to not have to later split RealI's smart constructors into multiple ones as well to create the typing information as well.

The smart constructor currently does not allow the extra flexibility. Would there be any issues if we were to allow it?

code/drasil-lang/Language/Drasil/Expr/Math.hs
balacij: From how we use the UID of the `RealI`, I didn't think this makes any large difference, it would just be bringing the computation of `sy (c ^. uid)` up here instead of later down in the code where it would render the symbol.

Sorry, I might be missing a bit of information here actually. Could you please explain what we want from `RealI`? I thought it was just a way of encoding `a < b`, `a < b < c`, and related information so that we can render it properly and display it nicely.

code/drasil-lang/Language/Drasil/Expr.hs
JacquesCarette: So I'm not sure we want this extra flexibility! I would want to rewrite such things to isolate the `x`. We want `RealI` to mean "bounding interval for a variable".

code/drasil-lang/Language/Drasil/Expr/Math.hs
JacquesCarette: The information it is meant to capture is that a particular quantity (which is crucially not a computed quantity, but a named quantity) belongs to a certain interval. That's what needs to be in `Expr`.  In `DisplayExpr`, we can be sloppier, as we won't try to reason about the meaning of it.

JacquesCarette:
Note that the change away from `UID` is ok. For now, `Expr` is not great, but hopefully we can tighten that later when things to typed. I'm still going to wait on merging until the conversations above 'settle'.

code/drasil-lang/Language/Drasil/Expr.hs
balacij: Ah, ok, that makes sense. In that case, holding the UID would definitely be better than carrying it in an `C`. I think we can cancel this PR on this alone then.

code/drasil-lang/Language/Drasil/Expr/Math.hs
balacij: I never knew that we had a logistical restriction on the UIDs allowed here. I think I fully understand now. Thank you!

balacij:
The change to `Expr` wouldn't change too much, so I think we can just reject this PR, perhaps revisiting it when we reconsider the encoding style later when we go typed. I was missing a bit of information when I created this PR. What do you think?

JacquesCarette:
Right - closing.

JacquesCarette:
Should the branch be deleted too?

balacij:
Yep, that's fine, it's fairly easy to redo later if need be.

Issue #98, opened 2016-12-19
Inventory / Re-design Data Structures which Implement Chunks


szymczdm:
We need to take a look at the existing data structures we have (the ones implementing our different "chunk" classes) and see what's good/bad for each. Things will need to be changed for the upcoming chunk hierarchy overhaul, so here goes.

# The current data structures we have (in master) are:

### Chunk.hs
- ConceptChunk
```
data ConceptChunk = CC String Sentence 

instance Eq ConceptChunk where
  c1 == c2 = (c1 ^. name) == (c2 ^. name)
instance Chunk ConceptChunk where
  name f (CC a b) = fmap (\x -> CC x b) (f a)
instance Concept ConceptChunk where
  descr f (CC a b) = fmap (\x -> CC a x) (f b)
```
- DefinedTerm
```
data DefinedTerm = DCC String Sentence Sentence

instance Eq DefinedTerm where
  c1 == c2 = (c1 ^. name) == (c2 ^. name)
instance Chunk DefinedTerm where
  name f (DCC n t d) = fmap (\x -> DCC x t d) (f n)
instance Concept DefinedTerm where
  descr f (DCC n t d) = fmap (\x -> DCC n x d) (f t)
instance ConceptDefinition DefinedTerm where
  cdefn f (DCC n t d) = fmap (\x -> DCC n t x) (f d)
```
- VarChunk
```
data VarChunk = VC { vname :: String
                   , vdesc :: Sentence
                   , vsymb :: Symbol}

instance Eq VarChunk where
  c1 == c2 = (c1 ^. name) == (c2 ^. name)
instance Chunk VarChunk where
  name f (VC n d s) = fmap (\x -> VC x d s) (f n)
instance Concept VarChunk where
  descr f (VC n d s) = fmap (\x -> VC n x s) (f d)
instance Quantity VarChunk where
  symbol f (VC n d s) = fmap (\x -> VC n d x) (f s)
```

### Unit.hs
- FundUnit
```
data FundUnit = UD { _vc :: ConceptChunk, _u :: USymb }

instance Chunk FundUnit where
  name = vc . name
instance Concept FundUnit where
  descr = vc . descr
instance Unit FundUnit where
  unit f (UD a b) = fmap (\x -> UD a x) (f b)
```
- DerUChunk
```
data DerUChunk = DUC { _uc :: FundUnit, _eq :: UDefn }

instance Chunk   DerUChunk where name  = duc . name
[...]

PR #2638, opened 2021-06-28
Transferring Website

### LINKED ISSUE #2524: Swift; missing doxygen generation

Ant13731:
Closes #2524
Website now generates similar to drasil-examples.

Edit: The website is currently up here: https://ant13731.github.io/Drasil/


JacquesCarette (approved):

balacij:
It doesn't look like the index.html file got updated with this PR? We should be on Build #98 but it's still on 96. Any ideas what happened @Ant13731?

https://github.com/JacquesCarette/Drasil/actions/runs/983598915
https://github.com/JacquesCarette/Drasil/blob/gh-pages/index.html

Ant13731:
Ohhh maybe, I don't think my website has the build number on it - that was something that hakyll added at the end of the HTML file when it was generated. That would also explain that first PR on my fork where the build had to be set to zero. 

I'm not sure how to encode that on a drasil-generated html file though. Hakyll used a footer for it.

Ant13731:
Does the deploy bot only activate if the build number is greater according to the html file?

balacij:
The build-num should be fine on this repository since the .build-num was less than the action id of the last run.

The deploy bot activates regardless of the build number, but the scripts change what they setup depending on the build num.

The build number isn't necessarily added by Hakyll on it's own, it was configured by the template + passing as a variable.

The PR on your fork didn't work because the copied over .build-num from gh-pages was at ~90, but you had no actions runs, so your "latest action run id" was < 90. So, we had to bring yours down to 0 to re-align it with your actions runs.

Ant13731:
Yea, I think I already have it there - I just need to append it to the generated file?

Ant13731:
Ah, wait that might be tricky. The drasil generator ends the html file before I can append it. But the variables are still available in the `Main.hs` file.

balacij:
I doubt that appending it to the generated file would fix this issue of deployment though. It doesn't look like it copied the generated index.html file over the existing one on gh-pages, let's try testing the CI steps locally to see what happened.

Ant13731:
Okay, I'll try running through the build.yaml file locally.

balacij:
Just created a new branch so we can sort this out. I pushed a mock deployment script. Let's keep it since we seem to need it often.

https://github.com/JacquesCarette/Drasil/compare/fixDeployScript

It doesn't look like the `build_website` function of `deploy_stage.sh` builds the website by the way, so locally, on my machine, there's nothing for it to copy over for the `  cp -r "$CUR_DIR"drasil-website/Website/. .` step.

balacij:
Oh, do we just need to add `make website` CI step?

balacij:
I think that's all we need actually. Let me push it, I'll test it on my fork.

Ant13731:
Ah okay, I'm not too sure. I ran the rn.sh script you gave me a while back on my drasil fork and it added the old version of the drasil website (as `Drasil_Website.html` instead of `index.html`)

balacij:
Testing here:
Waiting on this run: https://github.com/balacij/Drasil/runs/2945710491
Hopefully, this page will get updated: https://balacij.github.io/Drasil/

Using code from https://github.com/JacquesCarette/Drasil/compare/fixDeployScript

Though, I'm wondering if you already had a `make website` somewhere and there's something else going wrong?

Ant13731:
I made website one of the dependent targets for deploy_lite, so I had thought that would've built the website. Maybe Build.yaml doesn't run deploy_lite though?

balacij:
Oh, that's probably a better way to do it than adding a `make website` to the Build.yaml file. Though, I don't see where you added it as a dependency of the deploy_lite target?

`make deploy_lite` is what's run from `deploy.bash`, so it _should_ be run.

Ant13731:
Hmm, I thought I put it in. Let me check

balacij:
Sorry, why wouldn't the `make deploy_lite` be run? It's called in `try_deploy` of the `deploy.bash` file. These scripts could probably be shortened quite a bit now that we have drasil-website too.

Ant13731:
Sorry, I think I was looking at the wrong file. I didn't see that the scripts call make deploy_lite (I thought only the Build.yaml called the makefile targets).

balacij:
Ah, ok. Hopefully this works then.

Maybe we should do a round of cleaning up the deployment scripts to fix that? I like your idea of having the Build.yaml be the only thing call the Makefile targets.

Right now it seems the deploy_lite target really just gathers the files together, so we might want to alter things a little bit around that logic since the `make website` target already builds the website. In the old method, the `deploy_lite` target would also build the website code, the website, and then gather the files together.

Ant13731:
Should the deploy_lite stuff be put into a sh file then? That way the makefile target calls the script and any needed dependencies?

Ant13731:
It looks like the website is up and running on your fork as well!

balacij:
Hmm, I think we can remove `deploy_wrapper.bash`. I think we only use it in `mock_deploy.sh` now and the file is just 1 line, so we can probably just merge it into mock_deploy.sh. The $SHELL_OPTS_FILE is faked from `mock_deploy.sh` anyways.

Afterwards, I think it would be good for us to look into the Makefile and update it to create local deployments entirely. It would probably end up shortening the `make deploy`, `make deploy_lite`, Build.yaml/Test.yaml, and the `scripts/deploy*` files. Eventually, we can merge the `mock_deploy.sh` script into the Makefile too.

PR #2643, opened 2021-06-30
Fix typing of pattern matches for CodeExpr and Expr's add and mul functions

balacij:


code/drasil-example/Drasil/GlassBR/DataDefs.hs
```diff
@@ -216,7 +216,7 @@ probOfBreak = dd probOfBreakQD (map ref [astm2009, beasonEtAl1998]) Nothing "pro
 
 --DD14--
 calofCapacityEq :: Expr
-calofCapacityEq = mulRe (mulRe (sy nonFL) (sy glaTyFac)) (sy loadSF)
+calofCapacityEq = sy nonFL `mulRe` sy glaTyFac `mulRe` sy loadSF
```
balacij, line 219: `loadSF` here is a constant (1) but since we load it via a `sy loadSF`, we aren't able to optimize it away via the `mulRe`. The generated code still contains the "* 1.0";
https://github.com/JacquesCarette/Drasil/blob/b8da80982650c50c1abc97844fc35bd0a8211bc1/code/stable/glassbr/src/java/GlassBR/Calculations.java#L164

JacquesCarette: This is correct, i.e. according to the current design of Drasil.

JacquesCarette (approved):

PR #2648, opened 2021-06-30
Ch capitalization terms

### LINKED ISSUE #2635: `Ch` constructor should know if a term should be titleized
Working off of #2489, one issue I was also having (that just happened to pop up in the same bit of code) was the need for duplicate titleized version of combinators. For reference, here are combinators with "and" (both normal and titleized version).
```Haskell
and_ :: (NamedIdea c, NamedIdea d) => c -> d -> NP
and_ t1 t2 = nounPhrase''
  (phrase t1 `S.and_` phrase t2)
  (phrase t1 `S.and_` plural t2)
  CapFirst
  CapWords

and_TSP :: (NamedIdea c, NamedIdea d) => c -> d -> NP
and_TSP t1 t2 = nounPhrase'' 
  (phrase t1 `S.and_` plural t2)
  (phrase t1 `S.and_` plural t2)
  (Replace (atStart t1 `S.and_` plural t2))
  (Replace (titleize t1 `S.and_` titleize' t2))
```
These really shouldn't be needed, as the first one contains all the information needed to display the a `NamedIdea`. So I dug a little deeper, and I think I found the cause. In the instance declaration for `NP` as a `NounPhrase`, the sentence cases and title cases are defined as follows:
```Haskell
  sentenceCase n@(Phrase _ _ r _)   f = cap (f n) r
  titleCase n@(Phrase _ _ _ r)      f = cap (f n) r
```
And the `cap` function:
```Haskell
cap :: Sentence -> CapitalizationRule -> Sentence
cap _ (Replace s) = s
cap (S (s:ss)) CapFirst = S (toUpper s : ss)
cap (S s)      CapWords = capString s capFirstWord capWords
cap (S s1 :+: S s2 :+: x) r = cap (S (s1 ++ s2) :+: x) r
cap (s1 :+: s2) CapWords = cap s1 CapWords +:+ capTail s2
cap (s1 :+: s2) CapFirst = cap s1 CapFirst :+: s2
cap a _ = a
```
This works for all sentences that are made using the `S` constructor, but falls apart when combining concepts together. This is because `cap` doesn't account for the `Ch` `Sentence` constructor, which is used in concept-level combinators. I'll illustrate the process using an example with the terms `terminology` and `definition`. Our end goal is to display `Terminology and Definitions`:
```Haskell
--Combine them:
tAndD = terminology `and_` definition

--Then, we need the title case, so we call `titleizeNP`. 
--This breaks down into the `titleCase` method defined in the `NounPhrase` class (shown above).
-- Which then calls the `cap` function and breaks down as follows:
titleizeNP tAndD = cap (pluralNP tAndD) CapWords
                = cap (pluralNP (terminology `and_` definition)) CapWords
                = cap (phrase terminology +:+ S "and" +:+ plural definition) CapWords
                = cap (Ch TermStyle terminology +:+ S "and" +:+ Ch PluralTerm definition) CapWords
```
So a possible problem lies in the `cap` function, which doesn't know how to deal with a `Ch` constructor.

Two possible ideas for a fix could be:
-  allow `Ch` to take in a `CapitalizationRule` so that the printers could know whether or not to capitalize a term when the term is being looked up in the database. So then its constructor would become:
    ```Haskell
    Ch :: SentenceStyle -> CapitalizationRule -> UID -> Sentence
    ```
- change `SentenceStyle` to accept `CapPluralTerm`, `CapTermStyle`, and `CapShortStyle` in addition to the existing constructors (shown below)
    ```Haskell
    data SentenceStyle = PluralTerm
                   | SymbolStyle
                   | TermStyle
                   | ShortStyle
    ```

Ant13731:
- Removed unneeded combinators. 
- Ch now understands CapFirst and CapWords (as CapF and CapW).
- Made stable more cohesive (no titleized descriptions in the Table of Symbols section). However, GamePhys still has a lot of titleized descriptions that are probably best left for another PR.

Two weird things still: 
- The third-last pattern match of `cap` uses `+:+` instead of `:+:`:
     ```Haskell
    cap (s1 :+: s2) CapWords = cap s1 CapWords +:+ capTail s2
    ```
    Which really shouldn't be needed. Spaces are encoded into `NounPhrases` at the time they are being made, and every combinator function includes spaces as well. Using the latter breaks all the section titles for stable. I originally thought the definition of `+:+` could be the cause of the problem, but replacing it with the manual version (`:+: S " " :+:`) still has the same effect on combinators. I think something may be getting rid of the spaces in NounPhrases and I feel like `cap` should not have its own spaces added. I tried looking through the code (specifically `lang` and `printers` folders) but I couldn't find anything that would cause this to happen. Leaving it seems fine for now, but I thought I should mention it.
- The 5th pattern match of `cap` ignores all capitalization after a symbol:
    ```Haskell
    cap (P symb :+: x) CapWords = P symb :+: x
    ```
    This is used in descriptions of the form "x-component". Without this, it displays as "x-Component". This is expected, however its needed for the `CapWords` case when really it should only be used in the `CapFirst` case. Currently using a temp fix for now by ignoring everything after a P symbol.

Closes #2635


JacquesCarette:
The two "weird things" should be made into issues (if they have not already).

code/drasil-printers/Language/Drasil/Printing/Import/Helpers.hs
```diff
@@ -51,15 +53,28 @@ lookupC :: Stage -> ChunkDB -> UID -> Symbol
 lookupC Equational     sm c = eqSymb   $ symbResolve sm c
 lookupC Implementation sm c = codeSymb $ symbResolve sm c
 
--- | Look up a term given a chunk database and a 'UID' associated with the term.
-lookupT :: ChunkDB -> UID -> Sentence
-lookupT sm c = phraseNP $ termResolve sm c ^. term
+-- | Look up a term given a chunk database and a 'UID' associated with the term. Also specifies capitalization
+lookupT :: ChunkDB -> UID -> TermCapitalization -> Sentence
+lookupT sm c NoCap = phraseNP $ termResolve sm c ^. term
```
JacquesCarette, line 58: I think it would make sense to create a function from `TermCapitalization` to a `Sentence` transformer, i.e.
```haskell
foo NoCap = phraseNP
foo CapF = atStartNP
foo CapW = titleizeNP
```
(with a better name), and then various other functions here could go back to being 1-liners.  Probably a primed version will be needed too. [Can be done in a new PR]

JacquesCarette (approved):

PR #2656, opened 2021-07-05
Autogen traceability graphs

### LINKED ISSUE #801: Automatically generate traceability graphs
As mentioned in the comments of commit 2600f0a, the note that was present in the introduction for traceability graphs

> NOTE: Building a tool to automatically generate the graphical representation of the matrix by scanning the labels and reference can be future work.

is related to Drasil more than it's related to an SRS' content.

This issue is being made to ensure that the removal of the sentence from the current implementation doesn't lead to the future-work task being ignored.

Ant13731:
Closes #801 
I think this is close to what Nathaniel hand-made. Running `make traceygraph` will create pdfs of the graphs in the `build` folder. I still have to remove his hand-made generator before this merges with master, but I wanted to get your thoughts on the graphs so far. 

Should the dot files remain in stable? I don't think they will change much anyways.


code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -0,0 +1,84 @@
...
+mkGraphEdges cols rows si = makeTGraph (ensureItems "Traceability Graph" $ traceGRowHeader rowf si) (traceMColumns colf rowf cdb) $ traceMReferees colf cdb
+    where
+        cdb = _sysinfodb si
+        colf = layoutUIDs cols cdb
+        rowf = layoutUIDs rows cdb
+
+makeTGraph :: [String] -> [[String]] -> [String] -> [(String, [String])]
+makeTGraph rowName rows cols = zip rowName [zipFTable' x cols | x <- rows]
+  where
+    zipFTable' content = concatMap (\x -> if x `elem` content then [x] else [""])
```
JacquesCarette, line 63: I think you can use `filter` here and avoid singleton lists?

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -0,0 +1,84 @@
...
+
+getLabels :: TraceViewCat -> SystemInformation -> [UID]
+getLabels l si = mkGraphNodes [l] si
+
+-- | Checker for uids by finding if the 'UID' is in one of the possible data sets contained in the 'SystemInformation' database.
+checkUID :: UID -> SystemInformation -> UID
+checkUID t si
```
JacquesCarette, line 72: This is really inefficient. A `Map` uses a nice tree data-structure to make lookup fast O(log n), but `Map.keys` returns a list, which is O(n).  You are better off using pattern guards like
```haskell t si
  | Just x <- lookupIndex t (s ^. dataDefnTable) = x
```
which will fall through if they fail.

code/drasil-printers/Language/Drasil/DOT/Print.hs
```diff
@@ -0,0 +1,205 @@
...
+    outputSubAvsAll gi handle
+    hPutStrLn handle "}"
+    hClose handle
+
+outputSubAvsAll :: GraphInfo -> Handle -> IO ()
+outputSubAvsAll gi handle = do
+    mapM_ (mkDirections handle) (directionsAvsAll gi)
+
+    let labels = filterAndGI gi [assumpLabels, ddLabels, tmLabels, gdLabels, imLabels, rLabels, cLabels]
+        prefixLabels = ["A", "DD", "TM", "GD", "IM", "R", "C"]
```
JacquesCarette, line 103: I don't like these labels being hard-coded here (and duplicated in the file). These should be grabbed from the abbreviation of some chunk(s) instead.

code/drasil-printers/Language/Drasil/DOT/Print.hs
```diff
@@ -0,0 +1,205 @@
...
+        colours = map ($ gi) [assumpColour, ddColour, tmColour, gdColour, imColour, rColour, cColour]
+        
+        allLabels = zip3 labels prefixLabels colours
+
+    hPutStrLn handle "\n"
+    mapM_ (mkNodes handle) allLabels
+
+mkOutputRefvsRef :: GraphInfo -> IO ()
+mkOutputRefvsRef gi = do
```
JacquesCarette, line 113: These functions represent a lot of duplicated code. Could you factor them out?

code/drasil-printers/Language/Drasil/DOT/SWHS_TG.dot
```diff
@@ -0,0 +1,69 @@
+digraph SWHS_TG {
```
JacquesCarette: Generated files should not be in the repo (exception is 'stable' where these are marked as 'golden' versions).

code/drasil-printers/Language/Drasil/DOT/TraceabilityGraphGeneratorInfo.hs
```diff
@@ -0,0 +1,63 @@
+module Language.Drasil.DOT.TraceabilityGraphGeneratorInfo where
```
JacquesCarette: This looks like it should not have been checked in?

code/stable/dblpendulum/TraceyGraph/allvsall.dot
```diff
@@ -0,0 +1,94 @@
+digraph allvsall {
```
JacquesCarette, line 1: Now, here's a good question: do we want these in stable? I don't know!

JacquesCarette (changes requested):

code/stable/dblpendulum/TraceyGraph/allvsall.dot
Ant13731: Yea, I wasn't too sure about that either. I don't think it should be much of a bother (the traceability matrices barely change, so I'm hoping the same goes for the tracey graphs once everything is smoothed out). 

Although it could be a decent indicator of something changing unexpectedly if a lot of the directions change in the graphs. And with `make stabilize`, it's not that bad to compare and copy files.

It might also be nice to implement the graphs directly into the SRS, though I'm not too sure how that would be done. Maybe we could generate .png files for them and then include them into the SRS files as a figure? It could be its own section, just as the traceability matrix has its own section.

Ant13731:
Here's the update summary from these commits: 
- The actual labels themselves are extracted from the drasil database now. However, it is only those which are made up of `ConceptInstances` (so things like assumptions, likely changes, requirements, etc.). The definitions and models have their labels hardcoded, since it is a guarantee that if a `UID` is a data definition, its label should start with DD. This has the added benefit of also separating changes into likely and unlikely, and requirements into functional and non-functional (though their node colours are still the same)
- I couldn't completely get away from the hardcoded labels (the labelling for subgraphs in particular), but it doesn't really matter what those labels are, just as long as they exist. They do not show up on the pdfs of the tracey graphs, but prevent naming conflicts within the .dot files for now.
-  Currently, the graphs don't exactly match up with the matrices in terms of labelling. This is because the graphs need low-level `Strings` to work while the matrices use `References`. We can't really use sentences at the moment because we don't want things like references, symbols, and exprs popping up in a .dot file. However, this means that for items whose `UID` doesn't match their `Reference` display name, they will appear differently when comparing the graphs and matrices side-by-side. However, the traceability relations are consistent and still in tact.

Hopefully these can be resolved through future issues.

code/stable/dblpendulum/TraceyGraph/allvsall.dot
JacquesCarette: Generating the .png and including a reference to them in the SRS sounds like a good idea. The .png themselves should not be compared when doing a diff (that they exist should be checked).

Let's experiment with these being part of stable. We'll see if we like it.

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -23,14 +23,14 @@ mkGraphInfo si = GI {
...
 
-    , assumpLabels = (getLabels tvAssumps si,      "A")
-    , ddLabels     = (getLabels tvDataDefns si,    "DD")
-    , gdLabels     = (getLabels tvGenDefns si,     "GD")
-    , tmLabels     = (getLabels tvTheoryModels si, "TM")
-    , imLabels     = (getLabels tvInsModels si,    "IM")
-    , rLabels      = (getLabels tvReqs si,         "R")
-    , gsLabels     = (getLabels tvGoals si,        "GS")
-    , cLabels      = (getLabels tvChanges si,      "C")
+    , assumpLabels = (getLabels tvAssumps si,  "A")
```
JacquesCarette, line 26: I can't help but think that these strings are already in `data-drasil` somewhere.

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -49,7 +49,7 @@ mkGraphNodes entries si = (traceMReferees entryF cdb)
 -- Testing new graphnodes function
 -- | Gets the nodes of a graph based on a list of sections we want to examine and the system information.
 mkGraphNodes :: TraceViewCat -> SystemInformation -> ([UID], [String])
-mkGraphNodes entry si = (nodeContents, map checkUID' nodeContents)
+mkGraphNodes entry si = (nodeContents, map (\x -> checkUID' x si) nodeContents)
```
JacquesCarette, line 52: This is an indication that `checkUID'` should probably take its arguments in the opposite order.

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -90,20 +90,26 @@ checkUID t si
...
 checkUID' :: UID -> SystemInformation -> String
 checkUID' t si
-  | Just _ <- Map.lookupIndex t (s ^. dataDefnTable)        = refS $ datadefnLookup    t (s ^. dataDefnTable)
-  | Just _ <- Map.lookupIndex t (s ^. insmodelTable)        = refS $ insmodelLookup    t (s ^. insmodelTable)
-  | Just _ <- Map.lookupIndex t (s ^. gendefTable)          = refS $ gendefLookup      t (s ^. gendefTable)
-  | Just _ <- Map.lookupIndex t (s ^. theoryModelTable)     = refS $ theoryModelLookup t (s ^. theoryModelTable)
-  | Just _ <- Map.lookupIndex t (s ^. conceptinsTable)      = refS $ conceptinsLookup  t (s ^. conceptinsTable)
-  | Just _ <- Map.lookupIndex t (s ^. sectionTable)         = refS $ sectionLookup     t (s ^. sectionTable)
-  | Just _ <- Map.lookupIndex t (s ^. labelledcontentTable) = refS $ labelledconLookup t (s ^. labelledcontentTable)
+  | Just _ <- Map.lookupIndex t (s ^. dataDefnTable)        =  "DD: " ++ t --Directly encoded since these are unlikely to change
```
JacquesCarette, line 96: I still don't want the string `DD` to be hard-coded... (ditto for the others).

code/stable/dblpendulum/TraceyGraph/allvsall.dot
Ant13731: Okay, I think I'll do a bit of cleanup in terms of labels before merging this PR, and then start including the .png files in a new PR. We could also use pdfs similar to the module dependency graphs on the drasil website, where we just link the traceability graphs.

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -90,20 +90,26 @@ checkUID t si
...
+  | Just _ <- Map.lookupIndex t (s ^. conceptinsTable)      = (safeFromJust $ getA $ defResolve s $ sDom $ cdom $ conceptinsLookup  t $ s ^. conceptinsTable) ++ ": " ++ t
+  | Just _ <- Map.lookupIndex t (s ^. sectionTable)         = "Sec: " ++ t
+  | Just _ <- Map.lookupIndex t (s ^. labelledcontentTable) = "Content: " ++ t
   | t `elem` map  (^. uid) (citeDB si) = ""
   | otherwise = error $ t ++ "Caught."
   where s = _sysinfodb si
 
+-- | Safe version of 'fromJust', used in 'checkUID''.
+safeFromJust :: Maybe String -> String
+safeFromJust (Just x) = x
```
JacquesCarette, line 110:
```haskell
safeFromJust = maybe id ""
```

code/drasil-printers/Language/Drasil/DOT/Print.hs
```diff
@@ -121,20 +121,20 @@ mkOutputAllvsAll gi = do
...
-    hPutStrLn handle $ "digraph " ++ title ++ " {"
+mkOutput :: GraphInfo -> String -> (GraphInfo -> [(String, [String])]) -> [GraphInfo -> (([String], [String]), String)] -> [GraphInfo -> Colour] -> IO ()
+mkOutput gi ttl getDirections getLabels getColours = do
+    handle <- openFile (ttl ++ ".dot") WriteMode
+    hPutStrLn handle $ "digraph " ++ ttl ++ " {"
     let labels = filterAndGI gi getLabels
         colours = map ($ gi) getColours
 
-        allLabels = zipWith (\x y -> (fst x, snd x, y)) labels colours
+        allLabels = zipWith (\x y -> (fst $ fst x, snd $ fst x, snd x, y)) labels colours
```
JacquesCarette, line 131: use pattern-matching for x and y in the function so that you don't need to have this string of `fst` and `snd`.

Plus this is likely an abuse of tuples. Never ever use a 4-tuple, always make that a record. 3-tuples should be rather rare too. I don't think there's a good reason to use a tuple here, so I'd create a record.

JacquesCarette (changes requested):

Ant13731:
Update summary for this commit:
- Remove hardcoded labels. Data Definitions and the like all use the `abrv` function to get their respective abbreviations
- Match the labels to what appears in the SRS' traceability matrices by using reference addresses for display labels instead of UIDs
- Use a new `NodeFamily` type to hold UIDs, display labels, abbreviation, and colour

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -9,54 +9,41 @@ import Drasil.DocumentLanguage.TraceabilityMatrix (TraceViewCat, traceMReferees,
...
-mkGraphNodes :: TraceViewCat -> SystemInformation -> ([UID], [String])
-mkGraphNodes entry si = (nodeContents, map (\x -> checkUID' x si) nodeContents)
+-- | Gets the node family of a graph based on the given section
+-- and system information. Also applies a given colour to the node family.
+mkGraphNodes :: TraceViewCat -> SystemInformation -> String -> NodeFamily
+mkGraphNodes entry si col = NF {nodeUIDs = nodeContents, nodeLabels = map (checkUIDRefAdd si) nodeContents, nfLabel = checkNodeContents nodeContents, nfColour = col}
     where
+        checkNodeContents :: [String] -> String
+        checkNodeContents [] = ""
+        checkNodeContents xs = checkUIDAbbrev si $ head xs
```
JacquesCarette, line 40: `head` is always bad. Use `(x:xs)` pattern-matching instead.

code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -91,24 +73,37 @@ checkUID t si
...
-  | Just _ <- Map.lookupIndex t (s ^. insmodelTable)        =  "IM: " ++ t
-  | Just _ <- Map.lookupIndex t (s ^. gendefTable)          =  "GD: " ++ t
-  | Just _ <- Map.lookupIndex t (s ^. theoryModelTable)     =  "TM: " ++ t
-  -- Concept instances can range from likely changes to non-functional requirements, so use domain abbreviations for labelling.
-  | Just _ <- Map.lookupIndex t (s ^. conceptinsTable)      = (safeFromJust $ getA $ defResolve s $ sDom $ cdom $ conceptinsLookup  t $ s ^. conceptinsTable) ++ ": " ++ t
-  | Just _ <- Map.lookupIndex t (s ^. sectionTable)         = "Sec: " ++ t
-  | Just _ <- Map.lookupIndex t (s ^. labelledcontentTable) = "Content: " ++ t
+checkUIDAbbrev :: SystemInformation -> UID -> String
+checkUIDAbbrev si t
+  | Just _ <- Map.lookupIndex t (s ^. dataDefnTable)        = abrv $ datadefnLookup    t (s ^. dataDefnTable)
```
JacquesCarette, line 78: why are you throwing away the index that you just got? In fact, why not look up the value, instead of doing it twice?

JacquesCarette (changes requested):

JacquesCarette (approved):

Issue #1019, opened 2018-10-04
Enhance Document to include Chapters, Parts, etc.


szymczdm:
Currently our documents only support sections and subsections, but there are a few other partitioning schemes that we should also include (ex. Chapters, Parts). I think this would be a good addition when we get to larger recipes.

I've made some notes in the Document.hs file as of commit 47bd13d1eb7a938a59eb9fc174d57de3d833c4f7.

balacij:
I think this issue will naturally re-appear when we have a specific use-case for it. Until we have a north star, I don't think anyone will be motivated to work on this issue. So, another one should be opened when we have a specific need.

PR #2659, opened 2021-07-07
Adding More Partitions

Awurama-N:
Contributes to Issue #1019 .
I read the notes szymczdm left in commit [47bd13d](https://github.com/JacquesCarette/Drasil/commit/47bd13d1eb7a938a59eb9fc174d57de3d833c4f7)
I created a new branch which has what he described what should be added, but i definitely don't think this is all that has to be done.

### DIFF
--- a/code/drasil-lang/Language/Drasil/Document.hs
+++ b/code/drasil-lang/Language/Drasil/Document.hs
@@ -20,6 +20,10 @@ import Control.Lens (makeLenses, view)
 data SecCons = Sub   Section
              | Con   Contents
 
+data Partition = Sections
+                | Part
+                | Chapter
+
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
Right, this is probably step 1 of ~12 to do this. You'll then need to 'plug in' this new type somewhere, fix the printer to render things properly, and also have the front-end be able to generate parts and chapters.

Awurama-N:
hmm, Please how could i go about "Plugging it in somewhere"
Or since i'm replacing Sections with Partitions should i just follow through with were i see that?

JacquesCarette:
Replacing is a form of "plugging in".  Right now, nothing refers to values of `Partition` - plugging in means changing that.

PR #2662, opened 2021-07-07
Make `make hlint` and `make hot_hlint` return the error codes (no suppression with -)

### LINKED ISSUE #2660: `make hot_hlint` should return an error status code if there are suggestions
It looks like I introduced a bug in #2625!

balacij:
Closes #2660 

The prefixed `-` of the hlint runs suppressed the errors.

The actions run _should_ fail. We can fix them in this ticket as well, but I want to see it fail first to make sure.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -208,10 +208,12 @@ analysis: graphmod
 hlint: check_stack
 	- stack install hlint
 	- hlint .
+	@exit $(.SHELLSTATUS)
 
 # Use the latest HLint version, downloading the binary each time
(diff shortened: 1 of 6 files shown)


JacquesCarette (approved):

JacquesCarette:
However, this build failed too.

balacij:
Yep, that was intended, so we can validate that the errors will happen once again. I am going to clear the hlint errors here as well if that's okay?

JacquesCarette (approved):

PR #2669, opened 2021-07-09
Including Traceability Graph Links in SRS

### LINKED ISSUE #2658: Add Traceability Graphs to SRS
Once #2656 is merged, we can begin implementing the graphs into each of the SRS files. They could be displayed as a figure (using .png files) within the SRS or as a link to the pdf version of the graphs (similar to how the drasil website displays the module's dependencies)

Ant13731:
Closes #2658 

I had originally wanted to implement the traceability graphs as figures in the SRS, but the graphs only generate after the SRS has been made. Perhaps a future PR could make them generate at the same time, but with the current setup I don't think it is possible. That work is on the [`addTraceyGraphtoSRS`](https://github.com/JacquesCarette/Drasil/tree/addTraceyGraphtoSRS) branch, but has reached a dead end.

Then, I tried to use the completed `SystemInformation` data gathered in each example's `Body.hs` file to implement the needed references. Specifically, I just needed the name of the Project so that the references could have a file path to link to. I was planning on implementing this in the `mkTraceabilitySec` function so that we wouldn't need to hard-code any information. Getting this idea to work would fix the small hacks detailed below. The progress on this idea is detailed on [`addTraceyGraphSRSLinks`](https://github.com/JacquesCarette/Drasil/tree/addTraceyGraphSRSLinks). Really the only thing preventing this from working is a single function (whose type binding is detailed in `drasil-docLang/Drasil/DocLang/References.hs` of the `addTraceyGraphSRSLinks` branch). I think I need a getter and setter lens, but I don't think I understand enough about them to implement it. I tried looking up the documentation, but I couldn't find any setters that just take the `Lens` version of the functions (something like `^.` but in the other direction). In addition, I'm not sure how appending references to an existing `ReferenceMap` will affect the map.

Since the above wasn't really working, I ended up manually adding the references to the ReferenceMap for each section, hardcoding the file name. This also led to a couple of small hacks within `drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs` where the in-text references are written only using a UID, since the filepath isn't available at that time to actually create the reference. The hardcoded strings may have to stay, as some examples have different internal names than their folder (eg. `Pendulum` still has folder name `DblPendulum`, so using the data in Drasil would give the wrong filepath).


Ant13731:
The traceability graphs now generate as figures in an SRS. There is still a couple hacks that should be removed, but I think can be done in #2675.

There is no more `make traceability` target, instead the graphs will generate on a per-example basis (just using `make` will work). They also generate as `pdf, png, svg` files.

@balacij, the `svg` files worked really nicely on the HTML version of the SRS, but the TeX version wouldn't generate. I'm currently using `.png` files instead since TeX supports it (but it doesn't look very good on the HTML version). Also, when attaching the graphs as a link, only `png` files seem to work on LaTeX. Its probably easier to show than explain, but switching the file ending for these functions will change the behaviour. What do you think I should do?:

https://github.com/JacquesCarette/Drasil/blob/70fd8fc39919a50d203b77e47d64d665e8a01a13/code/drasil-docLang/Drasil/DocumentLanguage/TraceabilityGraph.hs#L164-L182

Also, `pendulumTitle` has been temporarily changed back to `DblPendulum`, as I need the abbreviation to match the folder name (which has some unwanted changes throughout the SRS).

balacij:
Nice!

I used `svg` files in LaTeX last semester, it required a bit of work to get it working -- you will need an InkScape installation + a bit of configuration for the pdf generator (allowing shell, so it can access inkscape). Did you try this already?

Alternatively, after a quickly google search, I wonder if we could use [dot2tex](https://github.com/kjellmf/dot2tex) to convert the dot files into TeX for us? If so, perhaps we can use that as the assets for the graphs?

> Also, when attaching the graphs as a link, only png files seem to work on LaTeX.

I'm not quite sure I understand what you mean here, sorry. Can you explain?

> Its probably easier to show than explain, but switching the file ending for these functions will change the behaviour. 

Are we currently using the same assets for the TeX/PDF and HTML output?

Ant13731:
I did a brief search before and I saw some people used InkScape but I didn't try it yet. Same with dot2tex, although a couple people on the [tex stackexchange](https://tex.stackexchange.com/questions/13675/use-graphviz-within-tex) said it seemed a little dated.

> I'm not quite sure I understand what you mean here, sorry. Can you explain?

The links in the generated pdf documents (from the tex files) will show up as links, but clicking on them does nothing. However, when using a link to the png version of the graphs (rather than svg or pdf), the links will open a new page and display the graph as expected.

> Are we currently using the same assets for the TeX/PDF and HTML output?

Yes, the graphs that show up in the SRS files are all `png`, but the links that display the graphs on a new tab are `pdf` files. These can be changed by changing the file endings inside the `traceyGraphPath` and `traceyGraphPaths` functions respectively.

balacij:
> I did a brief search before and I saw some people used InkScape but I didn't try it yet. Same with dot2tex, although a couple people on the tex stackexchange said it seemed a little dated.

The InkScape method worked for me, and it looks like the CTAN just has an outdated version of dot2tex, so it's possible that it works (but I don't know for sure -- didn't try it out myself). Do you want to try using the SVG method?

> The links in the generated pdf documents (from the tex files) will show up as links, but clicking on them does nothing. However, when using a link to the png version of the graphs (rather than svg or pdf), the links will open a new page and display the graph as expected.

That's interesting. How do the links change when you change the file types/encodings?

Ant13731:
For HTML, all the links work. I think the svg one displays the clearest as both a link and a figure in the document. It looks much better than using a png (which is really blurry when zooming in to read the labels, especially on larger documents like GlassBR).
For TeX, only the png link works, with pdfs and svg links doing nothing when clicked on. Only png seems to work for displaying as a figure as well.

I think it would be worth it to use svg files if possible but maybe save it for another PR after this one is merged. I can try the InkScape method and see if it works. I think the current png figures look good enough for a first pass. What do you think?

balacij:
> I think it would be worth it to use svg files if possible but maybe save it for another PR after this one is merged. I can try the InkScape method and see if it works. I think the current png figures look good enough for a first pass. What do you think?

Sounds good! It will probably also require people to install InkScape and for us to update the wiki for such, so it might be good to leave it for a second pass like you mentioned. I can also post my project on GitHub if you want a working setup for the SVGs in TeX.

Ant13731:
Yes, that would be great! Thanks!

balacij:
I posted the [project files](https://github.com/balacij/CS6TB3-Final-Project/tree/master/presentation).

Yep, we still need to build LaTeX instructions (I think I already have some that match our existing theme from my CS3FP3 tutorials, so I'll just post them first for #2685, and then you can add the additional stuff for InkScape?).

smiths:
I hadn't heard of InkScape before.  I had a quick look and I like what I see.  However, if we can avoid having another dependency, I think we should.  From my quick scan of the above instructions, it sounds like InkScape can be avoided using `dot2tex`.  The `dot2tex` would be done during the generation, right?  This would mean that users of the generated documentation wouldn't have to install another dependency.

If we need another dependency, like InkScape, that is fine, but minimizing dependencies makes our software easier to maintain.

Ant13731:
Okay, I'll try both methods and see which one is easier/looks better.

balacij:
Hmm, the PATH alteration step is definitely a pro-dot2tex point.

Regarding Git Bash / PowerShell, did you restart Git Bash before trying to use it? I believe Git Bash does something approximately like taking a static copy of the PATH when it starts, so if you make changes to your PATH after starting it, it requires a restart. This would also explain why a restart of Windows fixed it.

Ant13731:
Yea, I think that was it. Modifying the path itself wasn't too bad, there was a really good [guide](https://tex.stackexchange.com/questions/2099/how-to-include-svg-diagrams-in-latex) on stack exchange for it. If we do end up using inkscape, the documentation is already mostly written out already.

Ant13731:
Also, some weird behavior: `make tex` wouldn't run without errors unless I called `make` first. I haven't looked into it in-depth yet, but I think it has something to do with which stage the graphs are generated.

Ant13731:
I got all the previous errors worked out (it was an issue with how I got directory names for references). The links also work on the pdfs now too! Should I make a separate PR dependent on this one for the new inkscape stuff?

code/drasil-example/Drasil/GlassBR/Body.hs
```diff
@@ -386,7 +386,7 @@ blstRskInvWGlassSlab = phrase blastRisk +:+ S "involved with the" +:+
 bodyRefs :: [Reference]
 bodyRefs = map (ref.makeTabRef.getTraceConfigUID) (traceMatStandard si)
   ++ map ref [sysCtxFig, demandVsSDFig, dimlessloadVsARFig]
-  ++ map ref concIns ++ map ref section ++ map ref labCon
+  ++ map ref concIns ++ map ref section ++ map ref labCon ++ traceyGraphGetRefs "GlassBR"
```
JacquesCarette, line 389: you really shouldn't be hard-coding the name here - please grab it from the 'right' place.

code/drasil-example/Drasil/HGHC/Body.hs
```diff
@@ -75,7 +75,7 @@ usedDB = cdb ([] :: [QuantityDict]) (map nw symbols)
 
 -- References --
 bodyRefs :: [Reference]
-bodyRefs = map (ref.makeTabRef.getTraceConfigUID) (traceMatStandard si)
+bodyRefs = map (ref.makeTabRef.getTraceConfigUID) (traceMatStandard si) ++ traceyGraphGetRefs "HGHC"
```
JacquesCarette, line 78: and of course the same goes for all examples.

JacquesCarette (changes requested):
Also, needs a merge with master.

PR #2676, opened 2021-07-12
Move RefInfo, create DecRef, modify reference classes

### LINKED ISSUE #1261: Merge Reference.hs and RefProg.hs

### LINKED ISSUE #2657: `References` shouldn't have `RefInfo`
Since `References` are being held in the chunk database for deferred lookup by UID, all of the attached `RefInfo` information shouldn't be with the actual `Reference` itself. We have already made the `Ref` sentence constructor do the work of holding any needed `RefInfo`, similar to how `Ch` holds whether a term should be singular, plural, or short form. `RefInfo` is treated as Sentence-level display information, where the user explicitly writes in that they want `RefInfo` to display at a certain point in a sentence. 

As of right now, any reference that enters the `ChunkDB` gets the `RefInfo` value set to `None` (by calling `ref` on anything that is `Referable`). So it doesn't really make sense to keep it as a part of the `Reference` type because it should not be used inside a `Reference`. Leaving it may cause unwanted behaviour, as a user might unknowingly put `RefInfo` into a `ChunkDB`, causing all appearances of that reference to have the attached `RefInfo` when they only wanted it in one particular spot.

In addition, this has the added benefit of getting rid of `RefProg.hs`, since we can put that datatype into `Sentence.hs`. This design is similar to how `Ch` works, where all the Sentence-level stuff stays with the `Sentence` type.

Ant13731:
Closes #1261 
Closes #2657 

Summary of changes:
- moved `RefInfo` to `Sentence.hs`
- created new `DecRef` type. A decorated reference has a `Reference` and possibly some `RefInfo`
- `HasRefAddress` now returns `LblTypes` instead of `Strings`
- `HasDecRef` has essentially replaced the functionality and use of `HasReference`. I still left `HasReference` defined for now, but it is currently not used. I'm not sure if this change was for the better, but it seems to work okay.

This will probably have a couple merge conflicts with some of the other PRs that are currently open, but the fixes should be simple anyway.


JacquesCarette (approved):

JacquesCarette:
Needs a merge, but is otherwise good to go.

Issue #1216, opened 2019-04-18
Lists are Referable but References don't work


bmaclach:
I tried making a reference to the list of Goal Statements and saw that in the final PDF/HTML, the link did not work. The reference itself was created correctly, but it does nothing because a label for the list was never actually rendered. For example, in TeX, the reference `\hyperref[Lst:goalStmt]{Lst: goalStmt}` is properly generated but there is nowhere with the label `Lst:goalStmt`.

I did a little investigation and I think the issue is that we drop the `Label` information from the `List` when we make a `LayoutObj` out of it. Note how a `List` `LayoutObj` is not parameterized by a `Label`: 
https://github.com/JacquesCarette/Drasil/blob/27cf6da8ac0a0e3058e3e75dc1d4744ef697bad6/code/drasil-printers/Language/Drasil/Printing/LayoutObj.hs#L20-L31

Mornix:
Is the desired reference really to the list of goal statements (specifically) or to the concept of goal statements (i.e. the section)?

bmaclach:
To the list of goal statements specifically. Though what I really want to do is reference individual goal statements, but couldn't (#1215), so I tried this as another option. I suppose referring to the goal statements section would be just as effective, though.

JacquesCarette:
So I don't think we should be able to refer to a list in its entirety? An item in a list, definitely. The section in which the list lives, that too. Of course, on top of #1215 being a bug, I think the current state is a bug but in a different way: a list should not be `Referable` at all.

Awurama-N:
I initially thought I would have to add `Label` as a parameter for lists, from the first comment, but i see you said list should not be referrable at all. Please How could i stop this?
@balacij

balacij:
Hmm, I'm not quite sure if we even have any "Lst:" references at all. I tried to search for it to no avail. Do you have any specific locations they occur in @Awurama-N ? I see some small helper functions in docLang/Drasil.DocumentLanguage.Labels but I don't see any real usage of them. If the goal is to remove the functionality of references to Lists, a part of this will surely be to remove these helper functions (`grep "makeLstRef" -r * --include *.hs`). If the data constructors are completely hidden away, I don't think we would need to do much else other than that really.

Also, from the looks of it, the References/IDs for the items in the list are already working (according to the `makeList` function of both `drasil-printers/Language.Drasil.HTML.Print` & `drasil-printers/Language.Drasil.TeX.Print`). Did you validate this/find any working examples?

[...]

PR #2681, opened 2021-07-13
Deleting makeLst functions

Awurama-N:
Contributes to #1216

This is going to break stable, but i'm pushing the changes so you can also see why.
I deleted the makeLstRef function but it's used in defining other functions which are used in the Examples. These functions were defined in a Label.hs file under doc-lang. I tried to delete the file just to see what things it would affect in turn.

I'm thinking now that i don't touch these functions? or do you think there might be something else to do here?

### DIFF
--- a/code/drasil-docLang/Drasil/DocLang.hs
+++ b/code/drasil-docLang/Drasil/DocLang.hs
@@ -44,7 +44,7 @@ module Drasil.DocLang (
   -- Tracetable
   generateTraceMap,
   -- Labels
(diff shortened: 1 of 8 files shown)


Ant13731:
I think I have some possible ideas for a fix, so I'm just going to add a quick commit.

Ant13731 (approved):

Awurama-N:
Thank you!

balacij (approved):

JacquesCarette (approved):

PR #2684, opened 2021-07-14
Adding Table of Contents

### LINKED ISSUE #1018: HTML Table of Contents
I noticed that the HTML versions of the SRS don't include a table of contents. It'd probably be a useful feature for anyone that wants to review the documents.

I don't think it would be a very difficult feature to add, particularly since we can traverse the DocDesc and know exactly which sections are included.

Ant13731:
Closes #1018 
Summary of changes:
- Add Table of Contents Section to SRS
- Modified TeX printers to only print Table of Contents if the section is present in the SRS
- Rearranged `SRS.hs` functions to be in the same order as they appear on a SRS document
- Table of contents will not generate unless the section is included in the `SRSDecl`


JacquesCarette (approved):

JacquesCarette:
Needs a master merge, but is otherwise good to go.

JacquesCarette (approved):

PR #2688, opened 2021-07-15
LaTeX Use SVG for Tracey Graphs

### LINKED ISSUE #2658: Add Traceability Graphs to SRS

Ant13731:
Depends on #2669 
Closes #2658 

Changes the traceability graphs from just links into actual figures in the generated document using SVG. This also adds a dependency on inkscape.


Ant13731:
@balacij should I add in the instillation guide for inkscape to the New Workspace Setup guide or wait until this PR is done/merged?

balacij:
In past tickets, we did it before the PR was done/merged, so I'd tend towards it before, at least posted to this PR. Out of curiosity, did dot2tex not work?

Ant13731:
Ah I haven't really tried it yet. I was planning to see how the deploy worked on my forked repo and then going from there. 

I might be wrong in this, but I think I would have to change up the code a lot more as only the tex generator would use the .dot files while the html generator would still use the svg ones generated by `dot`.

JacquesCarette (approved):
The code looks fine. But I do wonder how portable this is, i.e. how hard will it be for people on Macs and/or Windows to run this? We ought to have our automated testing include all 3 platforms (at least on a daily or weekly basis, maybe not on all check-ins?)

code/drasil-example/Drasil/NoPCM/Body.hs
```diff
@@ -199,6 +202,7 @@ si = SI {
   _instModels  = NoPCM.iMods,
   _datadefs    = NoPCM.dataDefs,
   _configFiles = [],
+  _folderPath  = directoryName,
```
Ant13731, line 205: @JacquesCarette, the changes requested in #2669 have been sort-of resolved here. 

I manually had to add a `directoryName` for examples like NoPCM and DblPendulum, as the system's abbreviation is different from its folder path. I know that the pendulum example is currently being in the process of splitting into single and double examples, but NoPCM directly uses the same system name as SWHS, so I would have to manually add it as a string here. For the rest of the examples, I could grab their abbreviation and take away the spaces (as in GamePhysics, which has the abbreviation "Game Physics"). But this would require the folder name to be exactly the same as the abbreviation. 

Do you think I should keep the `directoryNames` for all the examples or modify their abbreviations to match the folder names and use those instead?

Ant13731:
> how hard will it be for people on Macs and/or Windows to run this? 

It does add another dependency on InkScape, but I didn't have too much trouble on windows (aside from adding the inkscape command to the PATH). The instructions are documented in the [New Workspace Setup wiki](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#inkscape).

balacij:
> We ought to have our automated testing include all 3 platforms (at least on a daily or weekly basis, maybe not on all check-ins?)

It looks like [all are supported operating systems](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and we can have [scheduling](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events). I guess we can have the workflows create an issue if an error occurs on stable. Should we create a ticket for it or did you want to block this ticket until the deployments are ready?

JacquesCarette:
Please create a ticket. I'll take the plunge on this ticket, and we'll see!

PR #2691, opened 2021-07-16
Data Table now generates graphs

Ant13731:
Contributes #2612 

Specifically, this takes the data Nathaniel's table collected and puts them in a graph format. These graphs are meant to show the relationship between data structures and classes (through instances). Types are coloured turquoise, classes defined in a given package are magenta, and classes that are used (but not defined in the given package) are pink.

There is also a `circo_` version for each graph (which I find to be a smidge more readable than the normal `dot` ones - especially `drasil-lang`). Run `make analysis` to see the new graphs.

Also changed the generated graphs from pdfs to svgs since they look a little clearer.


JacquesCarette:
Needs a merge.

JacquesCarette (approved):

JacquesCarette (approved):

PR #2697, opened 2021-07-19
Add Characteristics to Pendulum Example

### LINKED ISSUE #2632: Double Pendulum: Characteristics of the intended reader
In the introduction for the double pendulum, add a section about the characteristics of the intended reader.

cd155:
The following will be added to Pendulum Example

**_2.3 Characteristics of Intended Reader_**

The final version of Characteristics is a little different than what I proposed on #2632. 
1. I didn't use "advanced" adjective. As discussed on the ticket, it can be ambiguous. Instead, I changed "advanced" to "undergraduate".
2. "with knowledge of ODEs" is almost the same meaning as "having an understanding of ODEs", so I changed it to have a clear sentence.

close #2632

### DIFF
--- a/code/stable/dblpendulum/Website/DblPendulum_SRS.html
+++ b/code/stable/dblpendulum/Website/DblPendulum_SRS.html
@@ -321,6 +321,14 @@
             </p>
           </div>
         </div>
+        <div id="Sec:ReaderChars">
+          <div class="subsection">
+            <h2>Characteristics of Intended Reader</h2>
+            <p class="paragraph">
+              Reviewers of this documentation should have an understanding of undergraduate physics, undergraduate calculus, and ordinary differential equations. The users of Pendulum can have a lower level of expertise, as explained in <a href=#Sec:UserChars>Sec:User Characteristics</a>.
+            </p>
+          </div>
+        </div>
       </div>
     </div>
     <div id="Sec:SpecSystDesc">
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):
I approve, but @smiths should also review, as it involves specific wording in the examples.

code/stable/dblpendulum/Website/DblPendulum_SRS.html
```diff
@@ -388,6 +391,14 @@ <h2>Scope of Requirements</h2>
             </p>
           </div>
         </div>
+        <div id="Sec:ReaderChars">
+          <div class="subsection">
+            <h2>Characteristics of Intended Reader</h2>
+            <p class="paragraph">
+              Reviewers of this documentation should have an understanding of undergraduate physics, undergraduate calculus, and ordinary differential equations. The users of Pendulum can have a lower level of expertise, as explained in <a href=#Sec:UserChars>Sec:User Characteristics</a>.
```
smiths, line 398: As mentioned in #2632, I think we should say what level of undergraduate physics and calculus.  I think we should say level 2 physics and level 1 calculus.  I think in the future we should say what topics from physics and calculus we cover, but it is better if we don't get distracted by that right now.  At some point we should go through each of the examples and more carefully record the knowledge needed to understand the examples.

smiths:
@cd155, as mentioned in my above comment, please add information on the level of physics and calculus that is required.

smiths (changes requested):
Please make the suggested change to add the level information.

cd155:
Information on the level of physics and calculus be added to Characteristics.

_Reviewers of this documentation should have an understanding of **undergraduate level 2 physics, undergraduate level 1 calculus**, and ordinary differential equations. The users of Pendulum can have a lower level of expertise, as explained in Sec:User Characteristics._

smiths (approved):

smiths:
Thank you for making the requested changes @cd155.  If you are done with this branch, please delete it.

PR #2696, opened 2021-07-19
Website fix

Ant13731:
Should fix the website and allow it to compile.
Also added in the website folder when running `make clean`
Closes #2695

Issue #2554, opened 2021-06-02
`Space` needs to be able to represent function spaces 


Awurama-N:
There are currently no defined spaces to represent functions.
And Space needs to be able to represent function spaces also.

https://github.com/JacquesCarette/Drasil/issues/2395#issuecomment-853261914

JacquesCarette:
You should add a link to the discussion we had that pointed out the need for this.

balacij:
Nice. I think we'll need this for #2418

balacij:
This was done in the above mentioned PR.

Issue #2395, opened 2021-05-08
investigate units of items named qd* in PDController


JacquesCarette:
For example `qdLaplaceTransform` has type `Real`, which is incorrect. What it should be is less clear! `(Real -> Real) -> (Real -> Real)` perhaps?  `qdFxnTDomain` is also suspicious. All of them should be investigated.

Note that it is best if each change/fix becomes its own PR, instead of trying to fix them all at once. That way to good fixes can go in right away, and the fixes that need discussion won't hold that up.

Awurama-N:
Process Variable – Value: `Real` is sufficient
Process Error -
Set point – Value: `Real` is sufficient
Proportional control – seems like `Real` is insufficient 
Derivative Control - seems like `Real` is insufficient
Transfer Function - function: Real is insufficient 
Control variable - `Real` seems to be sufficient
Laplace transform – `Real` is insufficient
complex frequency domain parameter– `Real` is insufficient
function in the time domain - seems like `Real` is insufficient
negative infinity– value: `Real` is sufficient
positive infinity– value: `Real` is sufficient
inverse Laplace transform– `Real` is insufficient
damping coefficient – value: `Real` is sufficient
stiffness coefficient – value: `Real` is sufficient

JacquesCarette:
For the future:
- if you use `- ` in the first column in markdown text, it shows as bullets
- the words 'sufficient' and 'insufficient' are both long and thus hard to tell apart at a glance. Much better for communicating this information would have been 2 lists, one where you think the type is good, and one with not. Can you edit your comment thus please?
- there are also ways of doing tables in markdown, if you want to get fancier.
- your 'call' as to whether the type is good/bad is likely based on some resource (book, Wikipedia, etc); you should list what resources you used to come to your conclusions.

Awurama-N:
I have made the changes. For the resources would you like specific links? or this is good?

JacquesCarette:
In the future, specific links. For this time, let's not bother.

Awurama-N:
@JacquesCarette would you like me to do something further with this issue?
for example editing the qd* units that `Real` was not sufficient for.
And if so could you please direct me to a resource that could help me further understand the use of `Real` so I could make the edits correctly?
Thank you.
@smiths

JacquesCarette:
Try changing just Laplace transform first (I already gave you the signature), and let's see what happens.

Awurama-N:
@JacquesCarette 
[...]

PR #2700, opened 2021-07-19
Proto-Function definitions and function spaces, with a smaller example

### LINKED ISSUE #2418: Showing "f(x,y,z)"-like forms on the left-hand side of QDefinition equations
Related to #2371. 
~~[Related code](https://github.com/JacquesCarette/Drasil/compare/convDblPendulumIMod).~~ deleted it since it's already here below.

While trying to convert the only InstanceModel of the DblPendulum example, I came across an issue with getting the QDefinition to render the left-hand side of the equation as it does currently, with the left-hand side showing "theta_p as a function of time".
image

Presently, a few of the examples that don't generate code have `RelationConcepts` with `Relation`s (`Expr`s) that are of the form:
```
angularDisplacementRC :: RelationConcept
angularDisplacementRC = makeRC "angularDisplacementRC" (nounPhraseSP "calculation of angular displacement")
  EmptyS $ apply1 pendDisplacementAngle time $= sy initialPendAngle * cos ( sy angularFrequency * sy time)
```
(This is the `RelationConcept` used in the sole InstanceModel in DblPendulum)

Specifically, the left-hand side of the equality is `apply1 pendDisplacementAngle time`, `theta_p` as a function of time, currently displayed with `(t)` in the [equation](https://jacquescarette.github.io/Drasil/examples/DblPendulum/srs/DblPendulum_SRS.html#IM:calOfAngularDisplacement) of the InstanceModel information box.

Unfortunately, this is not currently easily convertible into a QDefinition using `relToQD` without changing the `apply1 pendDisplacementAngle time` into just a `sy pendDisplacementAngle` (e.g., `C "pendDisplacementAngle"`) because `relToQD` (and it's helper, `convRel`) only act on things of the form `EqBinaryOp Eq (C x) r` directly while the `apply1 pendDisplacementAngle time` is a `FCall` (not a `C`).

It's also not quite straightforward to convert this into a QDefinition directly either (at least from what I've tried at the moment, but I might be wrong!). The "left hand side" (`apply1 pendDisplacementAngle time`) that we want to be the target of the definition is currently of type `Expr` but we would need it to be of type `Symbol` (I believe) for when displaying. We have tons of nice smart constructors for writing Symbols.

Giving it a go (just using `p` with a subscript `theta` [intentionally backwards/incorrect], without the "t input" yet)...
```
angularDisplacementQD :: QDefinition
angularDisplacementQD = fromEqn' "pendDisplacementAngle"
  (nounPhraseSP "calculation of angular displacement")
  EmptyS (sub lP lTheta) Real (sy initialPendAngle * cos (sy angularFrequency * sy time))
```

And the left-hand side shows as `theta_p`, showing that this symbol override is not actually respected. It is just getting the symbol of the `pendDisplacementAngle` UID instead, which is:

```
pendDisplacementAngle :: UnitalChunk
pendDisplacementAngle = makeUCWDS "pendDisplacementAngle" (cn "displacement angle of pendulum")
        (phrase angle `the_ofThe` phrase pendulum)
        (sub lTheta lP) degree
```

At first, I thought this had to do with the `HasSymbol` instance for InstanceModels, so I changed
```
instance HasSymbol InstanceModel where symbol = symbol . view output
```

into:
```
instance HasSymbol InstanceModel where
  symbol (IM (EquationalModel q) _ _ _ _ _ _ _) = symbol q
  symbol im_                                    = symbol (view output im_)
```
where I set the symbol of an equational instance model to be that of the equational models' QDefinition. Unfortunately, it didn't change anything. However, I think this change still makes sense to do regardless because the equational models are definitional. ( please let me know your thoughts about this )

Moving on, I checked the same instance declaration for QDefinition, for which I think makes sense. It's the symbol of it's internal QuantityDict. However, if we wanted to make this change of having left-hand side of equations shown as functions of input variables, then _I believe_ this might be the area that we primarily edit (using the stage perhaps explicitly here to show when to have/not have the "function inputs tuple" shown), though I'm not 100% sure as I haven't tested it yet. However, I believe that altering this would affect many of the currently generated examples.

I'm not yet quite sure how to proceed with fixing this, but I have a feeling that this isn't something that should be fixed in isolation. I think that this would have a greater impact on how we generally show equations in InstanceModels. With the exception of [NoPCM's IM: heatEInWtr](https://jacquescarette.github.io/Drasil/examples/NoPCM/srs/NoPCM_SRS.html#IM:heatEInWtr), none of the current examples that generate code have this same area where a QDefinition has a left-hand side of an equation shown with the "function of X,Y,Z" syntax. With regards to the NoPCM heatEInWtr example, the `_definitions` of it's Body's `SystemInformation` is currently empty (is this normal? if we add it to `_definitions`, then it would cause a problem with the `relToQD` issue mentioned above) which is why this issue does not occur with it. It appears to be rendered differently, which I haven't yet fully figured out.

Do we want to have `f(a,b, ..., z)` or just `f` shown on the left-hand side of EquationalModel equations? If we want the former, do we also want these arguments to appear inside of the function calls used in expressions? If we want the latter, should I just change stable with this particular example, or do we want just this example in particular shown as a function of some inputs?

Should all InstanceModels' equations always be added to the `_definitions` lists in the `SystemInformation`s?

Is there any thoughts on this? Any help would be greatly appreciated :smile:

balacij:
Builds on #2696
Closes #2418 
Relates to #2554 
Contributes to #2371 

* Creates `FuncDefn` containers -- prototype function definitions, not yet used in code generation
  * First-order functions, using curried inputs
* Creates `FunctionalModel` ModelKinds -- a new variant of models for defining functions
* Convert DblPendulum's angular displacement IM into a `FunctionalModel` as a prototype

Leaving this as a draft because I'm unsure if we want named arguments in `Expr`/Function definitions.

Should we work on code generation next while converting more models? I suspect that we'll need to carry a bit more information about the inputs in the FuncDefns.


code/drasil-lang/Language/Drasil/Chunk/Function.hs
```diff
@@ -0,0 +1,87 @@
...
+import Language.Drasil.Chunk.Quantity (mkQuant', QuantityDict)
+import Language.Drasil.Expr.Display (defines)
+import Language.Drasil.Expr (Expr(FCall, C))
+import Language.Drasil.NounPhrase.Core (NP)
+import Language.Drasil.Space (mkPrimitiveMapping, Space)
+import Language.Drasil.Sentence (Sentence(EmptyS))
+import Language.Drasil.UID (UID)
+
+-- TODO: Should we have named arguments for functions? Probably in CodeExpr, but I don't think so in Expr. Perhaps the FCalls need to be different!
+data FuncDefn = FD {
```
JacquesCarette, line 23: So: do we really need this? What I mean is, do we need this at the Chunk level? What is different about functions that they need their own 'thing'?  In other words, could we re-use `QDefinition` for functions too?

code/drasil-lang/Language/Drasil/Space.hs
```diff
@@ -25,9 +25,20 @@ data Space =
   | Actor String
   | DiscreteD [Double]
   | DiscreteS [String] --ex. let Meal = {"breakfast", "lunch", "dinner"}
+  | Mapping [Space] Space -- TODO: this one shouldn't be exposed directly, only through a smart constructor
```
JacquesCarette, line 28: So we might want to use `NonEmpty` instead of a raw list - we don't want 0-ary functions.

code/drasil-lang/Language/Drasil/Space.hs
```diff
@@ -25,9 +25,20 @@ data Space =
...
   | Void
   deriving (Eq, Show)
 
+mkPrimitiveMapping :: [Space] -> Space -> Space
+mkPrimitiveMapping _   (Mapping _ _) = error "Target must be basic"
```
JacquesCarette, line 33: So rather than this, it would make sense to create a `Primitive` type that is just a wrapper for `Space`, but then use it in the above, i.e.
```haskell
   | Mapping [Primitive] Primitive
```
so that the invariant is also "in the type".

Plus, `mkFunction` is a better name...

code/drasil-lang/Language/Drasil/Space.hs
```diff
@@ -25,9 +25,20 @@ data Space =
...
+mkPrimitiveMapping ins trg
+  | any isMapping ins                = error "All inputs must be basic"
+  | otherwise                        = Mapping ins trg
+
+isMapping :: Space -> Bool
```
JacquesCarette, line 38: `isMapping` is probably an anti-pattern (see 'boolean blindness'), so you don't want to use it.

JacquesCarette (changes requested):
So I really expected more changes to `Expr` and fewer the chunks for adding functions?

code/drasil-lang/Language/Drasil/Chunk/Function.hs
balacij: I think we can get away with re-using `QDefinition`s for functions too. Let me try to do this and get back to you.

code/drasil-lang/Language/Drasil/Space.hs
balacij: Sounds good. What do you think of creating a PrimitiveSpace ADT for the primitives? Something along the lines of:
```
data PrimitiveSpace = Integer | Rational | Real | Natural | Boolean | ...
data Space = Primitive PrimitiveSpace | Function (NonEmpty PrimitiveSpace) PrimitiveSpace
```
I think this would also resolve the boolean blindness issue below. However, this would still be insufficient for the function spaces we need for #2395.

code/drasil-lang/Language/Drasil/Chunk/Function.hs
balacij: Changed this now.

code/drasil-lang/Language/Drasil/Space.hs
balacij: I renamed `Mapping` to `Function` as well.

JacquesCarette: `PrimitiveSpace` is a good idea.

JacquesCarette (approved):
This is looking good now.

balacij:
Thank you!

Issue #2642, opened 2021-06-29
Update Makefile targets related to deployment, and evaluate deployment scripts


balacij:
We should probably
* Add basic documentation (if not existing already) about usage of the scripts we have, at the top of each script
* Update the Makefile targets related to deployment 
* Re-evaluate deployment scripts now that we have our own website generator

balacij:
It looks like we need to change the default shell in the Makefile from `/bin/bash` to a more portable variant that looks up `bash` in the PATH, e.g., `/usr/bin/env bash`.

balacij:
@Ant13731 Does the ordering of the `make analysis` and `make graphs` CI steps matter? I'm just curious because right now they're not in the same order in the Test.yaml and Build.yaml scripts.

Ant13731:
No, I don't think `analysis` or `graphs` rely on each other at all, just as long as they are there before deploying. But it may be good to keep the same order.

balacij:
Thanks!

Ant13731:
Also the `rev` command doesn't work in git bash on Windows, so the `make deploy` target doesn't work. I've been using WSL to get around it, but I figured I should mention it.

balacij:
`rev` is just a command that reverses its inputs. I haven't been able to find a decent replacement for it yet for Windows.

As odd as it is, we will need to fix it for the test suite #2698.

Ant13731:
Hmm, I'm not too sure. I don't really know what the stubs are used for, but I can try and switch things around to see if something works.

Ant13731:
I think it might have to stay in some form, as something with the generated code fails. I got this error when I removed `src_stub` completely:
```
cd "build/GlassBR" && stack exec -- "glassbr" 
make[1]: Entering directory 'C:/Users//Documents/GitHub/Drasil/code/build/GlassBR/src/cpp'
doxygen doxConfig
process_begin: CreateProcess(NULL, doxygen doxConfig, ...) failed.
make (e=2): The system cannot find the file specified.
make[1]: *** [Makefile:22: doc] Error 2
```

balacij:
When #2768 is merged, we should add a note somewhere in the README + Wiki about `make help`.

JacquesCarette:
#2768 is merged now - so time to add a note?

Ant13731:
Yep, working on it in #2780

balacij:
[...]

PR #2709, opened 2021-07-20
Replace SHELL = `/bin/bash` in the Makefile with just `bash`, and add shebangs to `/usr/bin/env bash` to each script

balacij:
Contributes to #2642 

+ Order CI steps the same in Build.yaml and Test.yaml

@Ant13731 Do you still develop using Git Bash? If so, do you mind testing the scripts to validate that they work using Git Bash? I want to double check that Git Bash is still setting up `/usr/bin/env`.


Ant13731:
Ahh, none of the makefile targets work. I get this error:
```
$ make
rm -rf "./logs/"
/usr/bin/sh: C:/Program: No such file or directory
make: *** [Makefile:146: check_stack] Error 127
```

balacij:
I'll need to recreate it locally then and figure out what's going on. Thank you for testing it!

balacij:
I'll re-test this on a fresh Windows setup when I can. I think it will be fine, but it's better to double check.

Ant13731:
I did a quick check to see if `make`, `make tex`, `make analysis`, and `make graphs` would run and it worked without any problems.

balacij:
Ah, thank you very much @Ant13731!

JacquesCarette (approved):

PR #2711, opened 2021-07-21
Reorganize makefile targets

### LINKED ISSUE #2707: Is GraphViz a new dependency?
It seems we use `dot` with the default `make` now. Is GraphViz required for each `make` run? Are we able to get a partial run excluding the graphviz graphs?

If GraphViz is a new dependency, we should also add a section to the Wiki's New Workspace Setup page.

Ant13731:
Closes #2707
Contributes #2642


JacquesCarette:
I don't really like 'negative' targets. I would prefer names that reflect an incremental addition of more targets instead.

Ant13731:
Okay, should I instead change the plain `make` command to not generate graphs and add a new target `make traceygraphs`?

balacij:
I like the idea of a `make traceygraphs`, but I'm wondering if we removed the graph generation from the basic `make`, wouldn't we get `diff` errors saying that the graphs only exist in `stable/`?

Ant13731:
The dot files are Drasil-generated so they don't depend on graphviz (and stable doesn't need the graphs themselves, only the dot files).  The SRS links should just not work/not display, but I don't think it would break anything too much.

JacquesCarette:
So I'm kind of thinking of the following targets:
- **code** only the basic Drasil code (no examples, no web site, etc)
- **examples** = code + examples
- **test** = code + examples + test the examples
- **analysis** = code + examples + analyse
- **graphs** = code + examples + generated graphs
- **web** = needs everything?

The above might not be quite right, but hopefully it conveys the basic idea.

Ant13731:
Okay, I think I see the hierarchy we want. What should the plain `make` be then? Should it be everything or maybe the new `test` target?

- I think `code` is being used for gool, so maybe a `buildPkg` target just to compile the Drasil code? We might not be able to have a `build` target as one of our package names is "build".
- `analysis` doesn't need the Drasil code to be compiled as it only looks through the files, but we might want it to make sure it compiles anyway?
- The target name `graphs` is already being used by the module dependency graphs (used in the Drasil website). Should those be moved to a new `modulegraphs` target?
- The website itself doesn't need anything besides Drasil code to work (for now, it might need the examples later) so long as the links for everything else are in the right spot. I think `deploy_lite` does something similar to what `web` would do (in other words, make sure everything is in the right spot so the generated website doesn't have any broken links).

Should all of these targets be made with smaller "building block" targets for easy debugging? (It might take a lot longer to debug something from the analysis section if `analysis` ran the examples every time)

Also, should all of these be made using individual examples as their own targets (so `make glassbr` will still work) and then grouping that list of example targets into the larger targets defined above? I think that is what the Makefile is doing currently.

JacquesCarette:
Yes, I think `make` should go to `make test`.  Can you change the `code` target to `gool`? I think we want `analysis` to work on recent things, so it should "make" whatever is needed to do a proper analysis. If that means no compilation is needed, that's fine. On `graphs`: no keep the name as is; use `tracegraphs` for the new target. If `deploy_lite` already does the job of `web`, great, no need for a new target.

And yes, definitely made of smaller building block targers. Also yes for individual examples as targets.

balacij:
I continued a few changes related to this PR in #2722, namely changing the `make prog` (Build) step from the CI to a `make code`, and splitting up the examples in #2722, would this affect any of the changes in this? Could it perhaps help with anything?

Ant13731:
Oh! I completely forgot to change the CI steps (I'm surprised the tests didn't give me an error). Thanks for the catch. I'll check out #2722 now then.

JacquesCarette (approved):

JacquesCarette:
So is this ready?

balacij:
There's a few extra fixes related to this applied in #2722, so it might be best to only merge this code in via #2722.

Ant13731:
Yea, the `prog` target won't work here. It might just be best to merge @balacij's PR and drop this one.

Issue #2675, opened 2021-07-12
Automatically Include Common References in Examples


Ant13731:
Building off of @JacquesCarette's comments in #2279,

> Note that, over time, we want to automate these parts. These lists of documentation and adding them to the right places in the examples should not be a step that a human needs to do. In fact, for `doccon` and `doccon'`, that might be possible now.

We should be able to automate some references so that users do not have to manually place them in the `ChunkDB`. It might be too early to automate references created in specific examples, but references that are more universal (like `doccon`,`doccon'`, and those references listed in `DocLang/SRS/References.hs`) should be automatically placed in the `ChunkDB` for every example.

Completing this issue will also require modifying the instructions for [Creating Your Project in Drasil](https://github.com/JacquesCarette/Drasil/wiki/Creating-Your-Project-in-Drasil#adding-introduction-section).

Ant13731:
For the `IdeaDict` map in the chunk database (where things like `doccon` and `doccon'` are placed), should I include all of the concepts across all the domains we currently have?

 For example, should I include `softwarecon`, `educon`, `mathcon`, `compcon`, `physicCon`, `solidcon`, and `thermocon`? Everything except the last three almost always appear in the examples, so I think those should always be included. But should I automatically include `physicCon`, `solidcon` and `thermocon` the database, or let the user add them when they are needed?

JacquesCarette:
What we really should be doing is the opposite: only include the ones that are actually used. Automatically, not by hand.

All the currently hand-built lists of chunks in the examples should, over time, disappear. We don't want to fake this by moving the lists to other parts of Drasil either.

Ant13731:
Okay, I think I see what you mean, but I'm not too sure how to go about it fully automatically. 

My first thought was to have a second database running in the background (so that the user does not know about it) with all of the chunks from `drasil-data`. Then the example would check the `UIDs` used in the example and pull from the "background" database into its actual database. But this just feels like moving around the list of chunks rather than automatically including them when needed. It could also lead to unexpected behaviour if the same UID is used in `drasil-data` and the example when the user doesn't know about it.

My next idea would be to somehow tag all the named chunks in `drasil-data` with their respective domains (eg. Just make all the `NamedChunks` into `CIs` so the long list of chunks like `compcon` would all have the `compScience` tag) and have the user include which domains are needed for their project (maybe as a new field in `SystemInformation`), but I think that would still need a "background" database to hold all the chunks from `drasil-data`.

[...]

PR #2717, opened 2021-07-21
Automating References

Ant13731:
Contributes #2675 
I only ran into one weird thing: In GlassBR, I had to manually add the citations `astm2009` and `astm2016` to the reference database. I couldn't figure out why the citation database wasn't automatically including them, so I think it's best to make a new issue for it. Other than that, a lot of the manual referencing overhead from the examples has been removed.


JacquesCarette:
Need a master merge.

JacquesCarette (approved):
This is the kind of PR that I love - 0 changes to stable, but hundreds of lines of code deleted and the API of many modules many smaller.

JacquesCarette:
My guess about GlassBR: the only place those 2 references are used might be quite 'deep', and your checking doesn't quite dig into that part? A new issue is indeed fine.

Ant13731:
I think I found the source of the error: `codeSpec` was using the system information before the "fill" functions were called. Replacing `si` with `fullSI` seems to work.

JacquesCarette (approved):

Issue #2704, opened 2021-07-20
Finding all calls for dqd, dqdNoUnit, dqd', dqdWr and dqdQd


Awurama-N:
Searching Drasil for all calls of dqd, dqdNoUnit, dqd', dqdWr and dqdQd

Awurama-N:
## dqdNoUnit

in Data.Drasil.Quantitiies.Physics:
- restitutionCoef 

in Data.Drasil.Quatities.Math:
- gradient
- normalVect
- uNormalVect
- unitVect
- unitVectj
- petpVect
- euclidNorm

in Data.Drasil.Quantities.SolidMechanics:
- poissnsR

in Drasil.GlassBR.Unitals:
- aspectRatio
- glassTypeCon
- gTF
- loadSF

Drasil.PDController.Unitals:
- ipPropGain
- ipDerivGain
- ipSetPt
- opProcessVariable


## dqd’

in Data.Drasil.Quatities.Math:
- pi_
- posInf
- negInf

in Drasil.SSP.Unitals:
- constF
- fs
- fsMin
- cords
- earthqkLoadFctr
- normToShear
- scalFunc
- numbSlices
- minFunction
- mobShrC
- shrResC
- varblV
- index

in Drasil.Projectile.Unitals:
- launAngle
	
in Drasil.SWHS.Unitals:
- eta
- meltFrac
- fracMin
- consTol
- aspectRatio
- aspectRatioMin
- aspectRatioMax


in Language.Drasil.Chunk.Constrained:
- ConstrConcept constructor: cuc’’

in Language.Drasil.Chunk.Unital:
- UnitalChunk constructor ucStaged


## dqdQd

In Database.Drasil.ChunkDB.GetChunk:
- combine 
- combine’ 

in Drasil.GlassBR.IMods:
- symb

## dqd

in Drasil.GlassBR.Unitals:
- standOffDist

in Drasil.Projectile.Unitals:
- flightDur
- landPos
- launspeed
- offset
- targPos

in Drasil.PDController.Unitals:
- ipStepTim
- ipSimTime

in Language.Drasil.Chunk.Constrained:
- cuc’

in Language.Drasil.Chunk.Unital:
- UnitalChunk constructors ucs’, uc’, ucs,ucsWS,makeUCWDS

## dqdWR

in Drasil.GlassBR.IMods:
- symb

in Drasil.SSP.Requirements:
- inputDataTable
- inputsToOutput

in Drasil.SSP.Unitals:
[...]

Issue #2677, opened 2021-07-12
Chunks in `drasil-lang` that use `Maybe` instead of `Lenses`


Ant13731:
Here is a list of types that use `Maybe` (referring to the [Rationalizing data structures project](https://github.com/JacquesCarette/Drasil/projects)). I will update these later to include the types/classes above and below each.

- `ConstrainedChunk` (`Maybe Expr`)
- `ConstrConcept` (`Maybe Expr`)
- `DefinedQuantityDict` (`Maybe UnitDefn`)
-  `IdeaDict` (`Maybe String`)
- `QuantityDict` (`Maybe Expr`)
- `MayHaveUnit` (`Maybe UnitDefn`)
- `HasDerivation` (lens to `Maybe Derivation`)
- `HasReasVal` (lens to `Maybe Expr`)
- `ListType`, `ListTuple`, `RawContent` (not a chunk, but still has `Maybe String` or `Maybe Int`)
- `Uncertainty` (`Maybe Double` and `Maybe Int`)

balacij:
I believe that @Awurama-N has started work on the first 2 components (UncertainChunk + UncertQ). I'm not quite sure where the ticket went however.

JacquesCarette:
Yes, @Awurama-N was working on that. I see that maybe there is an easier set of things to change -- the `Maybe UnitDefn` in `DefinedQuantityDict` should simply be deleted. Of course, it's not that easy! The proper process will involve seeing what the actual calls to the various `dqd` constructors are.  But this should be doable.

So @Awurama-N can you open a new issue, where the first part will be to see all calls to `dqd`, `dqdNoUnit`, `dqd'`, `dqdWr` and `dqdQd` in all of Drasil (not just the examples). We'll see what should be done from there.  This is definitely a really awful wart that can be fixed.

balacij:
So, I believe this ticket was about analyzing which `Maybe` and trying to get rid of `Maybe` usage. However, I'm not sure I fully understand why we're trying to get rid of `Maybe`.

samm82:
I think we may have backtracked on trying to do this and will continue to use `Maybe`. However, the specifics of when this happens throughout the Drasil process is yet to be determined and requires more thought

JacquesCarette:
Right, I think I may have changed my mind on `Maybe`. We can and should use it for truly optional information.

We should never use it if we always expect to get a `Just` and getting a `Nothing` is an error.

balacij:
[...]

PR #2718, opened 2021-07-21
Switching dqd'

Awurama-N:
contrubutes to issue #2704 
contributes to issue #2677

### DIFF
--- a/code/drasil-example/Drasil/SSP/Unitals.hs
+++ b/code/drasil-example/Drasil/SSP/Unitals.hs
@@ -187,10 +187,10 @@ fsMin = dqd' (dcc "fsMin" (cn "minimum factor of safety")
 -- Once things are converted to the new style of instance models, this will
 -- be removed/fixed.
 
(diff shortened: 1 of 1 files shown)


Awurama-N:
dqd' phrases are in the format `ConceptChunk (Stage -> Symbol) Space (Maybe UnitDefn)` & ucStaged phrases are in the format `String NP String (Stage -> Symbol) u`.
I tried switching the dqd' in`coords` in Drasil.SSP.Unitals with `ucStaged` but errors kept coming up with having to change the way `coord` is currently structured.
Anthony helped and suggested i could use either `ucsWS` or `makeUCWDS` instead.
is this fix okay?

JacquesCarette (approved):

JacquesCarette:
This one is fine, because the `Stage -> Symbol` function is actually constant.

Any of the constructors for Unital should be fine, as long as there are no changes to stable.

JacquesCarette:
(I didn't delete the branch, you can continue to use it if you'd like - or delete it if you'll use a new one)

PR #2720, opened 2021-07-21
Double Pendulum: Add Organization of Document

### LINKED ISSUE #2633: Double Pendulum: Organization of the document
In the introduction for the double pendulum, add a section about the organization of the document.

cd155:
The Organization of Document will be added to the Pendulum example. Details on what's in Organization of Document listed at #2633

close #2633


JacquesCarette (approved):

cd155:
I just realized the location of figures may be different between .pdf and .html

In pdf, figure 1: System Context appears at top of the page before 2.4 section. In the html file, the figure locates in the middle of the system context.

PDF:
Screen Shot 2021-07-21 at 6 13 01 PM

HTML:
Screen Shot 2021-07-21 at 6 22 52 PM

JacquesCarette:
This is normal, that's how latex works. As long as the text doesn't assume where the figure actually is (by using words like 'above' or 'below'), it's fine.

cd155:
I see, thanks for the explanation.

PR #2722, opened 2021-07-22
Split examples into individual packages

### LINKED ISSUE #2708: Split up example into packages with libraries executables
So that we can import the examples as libraries for the drasil-website generator.

Maybe we can get away with using a manually written Cabal file for `drasil-example` (with multiple libraries in said Cabal file) instead of creating a package.yaml/stack.yaml file for each example...

balacij:
Closes #2708 
Builds on #2711 

* Splits drasil-example's examples into multiple subpackages with a library + executable -- this should assist with the website being able to automatically pull information from the examples.
* Renamed `pdController` targets and folders to `pdcontroller`
* Converts `gooltest`/`codegenTest` into a test-executable instead of a normal executable


balacij:
CI failed due to a bad dependency. I think we can just remove the dependency on drasil-example for drasil-website for now and work on adding the "auto pull information from the example libs" a bit later.

Ant13731:
Would we have to add each example individually as a dependency?

balacij:
Yes, we would.

Ant13731:
Okay, I think that should be all the stuff I wanted to add. I'm going to make a new issue on getting the website to take information from the examples.

balacij:
Nice, thank you! :smile:

JacquesCarette (approved):

PR #2724, opened 2021-07-22
Changing bold symbol command from mathbf to symbf

### LINKED ISSUE #2240: Add torque symbol in GamePhysics SRS pdf.
@smiths
Table 2 in GamePhysics SRS pdf is mising the torque symbol (tau).

Ant13731:
Closes #2240 

I checked through all the examples. Alpha, beta, and omega were not showing up as well (in SSP) but the fix works for them. The vector hats are now unbolded, but they align better (most of the time) anyways. 

Only one thing out out of the ordinary: the total base lengths of slices is still not bolded in SSP, but in the old version it did not show up at all.
In the HTML:
image
In the old pdf (mathbf):
image
In the new pdf (symbf):
image


smiths:
@Ant13731, what would be the consequences of changing the "fancy" lowercase l to just an uppercase l (ell).  That is what if we used L to represent the vector of base line lengths?  Could you have a look through the example to see if that use of L would clash with an existing use of L.  It might also make the example inconsistent with some of the diagrams.

If changing the symbol to L improve the situation, I'd be happy to do that.  That would be quicker than trying to figure out what is going on currently.

Ant13731:
Sorry, by the uppercase L, do you mean the curved one (ℒ )  or just a regular L?

smiths:
Just a regular L.

JacquesCarette (approved):

PR #2725, opened 2021-07-22
Add basic nix shell script

### LINKED ISSUE #2702: Add `shell.nix` configuration for developers working on Drasil
We would want to have the Nix expression install all CLI tools and compilers necessary for Drasil and Drasil-generated software.

It appears that Nix works well with Mac (many users seem to use it), it works very well on Linux, and it appears to run well on Windows under WSL.

The `shell.nix` name is the default name that a `nix-shell` command looks for when searching for a Nix expression file to create a shell environment.

We need:
* Stack v2.5.1
* `inkscape`
* Maybe `imagemagick`
* `graphviz`
* `texlive.combined.scheme-full`
* Fonts: `lmodern` & `lmmath`

balacij:
Closes #2702 

Unfortunately, installing fonts is considered a side-effect on systems, and, as such, is only allowed in the NixOS `configuration.nix` (or if users use `home-manager`, it's allowed through that too). So, this script will install the base requirements for a CLI user to develop Drasil and work with the generated artifacts.

(I also have a small change where I reorder a NOISY=yes parameter in both of the CI files.)

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: "Build"
         run: make code stackArgs="--no-terminal" GHCFLAGS="-Werror"
       - name: "Stable"
-        run: make stackArgs="--no-terminal" NOISY=yes GHCFLAGS="-Werror"
+        run: make stackArgs="--no-terminal" GHCFLAGS="-Werror" NOISY=yes
       - name: "TeX"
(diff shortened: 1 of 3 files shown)


balacij:
If anyone is interested in using this, it will likely take a bit of time on the first time you use `nix-shell`, but all subsequent usages _should_ be near instant.

JacquesCarette (approved):

Issue #2723, opened 2021-07-22
Get information for website based on examples


Ant13731:
A couple things to complete for this issue:
- [x] change folder names to be lowercase everywhere (under `build`, `datafiles`, `traceygraphs`, and `deploy` especially. (related to #1686)
- [x] make sure the `directoryNames` and `resourcePaths` for each example are correct given the new capitalization of folders (all lowercase)
- [x] start work on hooking info from the examples themselves rather than hardcoding the website (and make it more recipe-like)
- [x] Make some website sections clearer (DataTable and Package dependencies)
- [x] Remove unneeded manual references
- [x] Get Haddock docs for website

Ant13731:
@balacij, is there a way to import the `Main.hs` files from each example into the website? The `Body.hs` files all have different module names so those work, but all the `Main.hs` files have the same `Main` module name.

I tried changing the module names to `ExampleName.Main` but then I get an error that the examples can't find the file in `ExampleName/Main`.

Using `import Drasil.DblPendulum.Main`, or `import Main` don't work, as the compiler says they both don't exist. However, when I just try and load up GHCi for the website, it asks me which package I would like to take `Main from`. Is it possible to distinguish between the main files. Maybe something in the `stack` arguments?

balacij:
I placed the `Main.hs` file as an unexposed module, only in the executable. We can create a `Choices.hs` or `Artifacts.hs` file in the lib/ folder of each of them and place any information, that's currently only found in `Main`, in them, and then import them.

> Using import Drasil.DblPendulum.Main, or import Main don't work, as the compiler says they both don't exist. However, when I just try and load up GHCi for the website, it asks me which package I would like to take Main from. Is it possible to distinguish between the main files. Maybe something in the stack arguments?

If there is, I don't know of it. I think this is just the behaviour of GHCi but I might be wrong. What happens if we change back the module names of each `Main` file to the original `Drasil.X.Main`, expose them, and then import?

Ant13731:
> If there is, I don't know of it. I think this is just the behaviour of GHCi but I might be wrong. What happens if we change back the module names of each Main file to the original Drasil.X.Main, expose them, and then import?

I get the following error when manually changing the cabal file:
```
dblpendulum   > Preprocessing library for dblpendulum-0.1.24.0..
dblpendulum   > Cabal-simple_Z6RU0evB_3.0.1.0_ghc-8.8.4.exe: can't find source for
[...]

PR #2726, opened 2021-07-23
Using examples to generate the Case Study table on the drasil website

Ant13731:
Contributes #2723 

The website will now start to use the info from the examples. Here is the first part (focusing on `CaseStudy.hs`)
New files were needed in order to get at the code choices for each example. `Main.hs` is a hidden module, so they were transferred to `Choices.hs`


Ant13731:
Converting to a draft as I changed NoPCM's abbreviation to "NoPCM" from "SWHS". Is this change okay? (I think it affected a lot of the generated code, so I'm not sure)

smiths:
@Ant13731, it would be great if we could change NoPCM to actually use the abbreviation NoPCM throughout.  I cannot remember why off the top of my head, but I asked for that change a few years ago and I learned that it would break the NoPCM code.  Maybe that problem has been removed by other changes, or maybe you have fixed it?  Regardless, changing to use the name NoPCM would be a great change.

code/drasil-website/Drasil/Website/Example.hs
```diff
@@ -85,17 +105,16 @@ exampleDoxRefs path =[[(pendulum, [])],
...
+projectileC1, projectileC2, projectileC3, projectileC4, projectileC5 :: String
 pendulum = "DblPendulum"
 gamePhys = "GamePhysics"
 glassBR = "GlassBR"
 hghc = "HGHC"
 noPCM = "NoPCM"
 pdController = "PDController"
 projectile = "Projectile"
-projectileC1 = "Projectile_C_P_NoL_B_U_V_D"
+projectileC1 = "Projectile_C_P_NoLB_U_V_D"
```
balacij, line 117: Are we able to pull these strings from the examples SystemInformation/Choices? If not, we might want to figure out what information that the SIs are missing (si name, si description, choice names?) and add them (or tie everything together in something else that the drasil-website can better digest) so that we can pull them.

balacij: Everything is looking nice! Looking forward to seeing what's next for the website!

Ant13731: Oh, yes those strings should eventually move out - I've only really looked at the `CaseStudy.hs` file so far, but `Example.hs` will definitely be changed.

JacquesCarette (commented):
Looking good!

Ant13731:
>  I cannot remember why off the top of my head, but I asked for that change a few years ago and I learned that it would break the NoPCM code. Maybe that problem has been removed by other changes, or maybe you have fixed it? Regardless, changing to use the name NoPCM would be a great change.

I briefly looked through the code and it doesn't seem like it should break anything, but I'm not sure how to test it for sure. Other than that, I think these changes are good for now.

smiths:
The change in the name looks good to me.  I believe the problem was that the executable would always have the name SWHS, but your executable has the name NoPCM, right?

This is a nice improvement.

JacquesCarette (approved):

PR #2733, opened 2021-07-26
Remove Positioning Words from Figure

### LINKED ISSUE #2731: Figure position assumption
In the section of Physical System Description, it seems the generator will assume where the figure locates. The keyword "above" and "below" will be added. It works fine in html, but the pdf file has a different look.

It seems the Latex will insert the image at the beginning of the page, so the keyword "below" will most likely be wrong in some cases.

The miss-matched error can be reproduced in the double pendulum and projectile. To be honest, I think the problem occurs in all examples. Some examples just get lucky. When the image size is too big or the current page fills in with contents, the image will be inserted at the top of the next page. In that way, the "below" position still works fine. On the other hand, if the current page still has enough space, the image will be inserted at the beginning of the current page, then the "below" position will be wrong.

I think the quickest solution is to remove the position keyword after figures.

Screen Shot 2021-07-25 at 5 36 18 PM

@tingyuw Do you know how the keyword "above" and "below" be generated?

Ant13731:
Closes #2731


JacquesCarette (approved):
The fun of Drasil - 1 line change, 16 other documents change too.

PR #2735, opened 2021-07-26
switching dqdWr

Awurama-N:
contributes to Issue #2704


code/drasil-lang/Language/Drasil/Chunk/DefinedQuantity.hs
```diff
@@ -63,6 +63,9 @@ dqd' = DQD
 dqdWr :: (Quantity c, Concept c, MayHaveUnit c) => c -> DefinedQuantityDict
 dqdWr c = DQD (cw c) (symbol c) (c ^. typ) (getUnit c)
 
+dqdWr' :: (Quantity c, Concept c, HasUnit c) => c -> DefinedQuantityDict
+dqdWr' c = DQD (cw c) (symbol c) (c ^. typ) (Just $ findUnit c)
```
JacquesCarette, line 67: This should be marked as a temporary constructor that should not be used outside `drasil-lang`.

code/drasil-lang/Language/Drasil.hs
```diff
@@ -190,7 +190,7 @@ module Language.Drasil (
   , scale, shift
   , derUC, derUC', derUC''
   , fund, fund', compUnitDefn, derCUC, derCUC', derCUC''
-  , unitWrapper, getCu, MayHaveUnit(getUnit)
+  , unitWrapper, getCu, MayHaveUnit(getUnit), HasUnit(findUnit)
```
JacquesCarette, line 193: This should not be exported. This `findUnit` is a temporary fix as you change things around.

code/drasil-lang/Language/Drasil/Chunk/UnitDefn.hs
```diff
@@ -56,6 +56,9 @@ instance IsUnit        UnitDefn where
 class MayHaveUnit u where
    getUnit :: u -> Maybe UnitDefn
 
+class HasUnit u where
```
JacquesCarette, line 59: Can you change the name of this class? It clashes with another of the same name. It should be temporary too.  So `TempHasUnit`?

JacquesCarette (changes requested):

JacquesCarette (approved):
You'll need to do a master merge first, but this is good to go.

code/drasil-example/Drasil/GlassBR/IMods.hs
```diff
@@ -19,9 +19,9 @@ import Data.Drasil.Concepts.Documentation (goal)
 iMods :: [InstanceModel]
 iMods = [pbIsSafe, lrIsSafe]
 
-symb :: [DefinedQuantityDict]
-symb = map dqdWr [plateLen, plateWidth, charWeight, standOffDist] ++ 
-  [dqdQd (qw calofDemand) demandq]
+symb :: [UnitalChunk]
+symb = map ucuc [plateLen, plateWidth, charWeight, standOffDist] -- ++
+ -- [dqdQd (qw calofDemand) demandq]
```
balacij, line 24: I think this commented out dqdQd part still needed to be replaced.

JacquesCarette: Agreed.

JacquesCarette (changes requested):
As noted, and a master merge is needed too.

JacquesCarette:
Why hasn't this been worked on @Awurama-N ?

Awurama-N:
I was working on others, especially "addForAll", but i'm working on this now

code/drasil-example/Drasil/GlassBR/IMods.hs
```diff
@@ -20,7 +20,8 @@ iMods :: [InstanceModel]
 iMods = [pbIsSafe, lrIsSafe]
 
 symb :: [UnitalChunk]
-symb = map ucuc [plateLen, plateWidth, charWeight, standOffDist] -- ++
+symb =  [ucuc plateLen metre, ucuc plateWidth metre, ucuc charWeight kilogram, ucuc standOffDist metre, demand] -- this is temporary
```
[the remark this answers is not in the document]
JacquesCarette: I think that having units and having uncertainty are orthogonal, so I'm not sure either step is 'right'. The temporary fix actually looks ok for now.

JacquesCarette (approved):

balacij:
Sorry, can we revert this?

https://github.com/JacquesCarette/Drasil/pull/2735/commits/720bd4b9a86d6d2d38f8ab444a883ad6eadafad0 was meant to be a temporary commit while we were still figuring out how the code was working before the master merge. I think it reverted a few changes that were meant to stay in master.

Ant13731:
Yep, also I think one of the commits might have accidentally reverted the changes that separated all of the examples into their own subpackages.

JacquesCarette:
So it's going to be better to wait for #2787 rather than revert this one directly?

balacij:
I believe so, yes.

Issue #1800, opened 2019-08-02
Express constraints on concavity


bmaclach:
Currently in Drasil, constraints can either be expressed as a `Range`, `EnumeratedReal`, or `EnumeratedStr`.

Another type of constraint we need to express are constraints on concavity. An example in SSP is that the slip surface must be concave up (the slope between consecutive pairs of values from __x__<sub>slip</sub> and __y__<sub>slip</sub> must be monotonically increasing).

Issue #453 for monotonicity constraints may be relevant here.

JacquesCarette:
I removed "forall" from `Expr`, but it should be 'reborn' in `ModelExpr`, and then these constraints at the same time.

balacij:
Just a note, we don't have `ModelExpr` yet, it's still called `DisplayExpr`. However, I think we can still proceed with these 3 tickets without issue.

To start, we should add "forall" to DisplayExpr, approximately as follows:
1. Add the "ForAll" constructor to `DisplayExpr` (in drasil-lang). It should probably contain a UID for a thing to quantify over, maybe a space from which the variable belongs to (unless we can resolve it without explicitly mentioning it), and a statement DisplayExpr.
2. Run `make` and watch for the warnings, you will probably need to fix a few functions that pattern match on DisplayExpr.
3. For displaying "forall", we will need to introduce a new `Ops` constructor in the printing Expr (in drasil-printers) for "ForAll". The "forall" operator is the latex command "\forall". We would probably want to display as approximately `∀x ∈ R . statement` where x is the variable, R is the space it belongs to, and the statement is the carried DisplayExpr.

Afterwards, we can figure out how we want to encode concavity as a constraint, and go from there.

JacquesCarette:
This is a good one to put in a project to upgrade `DisplayExpr`. A few things were disabled from `Expr` that really need to come back. I think there are 4-5 issues like this (multiple of them linked already).

PR #2741, opened 2021-07-27
Adding `ForAll` to DisplayExpr

Awurama-N:
contributes to issue #1800


code/drasil-lang/Language/Drasil/DisplayExpr.hs
```diff
@@ -21,3 +22,4 @@ data DisplayExpr where
     SpaceExpr     :: Space -> DisplayExpr
     BinOp         :: DisplayBinOp -> DisplayExpr -> DisplayExpr -> DisplayExpr
     AssocBinOp    :: DisplayAssocBinOp -> [DisplayExpr] -> DisplayExpr
+    ForAll        :: UID -> Space -> DisplayExpr
```
balacij, line 25: With "ForAll"/universal quantification, we're saying that "given any variable that belongs to a space, this predicate function/relation will hold". Using [GADT syntax](https://en.wikibooks.org/wiki/Haskell/GADT#GADTs), the last component of the signature is the resulting type of the constructor, not a parameter of the constructor. So, it ends up that we're missing an extra DisplayExpr (for the predicate function) after the Space argument.

code/drasil-printers/Language/Drasil/Printing/Import/DisplayExpr.hs
```diff
@@ -34,5 +34,6 @@ dispExpr (AlgebraicExpr e)  sm = expr e sm
 dispExpr (SpaceExpr s)      sm = space sm s
 dispExpr (BinOp b l r)      sm = P.Row [dispExpr l sm, P.MO $ deBinOp b, dispExpr r sm]
 dispExpr (AssocBinOp b des) sm = P.Row $ intersperse (P.MO op) $ map (dispExpr' sm prec) des
+-- dispExpr (ForAll u _)       sm = _
```
balacij, line 37: The "where" statement is local to the `dispExpr (AssocBinOp b des) sm = P.Row $ intersperse (P.MO op) $ map (dispExpr' sm prec) des` line, so I think you will need to move this line after the `where` block.

Awurama-N:
i tried to make some changes but i'm still unsure of what to equate the functions to.

code/drasil-lang/Language/Drasil/Expr/Extract.hs
```diff
@@ -12,6 +12,7 @@ deNames (AlgebraicExpr e)  = eNames e
 deNames (SpaceExpr _)      = []
 deNames (BinOp _ l r)      = deNames l ++ deNames r
 deNames (AssocBinOp _ des) = concatMap deNames des
+deNames (ForAll u _)      = []
```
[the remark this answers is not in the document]
JacquesCarette: Do we really want to collect "u"? In some way, it's a bound name. It depends on what it's used for.

balacij:
> i tried to make some changes but i'm still unsure of what to equate the functions to.

For which functions, sorry?

For the deNames function, I wrote a note about it above. You can reuse the surrounding code too. The function is expected to return a list of the UIDs/names, so you will need to work with lists. The first parameter of ForAll will be one item that you need to collect. There's also a DisplayExpr parameter, so you will need to make a recursive call to `deNames` on it to get it's list of names, and then add the first parameter to it.

LearnYouAHaskell has a great primer for [working with lists](http://learnyouahaskell.com/starting-out#an-intro-to-lists) if you're interested too. In particular, `element : list` (Cons) will allow you to push an element to the top of a list.

Awurama-N:
for the ones that look incomplete, or have the unused `u` & `des`. 

oh okay

code/drasil-lang/Language/Drasil/Expr/Extract.hs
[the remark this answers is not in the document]
balacij: Ah, you're right! This is ultimately used for extracting/collecting UIDs to place in both the table of symbols and table of units, so it wouldn't need to be. It might be needed for the table of units if we find any issues later, but it probably shouldn't be put into the table of symbols, I imagine.

code/drasil-printers/Language/Drasil/Printing/Import/DisplayExpr.hs
```diff
@@ -36,4 +36,4 @@ dispExpr (BinOp b l r)      sm = P.Row [dispExpr l sm, P.MO $ deBinOp b, dispExp
 dispExpr (AssocBinOp b des) sm = P.Row $ intersperse (P.MO op) $ map (dispExpr' sm prec) des
   where prec = dePrecAssoc  b
         op   = deAssocBinOp b
--- dispExpr (ForAll u _ des)   sm = _        
+dispExpr (ForAll u _ des)   sm = P.Row [dispExpr des sm]
```
JacquesCarette, line 39: You're going to need to refine this display so that it actually shows the "forall". I think everything's already there in drasil-printers to support that.

JacquesCarette (changes requested):

code/drasil-printers/Language/Drasil/Printing/Import/DisplayExpr.hs
Awurama-N: like dispExprForAll? or deForAll?

JacquesCarette: It possibly doesn't need a new function. I meant displaying a proper forall, with the bound variable, etc.  There used to be something in `Import/Expr.hs` about a year ago that did that. Look through the history?

Awurama-N: hmm.. I can't seem to access the older commits for the git history of this file?

JacquesCarette: Not `Import/DisplayExpr.hs`, `Import/Expr.hs`.

balacij: The file got moved around, so it will be a bit difficult to search for it. You will need to check the original `Import.hs` file, and see from where the code came into that file, and continue the process until you found when it was removed from the file it was in.

To also give you an alternative route ... the goal is to display something that _approximately_ looks like `∀ x ∈ S • R`, where `x` is the variable, `S` is the space, and `R` is the proposition (here, it's just a DisplayExpr). The code looks like it currently only displays the `R`/proposition of the "ForAll". So, you need to add elements to the Row to also display (in order):
1. The `∀` symbol (you will need to define this one somewhere, probably the Ops datatype where the other ones are already)
2. The variable (see how similar `Expr`s are displayed -- `C` may be particularly helpful)
3. The `∈` symbol
4. The space from which the variable belongs (again, see how similar Spaces are displayed)
5. The `•` symbol
6. The proposition

There will probably be a few things further that we have to do to ensure that it is displayed properly, and there's also different styles of writing it, but I think this would be a good start.

balacij: Ah, just saw that you posted too. It looks like comments don't appear here in reviews as they're posted! It only shows active updates in normal conversations.

balacij:
There's a conflict now, so the CI won't run until you've merged in master and resolved the conflicts.

Awurama-N:
there was an error with naming this commit

balacij:
What was the error?

The [raw diff](https://github.com/JacquesCarette/Drasil/pull/2741/commits/c045a06405e32689248f38111f2cf968193ca1c9.diff) still looks good.

balacij:
@Awurama-N It looks like HLint is currently failing.

Did you figure out what the problem with the commit was, by the way?

Awurama-N:
nothing was showing up at first, but now i found out what was wrong. I'm trying to work on it, mostly inconsistencies with stable

balacij:
Nice! Is this PR ready now?

code/drasil-printers/Language/Drasil/HTML/Print.hs
```diff
@@ -225,6 +225,7 @@ pOps Point    = "."
 pOps Perc     = "%"
 pOps LArrow   = " &larr; "
 pOps RArrow   = " &rarr; "
+pOps ForAll   = " ForAll "
```
balacij, line 228: By any chance, did you try [`&forall;`](https://www.toptal.com/designers/htmlarrows/math/for-all/) here to see if that works?

Awurama-N: hmm i'll do that now

JacquesCarette:
Unsurprisingly, now needs a master merge.

JacquesCarette (approved):

JacquesCarette:
[No longer needs a master merge, so in it goes!]

PR #2744, opened 2021-07-28
Website Improvements Part 4

Ant13731:
Contributes #2723 
Contributes #2477 

Changes:
- New analysis section encases the Data Table and Package dependencies. The type and class-instance graphs also are linked to the website and show up in a table.
- Examples come before Case Study Table
- Added more description/detail to the section introduction
- Added more comments for the website and hooked into Haddock docs

@balacij, would you be able to take a look at the drasil-website package.yaml to see if the Haddock docs are generating properly? I looked at all the other packages and they seem to first define themselves as a library and then as an executable, so I did the same thing here. The docs wouldn't generate unless I included it as a library in the yaml file.


balacij:
Haddock, by default, looks only for __exposed modules__ of __libraries__ to show, it completely ignores executables and hidden modules. This is why, previously, our Haddock docs were missing lots of modules (they weren't exported by the library they belonged to), and drasil-examples didn't show up on our Haddock docs. Now that we've split up the examples into library + minimal executable, most of their code is shown on the [haddock docs](https://jacquescarette.github.io/Drasil/docs/full/index.html#n.4.13). The only file from which we're missing stuff is from Main.hs, which is fine I think.

Since you're changing it to library+executable, it might be good to further separate their source code folders like the drasil-example projects, otherwise we end up compiling files shared between the executable and the library twice. There are also some weird edge cases with cabal when they share the same source directory, cabal just spams 'your project needs to add X file into your deps for this executable'.

~:
```
drasil-website:
    lib/ <- source folder for all "library"-like code that we want shown on Haddock
    app/ <- contains "executable" code that won't show on Haddock, keeping it minimal is probably best for drasil-website
    ... <- the rest of the local files
```
with the `executable` website in the `package.yaml` having a dependency on the `drasil-website` library. This way, we avoid compiling the files twice, and we avoid the cabal edge case spam.

Ant13731:
Okay, thanks! I'll give it a go then.

Ant13731:
I switched all the files over but I'm stuck on getting it to compile. I have this error message:
```
Error: While constructing the build plan, the following exceptions were encountered:

Unknown package: website

Some different approaches to resolving this:


Plan construction failed.
```

balacij:
Oo, it looks like you moved the package 1 level deeper into `drasil-website`.

What happens if you cut out the deeper `website` folder and just unpack it's contents into `drasil-website`?

Also, the name of the package is still `drasil-website`, so I think you still should be installing via `stack install $(stackArgs) "drasil-website"`. I don't think Stack checks for the names of executables because `stack install` is really just `stack build --copy-bins` meaning "build this software package and copy it's binaries into ~/.local/bin".

Ant13731:
Ah, okay I see, yea that was it. I also needed to change the dependencies from `website` to `drasil-website`.

JacquesCarette:
Needs a master merge. Doing the review now.

JacquesCarette (approved):

PR #2746, opened 2021-07-28
Reducing manual website referencing

Ant13731:
Builds on #2744 
Contributes #2723


JacquesCarette (approved):
Looks good, but also needs a master merge.

PR #2754, opened 2021-07-30
Remove cabal files in favour of letting hpack regenerate them all the time + fix imports in each package

### LINKED ISSUE #2750: Differences in generated cabal files - remove autogenerated cabal files?
I recently ran `make` to build the Drasil examples locally and now the cabal files have been modified.  Now that the cabal files are generated, should we remove them from version control?

I had a look at my generated file versus the one in the repo.  The difference is in the comments at the start.  The hpack version number is different and the hash.  Here are the comments in my local version of `drasil-build.cabal`:

```
-- This file has been generated from package.yaml by hpack version 0.34.4.
--
-- see: https://github.com/sol/hpack
--
-- hash: 2e0c5df1143056f1218ea9befb510511e30bfb9f342e9b7b6481244e84233c6f
```
The version in the repo has hpack version of `0.33.0` and a different hash.

I will just reset my local cabal files, so this isn't a big deal.  However, if the cabal files are frequently updated locally, we might want to remove them from version control.

balacij:
Closes #2750


JacquesCarette (approved):
A master merge will be needed, even though the effect will be to immediately delete all of those files!

balacij:
After this is merged, everyone might need to run `find . -name "*.cabal" -type f -delete` on their machines to clear out the old cabal files, run `make clean`, and then a fresh `make`.

Issue #2646, opened 2021-06-30
Typing the Expression Language


balacij:
Long-term ticket for general discussion on, and tracking the progression of "adding typing to the expression language".

balacij:
@JacquesCarette 
Following up with discussion on https://github.com/JacquesCarette/Drasil/commit/ff34fe093b8898383232e2822f478a394243aef9#r52161989, the preface discussion of #2482 related to the `DiscreteS` and `DiscreteD` usage, and https://github.com/JacquesCarette/Drasil/issues/2118 about `Enumeration`s and `Discrete` spaces.

Currently, both of `EnumeratedStr` and `EnumeratedReal` are essentially unused (both used unused constraint lists), `DiscreteS` is unused, and `DiscreteD` is used (though, I believe it's used for a display hack due to the EnumeratedReal being essentially unused).

`EnumeratedStr` and `EnumeratedReal` are both used in exactly 1 spot each:
https://github.com/JacquesCarette/Drasil/blob/b8da80982650c50c1abc97844fc35bd0a8211bc1/code/drasil-example/Drasil/GlassBR/Unitals.hs#L108-L116
Note: In nomThick, `enumc = EnumeratedReal Physical`.

Removing both constraints does not alter stable because they go unregistered in the final constraint list used:
https://github.com/JacquesCarette/Drasil/blob/b8da80982650c50c1abc97844fc35bd0a8211bc1/code/drasil-example/Drasil/GlassBR/Unitals.hs#L67-L73
Note: the `inputDataConstraints` is what's registered for calculating the list of constraints for [Data Constraints Section of the SRS](https://jacquescarette.github.io/Drasil/examples/GlassBR/srs/GlassBR_SRS.html#Sec:DataConstraints).

`DiscreteD` is used as a display hack for displaying the values for which are allowed for the `EnumeratedReal`s allowed for `nomThick`. This occurs in the line defining `nomThick` as a unital, giving `nomThick` a noun phrase term/name with a `t element of {...}` suffix:
https://github.com/JacquesCarette/Drasil/blob/b8da80982650c50c1abc97844fc35bd0a8211bc1/code/drasil-example/Drasil/GlassBR/Unitals.hs#L108-L110
Note: `displayDblConstrntsAsSet` is a function that creates a `DisplayExpr` from the UID of `nomThick` and the list of doubles allowed.
There is no similar thing done for `DiscreteS` and the `glassType`.

[...]

PR #2760, opened 2021-08-03
Convert DisplayExpr into a 1-1 clone of Expr, renaming to ModelExpr

### LINKED ISSUE #2743: Implementing extensible expression language
As a component of #2646 

This is originally from an email I sent to Dr. Carette:

I've been trying to implement the "Datatypes à la carte" styled Expr/ModelExpr language for a while now, but I've been struggling to get pattern matching to work well when merging languages, and I've also been struggling to get a typed version of it working.

I found Sylvain Henry's continuation of the "Datatypes à la carte" work. He built a very similar system that relies on encoding the extensible sum types using lists of types (called "Variants"), where expressions of the language can take on any value of type X, where X is an element of the list of types designated. He then creates a variant using recursive sum types, naming these variants "Extensible ADTs" (EADTs).

He has a lot of documentation on EADTs: https://docs.haskus.org/eadt.html

And a very nice introduction/background page:  https://docs.haskus.org/eadt/background.html

I think Henry's approach is interesting. One issue that I had with the "Datatypes à la carte" approach I had is in creating values that contain type information for a whole language feature set (e.g., everything in Expr + everything in ModelExpr) when the term contained only one features from one of the languages. I believe this issue would occur if a TheoryModel expects ModelExprs but some instance of a  TheoryModel ends up being simple and ends up with something that only relies on things in Expr. Henry's approach appears to have no issues with this at all.

He also builds multiple libraries to support his work: https://hackage.haskell.org/packages/search?terms=haskus 

Specifically, the library containing EADTs, and supporting Variants: https://hackage.haskell.org/package/haskus-utils-variant 

I've been able to get pattern matching working nicely with EADTs (admittedly, his library and documentation makes a lot of the work straightforward). However, I haven't been able to get far in getting it typed, and I haven't been able to find any EADT-related resources online. The issues lie in missing instances for particular typeclasses in his library that I haven't looked into enough yet.

I have 2 more documents lined up for me to read that might be helpful for adding static typing via GADTs using the pattern functor/"Datatypes à la carte" approach:

https://www.researchgate.net/publication/228841104_Compositional_data_types
 
http://www.timphilipwilliams.com/posts/2013-01-16-fixing-gadts.html
However, I'm a bit concerned that the same pattern matching/deconstruction issues will occur (I wonder if we can merge the efforts?).

Aside: Henry has an interesting example of dynamic type checking: https://hsyl20.fr/home/posts/2018-05-22-extensible-adt.html#type-checking-example-paramorphism-example

Do you have any thoughts/advice on how I should proceed or what I should prioritize?

Specifically, do you think I should continue with building a prototype Expr/ModelExpr language using the "Datatypes a la carte" (or EADT style) despite not yet being able to create a typed variant of it? Though, these 2 papers lined up might resolve this issue.

balacij:
Closes #2743 
Contributes to #2646 

Builds on #2754


JacquesCarette:
needs a merge

JacquesCarette:
Still needs a merge, even after the other stuff got fixed.

balacij:
Thank you, it required a bit of work in the merge, moving things around from DisplayExpr to ModelExpr. Unfortunately, it gets lost in the merge commit.

@Awurama-N Please note this PR because it moves the ForAll to the new ModelExpr.

JacquesCarette (approved):

Issue #2759, opened 2021-08-03
Creating a log file for examples


Ant13731:
Building off of #1541 and @balacij's suggestions from a few meetings ago, I think it would be good to make a log file that would tell the user the following:
- all UIDs in the chunk database
- chunks that are not dependent on anything
- chunks that are not used (referenced) by anything

The first one should be easy enough to just print the UIDs (maybe term and abbreviation too if they exist?), and the other UIDs can be obtained through the traceability matrix/graph functions, but I'm not sure of the best way to implement it. 

It might be nice to have its own make target (something like `example_logs`), but I'm not sure how to get the `SystemInformation` from that point. Maybe we could create a haskell file in the `scripts` folder that depends on the examples, and just pull the system info from there.

Or we could just make a function and add it to each example's `Main.hs` file, so that the logs generate every time the file is run (similar to how the traceability graphs always generate up-to-date dot files). But I don't think we would want the logs in `stable`, or for the logs to run every time.

I was just thinking of making the three points into alphabetically ordered lists just to get the information down, but there's probably a better method of seeing the information.

JacquesCarette:
Good idea. It feels like this ought to be an option inside generating the examples itself, as opposed to a separate target. Some kind of `Debug` option in `Choice`?

Ant13731:
Okay, I like the idea of using the example itself (it will probably be much easier to work with). I'm not sure if we can use the existing choices type since that's meant for generated code, but I do have an idea for selecting options.

In #2751 we said that users should have a list of options to feed into the generator function for the kinds of files we want to output. We could instead use a record type to do this, something like:
```Haskell
data SRSChoices = SRSChc {
html :: Bool,
pdf :: Bool,
log :: Bool,
ntbook :: Bool }
```
Maybe we shouldn't use booleans, but something like that. Or perhaps:
```Haskell
data SRSChoices = SRSChc {
srsTypes :: [DocType]
showLog :: ShowLog}
```
where `ShowLog` is a type with two constructors. We could also add a `Debug` option to the Drasil generators, so the `Main.hs` file would look like:
```Haskell
main :: IO()
main = do
  setLocaleEncoding utf8
  gen (DocSpec SRS     "Projectile_SRS") srs printSetting
  gen (DocSpec Debug "Projectile_SRS") srs printSetting
  gen (DocSpec Jupyter "Projectile_SRS") srs printSetting
[...]

PR #2771, opened 2021-08-05
Allow users to choose generation file type

### LINKED ISSUE #2751: Allow Users to choose which file formats to generate SRS in
Based on #1732, from Dr. Carette and Dr. Smith,
> Having DocSpec mean the kind of document to generate makes sense. Now, we might want a bit more control, i.e. the choice to generate just the HTML or just the latex version of the SRS (for example). This shouldn't be controlled by a 'file type', but something more meaningful. So maybe a list of kinds? And then instead of `Filename`, perhaps `DirectoryRoot` ?

> We won't always want to generate all possible file types for all our outputs. Within our GitHub page, while we are developing Drasil, it makes sense for us to generate everything we can generate. However, when Drasil makes it out into the wild and someone is using it on a real project, they might not want everything. This becomes more noticeable if we extend the file formats for the SRS documents. For instance we could add a text version, or an rtf version. The person using the software might have no need for the tex version, and choose not to generate it. I don't think fine grained control on the types of files that are generated is a high priority, but we should allow for flexibility in the future.

Ant13731:
Also set up the generators to allow for logging from #2759 
Closes #2751 

Jupiter notebooks still need to be included in this new setup. For now, they just use a separate pattern match. Should they be included as an SRS option?


JacquesCarette (approved):
Looks good - but need a merge

JacquesCarette:
(and no longer needed merge, so in it goes)

Issue #1541, opened 2019-06-11
Automate Check That All Chunks Are Used


smiths:
With the reorganized traceability matrix (#1537), I feel like we are close to automating an exciting check.  Almost every chunk (concept?) should be used somewhere, or else, it doesn't make sense that is is included in the documentation.  For instance, if we introduce an assumption, then this assumption should be invoked somewhere.  If it isn't invoked, why is it even in the documentation?  We are in the process of discussing this in the Projectile example.  The exception to this is the requirements.  Requirements use the other pieces of information, but they themselves aren't used in the SRS.  They are used by the design and by the implementation.

Besides wanting to know whether every chunk is used, we also want to know if there are any chunks that do not use any other chunks.  There will be some pieces of information, like some theory models, that might not depend on anything else, but most chunks should reference some other piece of information.

I could envision a day when the compilation would fail, or at least give a warning, if our relationship rules were violated, but we don't want to be too hasty.  Since we are still working out the expected relationship between all of the chunks, I would like to take the first step of just highlighting the potentially odd cases. 

It would be great to have a function (or functions) that takes in the traceability matrix information and returns two things:

- a list of all chunks that are not referenced by anything (I believe this corresponds to an empty column in the traceability table.)
- a list of all chunks that do not depend on anything (I believe this corresponds to an empty row in the traceability table.)

I might have got the row and column labels wrong above.  I haven't verified against the current shape of the traceability table.  The main point is that empty rows or columns are interesting, and it would be nice to identify them.  If we had this information for each example, we could look at each instance and decide whether the result is correct, or, more likely, how we forgot to link the information.

In my thinking an automated test l like this would again highlight the potential power of Drasil.  

I've assigned this issue to @Mornix, but if it is bigger than I think it is, @JacquesCarette can reassign it, or we can label it as a future enhancement.

Mornix:
I think this is possible, but a little bit out from being a hard error. For example, with the new traceability matrices, I can see that GlassBR's glassType assumption is unused (at least as far as the SRS is concerned, however it is used the the codegen. I am open to flagging this as a "gen-time" warning as a current option.

Are requirements the only "roots" of traceability currently? (i.e. If we can't transitively reach a node starting from the requirements then it should be flagged.)

JacquesCarette:
[...]

PR #2774, opened 2021-08-09
Adding example log sheets for chunks and chunk UIDs

### LINKED ISSUE #2759: Creating a log file for examples

Ant13731:
Builds on #2771 
Closes #2759 (maybe #1541?)
Contributes #2675 

Logs now generate with every chunk from all maps of the chunk database. The first three tables are probably the more important ones. 

I also had to manually add some terms to the database for NoPCM and PDController, as they gave term errors, but I think we can revert that as we complete #2675.


JacquesCarette (approved):
Looks good, needs a merge

JacquesCarette:
And no longer needs a merge, so going in.

PR #2777, opened 2021-08-10
Symbol Decoration for Magnitude and Delta

### LINKED ISSUE #1605: Improve Symbol Decoration Infrastructure

Ant13731:
Closes #1605 

Currently, the symbols decorated with delta will appear at the end of that letter name. Specifically, `Δv` in GamePhysics will always order itself below all of the other `v` variables. @smiths, should this be changed so that `Δv` is placed right after `v` rather than after all of the other `v` values like `v_i` and `v_f`? The same thing will happen with the euclidean norm of vectors (`||v||`), but there is no use case of it in the examples.


smiths:
Yes, I would prefer delta v to be with v.  That is the way it would usually be written in a continuum mechanics textbook.  The same goes for the magnitude of v.  If this is possible, it would be a nice improvement.

JacquesCarette (approved):

JacquesCarette:
Needs a merge.

JacquesCarette:
No longer needs a merge - going in.

Issue #2783, opened 2021-08-12
Move literals out of Expr


balacij:
At times, we want just the Constant components of Expressions. We can move the constant components of Expr/ModelExpr into a new Constant data type that strictly contains them.

PR #2661, opened 2021-07-07
Remove Deriv from CodeExpr

balacij:
Contributes to #2628

Issue #2599, opened 2021-06-11
TM versus GD versus IM versus DD Discussion


smiths:
Discussion about the differences between TMs (Theoretical Models), GDs (General Definitions), IMs (Instance Models) and DDs (Data Definitions) come up all the time when working with Drasil.  I've created this issue so that I can present the current thinking on these definitions and propose my current idea on how to modify the definitions and improve the distinction.

The terminology and examples will all be from the viewpoint of an external reader of the SRS (Software Requirements Specification) documentation.  The distinctions made are from the point of view of the document reader.  I'm leaving for a separate discussion what information should be reflected in the internal Drasil representation of this knowledge and how best to encode this information.

I will first present the definitions I've been using for years.  I no longer think these definitions are adequate, but I won't get into that discussion until after I've presented the mental model that has been used to date.  I'll also define the supporting terminology of Goals and Assumptions.

**Goal**

A goal is a functional objective the system under consideration should achieve. Goals provide criteria for sufficient completeness of a requirements specification and for requirements pertinence. Goals are written abstractly, with a minimal amount of technical language and no equations. They should be understandable by non-domain experts. Goals are refined by the instance models. 

A sample goal from SWHS is "Predict the temperature of the water over time."

**Assumptions**

The assumptions are a refinement of the scope. The scope is general, where the assumptions are specific. All assumptions should be listed, even those that domain experts know so well that they are rarely (if ever) written down.  The assumptions will be used to refine the theoretical models and general definitions.

A sample assumption is "The only form of energy that is relevant for this problem is thermal energy. All other forms of energy, such as mechanical energy, are assumed to be negligible."

**Theoretical Model**

Theoretical models are abstract mathematical equations that will be refined into the instance models that are used to satisfy the goals for the project.  The theoretical models are general.  They will potentially come up in many different physical problems.

[Issue not done - submitting as a way to save it]

**General Definition**

**Instance Model**

**Data Definition**

JacquesCarette:
This is fantastic - definitely spells out lots of terminology. I need some time to have it sink in, and make a proposal.

One note: different disciplines abuse the words 'theory' and 'model' a lot. And these 'disciplines' can be sub-disciplines. The word 'model' means many many different things, and is fraught with danger because of that. Theory is not quite as bad, but also suffers from overloading.

smiths:
[...]

PR #2791, opened 2021-08-13
Parameterize QDefinition, reorganize DataDefinitions (for the QDef change), and move Deriv to ModelExpr

### LINKED ISSUE #2628: Move Expr's `Deriv` into `DisplayExpr`
We might need to convert DisplayExpr from the existing style of wrapping things into a 1-1 clone of Expr.

balacij:
Closes #2628 
Contributes to #2646 
Builds on #2824 + #2839

Creating a draft PR because it's quite far from completion, and it would require a few more changes in how we interact with `Model/Expr`s (which likely wouldn't end up being 'final' with this PR!).

Changes + discussion:
* `QDefinition`s are paremeterized on their `Expr` type because we need `QDefinition`s to be able to carry either `Expr`s or `ModelExpr`s for different purposes (e.g., a TM's equational model might be able to contain derivation, but an IM's equational model should not be allowed to contain a derivation [instead, the IM should be converted into some other model that allows for the derivation and would handle it appropriately]). Right now, I parameterize it on things that can be `Express`ed as a `ModelExpr`, which is essentially just `Expr` and `ModelExpr` at the moment. Note: This, as a means to an end, contradicts what we said we might want to do in [a comment on the original issue](https://github.com/JacquesCarette/Drasil/issues/2628#issuecomment-875937992) (see 4th item).
* `DataDefinitions` are parameterized on their `Expr` type as well because they contain `QDefinitions`. Our existing `DataDefinitions` are not quite fully used yet. For example, to my knowledge, the ones that contain derivatives are not yet used, they're strictly shown. Splitting them up based on the ones that contain `Expr`s vs `ModelExpr`s may prove helpful if we figure out a way to use those `ModelExpr`s appropriately. However, this may also be a sign that we need some form of `ModelKinds` in them as well (perhaps similar to how it's done in Instance Models).
* ModelKinds are becoming _partially_ typed based on their 'computability' (using Expr/ModelExpr at the moment, but that should change using a new type). However, I hope that this typing of them will allow us to create an idea of 'computable' models vs 'incomputable' models. For example, the `InstanceModel`s should be restricted to only `ModelKind`s that are 'computable', and `TheoryModel`s/`GenDefn`s may contain any of them (however, they should tend towards only containing the 'incomputable' models).
* Deriv is moved to `ModelExpr` strictly.
* At time of creating this PR, only GlassBR works, but I'm working my way through fixing the others.
* The code is still largely in-progress, so the CI should fail and there will be many dangling `TODO` notes.

Any feedback/notes would be much appreciated.

Thank you!


code/drasil-lang/Language/Drasil/Chunk/Eq.hs
```diff
@@ -20,125 +27,173 @@ import Language.Drasil.Chunk.DefinedQuantity (DefinedQuantityDict, dqd, dqd')
...
+  QD :: Express e => DefinedQuantityDict -> [UID] -> e -> QDefinition e
+
+qdQua :: Express e => Lens' (QDefinition e) DefinedQuantityDict
+qdQua = lens (\(QD qua _ _) -> qua) (\(QD _ ins e) qua' -> QD qua' ins e)
+
+qdInputs :: Express e => Lens' (QDefinition e) [UID]
+qdInputs = lens (\(QD _ ins _) -> ins) (\(QD qua _ e) ins' -> QD qua ins' e)
+
+qdExpr :: Express e => Lens' (QDefinition e) e
+qdExpr = lens (\(QD _ _ e) -> e) (\(QD qua ins _) e' -> QD qua ins e')
```
balacij, line 51: Unfortunately, the DataTypeContexts language extension is deprecated (and apparently does nothing anyways), so I switched to using GADTs here + DataDefinitions. `makeLenses` doesn't appear to work well with GADTs, so we manually create Lenses for them.

balacij: I originally wanted to, at the type-level, only allow Exprs/ModelExprs as the "RHS" of QDefinitions with minimal code reuse, but now I'm starting to doubt the need for the GADT syntax + Express constraint on the constructor. Perhaps we should revert it back to what it was originally?

JacquesCarette:
The main question is, at a design level, what invariants are you trying to capture in the types of the new/changed structures? What are you trying to achieve? [I believe what you're trying to achieve is good, the reason to have you spell it out in detail is to be able to match the technical solution to the requirement.]

code/drasil-lang/Language/Drasil/Chunk/Eq.hs
JacquesCarette: You might want to investigate going with
```haskell
data Exprs = Simple | Model

data QDefinition (e :: 'Exprs) = EC
  { _qua    :: DefinedQuantityDict
  , _inputs :: [UID]
  , _equat  :: e
  }
```
where I've used "promotion" to promote a data type to a kind. This limits `e` to be of kind `Simple` or `Model` only.  https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/promotion.html 

I'm pretty sure my code isn't quite right, but something like that should work.

Another option is to use GADTs sort of directly, as follows:
```haskell
data QDefinition e where
  QDE :: DefinedQuantityDict -> [UID] -> Expr -> QDefinition Simple
  QDM :: DefinedQuantityDict -> [UID] -> ModelExpr -> QDefinition Model
```
which requires fewer extensions and fancy stuff, so it likely preferable.

JacquesCarette (changes requested):

code/drasil-data/Data/Drasil/Equations/Defining/Physics.hs
```diff
@@ -16,11 +16,12 @@ import Data.Drasil.Concepts.Documentation (body, constant)
...
 
-newtonSLEqn, weightEqn, weightDerivAccelEqn, weightDerivNewtonEqn, weightDerivReplaceMassEqn,
+weightEqn, weightDerivAccelEqn, weightDerivNewtonEqn, weightDerivReplaceMassEqn,
   weightDerivSpecWeightEqn,
-  hsPressureEqn, speedEqn, velocityEqn, accelerationEqn  :: Relation
+  hsPressureEqn, speedEqn :: Relation
 
-newtonSLEqn               = sy QPP.mass `mulRe` sy QP.acceleration
+newtonSLEqn :: ModelExpr
+newtonSLEqn               = express $ sy QPP.mass `mulRe` sy QP.acceleration
```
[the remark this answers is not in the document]
balacij: I initially changed it because it was only used in a TheoryModel. At the time, it only allowed ModelExprs in it, but I think we can loosen that requirement and revert this to an Expr.

code/drasil-lang/Language/Drasil/Chunk/Eq.hs
balacij: Thank you!

The first solution is very interesting. I'll try to play around with it in a small test area. The second solution also looks good, it's definitely an improvement over the constraints method!

balacij: Just a small progress report.

I couldn't quite get the first solution working without GADTs, which then ends up looking like the second solution.

With the second solution, would we need to duplicate each of the smart constructors of the qdefinitions (to create simple and model constructors)? I don't seem to see an easy way to parameterize over the different Exprs nicely. Additionally, with #2783 in mind (assuming it's a good idea), we would potentially have `QDefinition Constant`s appearing, meaning that `QDefinitions` would have 3 kinds of smart constructors (1 for each language).

balacij: What do you think of creating an empty typeclass (like the `Quantity` one) that is constrained to `Express`ible things for which we only define for `ModelExpr`, and `Expr`? With this, we could use less `Express` constraints everywhere and have more of this one, and since we would only instantiate it for `Expr` and `ModelExpr`, we have essentially the same effect as `Exprs`.

balacij:
> The main question is, at a design level, what invariants are you trying to capture in the types of the new/changed structures? What are you trying to achieve? [I believe what you're trying to achieve is good, the reason to have you spell it out in detail is to be able to match the technical solution to the requirement.]

I think that one of the most important things we're trying to capture here is with regard to the conversion of various things (Models, QDefs, DataDefs, etc) into GOOL. Specifically, we want the conversion of `InstanceModels` into code to be total, and we want to restrict this at the `ModelKind` level.

Currently, the conversion of InstanceModels into code is not total because are instances where generation fails (for various reasons). One of these reasons is that there are terms of the `Expr` language that are a bit difficult to compute on their own/directly. For example, differentiation and integration isn't quite simple enough that we can directly compute it. As such, we are trying to make all of `Expr`'s vocabulary contain only things that are simple enough that they can be directly converted into `CodeExpr` (& GOOL). This involves moving the things that aren't directly computable into `ModelExpr`. For the things in `ModelExpr` that aren't directly convertible into GOOL, we might be able to write their conversion manually.

Finally, with how I changed the structures in this branch, we are using the type parameterized things to discuss their ability to be directly converted into code.

For example, the following things are expected to have a total conversion into code:
* `QDefinition Expr`
* `DataDefinition Expr`
* `ModelKind Expr`                 <--- I think this is a hacky type to indicate 'computable' here, because it may not _really_ be made computable directly by an `Expr` [more on this below]
* `InstanceModel` <- InstanceModels carry `ModelKind Expr`s, so we expect this to also have a total conversion

Meanwhile, the following will not have a total conversion:
* `QDefinition ModelExpr`
* `DataDefinition ModelExpr`
* `ModelKind ModelExpr`
* `GenDefn` <- this and `TheoryModel` should carry `ModelKind ModelExpr`s [This is bad, even as it is -- it really shouldn't care what kind of `ModelKind` it carries to be honest. This needs to be fixed.]
* `TheoryModel` <- ^^^

It does appear that I'm misusing Expr/ModelExpr to show this meaning. Specifically, as noted, `ModelKind Expr` is meant to be "models that are directly computable". Now, this may very well be through containing only simple `Expr`s, but it's different for each `ModelKind`. For example, for an `EquationalModel` to be a computable model/`ModelKind Expr`, it's `QDefinition` must be a `QDefinition` containing an `Expr` (e.g., `QDefinition Expr`). For `DEModels`, they contain `ModelExprs` to show whatever the DE is, and then must be paired with `ODEInfo` packets that contain direct `CodeExpr`s to be a `ModelKind Expr`. Note: DE part is not complete yet, we still need to move the `ODEInfo` packets into the constructors of `DEModels` I think.

Perhaps a better encoding of these are as follows:

Things that should have total conversion into computable things:
* `QDefinition Expr`
* `DataDefinition Expr`
* `ModelKind Computable`
* `InstanceModel` <- these will be forced to carry strictly `ModelKind Computable`s

Not-total conversion:
* `QDefinition ModelExpr`
* `DataDefinition ModelExpr`
* `ModelKind Incomputable`
* `GenDefn` <- no restriction on what kind of `ModelKinds` it can contain [e.g., Computable or Incomputable, it shouldn't care]
* `TheoryModel` <- ^^^

Alternatively, it might be even better to change the `Computability` (`Computable`/`Incomputable`) with either `Abstractness` (`Concrete`/`Abstract`) or `Obersvability` (`Observable`/`Unobservable`).

balacij:
There's currently some stable breakage due to needing to split up the DataDefinitions by Expr & ModelExpr. It reorders them in DblPendulum, but some disappear in other examples. I'm not quite sure yet, so I'll need to investigate the lists, and see if we can retain the ordering.

balacij:
It appears that DblPendulum's [DD:positionGDD](https://jacquescarette.github.io/Drasil/examples/dblpendulum/SRS/srs/DblPendulum_SRS.html#DD:positionGDD) and [TM:velocity](https://jacquescarette.github.io/Drasil/examples/dblpendulum/SRS/srs/DblPendulum_SRS.html#TM:velocity) carry the exact same information. However, when I removed it from the SI's DataDefinitions list, it removed the DD entirely.

code/drasil-docLang/Drasil/DocDecl.hs
```diff
@@ -133,7 +133,7 @@ mkDocDesc SI{_inputs = is, _sysinfodb = db} = map sec where
   scsSub Assumptions = DL.Assumptions $ fromConcInsDB assumpDom
   scsSub (TMs s f) = DL.TMs s f $ allInDB theoryModelTable
   scsSub (GDs s f dd) = DL.GDs s f (allInDB gendefTable) dd
-  scsSub (DDs s f dd) = DL.DDs s f (allInDB dataDefnTable) dd
+  scsSub (DDs s f dd) = DL.DDs s f (allInDB eDataDefnTable) dd -- TODO: What about the ModelExpr ones?
```
balacij, line 136: Well, I suspect it's around this area that's causing the `DataDefinition ModelExpr`s` to not appear. I'm not quite sure how to fix it yet however, still researching it.

balacij:
A lot of the regressions to stable are fixed, but the order that we show the DataDefinitions are now changed. Previously, we had:
```
DD SECTION
-------------------
DD1 Expr
DD2 Expr (contained Deriv term)
DD3 Expr (contained Deriv term)
DD4 Expr
...
```
all gathered from 1 list, with Exprs held in each DD. However, now we parameterized by the Expr-kinds in DDs/QDs because we've moved Deriv out of Expr and into ModelExpr. So, now, we have to group the DD Exprs and the DD ModelExprs separately. We've now ended up with something more like:
```
DD SECTION
-------------------
DD1 (Expr) <- all will be usable in code generation & by theories!
DD2 (Expr)
...
DDM (Expr)
DD1 (ModelExpr) <- these are not usable in code generation but may be used by theories
DD2 (ModelExpr) <- these are also where the 'Deriv' containing DDs have been moved to
...
DDN (ModelExpr)
```

Now I'm trying to figure out if this is potentially a good change and should do something extra, or if it's a regression and we should be striving to revert the change in ordering.

Do you have any thoughts on this @JacquesCarette @smiths ? This also heavily relates to past discussion on #2599 and #2661.

balacij:
Update: I spent some time trying to figure out how I can merge the 2 relevant kinds of DataDefinitions; `DataDefinition Expr` and `DataDefinition ModelExpr` together into 1 list & 1 map so that we can print them intermixed instead of "one pool of them after the other". I initially tried wrapping them into a `DataDefn` wrapper that removes the typing, but that's proving to cause a lot of noise on the code. Now I'm realizing that this may have been the wrong path to go down. A 'more correct' solution may be to add a list to the `SRSDecl` area (specifically, in the `SCSSub`) to designate ordering of the DDs.

@JacquesCarette What do you think about this solution instead? It would involve adding a `Maybe [UID]` as a parameter of each of the constructors in:
https://github.com/JacquesCarette/Drasil/blob/0bea97d81fdd7eb4e5c5aa74aaa5112dbdb2c40d/code/drasil-docLang/Drasil/DocDecl.hs#L72-L87
(Note: This data structure is used to designate what things are in the output, but currently don't order the fields, they just rely on the implicit ordering gathered from the ChunkDB's maps.)

Specific current example, from dblpendulum,
https://github.com/JacquesCarette/Drasil/blob/0bea97d81fdd7eb4e5c5aa74aaa5112dbdb2c40d/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs#L81-L89
It designates what things to display, but not the ordering.

JacquesCarette:
Sorry for being absent on this one, even though it's rather important. But you do seem to be on top of it.

I was going to suggest yet another option (but don't rush to implement this!) : 
- keep things typed
- when putting things in the SRS language, use 2 different lists
- use an explicit sorting criteria when rendering them (into a single list)

The option of using `Either (QD E) (QD ME)` did not seem so bad either.

However, it is also true that, once we put things into the SRS language, by and large, we don't care about the types anymore. I don't think we use them, and we're mostly going to print things. Oh, we might later do other things, but those should be done on things directly. Not necessarily on `SystemInformation`, but something pre-SRS.

I think this leads to some kind of design decision: an SRS (as a data-structure in Drasil) should be analyzable **as a document**, but not more than that. The components of the SRS can be analyzed too, but this should be done prior to them being assembled into the data-structure.

balacij:
No worries at all! Thank you!

If I'm understanding your suggestion correctly, I _think_ that what I tried yesterday was approximately that.

I had attempted to use a `type ModelOrder = Maybe [UID]` as a explicit model order for each of the TMs, GDs, DDs, and IMs individually. I made this attempt on the [`moveDerivToModelExprV3`](https://github.com/JacquesCarette/Drasil/compare/master...moveDerivToModelExprV3) branch, it was based off this branch but diverges after 0bea97d81fdd7eb4e5c5aa74aaa5112dbdb2c40d / last commit of Aug 30 (the [comparison](https://github.com/JacquesCarette/Drasil/compare/moveDerivToModelExpr...moveDerivToModelExprV3) to this branch is a little wonky on GitHub's web viewer).

It worked almost as well -- the list of DDs back went to normal, but the traceability tables were still out of order, and that appeared to require at least 1 more datatype alteration for passing around the ModelOrder as well. Ultimately, that attempt only required 1 datatype alteration in drasil-docLang to get at least half-way. When I came across the Traceability table issue, I realized that what I was doing was similar to your approach for creating ModelKinds for the 3 model types. Depending on our intended use case for DataDefinitions, the typing may be irrelevant, so I thought of the 
`Either (QD E) (QD ME)` as a sort of "ModelKinds" for DataDefinitions specifically too. Ultimately, it was also straightforward to fix. However, it still relies on the Maps in the ChunkDBs to be ordered (I'm not so sure if that's a guarantee from the standard Data.Map however, we might need to use Data.Map.Ordered's OMaps [but I'm not 100% sure on this]).

For this branch, if we're okay with this style for encoding DataDefinitions, then I think we have at least the following tasks:
1. The type synonyms you mentioned above.
2. To go back and make sure that each of the Exprs I converted into ModelExprs was strictly required.
3. To figure out if we can allow both Expr-typed and ModelExpr-typed things inside of GDs & TMs -- essentially need to re-evaluate the ModelKinds typing.
4. To figure out if we want the phantom types for indicating computability -- it may not be required, but it's at least something we can think about for now or for later.

code/drasil-code/Language/Drasil/CodeSpec.hs
```diff
@@ -147,8 +147,8 @@ asVC' (FData (FuncData n _ _))     = vc n (nounPhraseSP n) (Variable n) Real
 -- | Determines the derived inputs, which can be immediately calculated from the 
 -- knowns (inputs and constants). If there are DDs, the derived inputs will 
 -- come from those. If there are none, then the 'QDefinition's are used instead.
-getDerivedInputs :: [DataDefinition] -> [Input] -> [Const] ->
-  ChunkDB -> [QDefinition]
+getDerivedInputs :: [DataDefinition Expr] -> [Input] -> [Const] ->
```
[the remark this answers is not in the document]
balacij: The synonyms are added now.

JacquesCarette:
I'm assuming you'll explicitly ping me when this is ready for another review?

balacij:
> I'm assuming you'll explicitly ping me when this is ready for another review?

Yes. Though, I might ping for feedback on a potential resolution to the main problem with this PR:

The allowed type variables for a ModelKind is essentially just Expr or ModelExpr, with the types being derived from whether an Expr or a ModelExpr was used to construct a QDef/MultiDefn/ConstraintSet/etc for it's related EqModel, EqRealm, EqConstraints, etc. The intention was to have a `ModelKind ModelExpr` be required for TheoryModels and GenDefns, and `ModelKind Expr`s to be required for InstanceModels. However, I see no reason (and you also saw the problem earlier) why an `Expr` couldn't be used inside a TheoryModel if the related expressions can be written using the language of Expr instead of ModelExpr. To resolve this, I'm not quite sure what a good resolution is yet. I can't quite tell if having the typing discussed in the [discussion above](https://github.com/JacquesCarette/Drasil/pull/2791#issuecomment-902299095) would resolve it well enough.

balacij:
@JacquesCarette What do you think of introducing polymorphic expressions via a typed tagless final encoding for the smart constructors of Expr & ModelExpr? I [created a prototype](https://github.com/JacquesCarette/Drasil/compare/moveDerivToModelExpr...lessMonomorphicExprs) of this idea. It would resolve [earlier discussion regarding why an Expr needed to be converted into a ModelExpr](https://github.com/JacquesCarette/Drasil/pull/2791/files/3732f43e58323e7c2a0dafe413190da1f0b58cf0#r690661435) through making it polymorphic, and only binding it to Expr or ModelExpr when needed. This would also lessen our usage of `express` everywhere (less "rewrites" of constructed Exprs to ModelExprs because it would use the correct instance depending on where it's bound).

However, this PR may still not be friendly to the idea of "refining TMs into IMs" yet because we still force TMs & GDs to use ModelExprs, while IMs mostly use Exprs. Rewriting expressions from ModelExprs into Exprs will still prove difficult if we want to have it be done via refinements on TMs/GDs directly -- it will continue to require reconstruction with changes.

I didn't push the prototype directly into this branch directly in case @cd155 is using this branch as it could lead to another substantial change in his code which may or may not make it to `master`.

cd155:
@balacij my current change in this branch is relatively small. I think I can manage my changes if there is a large change.

JacquesCarette:
I'm going to come back to this asap, I'm just drowning in millions of things to do.

balacij:
Thank you & take your time! :smile:

I think I will create a PR from the prototype "lessMonomorphicExprs" branch to this branch ("moveDerivToModelExpr") so that reviewing can be a bit easier too.

JacquesCarette (approved):
I'm still not sure if this design is what we want (and I think you do have an alternate one) but, at the level of the code, this is all good.

balacij:
Do you think this is worth merging as it is right now (w/ #2837 merged in to this PR first)?

I still have doubts about the design as well, primarily because of a few lingering questions from the above discussions. However, I think this design is at least reasonably easy to re-write/re-work if the need arises.

Perhaps we should continue to work in PRs based on this one, and we can see how far we go with this as a separate branch from `master`?

JacquesCarette:
For sure #2837 should be merged in here first.  I don't want long-lived branches, so we should figure out the rest asap. But I do need to think about it just a bit more.

JacquesCarette:
So, let's ask ourselves again: should we just merge this in?

balacij:
I will try to do a pass on the code again to make sure I'm not regressing anything (or at least not in ways that we know we can easily fix later when things are built/modified), and then post a summary.

As of right now, I think that all issues that could arise from these code changes would also arise when we have typed expressions (notably including DataDefinitions in particular -- they needed to be split up into Expr and ModelExpr somehow). One thing that I'm not so sure about is `defnExpr`'s typeclass (DefiningExpr) -- how it will pan out with typed expressions will be interesting, or it might force us to remove it entirely!

balacij:
Now I'm working on recreating a miniature version of each component of Drasil that deal with Exprs in some way. I just posted some code over here: https://github.com/balacij/ExprTests/blob/main/src/Lib.hs

The "original"/3rd variant that I posted is the most similar to the one currently on this branch. It appears to work well with typed expressions being exposed as a type variable. I think it would be good to spend some time recreating miniature versions of ModelKinds, IMs, TMs, GDs, DDs, QDs, etc _can look_ with typed expressions. This _might_ have the potential to bring us to the 'right' (or at least a better) design than that on this branch.

For now, I don't see any real regressions to the codebase from this branch, but:
1. The `Express` constraints set on the data constructors for `ModelKinds` might be fruitless.
2. The way we adjusted DataDefinitions to accommodate Exprs and ModelExprs currently doesn't affect stable (not yet adding the stars, perhaps this should be done in a new ticket&PR?).
3. We need to be careful with where we are exposing the Expr/ModelExpr-expected type variables or else they may come back to haunt us. Wherever we currently have a `_ :: [Thing Expr]` or a `_ :: [Thing2 ModelExpr]` would not be compatible when the `Expr`s (or `ModelExpr`s) do not all have the same resulting expression type. We might just be able to hide these in an extra GADT, but I'm not entirely certain of how this would (just needs more testing, but it might just be that we need a well-placed constraint). Amongst these 3 points, this point would be the most important one as it could be the deciding factor of whether this design is good or not. This is also what I'm currently working on investigating on the new test repo I posted above.

I didn't see anything that would be too difficult to later revert/modify if need be, so I tend towards merging it in.

Do you have any thoughts/concerns regardings this design @JacquesCarette?

JacquesCarette:
I agree about the `Express` constraint. But that can come later. 

And the part where we sometimes want `Expr` (specifically), sometimes `ModelExpr` and sometimes be explicitly polymorphic over that, is definitely tricky. I think we'll discover what we really want there 'later'.

So merge?

balacij:
Yes, I agree, we should merge.

> And the part where we sometimes want Expr (specifically), sometimes ModelExpr and sometimes be explicitly polymorphic over that, is definitely tricky. I think we'll discover what we really want there 'later'.

Yes, I agree. I think we will get a better idea of what/when we want them soon. I'm currently working on restructuring my ExprTests repository, but I think I will return to this shortly after.

Issue #2788, opened 2021-08-13
Get rid of `mkQuantDef'`


Ant13731:
This issue will probably take a good chunk of time to complete, but I think we will want to start getting rid of functions that can overwrite information within a given UID. I originally set out to try and fill in more fields of the database automatically (#2775), but the concepts, quantities, and qdefinitions and modelkinds fields have the potential to give conflicting information to the same UID, which could result in some weird behaviour.

`mkQuantDef'` is defined as follows:
```Haskell
-- | Used to help make 'QDefinition's when 'UID' and 'Symbol' come from the same source, with the term separate.
mkQuantDef' :: (Quantity c, MayHaveUnit c) => c -> NP -> Expr -> QDefinition
mkQuantDef' c t = mkQDefSt (c ^. uid) t EmptyS (symbol c) (c ^. typ) (getUnit c)
```
Which gives it the ability to overwrite a lower-levelled chunk's `NP` when I don't think it should. It's used most often to make general chunks more precise (ex. The pendulum examples use a general `acceleration` chunk and replace its general term with one that makes sense in the context of a pendulum). Modifying the data assigned to a UID without modifying the UID itself could then result in unexpected terms popping up. To follow up the pendulum example, a user can no longer use the generalized definition of acceleration because that information has been lost and replaced with the specific one.

I think this sort of problem is very similar to #1116 and #91. There are also some similar constructors defined in `drasil-theory` for `ModelKind` (see #2775)

balacij:
> Which gives it the ability to overwrite a lower-levelled chunk's NP when I don't think it should.

This is very relevant to #2388. It makes certain chunks unreachable (overwritten in the UID maps).

Instead of getting rid of `mkQuantDef'` and similar functions, would it suffice, in general, to, when inheriting a UID, transform the originating UID in some way? For example, with `mkQuantDef'`, we could just append "QD" to the end of the UID. This would likely require us to go in and edit the existing code to make sure that our references are pointing to the right things. The UIDs are never really shown to end-users, so I'm not quite sure if we care about them being pretty.

JacquesCarette:
We indeed to not care about UIDs being pretty. Right now they are short, but I always thought they might 'morph' into something much more structured than just a `String`, as the need arose.

[...]

PR #2803, opened 2021-08-19
Implementing UID as a newtype.

Ant13731:
Contrtibutes #2788 

GlassBR also had a change, LDF now shows up as an acronym. I am not sure why (I haven't manipulated any of the UIDs around yet), but it doesn't seem to look wrong.

### DIFF
--- a/code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
+++ b/code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
@@ -13,7 +13,7 @@ module Data.Drasil.ExternalLibraries.ODELibraries (
 import Language.Drasil (HasSymbol(symbol), HasUID(uid), MayHaveUnit(getUnit),
   QuantityDict, HasSpace(typ), Space(..), narg, implVar, implVar', qw,
   compoundPhrase, nounPhrase, nounPhraseSP, label, sub, NamedArgument,
-  Idea(getA), NamedIdea(term), Stage(..))
+  Idea(getA), NamedIdea(term), Stage(..), UID(..))
 import Language.Drasil.Display (Symbol(Label, Concat))
 
 import Language.Drasil.Code (Lang(..), ExternalLibrary, Step, Argument,
@@ -293,7 +293,7 @@ solveFromToStep = quantfunc $ implVar "SolveFromToStep_oslo" (nounPhrase
   solT (label "SolveFromToStep")
 
 vecDepVar :: ODEInfo -> CodeVarChunk
-vecDepVar info = quantvar $ implVar (dv ^. uid) (dv ^. term) vecT
+vecDepVar info = quantvar $ implVar (uidToStr $ dv ^. uid) (dv ^. term) vecT
   (sub (symbol dv Implementation) (label "vec"))
   where dv = depVar info
 
@@ -301,7 +301,7 @@ vecDepVar info = quantvar $ implVar (dv ^. uid) (dv ^. term) vecT
 -- | Oslo's Vector type behaves like an array, so needs to
 -- be represented as one or else will hit type errors in GOOL.
 arrayVecDepVar :: ODEInfo -> CodeVarChunk
-arrayVecDepVar info = quantvar $ implVar (dv ^. uid ++ "vec") (dv ^. term)
+arrayVecDepVar info = quantvar $ implVar (uidToStr (dv ^. uid) ++ "vec") (dv ^. term)
   (dv ^. typ) (sub (symbol dv Implementation) (label "vec"))
   where dv = listToArray $ depVar info
 
@@ -605,7 +605,7 @@ odeMethodUnavailable = "Chosen ODE solving method is not available" ++
 
 -- | Change in @X@ chunk constructor (where @X@ is a given argument).
 diffCodeChunk :: CodeVarChunk -> CodeVarChunk
-diffCodeChunk c = quantvar $ implVar' ("d" ++ c ^. uid)
+diffCodeChunk c = quantvar $ implVar' ("d" ++ uidToStr (c ^. uid))
   (compoundPhrase (nounPhraseSP "change in") (c ^. term)) (getA c) (c ^. typ)
   (Concat [label "d", symbol c Implementation]) (getUnit c)
 
@@ -616,7 +616,7 @@ diffCodeChunk c = quantvar $ implVar' ("d" ++ c ^. uid)
 modifiedODESyst :: String -> ODEInfo -> [CodeExpr]
 modifiedODESyst sufx info = map replaceDepVar (odeSyst info)
   where
-    replaceDepVar cc@(C c) | c == depVar info ^. uid = C (c ++ "_" ++ sufx)
+    replaceDepVar cc@(C c) | c == depVar info ^. uid = C $ UID (uidToStr c ++ "_" ++ sufx)
                            | otherwise               = cc
     replaceDepVar (AssocA a es)           = AssocA a (map replaceDepVar es)
     replaceDepVar (AssocB b es)           = AssocB b (map replaceDepVar es)
--- a/code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
+++ b/code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
@@ -13,7 +13,7 @@ module Data.Drasil.ExternalLibraries.ODELibraries (
 import Language.Drasil (HasSymbol(symbol), HasUID(uid), MayHaveUnit(getUnit),
   QuantityDict, HasSpace(typ), Space(..), narg, implVar, implVar', qw,
   compoundPhrase, nounPhrase, nounPhraseSP, label, sub, NamedArgument,
-  Idea(getA), NamedIdea(term), Stage(..), UID(..))
+  Idea(getA), NamedIdea(term), Stage(..), (+++.), showUID)
 import Language.Drasil.Display (Symbol(Label, Concat))
 
 import Language.Drasil.Code (Lang(..), ExternalLibrary, Step, Argument,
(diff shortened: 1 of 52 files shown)


code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -293,15 +293,15 @@ solveFromToStep = quantfunc $ implVar "SolveFromToStep_oslo" (nounPhrase
...
+vecDepVar info = quantvar $ implVar (showUID dv) (dv ^. term) vecT
   (sub (symbol dv Implementation) (label "vec"))
   where dv = depVar info
 
 -- Hack required because 
 -- | Oslo's Vector type behaves like an array, so needs to
 -- be represented as one or else will hit type errors in GOOL.
 arrayVecDepVar :: ODEInfo -> CodeVarChunk
-arrayVecDepVar info = quantvar $ implVar (dv ^. uid ++ "vec") (dv ^. term)
+arrayVecDepVar info = quantvar $ implVar (showUID dv ++ "vec") (dv ^. term)
```
JacquesCarette, line 304: actually, should use `+++` here, then `showUID`. This will future-proof the code.

code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -605,7 +605,7 @@ odeMethodUnavailable = "Chosen ODE solving method is not available" ++
 
 -- | Change in @X@ chunk constructor (where @X@ is a given argument).
 diffCodeChunk :: CodeVarChunk -> CodeVarChunk
-diffCodeChunk c = quantvar $ implVar' ("d" ++ c ^. uid)
+diffCodeChunk c = quantvar $ implVar' ("d" ++ showUID c )
```
JacquesCarette, line 608: Hmm, I don't like this inconsistency. I think we should uniformly append. Hopefully fixing that now won't break things...

code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -616,7 +616,7 @@ diffCodeChunk c = quantvar $ implVar' ("d" ++ c ^. uid)
 modifiedODESyst :: String -> ODEInfo -> [CodeExpr]
 modifiedODESyst sufx info = map replaceDepVar (odeSyst info)
   where
-    replaceDepVar cc@(C c) | c == depVar info ^. uid = C (c ++ "_" ++ sufx)
+    replaceDepVar cc@(C c) | c == depVar info ^. uid = C $ c +++. ("_" ++ sufx)
```
JacquesCarette, line 619: `+++` here as well.

code/drasil-example/ssp/lib/Drasil/SSP/TMods.hs
```diff
@@ -82,7 +81,7 @@ mcShrStrgth = tm (equationalModelU "mcShrSrgth" mcShrStrgthQD)
 
 ------------------------------------
 mcShrStrgthQD :: QDefinition
-mcShrStrgthQD = fromEqnSt' (shrStress ^. uid) (nounPhraseSP "Mohr-Coulumb shear strength")
+mcShrStrgthQD = fromEqnSt' (showUID shrStress) (nounPhraseSP "Mohr-Coulumb shear strength")
```
JacquesCarette, line 84: Ah, no, we should never have `showUID` in examples. So this indicates a flaw somewhere. Maybe `fromEqnSt'` should take a `UID` ?

code/drasil-lang/Language/Drasil.hs
```diff
@@ -93,9 +93,10 @@ module Language.Drasil (
   -- Similar types are grouped together.
   
   -- *** Basic types
-  , UID
+  , UID, showUID
```
JacquesCarette, line 96: `showUID` should only be exported by `Development`, not part of the language.

code/drasil-lang/Language/Drasil/Chunk/Eq.hs
```diff
@@ -94,31 +94,33 @@ fromEqnSt' nm desc def symb sp =
...
-mkQDefSt :: UID -> NP -> Sentence -> (Stage -> Symbol) -> Space ->
+mkQDefSt :: String -> NP -> Sentence -> (Stage -> Symbol) -> Space ->
   Maybe UnitDefn -> Expr -> QDefinition
 mkQDefSt u n s symb sp (Just ud) e = fromEqnSt u n s symb sp ud e
 mkQDefSt u n s symb sp Nothing   e = fromEqnSt' u n s symb sp e
 
 -- | Used to help make 'QDefinition's when 'UID', term, and 'Symbol' come from the same source.
 mkQuantDef :: (Quantity c, MayHaveUnit c) => c -> Expr -> QDefinition
-mkQuantDef c = mkQDefSt (c ^. uid) (c ^. term) EmptyS (symbol c) (c ^. typ) (getUnit c)
+mkQuantDef c = mkQDefSt (uidToStr $ c ^. uid) (c ^. term) EmptyS (symbol c) (c ^. typ) (getUnit c)
```
JacquesCarette, line 104: This back-and-forth says that there should be a `mkQDefSt'` that takes a `UID` directly. This might propagate to needing other things below to support that.

code/drasil-lang/Language/Drasil/Chunk/Eq.hs
```diff
@@ -94,31 +94,33 @@ fromEqnSt' nm desc def symb sp =
...
+mkQuantDef' c t = mkQDefSt (uidToStr $ c ^. uid) t EmptyS (symbol c) (c ^. typ) (getUnit c)
 
 -- HACK - makes the definition EmptyS !!! FIXME
 -- | Smart constructor for QDefinitions. Requires a quantity and its defining 
 -- equation. 
 ec :: (Quantity c, MayHaveUnit c) => c -> Expr -> QDefinition
 ec c = EC (dqd' (cc' (nw c) EmptyS) (symbol c) (c ^. typ) (getUnit c)) []
 
--- | Factored version of 'QDefinition' functions
+-- FIXME: uidToStr shouldn't be used here.
```
JacquesCarette, line 117: Correct, do fix this...

code/drasil-printers/Language/Drasil/Printing/Import/Document.hs
```diff
@@ -101,7 +101,7 @@ layUnlabelled  _ (Bib bib)              = T.Bib $ map layCite bib
 
 -- | For importing a bibliography.
 layCite :: Citation -> P.Citation
-layCite c = P.Cite (c ^. citeID) (c ^. citeKind) (map layField (c ^. getFields))
+layCite c = P.Cite (showUID c) (c ^. citeKind) (map layField (c ^. getFields))
```
JacquesCarette, line 104: This feels wrong. Either because the `UID` is used for something real, or because it's assumed to be a `String`. Both are wrong.

code/drasil-utils/Utils/Drasil/Concepts.hs
```diff
@@ -348,19 +348,20 @@ a_Gen f t = nounPhrase'' (S "a" +:+ f t) (S "a" +:+ f t) CapFirst CapWords
 -- See 'compoundPhrase' for more on plural behaviour.
 -- /Does not preserve abbreviations/.
 compoundNC :: (NamedIdea a, NamedIdea b) => a -> b -> NamedChunk
-compoundNC t1 t2 = nc
-  (t1 ^. uid ++ t2^.uid) (compoundPhrase (t1 ^. term) (t2 ^. term))
+compoundNC t1 t2 = ncUID
+  ((t1 ^. uid) +++. showUID t2) (compoundPhrase (t1 ^. term) (t2 ^. term))
```
JacquesCarette, line 352: Ah, this shows the need for an extra combinator to combine two UIDs. This should then be used here.

JacquesCarette (changes requested):

code/drasil-lang/Language/Drasil/Chunk/Eq.hs
Ant13731: Ahh okay, I think we can actually just make `mkWDefSt` take a `UID` since its only used with UIDs anyway

JacquesCarette: Prefect.

code/drasil-printers/Language/Drasil/Printing/Import/Document.hs
Ant13731: I found that `Cite` from `Language.Drasil.Printing.Citation` is made using an `EntryID` rather than a `UID` which is just a direct synonym for `String`. Should it take a `UID` instead?

code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
Ant13731: Would we want a variant of `implVar` and `implVar'` that take in a `UID` rather than a `String`?

code/drasil-printers/Language/Drasil/Printing/Import/Document.hs
JacquesCarette: So I think the idea is that an `EntryID` might actually be different. In fact, I think it's a hack that we're using the UID for the EntryID. The EntryID will be user-visible, at least inside the sources. UIDs, other than in debugging (or the dependency graph) really should be invisible.

But let's not over-complicate: keep the hack, make an issue, it can be fixed after.

code/drasil-code/Data/Drasil/ExternalLibraries/ODELibraries.hs
JacquesCarette: We should. Then we can see if `implVar` is used at all in the remainders.

Ant13731: I originally didn't want to use `+++` for this since it doesn't really show a direct dependency from one chunk to the next, but it should be fine

Ant13731:
Once this PR is merged, the [`newtypeUID`](https://github.com/JacquesCarette/Drasil/tree/newtypeUID) branch can be deleted as well.

JacquesCarette (approved):

PR #2804, opened 2021-08-19
`performanceSpd` fix

### LINKED ISSUE #2800: Change UID and Term for `performanceSpd`
The definition for `performanceSpd` uses the same UID as `performance`, and doesn't properly use the `nounPhrase''` constructor
```Haskell
performance       = dcc "performance"       (nounPhraseSP "performance")
  "the action or process of carrying out or accomplishing an action, task, or function"

performanceSpd    = dcc (uidToStr $ performance ^. uid) (nounPhrase'' (phrase performance) (S "speed") CapFirst CapWords)
  "the action or process of carrying out or accomplishing an action, task, or function quickly"
```

Ant13731:
Closes #2800 
Builds on #2803


JacquesCarette:
Why does it build on #2803? This feels independent?

Ant13731:
I had to change the UID for it in #2803 to make the build work (since `nc` needs a String instead of a UID)

JacquesCarette (approved):

PR #2821, opened 2021-08-27
Remove Paths_drasil_website from build

balacij:


balacij:
When merged, the Build CI might fail due to HLint (#2823) updates. We should merge #2824 first before this one.

Issue #2789, opened 2021-08-13
Investigate what takes so long in full builds


JacquesCarette:
I was looking at https://github.com/JacquesCarette/Drasil/pull/2760/checks?check_run_id=3323104120 "live", and some things really take a lot of time & I don't know why they do.  For example, installing dependencies takes 5 minutes; the question would be, why are we asking for so many things that are not up-to-date in standard images?

Another is that it takes longer to build the docs than the code build itself, which seems quite odd. That warrants investigation too.

balacij:
Sounds good. I think I can contribute some things that I already know about these steps.

This will primarily target the "Build" workflow.

| Step | Approx. time spent | Discussion |
|------|--------------------|------------|
| Install system requirements | ~2min | Downloads all system dependencies (compilers for artifacts, tex, fonts, etc) |
| Install Stack | <1s | - | 
| Cache dependencies | ~2min on cache hit, 1s else  | The main issue with this is how often it *doesn't* get a cache hit. Needs investigation into why/why not. |
| Clean previous run | ~1min | Cleans the previous run, but Cabal also updates it's cache here -- this may need investigation into why it's not being cached too. |
| Update PATH | <1s | Adds Stack exe folder to $PATH | - |
| HLint | ~50s | Downloads HLint (fast), and runs it (slow). I don't think there's much we can do here unless we want to just remove it from the "Build" workflow, and just leave it in the "Test" one. |
| Install dependencies | ~5min | Will need investigation into which ones we need/which ones we don't. |
| Build | ~3min | We will need to analyze our construction plan, we _might_ be able to reduce it a bit by exporting more files in each package and via adding extra threads (though, GH Actions runner only have 2c/2t). |
| Stable | ~1min | We can merge this with the above, but it also does extra testing in "Build". Perhaps we can just remove it from "Build" in favour of leaving just in "Test"?  |
| TeX | ~3min | We might be able to do them concurrently via scheduling them as background jobs. This might be a bit difficult to figure out which artifacts broke however. |
| Code | ~1min | Compiling code, might be able to do the same as "TeX". |
| Doxygen | ~10s | - |
| Docs | ~5min | Takes a long time because it has to re-compile the originally compiled code with haddock docs (uses the originally compiled, so it's fast, most time is spend in Haddock generation), and then has to recompile (again, but from clean slate this time -- takes a long time) to generate the smaller version without all-exposed modules |
| Graphs | ~20s | - |
[...]

PR #2833, opened 2021-09-08
Allow less `docs` builds when unnecessary in CI

### LINKED ISSUE #2830: In test workflow, we only need to build fully exposed haddock docs, we can ignore the "normal" variant
Relates to #2789 

The Test workflow can safely ignore building the normal variant of the Haddock docs because with the fully exposed version, we will already have the "checks" done to make sure that the Haddock comments are well-formatted.

This should decrease total time spent in builds by (the time to compile drasil) + (the time to build Haddock docs).

balacij:
Closes #2830 
Builds on #2831 
Contributes to #2789 (arguably)

When building the Haddock docs for the PRs, we are doing so to test their formatting to ensure that we can generate public Haddock docs. Hence, in the CI runs for PRs, we only need to generate the "fully exposed" variants because that will test (& by extension, ensure) the ability to generate the lesser/normal variant.

This adds a `FULL` environment variable to the Makefile script, designating whether to _full_y build the Haddock docs or not. Passing a "1" represents "true", anything else is treated as "false".

This should shave off a few minutes off the build time for the PRs.


.github/workflows/Build.yaml
```diff
@@ -54,29 +61,35 @@ jobs:
         run: make stackArgs="--no-terminal" deps
       - name: "Build"
         run: make code stackArgs="--no-terminal" GHCFLAGS="-Werror"
-      - name: "Stable"
+      - name: "Test built artifacts against stable"
```
JacquesCarette, line 64: really like the improved names.

JacquesCarette (approved):

PR #2837, opened 2021-09-13
Polymorphic expressions for PR #2791

balacij:


code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/DataDefs.hs
```diff
@@ -34,7 +34,7 @@ positionXDD_1 = ddENoRefs positionXQD_1 Nothing "positionXDD1" [positionXRef_1,
 positionXQD_1 :: SimpleQDef
 positionXQD_1 = mkQuantDef xPos_1 positionXEqn_1
 
-positionXEqn_1 :: Expr
+positionXEqn_1 :: ExprC r => r
```
JacquesCarette, line 37: I must admit that I'm not at all sure about having this much polymorphism in the examples themselves.

In some sense, it's ok, as we should be moving a lot of this to an external DSL, and that will hide all of this. Not sure how soon that'll happen, so in the meantime, the examples will look more complex.

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/GenDefs.hs
```diff
@@ -284,7 +284,7 @@ yForceMD_2 = mkMultiDefnForQuant quant EmptyS defns
                     mkDefiningExpr "yForceWithMass2"
                       [] EmptyS $ express $ forceGQD ^. defnExpr,
                     mkDefiningExpr "yForceWithAngle2"
-                      [] EmptyS $ express E.yForceWithAngle_2]
```
JacquesCarette, line 287: having said that, being able to remove many many `express` calls is very nice!

code/drasil-example/projectile/lib/Drasil/Projectile/IMods.hs
```diff
@@ -97,7 +97,7 @@ landPosDerivSent2 = foldlSentCol [S "To find the", phrase landPos `sC`
   S "we want to find the", ch xPos, phrase value, sParen (ch landPos),
   S "at", phrase flightDur, fromSource timeIM]
 landPosDerivSent3 = foldlSentCol [S "From", refS speedIX,
-  sParen (S "with" +:+ E (defines iSpeed launSpeed)), S "we can replace", ch ixVel]
+  sParen (S "with" +:+ E (defines (sy iSpeed) (sy launSpeed))), S "we can replace", ch ixVel]
```
JacquesCarette, line 100: why does `defines` now take symbols? this seems to be a (very small) regression?

JacquesCarette (approved):

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/DataDefs.hs
balacij: Would we be able to create a type synonym that hides the polymorphism a bit? If so, would that change anything with this issue?

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/GenDefs.hs
balacij: Certainly! I think that a few well-placed `ExprC`s to replace a few `Express` constraints can also be useful in further removing "earlier" `express` calls.

code/drasil-example/projectile/lib/Drasil/Projectile/IMods.hs
balacij: `defines` previously had an `Express` constraint for its first parameter's inputs. With iSpeed, it was a quantity and was `express`ed as a `sy iSpeed` in the end (the Express instance is just `express = sy` for all quantity-like chunks).  This is bringing up an issue with `defines`. I would've fixed it immediately because `defines` should always be in the form of `x $= ...expr...` but we have one area (QDefinitions) where we have symbols that represent functions, for which we want to have the LHS of the equation be a `x(a,b,c)` instead. I'll take a second look at the code to see if we can get this fixed in this branch quickly before a merge.

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/DataDefs.hs
JacquesCarette: Not easily? Part of the point is that this is polymorphic, so that we do need/want a variable showing up.

Maybe a GADT, like
```haskell
data PExpr r where
  E :: ExprC r => r -> PExpr r
```
would let you replace `ExprC r => r` with `PExpr r` (which could be renamed `Expr r` if we want).  But we'd then have to have an easy way to add that `E` constructor.  Which is a lot like `express`!

Kind of feels like moving the deck chairs on the Titanic... but less bad.

code/drasil-example/projectile/lib/Drasil/Projectile/IMods.hs
JacquesCarette: For this particular instance of `defines`, what is going wrong is that we shouldn't be using this linguistic form inline at all. The relationship between `iSpeed` and `launSpeed` should be somehow captured in another piece of that, and the display form associated to that relationship should be what's used here.

So the first thing to do would be to fix those "bad" uses of `defines`. Then see those that are left.

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/DataDefs.hs
balacij: I see what you mean. What about with `RankNTypes`?
```
type PExpr = forall r . ExprC r => r
```

JacquesCarette: That might work!  I'd forgotten about that.

balacij: I just updated this, it looks like it worked. What do you think?

code/drasil-example/projectile/lib/Drasil/Projectile/IMods.hs
balacij: In this case, it looks like we're trying to substitute a variable with another in deriving a relation, `launSpeed` (launch speed) is an "instance" (of sorts -- I'm not quite sure what to call this relationship) of `iSpeed` (initial speed). Would this be a task for when we have refinements of models?

Related area in SRS (3rd sentence): https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#IM:calOfLandingDist

code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/DataDefs.hs
JacquesCarette: I think we've converged. Excellent.

JacquesCarette:
ok, merged. Should I delete this side branch?

balacij:
Yep, that's fine. Thank you very much!

code/drasil-example/projectile/lib/Drasil/Projectile/IMods.hs
JacquesCarette: what I'm wondering about is the source of that equality between `iSpeed` and `launSpeed`. It feels ad hoc to have it inline here. It feels like it ought to exist elsewhere, in dependently.

balacij: I agree, I believe it was just a temporary measure. I was thinking that this would be a part of the refinement's system you mentioned a few meetings ago. I think this will be a rather large design needed before we can do something about it.

I'll make a ticket so that we can discuss it more.

balacij: Moving discussion to #2852 so that we can more easily discuss and track progression of this.

PR #2838, opened 2021-09-15
Add more logging to `analysis` Makefile target

balacij:
Related to #2789 

It looks like much of the time in the `analysis` step is spent converting dot graphs into svgs and circo svgs.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -313,7 +313,7 @@ analysis: graphmod ##@Analysis Generate a table and some graphs to analyze Drasi
 		circo -Tsvg $$pack.dot > "./circo_$$pack.svg" ; \
 	done
 	@cd $(CLASSINST_GRAPH_FOLDER) && for pack in $(PACKAGES_E); do \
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

JacquesCarette:
So it would be good to know the relative time between `dot` and `circo`.

balacij:
As part of the regular CI or as an option in the Makefile? Or did you mean in general?

JacquesCarette:
I meant in general. Maybe one of `dot` and `circo` is just really slow, but the other is fast.

balacij:
Ah, ok. The total time spent on conversion of the `dot`s is less than 4s, while the `circo` ones together take > 2min. `circo` hangs on the typegraph generation for entry "drasil" (>1min spent here) and for "gool" (~1min spent here).

JacquesCarette:
I think we should drop `circo` entirely. I just looked at the results, and they were really not useful.

PR #2839, opened 2021-09-15
Rearrange code directories to conform to Stack/hpack expectations

### LINKED ISSUE #2835: Stack/hpack prefer projects to have source code split up into subdirectories
Stack + hpack prefers for source code directories of exes & libs to be split up. Hpack relies on this to generate cabal files where no hs files get compiled more than once. Right now, we have no problem with this.

Currently, we have the following structure:
```
code/
    drasil-example/*/
        package.yaml
        stack.yaml
        lib/
            <library code>
        app/
            <code related to creating executables>
    drasil-*/
        package.yaml
        stack.yaml
        ...
        Language/Drasil/... <- hs files
```

If we want to create more tests/executables, it may be helpful to further split up our `drasil-*` packages into the same form as the projects in `drasil-example/*/`. If we choose not to, then the package.yaml files have to manually specify to which executable each file belongs to. If we split them up by directory, then it will be able to tell that all files from within that directory belong solely to that exe/test/lib. As an example, our `drasil-code`'s [package.yaml](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-code/package.yaml) file contains a lot of explicitly named modules otherwise Cabal gives us errors noting "missing-home-modules". Taking [glassbr's package.yaml file](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-example/glassbr/package.yaml) as an example of the alternative structure, it does not need to list any files other than an entry point/Main module for executables/tests.

The main problem with reorganizing the `drasil-*` libs into the same structure as the examples is that we lose out on GitHub's commit log. From what it appears, GitHub does not track file movements in commit histories (e.g., it does not appear to use `git log --follow <file>` to generate the logs). I believe this exact issue was previously exhibited when Awurama was looking for the git log for some file via GitHub's git commit history web viewer.

Would this low visibility git commit history be okay?

balacij:
Closes #2835


code/drasil-gen/stack.yaml
```diff
@@ -34,7 +34,9 @@ packages:
 - ../drasil-build
 - ../drasil-code
 - ../drasil-code-base
+- ../drasil-data
```
JacquesCarette, line 37: why did drasil-gen gain these dependencies?

JacquesCarette (changes requested):
How much of a merge conflict will it generate if I merge this in now, ahead of the other PRs?

code/drasil-gen/stack.yaml
balacij: It was a bug that it wasn't there (if you were in `drasil-gen` and had run `stack build`, it would've failed due to a missing dep). One of the packages it relies on depends on drasil-data somewhere.

balacij:
I can pull it into `moveDerivToModelExpr` branch first (#2791). Since I used `git mv` instead of `mv` alone, it _should_ be fairly smooth, except for a few files that I completely removed in #2791.

JacquesCarette (approved):

JacquesCarette:
Either way is fine with me, just tell me which way you would prefer to proceed.

balacij:
This one first is probably best merged first. This way we can still leave #2791 alone in case it needs more work.

PR #2840, opened 2021-09-15
Only build dot & circo graphs for deployments, when required

balacij:
Contributes to #2789 

Removes dot and circo SVG conversions from tests, leaving them only performed for full deployments.


balacij:
This one doesn't appear to have any merge conflicts with  #2839 at the moment, so we can merge them in any order.

JacquesCarette (approved):

PR #2841, opened 2021-09-16
Script to find duplicate strings, frequency, location

peter-michalski:


code/scripts/duplicate_string_find.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash  
+
+#grep -rhEo (recursive, no filename, extended regular expressions, prints only matching parts
+#excludes stable, only includes .hs files
+#cuts outer double quotations
+#first awk searches for instances that appear 3 or more times
+#second awk removes instances that have double quotations anywhere by the ends
+#last two seds add back the double quotations
```
JacquesCarette, line 8: Note that, in theory a string like
```haskell
"some stuff and then \"an embedded quote\" is in there"
```
is actually different, because those escapes mean that it is a whole quote. Now, I'm pretty sure we don't have these (on purpose) except for somewhere in the printer to be able to output them. That would be worth checking.

JacquesCarette (approved):
This is beautifully documented!

A future improvement: the results of the script should be easy to process with other scripts (like sorting by number of occurrences, etc). Right now, it's optimized for human viewers, which is nice, but not necessarily best.

PR #2847, opened 2021-09-20
Avoid installing binaries on host computer on default `make`, create `install` and `X_install` targets to replace

### LINKED ISSUE #2845: Should Makefile be installing binaries on host computer?
The Makefile's basic `make` installs the project example executables in the host computer's Stack binaries path. Later, when using the software, we reference the executable using `stack exec` (the binary from the project-local `.stack-work` folder). As a result, it seems we don't use the binaries that we copied over into the Stack binaries path at all (e.g., on Linux machines, it appears in `~/.local/bin`), so I think we should remove them if not needed. If a user wants to install an example on their system, I think we should provide them with a `make <example>_install` (individual examples) and `make install` (all examples) commands to install the executables on their machines.

Do we require the executables to be installed on the host machine for any purposes? Regarding tests, I think we can get away with avoiding copying over the binaries.

Additionally, in our basic `make`, we use `stack install X` for each drasil package. I think this can be swapped with just a `stack build` to build the entire suite, and then following up with the diff tests. This should also allow for more of the packages to be built at once (might even help with build times very slightly).

balacij:
Closes #2845

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -76,6 +76,7 @@ DCP_E_SUFFIX = _deploy_code_path
 STABILIZE_E_SUFFIX = _stabilize
 BC_E_SUFFIX = _build_clean
 TRACE_GRAPH_SUFFIX = _trace_graph
+INSTALL_E_SUFFIX = _install
 
 # add example names with the appropriate suffixes to create per-example targets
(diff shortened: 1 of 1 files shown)


balacij:
Please ignore me converting back and forth between draft and "ready for review" status. This PR is ready for review.

JacquesCarette (approved):

Issue #2815, opened 2021-08-23
Add parenthesis to symbols in Expr


tingyuw:
In the notebook, if we would like to print something like this
![image](https://user-images.githubusercontent.com/47548045/130367628-351bc808-d357-43ba-b057-614dfb132905.png)
we have to add parenthesis to the unit because it has two ^ consecutively, so instead of ```v_y^i^2```, it should be ```(v_y^i)^2 ```.  

It's actually a printing thing, but if we add the parenthesis when printing, they will be added to all symbols, which I don't think is necessary for cases like ```y^2```. And we can't just hard code the ( ) like this ```square ((sy yVel))``` in the expression. 

Therefore, I think maybe we could add the parenthesis when we define the expression. I was thinking to add the constructor/operator in Expr.
I don't think we have this now since this kind of case doesn't appear in the existing example, but I might be wrong.

balacij:
That's an interesting issue. It looks like the "i" is a part of the symbol of  `yVel`. If this is the case, when printing, would it be possible to pattern match on the base terms to see if they're a symbol with a superscript? If they are, then we could blanket apply this as a fix for everything without having to manually designate the parentheses in the `Expr` language (e.g., instead of always having to manually write `square $ paren $ sy yVel` [assuming `paren` would be the smart constructor for the potential new term Parentheses term in Expr], we would automatically be able to determine if/when to place the parentheses when printing, based on `sy yVel`).

It also seems like this is this related to us not yet having proper list types yet in Drasil, which is why the "i" appears in the symbol (though I'm not 100% sure on this instance).

tingyuw:
Yes, the "i" is part of the symbol. I was thinking about the pattern matching, but we actually don't know whether the superscript is part of a symbol or an operation. We also don't know whether there is another superscript right after this one. We print the expression recursively and we only analyze what to print now.

balacij:
Are we unable to check for both cases? If there's a superscript after the 2, I imagine it should be rolled into the 2 assuming it was also a power operation. Are there other cases where a superscript appears?

tingyuw:
I think it is able to check if we define them differently since they both fall into the "Sup" category now. 
![image](https://user-images.githubusercontent.com/47548045/130501251-5cf8a5d7-7f50-4ce8-820d-4d5796f6b39b.png)

[...]

PR #2867, opened 2021-10-22
Fix notebook expressions display issues

tingyuw:
Contributes to #2815 
This is a temporary fix for the issue by replacing the vector components


JacquesCarette (approved):
Looks good. But just to be sure: this is ready to be merged in? Perhaps there should be an issue files to undo the temporary bits of this PR later?

tingyuw:
Even if we fix the superscript issue in the near future, I don't think we need to undo those changes.

PR #2870, opened 2021-10-28
Move Literals out of Expr/ModelExpr

### LINKED ISSUE #2783: Move literals out of Expr

balacij:
Closes #2783 

+ Create a basic Literal GADT
+ Create a TTF encoding of the same literals (so that it can be intermixed with Expr/ModelExpr)
+ Restrict ``constants'' to be defined by Literals, instead of Exprs.


balacij:
I think this is a good point to cut off work on this PR. I think there are other chunks that can be restricted to use Literals as well, but this might be getting too large as it is.

JacquesCarette (approved):

Issue #2842, opened 2021-09-16
Eliminate Duplicate Strings "# The default value is: NO." and "# The default value is: YES."


peter-michalski:
Details:

"# The default value is: NO."
Number of instances of string: 99
Number of files containing the string: 1
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs:"# The default value is: NO."

"# The default value is: YES."
Number of instances of string: 52
Number of files containing the string: 1
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Doxygen/Import.hs:"# The default value is: YES."

peter-michalski:
Proposing similar solution as in #2844. All such strings will be located in the same /code file.

JacquesCarette:
Again, too simple, but here for a different reason: you did not analyze the reason why this string is duplicated so many times. We need to find the reason for the duplication, so that the solution matches it.

All duplication elimination needs to come with a deeper analysis + explainable solution. The point is not to merely eliminate duplication, but to understand the deeper reason for the symptom.

peter-michalski:
Looks like each of the projects has doxConfig.txt files generated for each language (java, Python..). These files include "The default value is: " string for various parameters. 

comment in the file:
-- | A 'Doc' that holds optimized choices for configuring doxygen files.

Seems to me that it would be best to leave these strings as they are. They indicate optimized choices. They are simply being printed in a documentation file. The code is readable and understandable as is.

Is removing this duplication for the sake of removing duplication a good idea? My current thought is no, but please comment if you disagree.

smiths:
@JacquesCarette is the best person to comment on this, but when I look at [Import.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs) I can't help but think that it would be better if the doxygen config file wasn't hard-coded into the Haskell code.  A doxygen config file is relatively easy to generate using doxygen itself.  When Brooks did this he would have just used doxygen to create the config file and then converted it line by line to Haskell code.  There are a few spots (maybe only one?) where it is customized (the project name), but for the most part it is just standard boilerplate doxygen config info. 

[...]

PR #2874, opened 2021-11-01
Doxygen import: extracted duplicate strings

peter-michalski:
This pull request is for issue #2842. 

Tags for the duplicate strings are TagDefaultNO and TagDefaultYES.


code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs
```diff
@@ -23,6 +23,11 @@ yes, no :: Doc
 yes = text "YES"
 no = text "NO"
 
+-- | Renderings of comments regarding the default tag values for various Doxygen settings.
+tagDefNO, tagDefYES :: String
+tagDefNO = "# The default value is: NO."
+tagDefYES = "# The default value is: YES."
```
JacquesCarette, line 29: I was definitely thinking of splitting this into 3: "# The default value is: %s.", "NO" and "YES", where I'm using C-style printf formatting here (but this should be coded differently).  I would then create things like
```haskell
defNo, defYes :: Doc
defNo = defaultVal no
defYes = defaultVal yes
```
with
```haskell
defaultVal :: String -> Doc
```

code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs
```diff
@@ -107,15 +112,15 @@ makeDoxConfig prog s opt v =
   text "# option can be useful when feeding doxygen a huge amount of source files, where",
   text "# putting all generated files in the same directory would otherwise causes",
   text "# performance problems for the file system.",
-  text "# The default value is: NO.",
+  text $tagDefNO,
```
JacquesCarette, line 115: then this would be just `defNo`.

JacquesCarette (changes requested):
But the bigger meta-comment is that I was expecting a new `QDefinition` or something like that to be created to encapsulate the meaning of `"# The default value is:"`.

We need to teach *Drasil*, not just *Haskell*, about Doxygen concepts.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs
peter-michalski: Considering that yes and no and already declared as Doc in the file, is the following acceptable?
```
defNo, defYes :: Doc
defNo = defaultVal no
defYes = defaultVal yes
 
defaultVal :: Doc -> Doc
defaultVal val = text "# The default value is: " <+> val <+> text "."
```

Regarding QDefinition, I understand that this defines an expression with inputs, a definition, and a domain. I am currently reviewing its implementation. I have found some implementations in the code, including those below. Quite a few other QDefinitions are commented out. Can you please point me to an implementation that is similar to the problem in hand?

```
auxiliaryConstants :: [QDefinition]
auxiliaryConstants = assumptionConstants ++ gBRSpecParamVals

func_reqs_req6_pulledList :: [QDefinition]
func_reqs_req6_pulledList = [nonFL, glaTyFac, dimLL, tolPre, tolStrDisFac, strDisFac, hFromt]
```

JacquesCarette: Not that I'm pretty sure that `text "."` already has a combinator for it.  But otherwise, yes, the first block of code is good.

I'm probably asking for too much wrt `QDefinition` at this point. We'd need a proper generator and back-end printer to make this all work. So let's put that part on the back-burner for now.

peter-michalski: @JacquesCarette Helpers.hs. is very helpful indeed. I tried to import the relevant helper and got the error "Could not load module ‘Language.Drasil.Printing.Helpers’ it is a hidden module in the package ‘drasil-printers-0.1.10.0’" when compiling. 
The helpers weren't already imported somewhere else since ''Variable not in scope" (dot) if nothing is imported. Could you comment on this, why/how are these helpful helper functions hidden, and how are they supposed to be imported? 

I had the same issue with code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer.hs, and code/drasil-printers/lib/Language/Drasil/Plain/Print.hs, which also contain combinators for `text "."`

balacij: All packages in Drasil only export a limited number of their inner modules. Usually, they will have a base file that just re-exports a lot of it's functionality. To, generally, get a full list of the exported modules of a package, check out it's `package.yaml` file. The files listed under "exposed-modules" under the "library" section would be the ones that you can "import" from external modules. With both `drasil-printers` and `drasil-gool`, they have a limited number of, manually written, exported modules that you can use from external modules. It's only from that exposed modules that would be able to import from. If no "exposed-modules" list is listed in the `package.yaml` file, then it means that all module files in the `lib/` folder are exported, by default.

Since you found that combinator occurs in multiple areas, it might make sense for you to delete all of them in favour of making a single one in `drasil-utils`'s `Utils.Drasil.Document` module. Afterwards, you can import & re-export that new combinator you would have made via the main (and sole) `Utils.Drasil` module file that's exported by `drasil-utils`. Finally, you should be able to `import Utils.Drasil (yourCombinator)` from this file.

JacquesCarette: Thanks @balacij , perfect answer. And the right process: when we notice duplication in Drasil, we fix it (unless there are strong reasons not to).

peter-michalski: Thanks for the explanation, @balacij. Very informative. 

Regarding the deletion of duplicate combinators, upon deeper review,  there are similar combinators in `Helpers.hs`, `LanguageRenderer.hs`, and `Print.hs`, but there are slight differences between these, for good reason I'm sure, and many other combinators in these files are not duplicated. I am not sure if combining everything into one file is the best action at this time, maybe @JacquesCarette can comment. It might be worth looking into if a cleanup is needed, but I cannot say be certain of something like this at this time.

The appropriate combinator for what I need to do is `dot = text "."` from `LanguageRenderer.hs` in `drasil-gool`. The yaml package only exposes `GOOL.Drasil`, so I think I will simply expose `GOOL.Drasil.LanguageRenderer` as well.

peter-michalski (commented):
Split into YES and NO defaultVal cases, and imported dot

code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs
JacquesCarette: An even better combinator would be a "terminate sentence" (which does exist for `Sentence`) one. Finding a name is non-trivial.

As discussion on #2887 brought up too, we need to refactor the printing, not just expose stuff. `text" ."` has nothing to do with GOOL. It happens to be defined there (too) because GOOL used to be stand-alone.

peter-michalski: I do not see a terminate sentence combinator in `Utils/Drasil/Sentence.hs`, did you want me to add one there?

JacquesCarette: There's both `+:+.` and `!.`. But I think those work on `Sentence`, not `Doc`. You can indeed re-use those names, as code shouldn't be working on `Doc` and `Sentence` at the same time.

peter-michalski: @JacquesCarette okay, I will re-use the names for Doc. 

Obviously I wouldn't be defining this in Sentence.hs. Where should I add such a terminate sentence combinator for Doc?

JacquesCarette: @balacij is currently refactoring related matters - I think he's in the best position to make suggestions about that.

balacij: `drasil-utils`'s `Utils.Drasil.Document` looks like a good location to me. You might also need to re-import your combinator in `Utils.Drasil` and expose it in `Utils.Drasil` to get it working because `Utils.Drasil.Document` is a hidden module.

Depending on #2887, we might move the `Doc`-related combinators into a new package on their own, but that won't affect this much.

peter-michalski (commented):
updated based on request

PR #2880, opened 2021-11-03
Nier2021

smiths:
GitHub says that the NIER2021 branch can be automatically merged.  If that is the case, I think we should merge it, since it is easier to refer to it on the master branch.

It looks like the merge will include the white paper.  I don't think that is really a problem though, even though the white paper is not done.

@JacquesCarette, if you don't like this process, we could merge the NIER2021 branch into the white paper branch to at least prune one branch from our tree.


JacquesCarette (approved):
Yeah, it would have been nice to keep them separate, but that's more work than it's worth.

PR #2891, opened 2021-11-13
Some updates according to the lastest edition

Daisy-C:
delete information about code/docs and content inside it because they can't be found anymore.
add information about folders named drasil-database and drasil-drasil-website.


code/README.md
```diff
@@ -63,22 +63,6 @@ Also take a look at the Documentation section of the [Drasil Wiki](https://githu
...
-documentation set, and an internal documentation set with all modules fully exposed.
-
-Upon completion, you should receive several messages stating the Haddock for certain
-packages has been updated, as well as directory locations for the related generated Haddock docs.
-
-The generated Haddock documentation will be placed inside of a newly created `code/docs/` folder.
-You should open up the `code/docs/index.html` file with your web browser if you'd like to view 
-the external documentation set. Alternatively, if you would like to view the internal documentation set,
-you should instead open the `code/docs/full/index.html` file.
-
```
balacij, line 81: Hi @Daisy-C, nice to meet you!

The `code/docs/` folder isn't tracked in the repository because it's just generated artifacts. This text blurb was just discussing how to generate it so that users can inspect our [Haddock docs](https://jacquescarette.github.io/Drasil/docs/full/index.html) locally, on their machines. I _think_ we should keep this discussion here. However, it is out-of-date.

Currently, `make docs` builds only the "fully exposed modules" variant (so it only builds 1 variant). When we run `FULL=1 make docs`, it builds both as described in the first paragraph. So, it looks like the first paragraph just needs to be updated to note that `make docs` only builds 1 variant, while `FULL=1 make docs` builds both now.

balacij (approved):
Nice, thanks!

JacquesCarette (approved):

Issue #2885, opened 2021-11-10
Re-splitting drasil-lang


JacquesCarette:
(A lot of the analysis behind this issue is in #2883; specifically in the 'shallow analysis' for `drasil-lang`).

The analysis there accurately describes how `drasil-lang` defines a whole slew of encodings. Some of these are inter-dependent, some are not.  The [type dependency graph](https://jacquescarette.github.io/Drasil/analysis/TypeDependencyGraphs/lang.svg) and the [module dependency graph](https://jacquescarette.github.io/Drasil/graphs/drasil-lang.pdf) both illustrate this.

It would make a lot of sense to split this up into much smaller pieces. This can be done incrementally. The ones that are 'at the top' of the graph (in dependency order) can likely be hived off most easily. Also, I think there are things at the very bottom that can be split too.

To be precise, I think the following should be relatively easy to split:
- components of a notebook/document
- natural language
- people
- references & citations
- URIs

There will remain dependencies between
- mathematical languages
- constructs
- derivations

The usefulness of doing this split is definitely going to be that some of the dependencies on `drasil-lang` will shift to being dependencies on much smaller bits.

balacij:
Reminder to self: NamedArgument should likely be moved to drasil-code since it's "code-focused" and seemingly tightly bound to internal logic in drasil-code.

balacij:
Related to #3307: We have a few [code-related](https://github.com/JacquesCarette/Drasil/tree/master/code/drasil-lang/lib/Language/Drasil/CodeExpr) [chunks](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-lang/lib/Language/Drasil/Chunk/CodeVar.hs) in `drasil-lang` that I [started moving](https://github.com/JacquesCarette/Drasil/commit/05a9e473642136ad25843c5f44bf29f916cb30ee) to `drasil-code`, where I think they should belong. However, I'm blocked by a circular dependency when I try. I guess this is related to us previously having and then removing `drasil-code-base`.

Path to finding the issue:
* `drasil-lang`'s definition of a "document" language (another issue perhaps) that depends on `CodeExpr` for code blocks. I don't think it should depend on `CodeExpr` because `CodeExpr` is only for expressions, not whole code blocks. Temporarily changing the dependency to `Expr` should be fine, but a more appropriate input for code blocks is probably arbitrary text with a language tag or GOOL if we want to be restrictive to OO languages.
[...]

PR #2893, opened 2021-11-14
Move around typeclasses in `drasil-lang`

balacij:
Contributes to #2885

I want to send a PR early because I've significantly changed the module graph. It would be good for us to take another look at it, and figure out how we can split it. I want to move "Natural Language"-related components into a new `drasil-natural-language` package, but I do not want it to contain any imports on `drasil-lang` while doing so. However, Sentences have data constructors on `Expr`s/`ModelExpr`s/`USymb`s, so it won't be possible.

It's not immediately clear to me how we can reverse this dependency, but figuring this out will allow us to inline any data encodings (which we might be discussing in our own discussions) inside of Sentences.


JacquesCarette:
I think the trick would be to have `Sentence` have a parameter or two. Or at least `SentenceF` would, and then `Sentence` for external consumption would "close over" particular parameters of `SentenceF`.

JacquesCarette (approved):

PR #2899, opened 2021-11-22
Basic comments for the 3 *Expr languages.

### LINKED ISSUE #2897: The difference between CodeExpr, Expr, and Model Expr
When I was working on Drasil, I have some difficulties with the difference between CodeExpr, Expr, and ModelExpr. What's the mechanism of those files?

For example, in the imported phrase, the CodeExpr, Expr, and ModelExpr are looking almost identical, so I think it would be great to have some documentation to describe what are those files, and what's their relationship.

(Is that be ok assign the ticket to you? @balacij)

balacij:
Closes #2897

### DIFF
--- a/code/drasil-code-base/lib/Language/Drasil/Code/Expr.hs
+++ b/code/drasil-code-base/lib/Language/Drasil/Code/Expr.hs
@@ -63,10 +63,8 @@ data UFuncVN = Norm | Dim
 
 -- * CodeExpr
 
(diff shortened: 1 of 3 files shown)


cd155 (commented):
Thanks for adding explaining about three different Expr. They all make sense to me.

JacquesCarette:
See my comments on the issue about the definitions that I use when I think of these 3 languages.

JacquesCarette (approved):

PR #2903, opened 2021-11-30
Doxygen duplicate tag strings issue

peter-michalski:
This is a duplicate of the previous Doxygen duplicate strings issue. I needed to move this work from my master into a dev branch. The original pull request was deleted during this move.


code/drasil-utils/lib/Utils/Drasil/Document.hs
```diff
@@ -11,6 +11,14 @@ blank = text ""
 indent :: Doc -> Doc
 indent = nest 4
 
+-- | Helper which concatenates two Docs.
```
JacquesCarette, line 14: Why not use `<>` directly then?

JacquesCarette (changes requested):
While this does satisfy the raw requirement of removing duplicate strings, it also doesn't really move the needle towards making this a generator move very much. But that can be done later.

code/drasil-utils/lib/Utils/Drasil/Document.hs
peter-michalski: <> could certainly be used directly 

`a +:+. b = a <> b <> text "."`

Sine I was creating a helper which concatenates two Docs and appends a period, I thought that a situation may arise where a helper for concatenating two Docs would also be needed. Similar to `foldlSent` in `Fold.hs`

peter-michalski: Updated the PR, removing unnecessary functions.

JacquesCarette: The point is that `Doc` already has a rich API, which we should use directly. We perhaps under-use it. `Sentence` is our own type, so we're forced to provide combinators for it ourselves.

code/drasil-utils/lib/Utils/Drasil/Document.hs
```diff
@@ -11,14 +11,6 @@ blank = text ""
 indent :: Doc -> Doc
 indent = nest 4
 
--- | Helper which concatenates two Docs.
-(+:+) :: Doc -> Doc -> Doc
-a +:+ b = a <> b
-
--- | Helper which concatenates two Docs and appends a period.
-(+:+.) :: Doc -> Doc -> Doc
-a +:+. b = a +:+ b <> text "."
```
JacquesCarette: Except that you went too far! I did want this particular combinator to be created.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Doxygen/Import.hs
```diff
@@ -33,8 +32,8 @@ verbosityToDoc Quiet = yes
 
 -- | Renderings of comments regarding the default tag values for various Doxygen settings.
 defNo, defYes :: Doc
-defNo = defaultValSentence +:+. no
-defYes = defaultValSentence +:+. yes
+defNo = defaultValSentence <> no <> text "."
```
JacquesCarette, line 35: No, this particular change is a regression, it was better before.

JacquesCarette (changes requested):

code/drasil-utils/lib/Utils/Drasil/Document.hs
peter-michalski: If I add the following combinator, then I will also need to add `+:+`, as I had in the previous PR.

```
-- | Helper which concatenates two Docs and appends a period.
(+:+.) :: Doc -> Doc -> Doc
a +:+. b = a +:+ b <> text "." 
```

Or would you prefer the following for now.
```
-- | Helper which concatenates two Docs and appends a period.
(+:+.) :: Doc -> Doc -> Doc
a +:+. b = a <> b <> text "." 
```

JacquesCarette: I prefer the 2nd one.

peter-michalski: The PR is now updated to reflect the requested change.

JacquesCarette (approved):

Issue #2873, opened 2021-10-31
Merging chunk maps into a single map via `Data.Typeable`


balacij:
# Background
Presently, `ChunkDB`s are built as an enumeration of "chunk" maps;
https://github.com/JacquesCarette/Drasil/blob/10f9f9ed0e332c7a31875f6e035db96549c45945/code/drasil-database/lib/Database/Drasil/ChunkDB.hs#L165-L183

# Problem
i. We are unable to add (almost) arbitrary data into the ChunkDBs for various different systems.
ii. We manually track chunks in lists (by a single type, for later further processing) in `SystemInformation` when we already have them registered in the ChunkDB.
https://github.com/JacquesCarette/Drasil/blob/10f9f9ed0e332c7a31875f6e035db96549c45945/code/drasil-database/lib/Database/Drasil/SystemInformation.hs#L45-L53
iii. Adding arbitrary data to a ChunkDB is impossible without editing core `drasil-database` files, and difficult when editing core `drasil-database` files because it affects compilation of other examples.
iv. Working with similar maps requires redundant manual recreation (swhs and nopcm has a bit of this).
v. We have occurrences of conflicting UIDs because they are registered in different maps.

# Potential Solution : Merging Maps into a single one
I think we can solve these 5 or so problems by merging our maps into a single Map, and using more of the core `Data.Map` functionality.

We will need to hide type information to create a homogeneous Chunk type, and then merge our maps into a single Chunk Map.

The solution would, approximately, be as follows:
(1) Assume we have the following UID type, and typeclasses:
```haskell
type UID = String

-- | An interface to get the UID of a chunk.
class HasUID a where
    uid :: a -> UID

-- | Dump data to Strings for debugging-purposes.
class DrasilDumpable a where
    dump :: a -> String
```

(2) We build a `Chunk` type intended to carry arbitrary data in our final `ChunkDB`:
```haskell
import Data.Typeable -- this is an important design decision (as opposed to Type.Reflection) for simplicity
...
data Chunk where  -- Personally, I find GADTs to be more readable, but this can be done without them via ExistentialQuantification
    CHUNK :: (HasUID t, DrasilDumpable t, Typeable t) => t -> Chunk

instance HasUID         Chunk where uid  (CHUNK t) = uid t
instance DrasilDumpable Chunk where dump (CHUNK t) = dump t
...
```
[...]

PR #2904, opened 2021-11-30
Start appending to UIDs with a few variants of ModelKinds

balacij:
Contributes to #2873

When creating ModelKinds, we're adding information to some chunk. We should be updating it's UID instead of inheriting.

This brings up an issue of "How should we be sharing models between our examples when we only want small alterations?"


balacij:
The edited line counts are heavily skewed up because of the .dot files.

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
```diff
@@ -109,10 +111,11 @@ rocTempDerivArbVol :: [Sentence]
...
   atStartNP (the volHtGen), S "is assumed constant. Then", eqN 1, S "can be written as"]
 
-rocTempDerivConsFlx :: Sentence -> [ConceptInstance] -> [Sentence]
-rocTempDerivConsFlx s assumps = [S "Where", 
+rocTempDerivConsFlx :: (HasUID r, HasRefAddress r, HasShortName r) => r
+ -> Sentence -> [ConceptInstance] -> [Sentence]
+rocTempDerivConsFlx r s assumps = [S "Where", 
   foldlList Comma List (map ch [htFluxIn, htFluxOut, inSA, outSA]),
-  S "are explained in" +:+. refS rocTempSimp, s, S "Assuming", 
+  S "are explained in" +:+. refS r, s, S "Assuming",  -- FIXME: refS r is a hack to share the rocTempSimpRC instance without causing friction between nopcm and swhs
```
balacij, line 118: This `rocTempDerivConsFlx` reference expected the variant of `rocTempSimp` belonging to SWHS, but it was accidentally referenced in nopcm. So, when I changed the UIDs for the `rocTempSimp` in nopcm, this reference broke since the `rocTempSimp` of SWHS wasn't registered in NoPCM's ChunkDB. In other words, this was a reference bug that happened to not cause problems because the UIDs lined up unintentionally.

balacij, line 118: This `rocTempDerivConsFlx` reference expected the variant of `rocTempSimp` belonging to SWHS, but it was accidentally referenced in nopcm. So, when I changed the UIDs for the `rocTempSimp` in nopcm, this reference broke since the `rocTempSimp` of SWHS wasn't registered in NoPCM's ChunkDB. In other words, this was a reference bug that happened to not cause problems because the UIDs lined up unintentionally.

balacij: I'm not sure why this was posted twice.

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs
```diff
@@ -61,7 +61,7 @@ outputSymbols = map qw [QP.position, QP.velocity, QM.orientation,
 unitalChunks :: [UnitalChunk]
 unitalChunks = [QP.acceleration, QP.angularAccel, QP.gravitationalAccel, 
   QP.impulseV, QP.impulseS, iVect, jVect, normalVect, QP.distance, QP.displacement, 
-  QP.time, QP.angularDisplacement, posCM, posj, massj, mTot, accj, velj,
+  QP.time, QP.angularDisplacement, posCM, posj, massj, mTot, accj, angAccj, velj,
```
JacquesCarette, line 64: For example, how did you find out that `angAccj` was 'missing' from the list of unital chunks that should be "registered" for GamePhysics?

code/drasil-example/nopcm/lib/Drasil/NoPCM/GenDefs.hs
```diff
@@ -14,5 +14,5 @@ genDefs = [rocTempSimp, htFluxWaterFromCoil]
 
 rocTempSimp :: GenDefn
 rocTempSimp = gdNoRefs (othModel' rocTempSimpRC) (Nothing :: Maybe UnitDefn)
-  (Just $ rocTempSimpDeriv EmptyS [assumpCWTAT, assumpDWCoW, assumpSHECoW])
+  (Just $ rocTempSimpDeriv rocTempSimp EmptyS [assumpCWTAT, assumpDWCoW, assumpSHECoW])
```
JacquesCarette, line 17: This feels like an orthogonal (but good) change, where additional information is being attached to the derivation. Is that the case?

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
```diff
@@ -86,13 +86,15 @@ newtonLawNote u a c = foldlSent [ch u `S.is` S "found by assuming that",
 --  General Definitions Derivation  --
 --------------------------------------
 
-rocTempSimpDeriv :: Sentence -> [ConceptInstance] -> Derivation
-rocTempSimpDeriv s a = mkDerivName (S "simplified" +:+ phraseNP (rOfChng `of_` temp))
-  (weave [rocTempSimpDerivSent s a, map eS rocTempSimpDerivEqns])
+rocTempSimpDeriv :: (HasUID r, HasRefAddress r, HasShortName r) => r
```
JacquesCarette, line 89: So this seems like a non-trivial change, and it looks due to `rocTempSimpDerivSent` changing, but I'm not clear why?

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
JacquesCarette: Ah, I see how all the above changes 'line up' to here. What I don't quite understand is what is really going on here. At the implementation level, I see what's going on. It's at the "meaning" level that I'm not so sure.  What is the reference `r` supposed to point to? What's the knowledge that's being used / expected / woven together here?

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
```diff
@@ -14,13 +14,14 @@ import qualified Data.List.NonEmpty as NE
 
 -- | 'ConstraintSet's are sets of invariants that always hold for underlying domains.
 data ConstraintSet e = CL {
-    _con  :: ConceptChunk,
-    _invs :: NE.NonEmpty e
+    _csUid :: UID,
```
JacquesCarette, line 17: So this doesn't seem quite right [but I'm not sure]. It feels like the `ConceptChunk` should be the one that is created to exactly mean this set of constraints. So there shouldn't be 2 UIDs, because there shouldn't be 2 different ideas.

It could also be that this is truly a different idea, where we're associating a set of constraints to an existing idea, and so this association does form something new.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
```diff
@@ -36,4 +37,4 @@ instance Express e => Express (ConstraintSet e) where
 
 -- | Smart constructor for building ConstraintSets
 mkConstraintSet :: ConceptChunk -> NE.NonEmpty e -> ConstraintSet e
-mkConstraintSet = CL
+mkConstraintSet c = CL (c +++ "CS") c
```
JacquesCarette, line 40: This kind of hack strongly tells me that we need to revamp our `UID` implementation to something more structured. I know MMT uses URIs for doing something similar. I don't like that either, but it is (slightly) better than name concatenation.

I'm okay with keeping the hack for now, as we can easily find it by looking for `+++`.

code/drasil-theory/lib/Theory/Drasil/ModelKinds.hs
```diff
@@ -50,67 +50,67 @@ makeLenses ''ModelKind
 
 -- | Smart constructor for 'NewDEModel's
 newDEModel :: String -> NP -> DifferentialModel -> ModelKind e
-newDEModel u n dm = MK (NewDEModel dm) (mkUid u) n
+newDEModel u n dm = MK (NewDEModel dm) (mkUid u +++. "MK") n
```
JacquesCarette, line 53: That the string `"MK"` occurs multiple times in this file is bad. It should occur exactly once.

code/drasil-theory/lib/Theory/Drasil/MultiDefn.hs
```diff
@@ -45,41 +44,39 @@ makeLenses ''MultiDefn
...
 instance Express e => Express (MultiDefn e) where
-  express q = equiv $ sy q : NE.toList (NE.map (express . (^. expr)) (q ^. rvs))
+  express q = equiv $ sy (q ^. qd) : NE.toList (NE.map (express . (^. expr)) (q ^. rvs))
 
 -- | Smart constructor for MultiDefns, does nothing special at the moment. First argument is the 'String' to become a 'UID'.
-mkMultiDefn :: String -> QuantityDict -> Sentence -> NE.NonEmpty (DefiningExpr e) -> MultiDefn e
+mkMultiDefn :: UID -> QuantityDict -> Sentence -> NE.NonEmpty (DefiningExpr e) -> MultiDefn e
 mkMultiDefn u q s des
-  | length des == dupsRemovedLen = MultiDefn (mkUid u) q s des
+  | length des == dupsRemovedLen = MultiDefn (u +++. "MD") q s des
```
JacquesCarette, line 62: Although this embedded string does occur only once, the same pattern should be followed where this doesn't happen inline, but through a single spot.

JacquesCarette (changes requested):

code/drasil-example/nopcm/lib/Drasil/NoPCM/GenDefs.hs
balacij: It's an orthogonal issue, yes, but it's at least making the problem known (I don't think it was known before?). I think we're still losing out on information that we might want to retain later via refinement (here, showing that NoPCM's variant of rocTempSimp is that of SWHS with some alterations).

Since NoPCM's rocTempSimp originates from SWHS' but they use different ModelKinds, when I changed how `othModel'` alters UIDs, this caused an unknown UID "Reference" error for `rocTempSimpRC` (which is the existing UID of the same model from SWHS, but in NoPCM, it's UID is `rocTempSimpRCMK`) because SWHS' rocTempSimp's derivation had a reference to itself! However, since _that_ version of `rocTempSimp` is not the same one that NoPCM has, it caused a reference bug (since it's not registered in NoPCM's ChunkDB).

As such, I made his temporary measure to allow us to have self-referencing in chunk contents without causing friction. It might be better to create some sort of "SelfReference" data type, but that would need design. This should be better transcribed when we have a full refinement/specialization system.

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
balacij: Depending on how we view the above discussion, this can be reversed (in favour of creating at ticket to fix it later), document this a bit better, or we can do something else to improve it right now. 

The `r` parameter is to allow us to "pass the correct rocTempSimp" instance around without relying on both `rocTempSimp`s of NoPCM and SWHS happening to have the same UID.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
balacij: I agree. We need to form some sort of an algebra for our `UID`s. It would be particularly nice if we could use them to track down how chunks were formed, e.g., `data UID = Join (NonEmptyList UID) | Phenomena String | Pushout UID String` (except that this would need a lot more thinking done).

Sorry, what is "MMT"?

> I'm okay with keeping the hack for now, as we can easily find it by looking for +++.
Certainly, I think a better UID system would be a nice goal after the ChunkDBs are updated. This is also related to #2788

code/drasil-theory/lib/Theory/Drasil/ModelKinds.hs
balacij: Sounds good, we can fix that (and similarly below).

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs
JacquesCarette: So that's the "real" issue. We need to look at `angAccj`, `IM:rotMot` and the model create off of `angAccj` to see what's going on.

code/drasil-example/nopcm/lib/Drasil/NoPCM/GenDefs.hs
JacquesCarette: Can you also start a new issue for this? I feel like you're telling me about 1/2 of the information I really need to know to understand the issue. And definitely not enough information to understand if this solution is a fine hack or a horrible hack that shouldn't be let through.

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
JacquesCarette: Yes to creating a ticket. We'll see how that goes.

The question to understand will be whether `rocTempSimp` is the same idea, and so should be shared as-is and have the same UID, or is actually different between NoPCM and SWHS, and so shouldn't be regarded as the same.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
JacquesCarette: MMT ([system description paper](https://kwarc.info/people/frabe/Research/rabe_mmtsys_20.pdf) and [site](https://uniformal.github.io/doc/) ) is a very general system for describing theories that might sit in different logics.  It's massively over-engineered (for our purposes), and yet also doesn't contain some crucial knowledge capture mechanisms that we need.

We need to figure out what information should be embedded in our UIDs. Some kind of 'name spaces' comes to mind.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
balacij: I think this heavily relies on how we want to build chunks. In #2873, I viewed Chunks as self-contained units of knowledge and be denoted by a single type with properties shown through typeclasses. The units of knowledge might be built on other chunks (e.g., as it is here, forming a "constraint" relationship between a concept chunk "target" and a list of `e` "constraints". Another example might be QDefinitions where a QDefinition is a specific relationship between a QuantityDict and a singular defining expression), or pushed out from others (e.g., commonly IdeaDicts).

I guess we should pause work on #2873 because that assumes a single Chunk is a single self-contained unit of knowledge. If it isn't (and we want to have a dynamic map-like structure for chunks), then we _might_ need to change what the new ChunkDBs will look like.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
balacij: Ah, that's an interesting project!

> also doesn't contain some crucial knowledge capture mechanisms that we need.

That's very interesting. I'll need to look into it more.

>  We need to figure out what information should be embedded in our UIDs. Some kind of 'name spaces' comes to mind.

I'm a bit confused by our existing ConceptDomains; I'm having a hard time understanding which "level" we exactly want to call a conceptual domain for something. Do you think a new UID scheme could/should address this?

Perhaps we should move this discussion to a new ticket as well? I'll create a ticket.

code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
balacij: Ok, I will reset this first in a commit (before creating the ticket) so that we have a single self-contained commit we can reference in a new ticket.

code/drasil-example/nopcm/lib/Drasil/NoPCM/GenDefs.hs
balacij: Sounds good. I might not be explaining it well enough without being able to refer to specific code snippets. I'll try to write a code-snippet-oriented explanation in a new ticket.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
JacquesCarette: I agree, we need to get the idea and design of Chunks (and UIDs) straight first.

balacij:
It's time to sunset this PR until we figure out how we want to proceed.

Issue #2892, opened 2021-11-13
Fix direct use of Unicode in PDController


JacquesCarette:
https://github.com/JacquesCarette/Drasil/blob/10f9f9ed0e332c7a31875f6e035db96549c45945/code/drasil-example/pdcontroller/lib/Drasil/PDController/Unitals.hs#L15-L18

which was spotted while doing #2865 .

balacij:
What symbols belong in the "Table of Symbols" of the SRS document? Is +/- infinite something we want to put in it?

Should all functions be displayed as the invLaplace shown there (e.g., with it's parameters)?

smiths:
We wouldn't normally don't put mathematical symbols like +/- infinite in the table of symbols.  We think of them as part of the mathematical background knowledge.  In the terminology of theories that we discussed with @JacquesCarette, I think we can think of mathematical knowledge as a "context" theory.  There has to be a point where we stop defining things, or we'll also be defining the symbols +, -, d/dt, etc.

For functions, I like when they are given with their type, but I don't think that is necessary in the table of symbols.

balacij:
That makes more sense to me. Thank you! It's also interesting to think about how we could systematically retrieve information about the mathematical background knowledge. That seems it would need to be manually listed at the moment.

Does that mean we treat the "F(s)" component as part of L inverse's type (and that we should the parameter)?

smiths:
We don't currently show the types of the other symbols, so no I don't think (at least for the moment) that we should show the parameter for L inverse.  For gradient (∇) in SWHS and noPCM we don't show the arguments.

balacij:
The `syminvLaplace    = variable "L⁻¹[F(s)]"` is a bit more tricky. We'll figure this out soon  (hopefully), when we re-work functions and vectors.

PR #2918, opened 2021-12-15
Remove infinity QuantityDicts from PDController, and replace usage

balacij:
Contributes to #2892

Definite integration with infinite bounds becomes indefinite integration, which is why 'stable' changes slightly.

image

### DIFF
--- a/code/drasil-example/pdcontroller/lib/Drasil/PDController/TModel.hs
+++ b/code/drasil-example/pdcontroller/lib/Drasil/PDController/TModel.hs
@@ -20,23 +20,23 @@ theoreticalModels = [tmLaplace, tmInvLaplace, tmSOSystem]
 tmLaplace :: TheoryModel
 tmLaplace
   = tm (othModel' laplaceRC)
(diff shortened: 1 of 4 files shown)


JacquesCarette:
Except that seems wrong? We really do want this to display in the 'usual' way.

JacquesCarette:
Removing the infinity symbols from PDController is good (they should be in `drasil-data`!).

However, there should also be a display rule where if the integral is a definite integral on the whole domain, then it should display using infinite bounds.  This is different than an indefinite integral!

JacquesCarette:
We'd want infinity to be represented in the same way pi is.  We don't want to embed either of them too deeply into Drasil.  Take a look at `Range` for how I side-stepped having to 'write down' infinity in expressions, but how infinity can still appear in output if we so wish, because it is semantically present in the representation (even if it is not syntactically there).

balacij:
Since this PR is stale and I think what @JacquesCarette would require to rewrite this PR anyways, I'll go ahead and close this PR.

Issue #2843, opened 2021-09-16
Eliminate Duplicate String "get_input"


peter-michalski:
Details:

"get_input"
Number of instances of string: 12
Number of files containing the string: 6
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Descriptions.hs:"get_input"
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/DrasilState.hs:"get_input"
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/FunctionCalls.hs:"get_input"
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Generator.hs:"get_input"
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Modules.hs:"get_input"
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Parameters.hs:"get_input"

peter-michalski:
Proposing similar solution as in #2844. All such strings will be located in the same /code file.

JacquesCarette:
And my reaction is the same. This particular one is even trickier, as I think that "get_input" actually ends up appearing in the generated code, so it is a rather important name associated to a concept, so we need to better understand what that concept is.

peter-michalski:
./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/Descriptions.hs:

-- | Returns a description of the input constructor, checking whether each 
-- possible method that may be called by the constructor is defined, and 
-- including it in the description if so.
`inputConstructorDesc :: GenState Description`


-- | Returns a description of what is contained in the Input Format module,
-- if it exists.
```
inputFormatDesc :: GenState Description
inputFormatDesc = do
  g <- get
  let ifDesc False = ""
      ifDesc _ = "the function for reading inputs"
  return $ ifDesc $ "get_input" `elem` defList g
```
`elem` returns True if the list contains an item equal to the first argument


-- | Returns a description for the generated function that reads input from a file,
-- if it exists.
```
inFmtFuncDesc :: GenState Description
inFmtFuncDesc = do
  g <- get
  let ifDesc False = ""
      ifDesc _ = "Reads input from a file with the given file name"
  return $ ifDesc $ "get_input" `elem` defList g
```



./Drasil/code/drasil-code/Language/Drasil/Code/Imperative/DrasilState.hs:

-- | Get input format to be exported (for @get_input@).
-- See 'getExpDerived' for full logic details.
```
[...]

PR #2919, opened 2021-12-20
added Class to render Choices as String

peter-michalski:
Renamed the `RenderChoices` class to `RenderChoicesS` (`Sentence`) and added `RenderChoicesStr`, as suggested in #2843, to render `ComponentName`.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
@@ -74,7 +74,7 @@ inputConstructorDesc = do
       icDesc True = "checking " ++ pAndS ++ " on the input"
       dl = defList g
   return $ "Initializes input object by " ++ stringList [ 
-    ifDesc ((showChsStr GetInput) `elem` dl),
+    ifDesc (showChsStr GetInput `elem` dl),
     idDesc ("derived_values" `elem` dl),
     icDesc ("input_constraints" `elem` dl)]
 
@@ -85,7 +85,7 @@ inputFormatDesc = do
   g <- get
   let ifDesc False = ""
       ifDesc _ = "the function for reading inputs"
-  return $ ifDesc $ (showChsStr GetInput) `elem` defList g
+  return $ ifDesc $ showChsStr GetInput `elem` defList g
 
 -- | Returns a description of what is contained in the Derived Values module,
 -- if it exists.
@@ -169,7 +169,7 @@ inFmtFuncDesc = do
   g <- get
   let ifDesc False = ""
       ifDesc _ = "Reads input from a file with the given file name"
-  return $ ifDesc $ (showChsStr GetInput) `elem` defList g
+  return $ ifDesc $ showChsStr GetInput `elem` defList g
 
 -- | Returns a description for the generated function that checks input constraints,
 -- if it exists.
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
@@ -14,7 +14,7 @@ import Language.Drasil.Code.Imperative.DrasilState (GenState, DrasilState(..),
   inMod)
 import Language.Drasil.Chunk.Code (CodeIdea(codeName), quantvar)
 import Language.Drasil.Choices (ImplementationType(..), InputModule(..), 
-  Structure(..))
+  Structure(..), ComponentName(..), showChsStr)
 import Language.Drasil.CodeSpec (CodeSpec(..))
 import Language.Drasil.Mod (Description)
 
@@ -74,7 +74,7 @@ inputConstructorDesc = do
       icDesc True = "checking " ++ pAndS ++ " on the input"
       dl = defList g
   return $ "Initializes input object by " ++ stringList [ 
-    ifDesc ("get_input" `elem` dl),
+    ifDesc ((showChsStr GetInput) `elem` dl),
     idDesc ("derived_values" `elem` dl),
     icDesc ("input_constraints" `elem` dl)]
 
@@ -85,7 +85,7 @@ inputFormatDesc = do
   g <- get
   let ifDesc False = ""
       ifDesc _ = "the function for reading inputs"
-  return $ ifDesc $ "get_input" `elem` defList g
+  return $ ifDesc $ (showChsStr GetInput) `elem` defList g
 
 -- | Returns a description of what is contained in the Derived Values module,
 -- if it exists.
@@ -169,7 +169,7 @@ inFmtFuncDesc = do
   g <- get
   let ifDesc False = ""
       ifDesc _ = "Reads input from a file with the given file name"
-  return $ ifDesc $ "get_input" `elem` defList g
+  return $ ifDesc $ (showChsStr GetInput) `elem` defList g
 
 -- | Returns a description for the generated function that checks input constraints,
 -- if it exists.
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
@@ -13,7 +13,8 @@ import Language.Drasil.Code.ExtLibImport (ExtLibState)
 import Language.Drasil.Choices (Choices(..), AuxFile, Modularity(..), 
   ImplementationType(..), Comments, Verbosity, MatchedConceptMap, 
   ConstantRepr, ConstantStructure(..), ConstraintBehaviour, 
-  InputModule(..), Logging, Structure(..), inputModule)
+  InputModule(..), Logging, Structure(..), ComponentName(..), inputModule,
+  showChsStr)
 import Language.Drasil.CodeSpec (Input, Const, Derived, Output, Def, 
   CodeSpec(..),  getConstraints)
 import Language.Drasil.Mod (Mod(..), Name, Version, Class(..), 
@@ -260,14 +261,14 @@ getExpInputFormat n chs _ = fMod (modularity chs) (inputStructure chs)
         fMod _ Bundled = []
         fMod Unmodular _ = [(giNm, n)]
         fMod (Modular Combined) _ = [(giNm, "InputParameters")]
-        giNm = "get_input"
+        giNm = showChsStr GetInput
 
 -- | Get input format defined in a class (for @get_input@).
 -- See 'getDerivedCls' for full logic details.
 getInputFormatCls :: Choices -> [Input] -> [ClassDef]
 getInputFormatCls _ [] = []
 getInputFormatCls chs _ = ifCls (inputModule chs) (inputStructure chs)
-  where ifCls Combined Bundled = [("get_input", "InputParameters")]
+  where ifCls Combined Bundled = [((showChsStr GetInput), "InputParameters")]
         ifCls _ _ = []
 
 -- | Gets exported calculations.
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
@@ -268,7 +268,7 @@ getExpInputFormat n chs _ = fMod (modularity chs) (inputStructure chs)
 getInputFormatCls :: Choices -> [Input] -> [ClassDef]
 getInputFormatCls _ [] = []
 getInputFormatCls chs _ = ifCls (inputModule chs) (inputStructure chs)
-  where ifCls Combined Bundled = [((showChsStr GetInput), "InputParameters")]
+  where ifCls Combined Bundled = [(showChsStr GetInput, "InputParameters")]
         ifCls _ _ = []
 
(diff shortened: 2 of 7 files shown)


JacquesCarette:
This seems right, but I'd want to see it used to make sure it will actually work. So please continue with the implementation.

peter-michalski:
@JacquesCarette, I have implemented the class, replacing the "get_input" string.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
```diff
@@ -74,7 +74,7 @@ inputConstructorDesc = do
       icDesc True = "checking " ++ pAndS ++ " on the input"
       dl = defList g
   return $ "Initializes input object by " ++ stringList [ 
-    ifDesc ("get_input" `elem` dl),
+    ifDesc (showChsStr GetInput `elem` dl),
```
JacquesCarette, line 77: So this now becomes a 'hack' at the String level. So the question is, what is being represented by `dl`, and could it be done otherwise?

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
```diff
@@ -260,14 +261,14 @@ getExpInputFormat n chs _ = fMod (modularity chs) (inputStructure chs)
         fMod _ Bundled = []
         fMod Unmodular _ = [(giNm, n)]
         fMod (Modular Combined) _ = [(giNm, "InputParameters")]
-        giNm = "get_input"
+        giNm = showChsStr GetInput
```
JacquesCarette, line 264: Here the question becomes: could the list of pairs being returned not have something more meaningful in its first argument? If it was changed, how many places would be affected?

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
```diff
@@ -35,7 +36,7 @@ getAllInputCalls = do
 
 -- | Generates a call to the function for reading inputs from a file.
 getInputCall :: (OOProg r) => GenState (Maybe (MSStatement r))
-getInputCall = getInOutCall "get_input" getInputFormatIns getInputFormatOuts
+getInputCall = getInOutCall (showChsStr GetInput) getInputFormatIns getInputFormatOuts
```
JacquesCarette, line 39: Why is the first argument to `getInOutCall` still a `String` instead of being something renderable?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -266,7 +267,7 @@ genInputConstructor = do
         ctor <- genConstructor "InputParameters" cdesc (map pcAuto cparams)
           [block ics]
         return $ Just ctor
-  genCtor $ any (`elem` dl) ["get_input", "derived_values", 
+  genCtor $ any (`elem` dl) [showChsStr GetInput, "derived_values",
```
JacquesCarette, line 270: So this line makes me think that "derived_values" and "input_constraints" are of a similar nature to "get_input", and maybe should be dealt with at the same time. [Or maybe not, because things spin out of control if you do it that way. Investigation needed.]

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -422,9 +423,9 @@ genInputFormat s = do
         outs <- getInputFormatOuts
         bod <- readData dd
         desc <- inFmtFuncDesc
-        mthd <- getFunc s "get_input" desc ins outs bod
+        mthd <- getFunc s (showChsStr GetInput) desc ins outs bod
```
JacquesCarette, line 426: same as earlier: shouldn't `getFunc` be upgraded to have its 2nd argument be renderable?  Note that you can make String renderable with an identity function.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Parameters.hs
```diff
@@ -56,13 +56,13 @@ getInputFormatIns = do
   let getIns :: Structure -> InputModule -> [CodeVarChunk]
       getIns Bundled Separated = [quantvar inParams]
       getIns _ _ = []
-  getParams "get_input" In $ quantvar inFileName : getIns (inStruct g) (inMod g)
+  getParams (showChsStr GetInput) In $ quantvar inFileName : getIns (inStruct g) (inMod g)
```
JacquesCarette, line 59: same with `getParams`'s first argument.

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
peter-michalski: Is the `defList` here part of the [Text.Html](https://hackage.haskell.org/package/html-1.0.1.2/docs/src/Text-Html.html#defList) module?  [DrasilState.hs](https://github.com/JacquesCarette/Drasil/blob/92dddf7a545ba5029f99ad5c5eddcd8dad56a2d8/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs) has this as `defList :: [Name]`, I'm not quite sure what is happening with `dl = defList g`

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
peter-michalski: You are referring to the `n` in `getExpInputFormat n chs _`, correct?

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
peter-michalski: Could you please explain what is meant by 'something renderable'?  Does the new `RenderChoicesStr` not now render the `get_input` string?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
peter-michalski: Noted, and will come back to this after the other feedback to this pull request has been resolved.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
peter-michalski: Noted that this is similar to your feedback on the proposed changes to FunctionCalls.hs. I will leave my related question for that feedback as I would appreciate some clarification on what is meant by renderable in these cases. Could you please briefly note or link to some information or code that helps explain 'identity functions'? A search of the repo for those keywords did not immediately provide any leads.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
JacquesCarette: I'm referring to `giNm` still being a `String`. Could it be `GetInput`, for example, instead? Why make it into a `String` so eagerly?

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
JacquesCarette: Exactly: it renders it *here*. What if you waited to render things 'later'. Then point is that `GetInput` is more informative, because it is 'typed' and known.  So what you need is for this argument to be of class `RenderChoicesStr` rather than being rendered right away.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
JacquesCarette: I mean
```haskell
instance RenderChoicesStr String where
  showChsStr = id
```

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
peter-michalski: `getInOutCall` is expecting a `Name` (`String`) as its first argument. It does not seem right to change the signature of `getInOutCall`, which is used in several places, and a `Name` seems like a good argument type. 

AM I wrong and should I change the signature of `getInOutCall` along with any `Name`'s that are passed into it, or should I leave the signature of `getInOutCall` as is and update Choices.hs to have `GetInput` be a `Name` (`String`)? 

I'm trying to understand if the problem is the code that I had submitted for this PR, or if the issue is deeper than that and this  requires an update of the current `getInOutCall` function and any code that is passed in as its arguments.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
peter-michalski: `getExpInputFormat` is expecting to return `[ModExp]`, a (String, String). 

Setting `giNm = GetInput` would result in the following error:

```
Couldn't match type ‘ComponentName’ with ‘[Char]’
Expected type: [ModExp]
Actual type: [(ComponentName, [Char])]
```

This would work if `GetInput` were a `String` or `Name`. We would not need to print using `showChsStr GetInput`. Is this the direction to go?

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
JacquesCarette: That's exactly the discussion: should `getInOutCall` be some kind of very simple function that can't do any reasoning at all, or should it have the capability to do complex things because it knows more about its inputs?  I'm thinking the latter.

So yes, that would involve changing the signature of `getInOutCall` so that its first input be an instance of `RenderChoicesStr` instead of being a `Name`. I think of something of type `RenderChoicesStr a => a` as something that can have a name (in the future), but right now has a different, hopefully more informative, representation.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
JacquesCarette: The same applies here as below: maybe the type should be `RenderChoicesStr a => [(a, String)]` (where, of course, the second `String` will likely have its existence as a string questioned in the future.)

Another possibility is to have the type be `[(ComponentName, String)]`. In the short-term, that might be simpler.

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
peter-michalski: I understand the high level idea what is wanted but am unsure of the implementation based on the above. Currently we have:

```
getInOutCall :: (OOProg r) => Name -> GenState [CodeVarChunk] -> 
  GenState [CodeVarChunk] -> GenState (Maybe (MSStatement r))
```

If I import RenderChoicesStr and update the function signature to the following, I can indeed  just use GetInput as the first input to the function.

```
getInOutCall :: (OOProg r, RenderChoicesStr a) => a -> GenState [CodeVarChunk] -> 
  GenState [CodeVarChunk] -> GenState (Maybe (MSStatement r))
```

But this creates more problems since a `Name` is needed for `fAppInOut` (in GenerateGOOL.hs).
`Couldn't match expected type ‘Name’ with actual type ‘a’`

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
peter-michalski: I'm not familiar with what is suggested here. Should this be using an `Either` instead somewhere? The following has more errors than the `RenderChoicesStr a => a` that was suggested below, so I must be misunderstanding how this is to be applied, or maybe it is not code but the context of what should happen.

```
getInOutCall :: (OOProg r, RenderChoicesStr a) => [(a, String)] -> GenState [CodeVarChunk] -> 
  GenState [CodeVarChunk] -> GenState (Maybe (MSStatement r))
```

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
peter-michalski: To follow up on this and the other comment above, do we want to do-away with `Name` in (all?) other functions as well? Otherwise, if you were suggesting in the other comment that we have _either_ `RenderChoicesStr` or `Name/String` then that would probably make things easier for now, but it is not clear to me yet at this point how to implement that.

JacquesCarette: Just to record our conversation and apply it to here: we want to do away with `Name` as an argument to `getInOutCall` as a first step, along with whatever else is need to make that work. We don't want to do away with all of them in one big swoop. But we do want to (eventually) examine them all.

Also: for now, rather than going with the class-based, very general approach, let's go with just an algebraic datatype replacing `Name`.

peter-michalski:
Removing duplicate Strings, rendering them as Choices is addressed in  #2966. I will close this PR in favour of the other one.

PR #2951, opened 2022-03-10
Categorizing Choice base on new feature diagram

### LINKED ISSUE #2947: Categorize Choice attributes
This is the new tree I propose for the `Choice` hierarchy tree. Here is what I changed.
- **Moved Implementation Type into Modularity**. The Implementation Type is about whether the generated code is an actual program or library. In the root of code, the difference is whether there is a `Controller`. By choosing `Modular` in modularity, it will divide the program into four modules, controller, input, calculations, and output. Therefore, it makes sense to move Implementation Type into Modularity
- **Combined inputStructure, constStructure, and constRepr**. I think those three are about how information should be encoded. The difference between Modularity and Data is that Modularity builds the framework, and Data enforce the details of how information should be written down. The Input information is data, and the constant itself is data, so I choose the name `Data`.
- Combine ConceptMatchMap and SpaceMatch as `Map`
- Combine Doxygen configuration as `Documentation`
- Combine logging configuration as `Log`
- Combine Software Constraint  and Physics Constraint as `Requirement`
- Combine Documentation, Log, Auxiliary files as `OptionalFeatures`

The following is the details on the tree diagram

1. modularity: Modularity
    - Unmodular
    -  Modular
        - Input Module
            - Combined: all input functions will be combined in one module/file.
            - Separated: all input functions will have their own module/file. eg, InputConstraints.py and InputFormat.py are in their own file.
        - Control Module
            - Library: no controller component or main function will be generated
            - Program: a controller component or main function will be generated

2. Data: how information be encoded
    - inputStructure
        - Unbundled: each input variable was directly passed to the function as its own parameter
        - Bundled: all input variables stored in a separated object.
    - constStructure: ConstantStructure
        - Inline: macro expansion on variables. All variables will be replaced by their value.
        - WithInputs: constants are stored in variables within the input object
        - Store Unbundled: constants are stored in variables that were directly passed to the function as its own parameter.
        - Store Bundled: constants are stored in variables within a separated object
    - constRepr
        - Var: all constants represented as regular variables.
        - Const: use selected language's mechanism for defining constants. eg. C# has a build-in constant called `const`

3. Map: maps for which concept/space to use
    - conceptMatch: specifies matches between chunks and `CodeConcept`
    - spaceMatch: specifies matches mathematical spaces and type in the targeted language.

4. Optional Features: those features can be toggled on/off
    - Documentation: Doxygen configuration
        - comments: comment in function, class, module
        - doxVerbosity: types of documentation
        - dates: Doxygen date-field visibility options.
    - Log: logging configuration
        - LogFunc: log messages for function calls
        - LogVar: log messages for variable assignments
    - Auxiliary files: other files, 
        - SampleInput
        - ReadME

5. Requirement: those come from SRS
    - Software Constraint
    - Physics Constraint

6. ODE: ode should be built from Instance Model. Putting ODE in `Choice` is a temporary solution. I inclined to just leave it alone for now, because eventually ODE will be removed from `Choices`.

Detail definitions on every attribute
https://github.com/JacquesCarette/Drasil/wiki/The-Code-Generator

cd155:
This is the new feature diagram about `Choices`. This PR covered all the changes of `Choices`. Since I only changed the internal representation of `Choices`, so no softifacts changed.

close #2947


code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -74,60 +39,101 @@ class RenderChoices a where
...
 
 -- | Renders the modularity of a program.
-instance RenderChoices Modularity where 
+instance RenderChoices Modularity where
   showChs Unmodular = S "Unmodular"
   showChs (Modular Combined) = S "Modular Combined"
   showChs (Modular Separated)= S "Modular Separated"
 
 -- | Options for input modules.
-data InputModule = Combined -- ^ Input-related functions combined in one module.
```
JacquesCarette, line 89: Here (and above), you're removing comments from the code. Why? This is usually not a good idea?

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -250,48 +266,77 @@ hasSampleInput [] = False
...
-  , (S "Software Constraint Behaviour", showChs $ onSfwrConstraint chs)
-  , (S "Physical Constraint Behaviour", showChs $ onPhysConstraint chs)
-  , (S "Comments", showChsList $ comments chs)
-  , (S "Dox Verbosity", showChs $ doxVerbosity chs)
-  , (S "Dates", showChs $ dates chs)
-  , (S "Log File Name", S $ logFile chs)
-  , (S "Logging", showChsList $ logging chs)
-  , (S "Auxiliary Files", showChsList $ auxFiles chs)
-  ] 
+  , (S "Modularity", showChs $ modularity $ architecture chs)
```
JacquesCarette, line 326: Maybe like up into 2 columns? That will make it visually easier to parse what's going on here.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -23,7 +23,7 @@ import Utils.Drasil (subsetOf)
 import Control.Lens ((^.))
 import Data.List (intercalate, nub, (\\))
 import qualified Data.Map as Map
-import Data.Maybe (mapMaybe)
+import Data.Maybe (mapMaybe, fromJust, isNothing)
```
JacquesCarette, line 26: `fromJust` and `isNothing` are typically "code smells" in Haskell. You should use pattern-matching instead.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -93,11 +99,12 @@ codeSpec SI {_sys         = sys
            , _sysinfodb   = db} chs ms =
   let n = programName sys
       inputs' = map quantvar ins
-      const' = map qtov (filter ((`Map.notMember` conceptMatch chs) . (^. uid))
+      const' = map qtov (filter ((`Map.notMember` conceptMatch (maps chs)) . (^. uid))
         cnsts)
       derived = map qtov $ getDerivedInputs ddefs inputs' const' db
       rels = (map qtoc (getEqModQdsFromIm ims ++ mapMaybe qdEFromDD ddefs) \\ derived)
-        ++ map odeDef (odes chs)
+        ++ if isNothing ode then [] else map odeDef (odeInfo $ fromJust ode)
```
JacquesCarette, line 106: here you should add an auxilliary function that does the work, and can do it by pattern-matching on `ode`.

JacquesCarette (changes requested):
The code changes are excellent. There are only some very small details that need to be fixed before this can be merged.

code/drasil-code/lib/Language/Drasil/Choices.hs
cd155: I think I misunderstood the reason add the wiki link to the comments. I thought I have to remove comments which are already documented in the wiki. 

All removed comments occurred in the Choices file, I can add them back.

JacquesCarette: If we had Drasil fully working, we wouldn't want such redundancy. But in the meantime, we don't want to have people have to hunt for documentation.

cd155: I added previous comments back

code/drasil-code/lib/Language/Drasil/Choices.hs
cd155: Here is what I changed for this part. Is that what we want?
```haskell
    (S "Languages", 
      foldlSent_ $ map (S . show) $ lang chs), 
    (S "Modularity", 
      showChs $ modularity $ architecture chs), 
```

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
cd155: Removed fromJust and isNothing

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
cd155: Removed the if statement. An auxiliary function was created with pattern matching.

code/drasil-code/lib/Language/Drasil/Choices.hs
JacquesCarette: No, I meant 2 physical columns on the same line, so the above should look like
```haskell
(S "Languages",       foldlSent_ $ map (S . show) $ lang chs), 
(S "Modularity",      showChs $ modularity $ architecture chs), 
(S "Input Structure", showChs $ inputStructure chs)
```
where I used the next "longer" line for better illustration.

cd155: Changed the format to 2 physical columns on the same line.

JacquesCarette (approved):

Issue #1166, opened 2019-03-20
Abstract Data Definitions


bmaclach:
We need a better way to describe data definitions that are defined by some abstract function that will be specified at the implementation stage. Currently we use a hack where, in cases where we need to display a symbol for an implementation-stage function, we use `asExpr'` and/or `asVC'` to create a stage-independent version of that function.
https://github.com/JacquesCarette/Drasil/blob/193a0337512b1da40045b52c6a46d8a6f28f8e33/code/drasil-code/Language/Drasil/CodeSpec.hs#L216-L224

See the origin of the need for this in #1135 and #1139.

peter-michalski:
`asExpr'` is no longer in the code base, and was replaced by `funcUID`' in Drasil/code/drasil-code/lib/Language/Drasil/CodeSpec.hs :

```
-- FIXME: hack. 
-- | Used for implementation-stage functions that need to be displayed in the SRS.
funcUID' :: Func -> UID
funcUID' f = asVC' f ^. uid

-- FIXME: Part of above hack
-- | Helper for 'funcUID''.
asVC' :: Func -> QuantityDict
asVC' (FDef (FuncDef n _ _ _ _ _)) = vc n (nounPhraseSP n) (Variable n) Real
asVC' (FDef (CtorDef n _ _ _ _))   = vc n (nounPhraseSP n) (Variable n) Real
asVC' (FData (FuncData n _ _))     = vc n (nounPhraseSP n) (Variable n) Real 
```

`asVC'` is referenced in:

1.  Drasil/code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs 
```
symbolsForTable :: [QuantityDict]
symbolsForTable = inputs ++ outputs ++ tmSymbols ++ map qw specParamVals ++ 
  map qw symbolsWithDefns ++ map qw symbols ++ map qw unitless ++
  map qw inputDataConstraints ++ map asVC' [interpY, interpZ]
```

2.   Drasil/code/drasil-example/glassbr/lib/Drasil/GlassBR/DataDefs.hs 
```
strDisFacEq :: Expr
-- strDisFacEq = apply (sy stressDistFac)
--   [sy dimlessLoad, sy aspectRatio]
strDisFacEq = apply (asVC' interpZ) [str "SDF.txt", sy aspectRatio, sy dimlessLoad]
```

JacquesCarette:
It is unsurprising that both uses are in GlassBR. 

[...]

PR #2952, opened 2022-03-10
Removing asVC' hack with replacement QuantityDicts

peter-michalski:
For issue #1166

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
@@ -14,7 +14,6 @@ import Language.Drasil.Code (($:=), Func, FuncStmt(..), Mod,
 import Language.Drasil.CodeExpr
 import qualified Drasil.GlassBR.Unitals as U
 import Language.Drasil.Printers
-import Text.PrettyPrint.HughesPJ (render)
 
 
 allMods :: [Mod]
@@ -163,7 +162,7 @@ extractColumnCT = funcDef "extractColumn" "Extracts a column from a 2D matrix"
   ]
 
 interpY :: Func
-interpY = funcDef (render $ symbolDoc (symbol U.interpY Implementation))
+interpY = funcDef (showSymb $ symbol U.interpY Implementation)
   "Linearly interpolates a y value at given x and z values" 
   [filename, x, z] Real (Just "y value interpolated at given x and z values")
   [
@@ -188,8 +187,14 @@ interpY = funcDef (render $ symbolDoc (symbol U.interpY Implementation))
     FRet $ linInterp [ vLook zVector i (int 0), sy y_1, vLook zVector i (int 1), sy y_2, sy z ]
   ]
 
+-- | showSymb :: Symbol -> String
+-- | showSymb a = render $ symbolDoc a
+
+-- | interpZ :: Func
+-- | interpZ = funcDef (showSymb $ symbol U.interpZ Implementation)
+
 interpZ :: Func
-interpZ = funcDef (render $ symbolDoc (symbol U.interpZ Implementation))
+interpZ = funcDef (showSymb $ symbol U.interpZ Implementation)
   "Linearly interpolates a z value at given x and y values" 
   [filename, x, y] Real (Just "z value interpolated at given x and y values")
   [
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
@@ -12,6 +12,10 @@ import Language.Drasil.Code (($:=), Func, FuncStmt(..), Mod,
   asVC, funcDef, fDecDef, ffor, funcData, quantvar, 
   multiLine, packmod, repeated, singleLine)
 import Language.Drasil.CodeExpr
+import Language.Drasil.Development (showUID)
+
+
+import qualified Drasil.GlassBR.Unitals as U
 
 allMods :: [Mod]
 allMods = [readTableMod, interpMod]
@@ -159,7 +163,7 @@ extractColumnCT = funcDef "extractColumn" "Extracts a column from a 2D matrix"
   ]
 
 interpY :: Func
-interpY = funcDef "interpY" 
+interpY = funcDef (showUID U.interpY)
   "Linearly interpolates a y value at given x and z values" 
   [filename, x, z] Real (Just "y value interpolated at given x and z values")
   [
@@ -185,7 +189,7 @@ interpY = funcDef "interpY"
   ]
 
 interpZ :: Func
-interpZ = funcDef "interpZ" 
+interpZ = funcDef (showUID U.interpZ)
   "Linearly interpolates a z value at given x and y values" 
   [filename, x, y] Real (Just "z value interpolated at given x and y values")
   [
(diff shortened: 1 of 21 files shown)


code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -185,7 +189,7 @@ interpY = funcDef "interpY"
   ]
 
 interpZ :: Func
-interpZ = funcDef "interpZ" 
+interpZ = funcDef (showUID U.interpZ)
```
balacij, line 192: To continue the discussion we had on Teams here a bit, `showUID` (something that shouldn't really exist outside of ChunkDB debugging code) usage here is only desired because we want it to rely on the information from the newly created QuantityDicts (I guess a more appropriate way to get the function name would be a "showSymbol"-like function for the implementation-stage name of the interpZ QuantityDict), but this _should_ be best "cleaned up" if/when we place `Func`s inside of the ChunkDBs. As they are right now, we aren't able to palce them in the ChunkDBs because (a) they don't have UIDs, and (b) the ChunkDBs would require a bit more effort for us to be able to palce them inside of them (maybe #2873).

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -159,7 +163,7 @@ extractColumnCT = funcDef "extractColumn" "Extracts a column from a 2D matrix"
   ]
 
 interpY :: Func
-interpY = funcDef "interpY" 
+interpY = funcDef (showUID U.interpY)
```
JacquesCarette, line 166: Surely there's another field than `UID` in `U.interpY` which also have value "interpY" that could be used? Of this whole change, this is the one bit that sticks out as a hack, all the rest seems great.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
JacquesCarette: Using a hacked up `showSymbol`-like function would be quite a bit superior to using `showUID`. It would at least be pulling from the 'right' bucket of information.

JacquesCarette (changes requested):

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
balacij: I agree, that makes sense.

Regarding the type of the symbol, should it really be a String? Specifically, is there any worth in changing it to the "Implementation-level Symbol"?

If we choose String, I think we have ~2 options:
1) figure out a way to convert the "Implementation-level Symbol" into a Doc/Text (this should already exist somewhere I believe), and then into extract it into a String, or
2) roll our own "Symbol -> String" function (of course, this creates potential room for misalignment with the existing Symbol -> Doc/Text functions)

JacquesCarette: No, it doesn't have to be String. Thinking about it, it should indeed be implementation-level symbol. Well spotted!

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
peter-michalski: @JacquesCarette @balacij  I have made the requested changes in 17d3425547ecceaa52a4d347860ab93713b26c01

JacquesCarette: You have - but the changes are too "brute force". You most definitely should not be importing HughesPJ into any example! The right thing to do is to write an appropriate function (that does what you just coded up) somewhere in an appropriate Drasil package, and then call it here.  Figuring out the right signature is going to be a fun design problem.

peter-michalski: @JacquesCarette would a showSymb-like function be helpful elsewhere as well (not sure of which package to place into yet)?

```
showSymb :: Symbol -> String
showSymb a = render $ symbolDoc a
```

This would allow for the removal of HughesPJ from the example

```
interpZ :: Func
interpZ = funcDef (showSymb $ symbol U.interpZ Implementation)
```

JacquesCarette: Yes, such a function would be good. Probably belongs in drasil-printers. I'd be tempted to put it under the ASCII printer.

peter-michalski: @JacquesCarette I had just edited the above comment to suggest placing it in Drasil/Symbol/Helpers.hs, but I will look into placing it into drasil-printers.

peter-michalski: @JacquesCarette The requested change to add a new function has been applied in the latest commit. The PR is updated.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -163,7 +162,7 @@ extractColumnCT = funcDef "extractColumn" "Extracts a column from a 2D matrix"
   ]
 
 interpY :: Func
-interpY = funcDef (showUID U.interpY)
+interpY = funcDef (showSymb $ symbol U.interpY Implementation)
```
JacquesCarette, line 165: These details should not be in GlassBR. The function that does
```haskell
foo :: HasSymbol x => x -> String
foo x = showSymb $ symbol x Implementation
```
should be in drasil-printers and then called from here.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -188,8 +187,14 @@ interpY = funcDef (showUID U.interpY)
...
 
+-- | showSymb :: Symbol -> String
+-- | showSymb a = render $ symbolDoc a
+
+-- | interpZ :: Func
+-- | interpZ = funcDef (showSymb $ symbol U.interpZ Implementation)
+
 interpZ :: Func
-interpZ = funcDef (showUID U.interpZ)
+interpZ = funcDef (showSymb $ symbol U.interpZ Implementation)
```
JacquesCarette, line 197: and here

code/drasil-printers/lib/Language/Drasil/Plain/Print.hs
```diff
@@ -180,3 +180,7 @@ fenceDocR Paren = text ")"
 fenceDocR Curly = text "}"
 fenceDocR Norm = text "\\|"
 fenceDocR Abs = text "|"
+
+-- | Helper for printing Symbols
+showSymb :: Symbol -> String
```
JacquesCarette, line 185: good start, but this only does 1/2 the job

JacquesCarette (changes requested):

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
peter-michalski: @JacquesCarette The requested change is added in 243618fd3324e77d864c839b82d6ba4b20b0af52

JacquesCarette (approved):

Issue #2915, opened 2021-12-08
Nested list in Table of Contents


tingyuw:
The current way to build the table of contents is by passing the ``DocDesc``, 
```
toToC :: DocSection -> ItemType
toToC TableOfContents      = mktToCSec
toToC (RefSec rs)          = mktRefSec rs
toToC (IntroSec i)         = mktIntroSec i
toToC (StkhldrSec sts)     = mktStkhldrSec sts
toToC (GSDSec gs')         = mktGSDSec gs'
toToC (SSDSec ss)          = mktSSDSec ss
toToC (ReqrmntSec r)       = mktReqrmntSec r
toToC (LCsSec lc)          = mktLCsSec lc
toToC (UCsSec ulcs)        = mktUCsSec ulcs
toToC (TraceabilitySec t)  = mktTraceabilitySec t
toToC (AuxConstntSec acs)  = mktAuxConsSec acs
toToC Bibliography         = mktBib
toToC (AppndxSec a)        = mktAppndxSec a
toToC (OffShelfSolnsSec o) = mktOffShelfSolnSec o
```
and since the subsections are contained in the main section, we can easily map the subsections to build a nested list. 
```
-- | Helper for creating the 'Reference Material' section ToC entry
mktRefSec :: RefSec -> ItemType
mktRefSec (RefProg _ l) =
  mkHeaderItem (namedRef SRS.refMatLabel $ titleize Doc.refMat) $ map mktSubRef l
  where
    mktSubRef :: RefTab -> Sentence
    mktSubRef TUnits        = namedRef SRS.tOfUnitLabel   $ titleize' Doc.tOfUnit
    mktSubRef (TUnits' _ _) = namedRef SRS.tOfUnitLabel   $ titleize' Doc.tOfUnit
    mktSubRef (TSymb _)     = namedRef SRS.tOfSymbLabel   $ titleize' Doc.tOfSymb
    mktSubRef (TSymb' _ _)  = namedRef SRS.tOfSymbLabel   $ titleize' Doc.tOfSymb
    mktSubRef TAandA        = namedRef SRS.tOfAbbAccLabel $ titleize' Doc.abbAcc
```
[...]

Issue #2693, opened 2021-07-17
Changing the representation of Section


tingyuw:
Currently, we use ``Section`` for sections and subsections, which makes the structure recursive. 
```haskell
-- | Section Contents are split into subsections or contents, where contents
-- are standard layout objects (see 'Contents').
data SecCons = Sub Section
             | Con Contents
-- | Sections have a title ('Sentence'), a list of contents ('SecCons')
-- and a shortname ('Reference').
data Section = Section 
             { tle  :: Title 
             , cons :: [SecCons]
             , _lab :: Reference
             }
makeLenses ''Section
```
To make it not recursive, we need to use a different data for subsections and add level information. I was thinking of something like this:
```haskell
data Section = Section Title [SecCons] Reference
data SecCons= 
     Sub Depth Title [Contents] Reference
   | Con Contents
```
We will need to use ``SecCons`` instead of ``Section`` to represent subsections and add level information to it. It works in a similar way as the printer currently does, maybe it would be more straightforward to wrap the sections and print headers based on the level. But the original structure makes more sense to HTML. Not sure whether should I modify the ``Section`` to make it work in all types of documents or introduce a new type of ``Section`` for notebook.

I'm also considering the possibility of introducing a new type in ``SecCons`` for cells because a markdown cell seems to be something between ``Sub`` and ``Con``. However, it also seems unnecessary because I think a cell could be either a ``Sub`` or a ``Con``. 

@balacij @Ant13731 Any ideas and suggestions will be greatly appreciated!

Ant13731:
I think the new idea for sections could work out really well. I'm not sure if this would be able to work in a notebook, but would we be able to make `SecCons` recursively hold information? Maybe something like:
```Haskell
data Section = { tle  :: Title 
             , cons :: [SecCons]
             , _lab :: Reference
             } -- leaving as a record for the lens to the reference
data SecCons = 
     Sub Depth Title [SecCons] Reference
   | Con Depth Contents -- might need to know this depth but not sure
-- Would it be able to flatten itself before being put into a Section?
-- That way we wouldn't need to know the depth?
```
That way the sections themselves are not recursive (and would match up with cells), but we can still keep the idea of recursively adding content (which might match with HTML nicely).

balacij:
It sounds like you specifically want to remove the recursive structure, but I'm not quite sure I understand why. Can you please explain?

tingyuw:
[...]

PR #2954, opened 2022-03-15
Changing the representation of Section

tingyuw:
Contributed to #2693, flatten the Section representation and removed nested structure. Stable files are changed because html is generated in a slightly different way. More changes need to be done later. Future works include: 
1. `TermsandDefs` section consistency for GlassBR #1432 - I removed this section in GlassBR for now since we aren't able to generate it in the new structure, will add it back later after we address the issue. 
2. Adjust higher level structure and language and use tree structure for `Table of Contents`, as discussed in #2915.


tingyuw:
I'll fix the build failed in the next few days.

balacij:
Was this just outdated? Do we need to keep the old branch?

tingyuw:
@balacij Some of these changes are in another branch. I'm currently cleaning up my code, and I will delete this branch (all my branches) after I have merged everything into the master. I will clean up my tickets as wll, don't worry:smile:

PR #2966, opened 2022-04-22
Choices internal concept

peter-michalski:
PR for removing some duplicate Strings. In https://github.com/JacquesCarette/Drasil/commit/c53be7b26a466e3a569589201d80a1a77ba31378 I set up the InternalConcepts in an associative list of (InternalConcept, Name) as part of the Choices record syntax.This provides some ground work for allowing the user to choose their own function names.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -8,13 +8,14 @@ module Language.Drasil.Choices (
   CodeConcept(..), matchConcepts, SpaceMatch, matchSpaces, ImplementationType(..),
   ConstraintBehaviour(..), Comments(..), Verbosity(..), Visibility(..),
   Logging(..), AuxFile(..), getSampleData, hasSampleInput, defaultChoices,
-  choicesSent, showChs) where
+  choicesSent, showChs, InternalConcept(..), genICFuncName) where
 
 import Language.Drasil hiding (None)
 import Language.Drasil.Code.Code (spaceToCodeType)
 import Language.Drasil.Code.Lang (Lang(..))
 import Language.Drasil.Data.ODEInfo (ODEInfo)
 import Language.Drasil.Data.ODELibPckg (ODELibPckg)
+import Language.Drasil.Mod (Name)
 
 import GOOL.Drasil (CodeType)
 
@@ -38,7 +39,9 @@ data Choices = Choices {
   -- | Constraint violation behaviour. Exception or Warning.
   srsConstraints :: Constraints,
   -- | List of external libraries what to utilize 
-  extLibs :: [ExtLib]
+  extLibs :: [ExtLib],
+  -- | List of modifiable function names
+  functionNames :: [(InternalConcept, Name)]
 }
 
 -- | Renders program choices as a 'Sentence'.
@@ -338,7 +341,8 @@ defaultChoices = Choices {
     (makeLogConfig [] "log.txt") 
     [ReadME],
   srsConstraints = makeConstraints Exception Warning,
-  extLibs = []
+  extLibs = [], 
+  functionNames = fnList
 }
 
 -- | Renders 'Choices' as 'Sentence's.
@@ -363,3 +367,26 @@ choicesSent chs = map chsFieldSent [
 -- | Helper to combine pairs of 'Sentence's for rendering 'Choices'.
 chsFieldSent :: (Sentence, Sentence) -> Sentence
 chsFieldSent (rec, chc) = rec +:+ S "selected as" +:+. chc
+
+-- | List of user defined function names.
+-- | List is populated with default values.
+fnList :: [(InternalConcept, Name)]
+fnList = [
+  (GetInput, "get_input"), 
+  (DerivedValues, "derived_values"), 
+  (InputConstraints, "input_constraints"), 
+  (WriteOutput, "write_output")]
+
+-- | Data type of user defined concepts
+data InternalConcept = InputConstraints | WriteOutput | DerivedValues 
+  | GetInput | InputParameters | InputFormat deriving Eq  
+
+-- | Returns user defined function Name
+genICFuncName :: InternalConcept -> Name
+genICFuncName ic = existsFuncName(lookup ic (functionNames defaultChoices))
+  
+-- | Helper function for genICFuncName
+existsFuncName :: (Maybe Name) -> Name
+existsFuncName (Just x) = x
+existsFuncName Nothing = error ("InternalConcept missing from defaultChoices functionNames")
+  
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -17,12 +17,12 @@ import Language.Drasil.Data.ODEInfo (ODEInfo)
 import Language.Drasil.Data.ODELibPckg (ODELibPckg)
 import Language.Drasil.Mod (Name)
 import Data.Maybe (mapMaybe)
+import Data.Map (Map, fromList, (!), lookup, toList)
+import Data.Tuple (swap)
 
 import GOOL.Drasil (CodeType)
 
 import Control.Lens ((^.))
-
-import Data.Map (Map, fromList, (!), lookup, toList)
 
 -- Full details of Choices documentation 
 -- https://github.com/JacquesCarette/Drasil/wiki/The-Code-Generator
@@ -384,10 +384,6 @@ fnList = fromList [
 -- | Helper function for InternalConcept String Map  
 listStrIC :: [String] -> [InternalConcept]
 listStrIC = mapMaybe (flip Data.Map.lookup . fromList . map swap $ toList fnList)
-
--- | Helper function for listStrIC
-swap :: (a, b) -> (b, a)
-swap (a, b) = (b, a)
 
 -- | Data type of user defined concepts
 data InternalConcept = InputConstraints | WriteOutput | DerivedValues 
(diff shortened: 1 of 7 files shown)


code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -362,4 +366,27 @@ choicesSent chs = map chsFieldSent [
 
 -- | Helper to combine pairs of 'Sentence's for rendering 'Choices'.
 chsFieldSent :: (Sentence, Sentence) -> Sentence
-chsFieldSent (rec, chc) = rec +:+ S "selected as" +:+. chc
\ No newline at end of file
+chsFieldSent (rec, chc) = rec +:+ S "selected as" +:+. chc
+
+-- | List of user defined function names.
+-- | List is populated with default values.
```
balacij, line 372: Very minor, but I think the second item doesn't need a `|`?

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -362,4 +366,27 @@ choicesSent chs = map chsFieldSent [
...
+fnList :: [(InternalConcept, Name)]
```
balacij, line 373: So as long as `InternalConcept` has an Eq instance, I think we can also make this a `Map InternalConcept Name` to avoid conflicting entries. Alternatively, it could be a provided Haskell-level function.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -362,4 +366,27 @@ choicesSent chs = map chsFieldSent [
...
+  (InputConstraints, "input_constraints"), 
+  (WriteOutput, "write_output")]
+
+-- | Data type of user defined concepts
+data InternalConcept = InputConstraints | WriteOutput | DerivedValues 
+  | GetInput | InputParameters | InputFormat deriving Eq  
+
+-- | Returns user defined function Name
+genICFuncName :: InternalConcept -> Name
+genICFuncName ic = existsFuncName(lookup ic (functionNames defaultChoices))
```
balacij, line 386: I _think_ the `defaultChoices` should be replaced with a function argument in `genICFuncName`, or else the Choices option you added would always be overridden by the basic `defaultChoices` Choices.

peter-michalski: @balacij I'm not sure I follow. `functionNames = fnList` would be overridden by a basic `defaultChoices` Choices? Why would this happen? Could you also please elaborate on the proposed solution to fix such an issue?

balacij: By "basic `defaultChoices`", I meant the `Choices` value that we have defined and commonly use -- `defaultChoices` is the variable containing it.

You added a mapping of InternalConcepts to Names in the `Choices` record, but, when grabbing the names, I _think_ this is pulling from the built-in `defaultChoices` instead of the one the user provided, which is why I suggested that it might need to be an argument, so that it can grab from the `Choices` instance the user provided instead of strictly. Otherwise, the user provided information for the function names wouldn't be used.

code/drasil-code/lib/Language/Drasil/Choices.hs
peter-michalski: @balacij Are you suggesting a helper function for editing the entries in `fnList`? How about the following:

```
-- | Helper for list entries
fnListEntry :: Eq k => k -> v -> [(k, v)] -> [(k, v)]
fnListEntry key value list = (key,value):(filter ((key /=).fst) list)
```

balacij: That does look like it would do what I suggested, but I was suggesting we can use [Data.Map's Map](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Lazy.html) (Lazy/Strict, probably doesn't matter here since it's small) to contain the mapping, and then we can use their tools a bit. Then, the fnList type signature would change to `Map InternalConcept Name` (e.g., a mapping from InternalConcepts to Name). The `Map k v` requires the type `k` to have an `Eq` instance defined (which, I now see you have defined through the `deriving Eq` bit in the data definition). Alternatively, a haskell-level function might be okay too (e.g., `fnList :: InternalConcept -> Name`).

JacquesCarette: Yes, `Data.Map` would be better here. Better reuse.

code/drasil-code/lib/Language/Drasil/Choices.hs
peter-michalski: Fixed in 4900c9d014d56227f556a60b0e365d3b4bf545c1

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: > Are the built-in defaultChoices not overwritten by the user's Choices?

`defaultChoices` (and all data in Haskell) is immutable. The "dynamic"/mutated data has to be provided through the input arguments, unless, I believe, you're using `ST` (where we can have mutable variables, and which I don't think we want to use).

Sounds good to me! :smile:

peter-michalski:
> HLint is causing problems for this, but it looks like a separate issue.
> 
> EDIT: I created #2969, it fixes the HLint problem locally, and I haven't waited yet to make sure it works on the Actions run, but if you want, you can merge #2969 in to your remote branch to also get the fix here.

Thanks for this. The checks of this PR have now passed.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -362,4 +367,27 @@ choicesSent chs = map chsFieldSent [
...
+-- | Data type of user defined concepts
+data InternalConcept = InputConstraints | WriteOutput | DerivedValues 
+  | GetInput | InputParameters | InputFormat deriving (Eq, Ord)  
+
+-- | Returns user defined function Name
+genICFuncName :: InternalConcept -> Name
+genICFuncName ic = existsFuncName(Data.Map.lookup ic (functionNames defaultChoices))
+  
+-- | Helper function for genICFuncName
+existsFuncName :: Maybe Name -> Name
```
JacquesCarette, line 390: This is a Haskell anti-pattern.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -362,4 +367,27 @@ choicesSent chs = map chsFieldSent [
...
+  (InputConstraints, "input_constraints"), 
+  (WriteOutput, "write_output")]
+
+-- | Data type of user defined concepts
+data InternalConcept = InputConstraints | WriteOutput | DerivedValues 
+  | GetInput | InputParameters | InputFormat deriving (Eq, Ord)  
+
+-- | Returns user defined function Name
+genICFuncName :: InternalConcept -> Name
+genICFuncName ic = existsFuncName(Data.Map.lookup ic (functionNames defaultChoices))
```
JacquesCarette, line 387: May as well use `(!)` from `Data.Map` here, if you're just going to error out. 

We need a better way to error out, but forcing that on this PR is too much.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
```diff
@@ -74,9 +74,9 @@ inputConstructorDesc = do
       icDesc True = "checking " ++ pAndS ++ " on the input"
       dl = defList g
   return $ "Initializes input object by " ++ stringList [ 
-    ifDesc ("get_input" `elem` dl),
-    idDesc ("derived_values" `elem` dl),
-    icDesc ("input_constraints" `elem` dl)]
+    ifDesc (genICFuncName GetInput `elem` dl),
```
JacquesCarette, line 77: So this code is a tiny improvement over what was there before. Why is `dl` still a `String` ? Does it have to be?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
```diff
@@ -85,7 +85,7 @@ inputFormatDesc = do
   g <- get
   let ifDesc False = ""
       ifDesc _ = "the function for reading inputs"
-  return $ ifDesc $ "get_input" `elem` defList g
+  return $ ifDesc $ genICFuncName GetInput `elem` defList g
```
JacquesCarette, line 88: same here - it feels like `defList` should have its type "improved".

JacquesCarette (changes requested):
You have gotten rid of the strings, but there's more to do to improve the design.

code/drasil-code/lib/Language/Drasil/Choices.hs
peter-michalski: This is addressed in a986e0cf111cee60ab034fb92462537589e8bdcd

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
peter-michalski: The type of `defList` type has been changed in a986e0cf111cee60ab034fb92462537589e8bdcd but feels a bit hacky with my use of `listStrIC` and `swap` . Please provide some feedback on improvements and if they are needed in this PR.

JacquesCarette:
@hrzhuang could you try to take over this PR please?

hrzhuang:
I continue this in #3642 since I cannot push to @peter-michalski's fork.

JacquesCarette:
Once you've 'mined' this, do close this PR please.

hrzhuang:
Closing this PR now since work is continuing in #3642.

PR #2969, opened 2022-04-26
Disable single HLint false-positive: 'Evaluation' on `const`

### LINKED ISSUE #2968: HLint updated, and Drasil has new suggested changes
HLint is causing problems for this, but it looks like a separate issue.

_Originally posted by @balacij in https://github.com/JacquesCarette/Drasil/issues/2966#issuecomment-1110004274_

balacij:
Fixes #2968


JacquesCarette (approved):
Fine for now, but I think the better change would be to rename `const` in our classes to something else. Too much danger.

Issue #2906, opened 2021-12-01
Solve a first order system ODE with two dependent variables


cd155:
### Current situation
The front-end (interface) of the ODE solver only takes one dependent variable. When we solve the second-order ODE, we will rewrite it as two first-order ODEs.

For example:
```
The original equation is
z'' = -(1+K) z' - (20+K) z + rK

In order to solve this second-order ODE, it be rewritten into two equations
x' = y
y' = -(1+K) y - (20+K) x + rK
```

In this case, there are two dependent variables after rewriting. The current front-end only takes one dependent variable (Magically, the current ODE solver is able to solve two first-order equations by just using one dependent variable.)

However, the problem is raised when we want to solve a first-order system ODE with two dependent variables without any re-write.

For example:
x' = 4 x + 7 y
y' = -2 x - 5 y

Both x and y are dependent variables.

### Purposed change
Front-end:
- The front-end should allow taking two dependent variables. 

Back-end:
- After taking two dependent variables and multiple `Expr`, that information needs to be properly sent to External Library.

### Result
Solve a first-order system ODE with two dependent variables

### My thought and questions
When I was checking Python library scipy, it is fairly common to solve a system ODE with two variables. The original design only takes one dependent variable, so I just wonder to know why we take one dependent variable at that moment. @JacquesCarette 

Mentioned #2902

smiths:
@cd155, when you solve a second order ODE you do add another dependent variable.  An example is shown here:

https://tutorial.math.lamar.edu/classes/de/systemsde.aspx

Your wording above makes it sound like the system of first order ODEs from solving a second order ODE only involves one dependent variable.

cd155:
@smiths Thanks for the correction. Indeed, to solve a second-order ODE, a new dependent variable is introduced. I will edit my post once I am ready. I think my confusion is coming from the interface of `ODEInfo`, it only has one dependent variable.

```haskell
data ODEInfo = ODEInfo {
  -- | Independent variable.
  indepVar :: CodeVarChunk,
  -- | Dependent variable.
  depVar :: CodeVarChunk,
  -- | Other variables in the ODE.
  otherVars :: [CodeVarChunk],
  tInit :: CodeExpr,
  tFinal :: CodeExpr,
  -- | Initial value of an ODE.
  initVal :: CodeExpr,
  -- | ODE equations.
  odeSyst :: [CodeExpr],
  odeOpts :: ODEOptions
}
```
For example: 
[...]

PR #2979, opened 2022-05-10
Solve a second order ODE

cd155:
Completed solve PDcontroller in C++ and C#. Java library has a different interface to solve a second-order ODE, and it might take a longer time to complete.

In order to solve a second-order ODE, we have to reduce it to a system of first-order ODE in two equations. Therefore, this PR will complete #2906

Note: the branch name is misleading, please ignore it.


cd155:
I figured out how to deal with Java in a second-order order. Since this PR hasn't been merged, I will add the Java part to this PR.

cd155:
This PR addresses the issue in #2981

This PR did not address the issue in https://github.com/JacquesCarette/Drasil/issues/2980#issuecomment-1126542363, I added the boost library to datafiles/pdcontroller. Ideally, we don't want to have to duplicated library.

JacquesCarette:
Boost is huge - we really don't want to have it in here twice. In fact, ideally, we wouldn't even have it once! [We would do things the same way CI does, and download it as necessary when testing]

cd155:
Does this require to change in the make file? Any example I can check out to see how this can be done?

JacquesCarette:
Yes, this would require a change to the make file. And the results would be hard to make portable.

I really don't want to add Boost to the Drasil repo. It's too big. Maybe @balacij has some clever ideas?

balacij:
I agree that ideally we shouldn't/wouldn't have the entire Boost library (or any other library other than our own for that matter) placed in the repository. However, I doubt that we would want to resolve the "'required libraries management' for our generated case studies" issue on this ticket.

So, I think an alternative temporary solution is to use symbolic links. We might want to create a `datafiles/commons/` directory with all the external library packages, moved in from `datafiles/nopcm/` (which I think is where all the manually carried libraries exist), and then create [relative symbolic links](https://mokacoding.com/blog/symliks-in-git/) linking the required libraries from `datafiles/commons/...` to `datafiles/pdcontroller/...`, `datafiles/nopcm/...`. Symbolic links work really well on both Linux and macOS, but they don't exist out-of-the-box for Windows. For Windows users using Git Bash, they will need to re-install Git Bash enabling Symbolic Link support to get it working. For Windows users not using Git Bash, we will need to do a bit more research on the topic.

smiths:
Although symbolic links shouldn't be our long term solution, I agree that a temporary solution along these lines seems reasonable.  I thought we had an existing issue about improving how we handle external libraries, but my search through the issues didn't find anything.

cd155:
#2980 is the ticket try to address this issue. @smiths

smiths:
Yes, #2980 is part of what I was thinking about, but I also thought there was an older issue asking for a solution that didn't involve storing any external libraries in our repo.  It is unusual for a project to keep copies of other projects.  In the long term we want to have a more elegant solution.  Maybe we can generate installers for each of the case studies, or connect with some kind of package manager?

cd155:
Removed duplicated libraries by adding symbolic links.

JacquesCarette:
So it's going to be tricky of me to review this - I'm going to have to do it on the commits themselves. github's interface for PRs with lots of files changed is awful and slow.

JacquesCarette (approved):

cd155:
Yeah, once the chang file is over 3000, it just shows the top 3000 files.

PR #2993, opened 2022-05-24
Replace '.'s with '_'s in dot ID names

### LINKED ISSUE #2988: Some dot graphs broken on website
When looking at the Class Instance dot graphs on the Drasil website, some do not render at all. For example, on Chrome, clicking the [drasil-lang](https://jacquescarette.github.io/Drasil/analysis/ClassInstDep/packagegraphs/lang.svg) graph outputs the following message:
image

However, other packages like [drasil-doclang](https://jacquescarette.github.io/Drasil/analysis/ClassInstDep/packagegraphs/docLang.svg) work as expected.

After a bit of looking around, I found that the dot files don't like qualified class instances. For example, the `M.` of this line causes a syntax error in the generated dot files: https://github.com/JacquesCarette/Drasil/blob/adfc15f13c7cf81101c1dbab97a6aa61f10a4ec6/code/drasil-lang/lib/Language/Drasil/Expr/Class.hs#L406

I think this would need a change in the dotfile generators to filter instances of the format `Qualifier.Class` to something like `Qualifier_Class`.

balacij:
Closes #2988

### DIFF
--- a/code/scripts/ClassInstDepGen.hs
+++ b/code/scripts/ClassInstDepGen.hs
@@ -209,9 +209,16 @@ isEntryEmpty :: SmallEntry -> Bool
 isEntryEmpty (SE [] [] [] []) = False
 isEntryEmpty _ = True
 
+escapeDotID :: String -> String
+escapeDotID = map go
+  where
+    go :: Char -> Char
+    go '.' = '_'
+    go c = c
+
 -- Helper to convert class instances into graph edges.
 mkPkgEdges :: [ClassInstance] -> [Edges]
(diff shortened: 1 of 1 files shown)


balacij:
As described by @Ant13731, this _should_ resolve the immediate issue. 

However, I think we have at least 2 residual issues from this:
1) Why is the intended order of analysis creation: `make graphs`, `make analysis`, and then `make convertAnalyzed`?
2) Why doesn't `make analysis` work on its own? It seems we **need** the `split` package implicitly, and it only works on our machines (and the CI) because we typically run `make code` first (which pulls in the `split` package into the impure work cache that it pulls from).

JacquesCarette (approved):

JacquesCarette:
Merged - but those residual issues should be made into actual issues.

Issue #2985, opened 2022-05-19
neg should generate parenthese


cd155:
https://github.com/JacquesCarette/Drasil/blob/adfc15f13c7cf81101c1dbab97a6aa61f10a4ec6/code/drasil-lang/lib/Language/Drasil/Expr/Class.hs#L98-L99

When we apply an `Expr` to `neg`, it doesn't generate a parenthesis in the code. For example, if we apply an `Expr` in `neg` it just generates a `-` in the code. This becomes a problem when the `Expr` is `a+b`, rather than a single individual `Expr`. 

Right now `neg (a + b)` will be `- a + b` in code rather than `- (a + b)`

I think we saw the issue related to `neg` multiple times.

Previously, we discussed that
1. a + -b should be a - b
2. a - -b should be a + b

Without the fix, it really hard to do matrix transformation to a certain format. There is one step that requires applying `neg` to Expr. https://github.com/JacquesCarette/Drasil/issues/2973#issuecomment-1120442094

smiths:
@cd155, are you sure about this?  I'm not surprised that Drasil has awkward notation like `a--b`, as we have previously discussed, but `neg (a + b)` being expanded as `- a + b` surprises me.  This isn't just awkward notation; it is wrong.  Can you create a working example that you can point to that shows this behaviour?  The example will verify the problem in the existing implementation, and it will give us a path toward fixing the problem.

cd155:
A quick example can be created by editing in NoPCM, @smiths 

In this file
https://github.com/JacquesCarette/Drasil/blob/adfc15f13c7cf81101c1dbab97a6aa61f10a4ec6/code/drasil-example/nopcm/lib/Drasil/NoPCM/ODEs.hs#L19

Change
```haskell
[recip_ (sy tauW) `mulRe` (sy tempC $- idx (sy tempW) (int 0))]
```
 to 
```haskell
 [neg (sy tauW `addRe` sy tauW) `mulRe` (sy tempC $- idx (sy tempW) (int 0))] 
```

In the python code it has (the problem occurs in every language)
```python
-tau_W + tau_W * (T_C - T_W[0])
```

but it supposes to be 
```python
-(tau_W + tau_W) * (T_C - T_W[0])
```

smiths:
Thank you @cd155.  That is indeed incorrect.  It looks like we are too aggressive in removing parentheses.  

[...]

Issue #2194, opened 2020-06-24
Build ODEInfo from InstanceModel


bmaclach:
Currently, the `InstanceModel` type is not sophisticated enough to be able to detect when an instance model is an ODE and extract the required information from the instance model to solve the ODE. Thus, users must compile the `ODEInfo` object on their own, meaning the ODE-related information must be specified in two places (`InstanceModel` and `ODEInfo`).

We should instead build the `ODEInfo` automatically based on the `InstanceModel`. This requires changing the `InstanceModel` type such that the required information can be extracted. Some steps in this direction were taken in the `QDef-In-InstMod` branch.

It may make sense for the `ODEOptions` object, which is contained in `ODEInfo`, to still be defined by the user, instead of being automatically generated, because the information contained in `ODEOptions` may not end up being included in the `InstanceModel` representation for ODEs. The goal is just for the user to not have to specify the same information more than once.

smiths:
Great issue @bmaclach.  I agree that the user will still specify some information in `ODEOptions`.  The Instance model captures the physics, but it should be written abstractly.  How to satisfy the physics, in this case using an ODE solver, is a separate stage in problem solving.  We will always have to add information as we go from abstract requirements to concrete implementations.  As you have stated though, we don't want to have to repeat the same information in more than one place.  We can add knowledge, but any transformation of knowledge should be automated.

JacquesCarette:
There are a few issues related to this already - I think all of them are assigned to me. It is still on my list to try to recapture the ideas in the branch back into main. The basic idea is to have our 'models' have more static information in them. Right now, too much of the information is 'buried', or 'encoded'. This is sufficient for printing and some amounts of code generation, but not enough to do anything "smart" with the information.  We need to teach Drasil that ODE Models are a thing.

cd155:
To record meetings in #2943
With the work of splitting out the model kinds, we are one step further knowing when a model is an ODE.
Currently, we manually build `ODEInfo` and `ODEOptions`. Some information should available in Instance Model. Instead of building the ODE model manually, some information should retrieve from Instance Model.

cd155:
Here are two ways to make `ODEInfo` automatically created from Instance Model. I think both of them seem fine, but any suggestion would be really appreciated. Here are potential changes. @JacquesCarette 

1. Build the `ODEInfo` base on `DifferentialModel`,
[...]

PR #2995, opened 2022-05-26
Automate ODEInfo in a single ODE

### LINKED ISSUE #2973: ODE equation transform, how to deal with transform
In order to solve a high order of ODE, it has to transform into a system of first-order odes. To take a second-order ODE as an example

The equation `2y′′−5y′+y=0`
will be transfer to (we introduce two variable **x1 = y**  and **x2 = y'**)
- x1'=x2
- x2'=−1/2 * x1 + 5/2 * x2

For a 2nd order ODE, we can write it as a system of first-order ODEs in **two** equations
For a 3rd order ODE, we can write it as a system of first-order ODEs in **three** equations
For a 4th order ODE, we can write it as a system of first-order ODEs in **four** equations
.... and so on

In the new system of first-order ODEs, on the left-hand side are the first-order derivatives of newly introduced variables. Take `2y′′−5y′+y=0` as an example. It was re-written and become 
- x1'=x2
- x2'=−1/2 * x1 + 5/2 * x2
The left-hand side becomes x1' and x2'.

This rewriting is sufficient to allow the external libraries (scipy python) to solve a high-order ODE. The question comes what is the user input, and how to convert the user input into a well-form format which external libraries can solve.

**I am not quite sure whether Drsail is capable of this kind of transformation**. Once the input is given, Drasil has to do some internal transformation to generate the expression of the new system of first-order odes.

A temporary solution would be manually doing the transformation in `ODEInfo`. By manually creating the well-form format, the external libraries can solve the high-order ODE. However, we are currently working on getting rid of `ODEInfo`, so automating the transformation seems inevitable.

cd155:
For a single ODE, this PR will automate generating the `ODEInfo` part. A user can put a single ODE, and no longer manually build the `ODEInfo`.

Close #2973

This PR is partly complete #2194, because it only automates a single ODE, not a system ODE.

This PR was blocked by #2985, the `neg` need to be fixed to generate the right equation.


JacquesCarette (approved):
Very nice - all the added documentation is wonderful.

JacquesCarette:
Before I merge this: the changes to the code are correct, right? That's a pretty big change to the equations, I want to make sure it's correct before doing this.

smiths:
This commit really does highlight that we eventually need test cases for our generated code.  We would be more comfortable with a change like this if we could rerun our regression tests on the case studies.  I did check a few at random and the new equations appeared equal to the old equations.  A more systematic review would be a good idea though.

cd155:
@JacquesCarette @smiths before we merge it, the code is correct if #2985 is fixed. Currently, the neg didn't add parenthesis for an `Expr`. The equation can be wrong if we have an `Expr` a + b, then neg (a + b), it will be just - a + b, which is wrong. I test a temporary fix by changing language renderers (it adds parenthesis), and all equations are right. That's why I comment this PR is blocked by #2985.

JacquesCarette:
The way to indicate that a PR is not ready to be merged is to make it a Draft (I've just done that now).

cd155:
Close #2985

[40feca4](https://github.com/JacquesCarette/Drasil/pull/2995/commits/40feca404b998062529db1a4af19f7495f5f14d9) fixed the bug in #2985

PR #3004, opened 2022-05-30
Add generated code for the double pendulum 

### LINKED ISSUE #2619: Double Pendulum: Solve a system of ODEs
To generate code for the double pendulum example, we'll need to extend Drasil's interface to external libraries so that we can solve a system of ODEs.

cd155:
Currently, it only outputs a list of θ1. This can be fixed if we successfully add the output option for the function.

The example solves a non-linear system of ODE. Target languages: python, java, c#, and c++

Close #2619


JacquesCarette (approved):
Very nice!

PR #3023, opened 2022-06-20
Add preferred-citation section to CITATION.cff file

balacij:
This will make users who cite Drasil use "The Drasil Research Team" instead of our names, which is more inclusive, as mentioned by @JacquesCarette.


smiths:
Good idea.

JacquesCarette (approved):

PR #3033, opened 2022-07-18
Package dependencies

peter-michalski:
Adds all-package dependency graph in /graph folder, and renders the graph in the Package Dependency Graphs section of the website.


code/Makefile
```diff
@@ -232,6 +232,14 @@ stabilize: $(STABILIZE_EXAMPLES) ##@Examples Overwrites the stable folder with u
 # then make pdfs from the generated TeX files.
 tex: $(TEX_EXAMPLES) ##@Examples Generate all example pdfs. Needs Graphviz to work.
 
+# Generate graph of package dependencies
+packagedeps:
```
JacquesCarette, line 236: I think the `packagedeps` target should explicitly depend on having `dot` and `tred` installed. I think the `graph` one does that already? [We want to be precise in our dependencies so as to future-proof the Makefile.]

JacquesCarette (changes requested):
Only one thing to tweak, rest looks great.

JacquesCarette (approved):

PR #3061, opened 2022-09-26
Move Dong's final report to the Main branch

### LINKED ISSUE #3060: Move Dong's final report to the Main branch.
@smiths I made a brief literature review. I don't have a new chapter for the literature review, so I put it in Chapter 2 - introduction paragraph 2. Since we face many problems in this research, I choose to find the problem of how other projects deal with manually creating code. Here is the newly added content:

In many projects, developers want to remove manually created code as much as possible because they are tedious and error-prone. For example, the FEniCS is a project for solving differential equations. While solving differential equations with the finite element method, two variables, $A_T$ and $l_T$, are needed and usually calculated by hand (227)(cite loggetal2012). To avoid creating special-purpose code, FEniCS developers decided to solve the problem by creating a form complier. Another example is the ATLAS project, a software tool for solving linear algebra. Researchers can implement many optimizations on specific platforms, but those optimizations may cause a slowdown on other platforms (3)(cite whaleyetal2001). A traditional way of handling this problem is manually creating code for each platform. To avoid creating manually created code, ATLAS researchers decided to create a new paradigm called Automated Empirical Optimization. The Drasil team faces a similar problem: we manually extract necessary information from the original ODE. One solution is to explore a new way to store ODEs in a new data structure. The new structure must allow the different forms to be isomorphic, which means we can map one ODE form to another.

After that, I will move my final report to the Main branch.

cd155:
- Added literature review
- Moving my final report to the main branch 

close #3060


cd155:
Hi @balacij, I got 28 errors from HLint, it seems a new rule somehow prevents me from merging the branch that has hlint warning.

PR #3092, opened 2022-11-11
Add hie.yaml to fix "multi-cradle" issues for Haskell Language Server

balacij:
HLS broke for me with their latest update, but I was able to fix it by adding this "hie.yaml" file. It is generated by [`implicit-hie`](https://hackage.haskell.org/package/implicit-hie).

@peter-michalski I remember you mentioning that you always had problems with HLS. Do you want to check if this branch makes it work for you? You will need to do a `make code` first, and then open VSCode. If VSCode fails to load, running `haskell-language-server` manually in the `code/` directory might help us understand what's broken with yours.

### DIFF
--- a/code/hie.yaml
+++ b/code/hie.yaml
@@ -1,0 +1,118 @@
+cradle:
+  stack:
+    - path: "drasil-build/lib"
+      component: "drasil-build:lib"
+
+    - path: "drasil-code/lib"
+      component: "drasil-code:lib"
+
+    - path: "drasil-code/test"
+      component: "drasil-code:test:codegenTest"
+
+    - path: "drasil-code-base/lib"
+      component: "drasil-code-base:lib"
+
+    - path: "drasil-data/lib"
+      component: "drasil-data:lib"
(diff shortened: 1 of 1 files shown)


peter-michalski:
@balacij I can't install `haskell-language-server` on this PC because I am running a KDE Neon version that is based on Ubuntu 18.04 (and at this time will not be updating it for reasons), which [does not contain `libtinfo6`.](https://forum.cardano.org/t/ghc-install-libtinfo-so-6-cannot-open-shared-object-file/61888/3)

ghc-make.log:
```
... error while loading shared libraries: libtinfo.so.6: cannot open shared object file: No such file or directory
ghc.mk:983: recipe for target 'install_packages' failed
...
```

I want to set up a Drasil dev environment on my Windows PC soon. I will probably do the same on a VM running a newer Linux.

JacquesCarette (approved):

balacij:
@peter-michalski Ah, ok. Sounds good then!

PR #3145, opened 2022-11-27
Independently fix typechecker

JacquesCarette:
@balacij you might need to backport this to your type rules too?


balacij:
Ah, thank you!

balacij (approved):

PR #3146, opened 2022-11-27
Progress on making SSP typecheck

JacquesCarette:
This can (and should be) integrated as it is 'better', but it does not fully fix any open Issue. Fixing more requires changes to Drasil itself.

### DIFF
--- a/code/drasil-example/ssp/lib/Drasil/SSP/DataDefs.hs
+++ b/code/drasil-example/ssp/lib/Drasil/SSP/DataDefs.hs
@@ -48,7 +48,9 @@ intersliceWtrFEqn = completeCase [case1,case2,case3]
           sy waterWeight), inxi waterHght $>= inxi slopeHght)
 
         case2 = (half (square (inxi waterHght $- inxi slipHght))  `mulRe` sy waterWeight,
-                inxi slopeHght $> inxi waterHght $> inxi slipHght)
+                (inxi slopeHght $> inxi waterHght)
+                $&&
+                (inxi waterHght $> inxi slipHght))
 
         case3 = (exactDbl 0, inxi waterHght $<= inxi slipHght)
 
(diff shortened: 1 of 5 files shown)


code/drasil-example/ssp/lib/Drasil/SSP/DataDefs.hs
```diff
@@ -48,7 +48,9 @@ intersliceWtrFEqn = completeCase [case1,case2,case3]
           sy waterWeight), inxi waterHght $>= inxi slopeHght)
 
         case2 = (half (square (inxi waterHght $- inxi slipHght))  `mulRe` sy waterWeight,
-                inxi slopeHght $> inxi waterHght $> inxi slipHght)
```
balacij, line 51: For later, we can probably make this a syntactic sugar of *Expr. I will open a ticket referencing this.

code/stable/ssp/SRS/HTML/SSP_SRS.html
```diff
@@ -3071,7 +3071,7 @@ <h3>Data Definitions</h3>
                     <tr>
                       <th>Equation</th>
                       <td>
-                        \[\symbf{h}=\frac{1}{2} \left({\symbf{h}^{\text{R}}}+{\symbf{h}^{\text{L}}}\right)\]
+                        \[\symbf{h}=\frac{1}{2} \left({{\symbf{h}^{\text{R}}}}_{i}+{{\symbf{h}^{\text{L}}}}_{i}\right)\]
```
balacij, line 3074: This makes sense to me. However, it looks like the LHS of the `=` should be `h_i` too. Since we use a `QDefinition` to define `h`, it might not be quite fit well to say "forall i. h_i equals to...". For this type error, it might be best that we add vector scaling and vector addition.

I've got a bunch of code work to do today regarding cleaning up some typing rules today, but I'm not sure if I'll be able to get to this. I'll open a ticket as blocking this.

balacij (approved):

PR #3150, opened 2022-11-28
Remove 'Radians' from 'Space' enumeration (type universe)

### LINKED ISSUE #3136: Remove `Radians` from `Space`
Yes, I agree that radian is a unit.  I was actually looking at the symbol rad the other day to make sure that it is an SI symbol.  Our table of units says that rad is SI, but I wasn't actually sure, since rad is a derived unit.  It is really m/m (arc length/radius).  I seemed to remember that SI didn't use any symbol for rad, since one isn't technically necessary.  The table "SI derived units with special names and symbols" on the Wikipedia page for [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units) confirmed for me that rad is a symbol used in SI.  

We would not consider metres to be a space, so radians should definitely be out of Space.  :-)

I could see us having an angle space or a length space (etc), if we wanted to attach extra meaning to a specific Real.  We wouldn't name this space after the units though.

_Originally posted by @smiths in https://github.com/JacquesCarette/Drasil/issues/3132#issuecomment-1327684757_

balacij:
Closes #3136 

It doesn't appear to have caused any type-checking errors with the few things I changed to `Real`s.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Code.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Code.hs
@@ -18,7 +18,6 @@ newtype Code = Code { unCode :: [(FilePath, Doc)]}
 spaceToCodeType :: S.Space -> [CodeType]
 spaceToCodeType S.Integer        = [Integer]
 spaceToCodeType S.Natural        = [Integer]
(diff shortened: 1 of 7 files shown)


JacquesCarette (approved):
Wow, that was easy!

Issue #3104, opened 2022-11-22
Type checking error in pdControllers DifferentialModel: imPDRC


balacij:
Spawned from #3093 

```
`imPDRC` exposes ill-typed expressions!
  - Inferred type `Boolean` does not match expected type `Rational`
  - Associative arithmetic operation expects all operands to be of the same type.
    Received:
      - ERROR: Associative arithmetic operation expects all operands to be of the same type.
      Received:
        - Real
        - Rational
        - ERROR: Associative arithmetic operation expects all operands to be of the same type.
        Received:
          - Real
          - Rational
          - Real
```

There are 2 equations exposed by `imPDRC` because it's a 'DifferentialModel'. We might need to re-evaluate what equations the differential models expose for type checking. 

1st error: To my understanding, from the comments, DifferentialModels expose a series of equations, so they should all be boolean-typed Exprs:

https://github.com/JacquesCarette/Drasil/blob/3587dd15a468e3d9d7b90d344753d0fe3a0702c5/code/drasil-lang/lib/Language/Drasil/Chunk/DifferentialModel.hs#L129-L131

https://github.com/JacquesCarette/Drasil/blob/0a240c0231b6f737011bc090962ae1925fbc83e2/code/drasil-example/pdcontroller/lib/Drasil/PDController/IModel.hs#L23-L52

JacquesCarette:
Current error is better:
```
`imPDRC` exposes ill-typed expressions!
  - ERROR: Inferred type `Boolean` does not match expected type `Real`
  - ERROR: Inferred type `Boolean` does not match expected type `Real`
 ```
So I'm not quite sure where the expected type is coming from. The `Boolean` is quite explicit in there (because it's an equation). Might this be a bug in the typechecker? @balacij

**Comment by JacquesCarette** [comment]
This does make more sense. And in current master, all those `Rational` are gone. I'm testing a fix now.

**Comment by JacquesCarette** [comment]
The error has now changed quite a lot:
```
`imPDRC` exposes ill-typed expressions!
  - ERROR: Inferred type `Boolean` does not match expected type `Real`
  - ERROR: Inferred type `Boolean` does not match expected type `Real`
```
More debugging to do!

**Comment by JacquesCarette** [comment]
Now that the fix is in, now the error is
```
`imPDRC` exposes ill-typed expressions!
  - ERROR: Inferred type `Vect Real` does not match expected type `Real`
[...]

PR #3157, opened 2022-11-29
Change type-checking expectation for DifferentialModel

balacij:
Related to #3104 

I might as well commit it, I'm not entirely sure if it's right yet, but I think it makes more sense.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Chunk/DifferentialModel.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Chunk/DifferentialModel.hs
@@ -28,7 +28,7 @@ import Language.Drasil.Chunk.Quantity (qw)
 import Language.Drasil.Literal.Class (LiteralC(exactDbl, int))
 import Data.List (find)
 import Language.Drasil.WellTyped (RequiresChecking (requiredChecks))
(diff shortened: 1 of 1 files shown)


JacquesCarette:
I'll take care of fixing this (the build failed with an hlint warning)

balacij:
Ah, thank you very much!

JacquesCarette:
So now the error is
```
`imPDRC` exposes ill-typed expressions!
  - ERROR: Inferred type `Vect Real` does not match expected type `Real`
  - ERROR: Inferred type `Vect Real` does not match expected type `Real`
```
Problem is, I have no idea where this inferred type comes from. As far as I can tell, there are no vector operations involved, and that seems to be the only way `Vect` gets inferred.

JacquesCarette:
Aha, a clue: `formEquations`, which is used to create what is actually checked, uses `idx`. So it assumes some things are vectors, when this is not actually a good assumption. So it looks like it is `formEquations` which has a problem.

JacquesCarette (approved):

JacquesCarette:
So I'm going to merge this in, even though it's only a partial fix. But this is definitely better.

Issue #3107, opened 2022-11-22
Type checking error in ssp EqModel: crtSlpId


balacij:
Spawned from #3093

```
`crtSlpIdIM` exposes ill-typed expressions!
  - Function application on non-function: Upsilon
```

https://github.com/JacquesCarette/Drasil/blob/0a240c0231b6f737011bc090962ae1925fbc83e2/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs#L590-L608

Issue #3105, opened 2022-11-22
Type checking error in EqModel: fctSfty


balacij:
Spawned from #3093 

```
`FS` exposes ill-typed expressions!
  - Associative arithmetic operation expects all operands to be of the same type.
    Received:
      - ERROR: Associative arithmetic operation expects all operands to be of the same type.
      Received:
        - ERROR: List accessor expects a list/vector
        - ERROR: List accessor expects a list/vector
        - ERROR: List accessor expects a list/vector
```

It looks like it access non-list/vector variables using `idx`.

https://github.com/JacquesCarette/Drasil/blob/0a240c0231b6f737011bc090962ae1925fbc83e2/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs#L53-L68

PR #3158, opened 2022-11-29
Fix a bug in typechecker and some types in SSP

### LINKED ISSUE #3106: Type checking error in ssp EqModel: nrmShrFor
Related to #3093, similar to #3105 

```
`nrmShrForIM` exposes ill-typed expressions!
  - List accessor expects a list/vector
```

https://github.com/JacquesCarette/Drasil/blob/0a240c0231b6f737011bc090962ae1925fbc83e2/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs#L374-L388

JacquesCarette:
But there are many more left!  Also, remove some awful hacks, but leave in place at least one hack.

Fixes #3106 , #3107 , #3105 .

### DIFF
--- a/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs
+++ b/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs
@@ -588,15 +588,25 @@ intrSlcDerivEqn = inxi intNormForce $=
 
 --------------------------------------------------------------------------
 crtSlpId :: InstanceModel
-crtSlpId = imNoDeriv crtSlpIdMK [qwUC slopeDist, qwUC slopeHght, qwUC waterDist,
+crtSlpId = imNoDeriv (othModel' crtSlpIdRC) [qwUC slopeDist, qwUC slopeHght, qwUC waterDist,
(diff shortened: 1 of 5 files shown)


code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs
```diff
@@ -605,7 +615,8 @@ crtSlpIdQD = mkQuantDef fsMin crtSlpIdExpr
 crtSlpIdExpr :: Expr
 crtSlpIdExpr = apply minFunction [sy slopeDist,
```
balacij, line 616: Out of curiosity, what needs to happen to make this right?

balacij (approved):

balacij:
image

The rendering of characters is noticeably different with this one too.

code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs
JacquesCarette: I think what this "function" is supposed to do is to implement lots of the checks that are in the words that come with this instance model. Not sure what it is really supposed to return.

PR #3170, opened 2022-12-01
Fix #3161 (kEnergy in GamePhysics)

### LINKED ISSUE #3161: Typechecking error in GamePhysics: kEnergy
Error is reported as
```
`kEnergy` exposes ill-typed expressions!
  - ERROR: Associative arithmetic operation expects all operands to be of the same expected type (Real).
    Received:
    - Real
    - ERROR: Powers should only be applied to the same numeric type in both operands, or real base with integer exponent. Received `Vect Real` ^ `Real`.
```
The error seems to arise from the expression
```Haskell
kEnergyEqn :: Expr
kEnergyEqn = sy QPP.mass `mulRe` half (square (sy QP.velocity))
```
where `QP.velocity` is a vector. Probably missing a `norm` call.

JacquesCarette:
Definition of kEnergy was missing a norm call. Was found through type checking. Fixes #3161.


balacij (approved):
That was a smooth fix :smile:

PR #3173, opened 2022-12-01
Upgrade various things in Build.yaml

JacquesCarette:
let's see if this helps with the timeouts.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -31,9 +31,8 @@ jobs:
       - name: "Update apt package list"
         run: |
           sudo apt-get update 
-          sudo apt upgrade -y
       - name: "Install system requirements"
-        run: sudo apt-get --no-install-recommends install -y libgmp-dev python3 graphviz doxygen fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-xetex texlive-luatex g++ default-jdk mono-devel inkscape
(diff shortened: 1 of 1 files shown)


.github/workflows/Build.yaml
```diff
@@ -19,17 +19,17 @@ jobs:
     runs-on: ubuntu-latest
```
balacij, line 19: By any chance, can we just revert this back to 20.04?

I'm not sure if Runner upgrades are persistent, but, if they're not, then we might be using up a lot more CI minutes.

JacquesCarette: That's possibly what I will try next.

JacquesCarette:
So the other PR, when restarted, worked. So maybe it's best to not merge this PR and only make the minimal changes?

balacij:
Wow, that's great! Any idea what finally fixed it? In any case, we should for sure update the checkout and cancel-workflow-action tools, so we can at least keep those.

balacij (approved):
As the changes are now, I like them and I think they should be merged. It also looks like the CI worked well.

I will let you analyze the CI as you mentioned as well, so I will just leave "Approval" without merging.

JacquesCarette:
So the weird part is that it worked, and now it's failing again, somewhere else. I'm guessing something else has updated itself, and now we're encountering a problem. Sigh. There's a new PR to revert things back a bit, for now. I'll open an issue for the test.

PR #3180, opened 2022-12-03
Fix GI again

JacquesCarette:
Looks like there's something weird going on with lualatex on ubuntu-latest. So best to revert to 20.04 for now while this is being debugged.


balacij (approved):
I agree.

PR #3183, opened 2022-12-05
Implement vector + and -

### LINKED ISSUE #3162: Typechecking error in GamePhysics: v_i^AB
Meta-comment: super-hard to track these down, as it's only detected through chasing down a long tunnel starting at `v_i^AB`, but really happening somewhere else.
```
`v_i^AB` exposes ill-typed expressions!
  - ERROR: Both operands of an (in)equality (=/≠) must be of the same type. Received `Vect Real` & `Real`.
```

So first thing to figure out is what this is (from Unitals)
```haskell
initRelVel = ucs' (dccWDS "v_i^AB" (compoundPhrase'
                 (compoundPhrase' (cn "initial relative") (QP.velocity ^. term))
                 (cn "between rigid bodies of A and B")) (phrase QP.velocity))
                 (sup (sub (eqSymb QP.velocity) QP.initial) (Concat [lBodyA, lBodyB])) (Vect Real) velU
```
Note that the UID is very different than `initRelVel`, which is now what must be used to really track things down. 

From DataDefs:
```haskell
reVelInColl :: SimpleQDef
reVelInColl = mkQuantDef initRelVel reVelInCollEqn

reVelInCollEqn :: Expr
reVelInCollEqn = sy velAP $- sy velBP
```
So indeed, that `SimpleQDef` is ill-typed since the lhs is `Vect Real` and the rhs is `Real`. I think the rhs should be a vector too? @smiths ?

JacquesCarette:
Also fixes a minor annoyance in the Makefile. Fixes #3162. Also fixes a bug in the typechecker on cross products.

Will need some refactoring, as this introduces some helper functions (and a class) in the 'wrong' place. Can be done later.


code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -206,6 +216,17 @@ assocArithOperToTy MulI  = S.Integer
 assocArithOperToTy AddRe = S.Real
 assocArithOperToTy MulRe = S.Real
 
+-- helper function for typechecking to help reduce duplication
+vvvInfer :: TypingContext Space -> VVVBinOp -> Expr -> Expr -> Either Space TypeError
+vvvInfer ctx op l r = case (infer ctx l, infer ctx r) of
+    (Left lt@(S.Vect lsp), Left (S.Vect rsp)) -> if lsp == rsp && S.isBasicNumSpace lsp
+                                                      && lsp /= S.Natural
```
balacij, line 223: Naturals are always an interesting case for us. We can allow vector addition of naturals, but I think vector cross product of naturals is a bit dubious, same as subtraction (which would really be monus).

This is the only thing there is to _really_ question (everything else seems spotless).

balacij (changes requested):

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: Agreed. Done in next check-in.

JacquesCarette:
So I think the build might fail (it did for me locally) because of a change in GlassBR's PDF. Let's see.

JacquesCarette:
Hey, the build worked!

balacij (approved):

balacij:
I find it odd that I need to manually re-request a review to myself (after all threads were resolved) to accept the PR.

Anycase, everything looks good to me!

PR #3192, opened 2022-12-20
Fix `d_offset` constraint in Projectile

### LINKED ISSUE #3187: Fix d_offset constraint in the Projectile Example
In `IM:messageIM` in the [Projectile Example](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:IMs). The input constraint says:

`d_offset > - p_land` 

However, the constraint should be correct to read:

`d_offset > - p_target`

This is derived from IM:offsetIM, where we know `d_offset = p_land - p_target`.  We also know that `p_land` is greater than 0.  If we substitute `p_land = 0` into the expression for `d_offset`, we find `d_offset > - p_target`.

This corresponds to the case where the projectile goes straight up and lands at the origin (lands at the launcher).  We have constrained the launch angle to be less than pi/2, so it cannot land to the left of the launcher. The offset at the origin is `-p_target`.  The offset cannot be smaller (more negative, further to the left) than this.

This issue was raised in #2264.

@samm82, I've assigned you to correct this, since you are the projectile expert.  :smile:

samm82:
Closes #3187 

This change did not affect the generated code, but removed the line "Successfully matched Rational with Float." (or its equivalent) from the designLogs for projectile_s_l_nol_u_u_v_f.


smiths (approved):
The changes look good to me, with the idea that in the future we will improve the implementation as discussed in #3187.

JacquesCarette (approved):
I'm guessing that the change wrt Rational is likely that you removed the last reference to something typed `Rational`. I had already removed a lot (it was a hack - it really was meant for things that are rational, not real). I am surprised that it was your change that triggered this, but hey, it is an expected improvement.

Issue #3078, opened 2022-10-19
Generate SRS in .ipynb format


tingyuw:
Generate .ipynb file for all examples and fix the formatting issues so it can be read by the Jupyter Notebook.

tingyuw:
JSON format for all SRS examples will now be generated. Formatting issues are fix as well. We can view and open the SRS in Jupyter Notebook. The file for each example can be found under path ../SRS/JSON/example.ipynb. There are still some references and equations need to be fixed. 

Reminder: Add .ipynb_checkpoints to gitignore.

tingyuw:
Fix this issue #2761 in SRS.

JacquesCarette:
Nice progress. Looking forward to the PR.

tingyuw:
Closing this as it's been merged to master.

PR #3201, opened 2023-01-03
Generate SRS in JSON format

tingyuw:
Contributed to #3078.
This branch is used to work on code generation but I moved the updated code for generating SRS in JSON format to this branch so this part can be merged first if everything looks good. 
 
Key changes:
1. Generate all examples' SRS in JSON format (file locate in example/SRS/JSON/example_SRS.ipynb)
2. Update the stable folder accordingly

A quick question: if we open the .ipynb file in the Jupyter lab/notebook, a .ipynb_checkpoints folder will be generated, should we include this folder in the .gitignore to avoid miscommitting?


code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -87,11 +88,11 @@ prntCSS docType fn body = do
     getFD dtype = show dtype ++ "/HTML/"
 
 -- | Renders the documents.
-writeDoc :: PrintingInformation -> Format -> Filename -> Document -> Doc
-writeDoc s TeX  _  doc = genTeX doc s
-writeDoc s HTML fn doc = genHTML s fn doc
-writeDoc s JSON _ doc  = genJSON s doc
-writeDoc _    _  _   _ = error "we can only write TeX/HTML (for now)"
+writeDoc :: PrintingInformation -> String -> Format -> Filename -> Document -> Doc
```
JacquesCarette, line 91: you should specify in the comment what the `String` argument is for

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -87,11 +88,11 @@ prntCSS docType fn body = do
...
+writeDoc s _  TeX  _  doc = genTeX doc s
+writeDoc s _  HTML fn doc = genHTML s fn doc
+writeDoc s st JSON _  doc = genJSON s st doc
+writeDoc _ _   _   _  _   = error "we can only write TeX/HTML (for now)"
```
JacquesCarette, line 95: you should fix the error message

code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
```diff
@@ -27,33 +27,43 @@ import Language.Drasil.JSON.Helpers (makeMetadata, h, stripnewLine, nbformat,
  tr, td, image, li, pa, ba, table, refwrap, refID, reflink, reflinkURI, mkDiv)
 
 -- | Generate a python notebook document (using json).
-genJSON :: PrintingInformation -> L.Document -> Doc
-genJSON sm doc = build (makeDocument sm doc)
+-- build : build the general Jupyter Notbook document
+-- build': build the SRS example in JSON format
+genJSON :: PrintingInformation -> String -> L.Document -> Doc
+genJSON sm "Jupyter" doc = build (makeDocument sm doc)
```
JacquesCarette, line 33: Matching on a `String` is a bad idea. Create a data type (whose contents can be, for example `Jupyter` and `Other`).

JacquesCarette (changes requested):
Just a few minor things. I was a little surprised that the output was .ipynb since this seems to had JSON output, not Jupyter output?

code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
tingyuw: We have a ``DocType`` data in drasil-gen but it's not a dependency of the printer (drasil-printer is a dependency of drasil-gen though). Should we move the ``DocType`` to printer so we can use the information in both places?

tingyuw:
Jupyter Notebook is a JSON format document with .ipynb file extension. I guess is because it usually mix with Python code.

JacquesCarette (approved):

code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
JacquesCarette: That's likely not going to be the final resting place for `DocType`, but this will work for now.

Issue #3224, opened 2023-01-07
Node.js deprecations


JacquesCarette:
We're getting a bunch of warnings ([see this build run for example](https://github.com/JacquesCarette/Drasil/actions/runs/3858435698)). @balacij , might you be able to take care of this?

balacij:
Yes, I will be able to. I think it's just the github-pages-deploy-action script that needs to be updated.

PR #3226, opened 2023-01-09
Update GH Actions: JamesIves/github-pages-deploy-action

balacij:
This _should_ do the trick for #3224, but let's merge this first to find out, and then manually handle (close) #3224 if it worked.


balacij:
More updates are needed, and I already moved to another branch. Sorry for the dead PR!

Issue #3204, opened 2023-01-04
Update the used Stackage & GHC version


balacij:
The latest Haskell Language Server (HLS) release, [1.9.0.0](https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html), has deprecated support for GHC 8.8.4 (which we use). I'm not fully sure why they've chosen to do this, but I imagine it's just an issue of maintenance. 

If new users come across Drasil and want to use it, they _might_ want to use HLS (as I do). The typical way to install HLS is using the prebuilt binaries and currently supported GHC versions. As such, users would need to install from source to use HLS with Drasil.

It seems that the Haskell community navigates HLS developments against "current" stackage & GHC versions. So, I wonder: should we upgrade our focused GHC version? If so, should we also upgrade from the Haskell2010 language spec. to GHC2021?

The current Stackage version is [20.5](https://www.stackage.org/lts-20.5) and uses GHC version 9.2.5.

Naively updating the Stackage version to 20.5 doesn't appear to cause any errors, but it does result in some build warnings that we would need to resolve.

smiths:
@JacquesCarette has more expertise than me, but my instinct is that we should update.  It won't get easier over time.  Maybe we could wait until the summer, with the hope that we'll have a summer student that can help with any migration issues that arise?

balacij:
Sounds good, thank you.

Regarding the changes, I'm more than happy to make the changes. I don't there are too many needed changes. I changed the logs and I'm having the [CI run](https://github.com/JacquesCarette/Drasil/actions/runs/3843220511/jobs/6545283072) to show all the warnings that pop up now.

JacquesCarette:
If it's easy, we should upgrade.

balacij:
I'll handle it.

Aside: "difftastic" is a nice drop-in replacement to "diff"!

![Screenshot from 2023-01-06 09-50-52](https://user-images.githubusercontent.com/1627302/211036229-7606cc15-c513-4c02-962b-2e295c40d242.png)

PR #3233, opened 2023-01-10
Resolve conflicting instances for Semigroup & Monoid for `PrintLaTeX Doc`

### LINKED ISSUE #3230: Conflicting Semigroup and Monoid instance for `PrintLaTeX Doc`
Blocking #3204 because GHC 9.2.5 now adds a warning for `Monoid` instantiation with instances of `mappend` conflicting with `Semigroup`'s `<>`

https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs#L79-L87

[Related Monoid Wiki page on Haskell docs.](https://wiki.haskell.org/Monoid)

balacij:
Closes #3230 

I'm not quite sure I'm happy with this resolution yet. The 'code' part of the solution isn't difficult now that it's understood. The only questionable component is: what should `<>` (and hence `mappend`) be doing?

For now, I've defaulted to `mappend` to `<>` as per current recommendations by the Haskell community, as `mappend` will eventually be removed. Then, I replaced usage of `mconcat` (which relied on the old `mappend` definition) with a copy of the old definition.

So, now the question goes to: was this the right solution?

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs
@@ -80,11 +80,9 @@ getCtx = PL id
 instance Semigroup (PrintLaTeX TP.Doc) where
   (PL s1) <> (PL s2) = PL $ \ctx -> s1 ctx TP.<> s2 ctx
 
--- very convenient lifting of $$
--- | D is a monad.
+-- | D is a monoid.
 instance Monoid (PrintLaTeX TP.Doc) where
   mempty = pure TP.empty
-  -- (PL s1) `mappend` (PL s2) = PL $ \ctx -> s1 ctx $$ s2 ctx
 
 -- may revisit later
 -- | Since Text.PrettyPrint steals <>, use %% instead for mappend.
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs
@@ -84,12 +84,12 @@ instance Semigroup (PrintLaTeX TP.Doc) where
 -- | D is a monad.
 instance Monoid (PrintLaTeX TP.Doc) where
   mempty = pure TP.empty
-  (PL s1) `mappend` (PL s2) = PL $ \ctx -> s1 ctx $$ s2 ctx
+  -- (PL s1) `mappend` (PL s2) = PL $ \ctx -> s1 ctx $$ s2 ctx
 
 -- may revisit later
 -- | Since Text.PrettyPrint steals <>, use %% instead for mappend.
 (%%) :: D -> D -> D
-(%%) = mappend
+(%%) = liftA2 ($$)
 
 -- | Lifts Text.PrettyPrint's $+$. Above, with no overlapping. Associative.
 ($+$) :: D -> D -> D
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
@@ -247,7 +247,7 @@ makeHeaders ls = hpunctuate (text " & ") (map (bold . spec) ls) %% pure dbs
 
 -- | Creates the rows for a table.
 makeRows :: [[Spec]] -> D
-makeRows = mconcat . map (\c -> makeColumns c %% pure dbs)
+makeRows = foldr (%%) mempty . map (\c -> makeColumns c %% pure dbs)
 
 -- | Creates the columns for a table.
 makeColumns :: [Spec] -> D
@@ -460,7 +460,7 @@ makeBib sm bib = mkEnvArgBr "filecontents*" (bibFname ++ ".bib") (mkBibRef sm bi
 
 -- | Renders a bibliographical reference.
 mkBibRef :: PrintingInformation -> BibRef -> D
-mkBibRef sm = mconcat . map (renderF sm)
+mkBibRef sm = foldr (%%) mempty . map (renderF sm)
 
 -- | Helper that renders a citation.
 renderF :: PrintingInformation -> Citation -> D
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
@@ -247,7 +247,7 @@ makeHeaders ls = hpunctuate (text " & ") (map (bold . spec) ls) %% pure dbs
 
 -- | Creates the rows for a table.
 makeRows :: [[Spec]] -> D
-makeRows = foldr (%%) mempty . map (\c -> makeColumns c %% pure dbs)
+makeRows = foldr ((%%) . (\c -> makeColumns c %% pure dbs)) mempty
 
 -- | Creates the columns for a table.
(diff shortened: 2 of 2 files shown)


JacquesCarette:
[Drasil: Semigroup instance](https://trello.com/c/ipl8QOen/280-drasil-semigroup-instance)

balacij:
In this PR, I forced the shared bin. operation for `PrintLaTeX Doc`'s Semigroup and Monoid instances to be defined the way that the Semigroup instance had it (basic horizontal concatenation, no spacing) because the opposite way (Monoid's over Semigroup's definitions) broke a considerable amount more. For example, it broke sentence building, *TeX commands, and lots more:

```
266,268c1733,1779
< \subsubsection{Goal Statements}
< \label{Sec:GoalStmt}
< Given the mass and length of the rod, initial angle of the mass and the gravitational constant, the goal statements are:
---
> \subsubsection
> {
> Goal
>  
> 
> Statements
> }
> \label
> {
> Sec:GoalStmt
> }
> Given
>  
> the
>  
> mass
>  
> and
>  
> length
>  
> of the
>  
> rod
> ,
>  
> initial angle
>  
> of the
>  
> mass
>  
> and
>  
> the
>  
> gravitational constant
> ,
>  
> the
>  
> goal
>  
> statements
>  
> are
> :
```

Meanwhile, the Semigroup definition over the Monoid definition only broke _some_ tables (table of units, traceability matrix, etc. -- it seemed that other tables might have been built in different ways).

So, for the reason of "less immediate breakage when copying one definition over the other," I think it's fair to default the assoc. bin. op. of the Semigroup & Monoid instance of `PrintLaTeX Doc` to the definition used in the original Semigroup instance's `<>`. Other than this, I can't think of any real reason to use any one over the other.

JacquesCarette:
I'm having a hard time visualizing all of this. Could you add links to what the old code was, and what it is now? There shouldn't be any breakage at all, so I'm a little puzzled.

balacij:
Yes, of course.

# Greater Objective: #3204 

We wanted to upgrade the used GHC + Stack LTS versions.

# Problem

After updating the GHC version (through updating the used Stack LTS) from 8.8.4 to 9.2.5, Drasil's compilation throws the following _warning_ (not error):

```console
drasil-printers > /home/jasonbalaci/Programming/Drasil/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs:87:3: warning: [-Wnoncanonical-monoid-instances]
drasil-printers >     Noncanonical ‘mappend’ definition detected
drasil-printers >     in the instance declaration for ‘Monoid (PrintLaTeX TP.Doc)’.
drasil-printers >     ‘mappend’ will eventually be removed in favour of ‘(<>)’
drasil-printers >     Either remove definition for ‘mappend’ (recommended) or define as ‘mappend = (<>)’
drasil-printers >     See also: https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/semigroup-monoid
drasil-printers >    |         
drasil-printers > 87 |   (PL s1) `mappend` (PL s2) = PL $ \ctx -> s1 ctx $$ s2 ctx
drasil-printers >    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The [link](https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/semigroup-monoid) it provides gives a bit of background information about the change in GHC, as well as Semigroup and Monoids in Haskell.

To be clear, the `base` GHC library now expects that instances of `Monoid` can only occur for things that are `Semigroup`s already, and that all `Monoid`s are expected to share the same associative binary operation with their contained `Semigroup`.

The problematic area of code it found is:

https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-printers/lib/Language/Drasil/TeX/Monad.hs#L79-L87

Here, the definitions for `<>` and `mappend` are different, despite denoting, what should be, the same associative binary operation. Specifically, the Semigroup instance uses `TP.<>` (Text.PrettyPrint's horizontal document concatenation) while the Monoid instance uses `$$` (Text.PrettyPrint's horizontal document concatenation with spacing).

So the issue lies in the conflicting definitions of what should be the same associative binary operator.

# Solutions

The "solution" is partially given to us. `mappend` will eventually be removed from the `Monoid` typeclass definition and they've added a warning notification for when `mappend` is defined in ways other than expected (e.g., using `Semigroup`'s `<>`). So, we ultimately need to remove the `mappend` definition. However, we have a few options related to the "how" and "when":

## A: Later

We can kick the can down the road and suppress this warning with a GHC flag, but I think it's best to get it out of the way while we know what the issue is. 

## B: Now

Assuming we want to resolve it now, then we need to remove the definition of `mappend` from `Monoid` (GHC will default it to Semigroup's `<>`) and commit the definition of the associative binary operation (`<>` of `Semigroup`) to one of those currently defined (or something completely different, I suppose). Thus, we have two reasonable options:

1. Prefer the original `Semigroup`'s `<>` definition, as done in this PR
2. Prefer the original `Monoid`'s `mappend` definition

If  we go down the route of (1), then there is only a bit of breakage in expectations (hence breaking `stable/`). Thankfully, the issues all lie in `mconcat` usage (note that [`mconcat` is defined using `mappend`](https://hackage.haskell.org/package/base-4.17.0.0/docs/src/GHC.Base.html#mconcat)).

If we go down the route of (2), then there is [a lot more breakage in expectations](https://github.com/JacquesCarette/Drasil/pull/3233#issuecomment-1378105489) (hence breaking `stable/`). We would need to re-evaluate all of our `<>` usage in the document printing, which is rather frequent and often. Notably, it breaks how Sentences are printed.

Assuming we wanted to fix the issue "now," then I can't think of any reasoning to choose between (1) and (2) other than the amount of breakage it would result in. Thus, I've defaulted this PR to option (1), which has a fairly minimal amount of breakage.

JacquesCarette:
I agree that we shouldn't kick this down to 'later'. And it's quite, quite sneaky to have different instances for `Semigroup` and `Monoid`! (more than likely, `git blame` would point the finger at me for that...)

So it makes most sense to define things using the 'no spacing' version, as this is the lowest-level "stick things together" operation, and we're really not ready for that. And thus, indeed, analyzing in more depth the breakage from that version is the way to go.  So down (1) we go.

(The operator names are cryptic, I'm not sure they are mnemonic to anyone except maybe the original author; well, except for `mappend` which is short for `monoid append` as 'appending' is what the 'free monoid' (i.e. monoid with no extra equations) does).

So now I'll dig into the PR proper, to see if I have comments there, now that I understand the problem and the design space. As I said yesterday, that's necessary to be able to judge if the solution is a good one.

JacquesCarette (changes requested):
Wow, that was a lot of discussion for so few lines of code changed!

However, this has revealed a design decision that isn't commented: `makeRows` and `mkBibRef` do some spacing control as well as 'their job'. That should be added to the comment for each. And then this will be ready to go.

balacij:
Indeed. Figuring out the issue was `mconcat` was a fun maze to navigate. Figuring out which it should be was also a bit confusing because of qualified imports between files being different.

I've updated the comments now as you've suggested. Sorry for the delay!

Issue #3234, opened 2023-01-10
Duplicate "the" (coming from `the_ofThe`, etc.?)


samm82:
From my investigation in #3199, I noticed some `UnitalChunk`s with sentences built using `the_ofThe`, etc. but also having "the" prepended (these specific examples from DblPendulum):

https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Unitals.hs#L90-L92
https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Unitals.hs#L98-L100
https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Unitals.hs#L58-L60
https://github.com/JacquesCarette/Drasil/blob/1831438948f4bb92ba7050a27ccbff6489904784/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Unitals.hs#L146-L148

I believe this results in "the" being duplicated:
![image](https://user-images.githubusercontent.com/35857611/211656239-fbccdcab-18ef-4ab0-a5a2-053ef6fb476f.png)

JacquesCarette:
Nicely spotted!

cd155:
The problem occurs while using the combination of `phraseNP` and `"the_ofThe"`. 

For example, 
``` nounPhraseSent $ phraseNP (horizontalVel `the_ofThe` firstObject) ```

`the_ofThe` will add "the" at the beginning, but the whole word is treated as an `NP`. The second "the" will be added when Drasil thinks "the whole word" is singular.

If we change the `the_ofThe` to `ofThe`, will remove the extra "the" from the definition. However, this will also remove "The" from the Table of Symbols.

JacquesCarette:
This code seems to be quite cursed: it really visibly adds `the` twice. Worse, it does its own capitalization. And it duplicates things too.

smiths:
[...]

PR #3237, opened 2023-01-14
`UnitalChunk` Constructor Redesign

### LINKED ISSUE #3199: Redesigning `UnitalChunk` constructors
There has been discussion throughout a few issues that seems to hinge on the constructors for `UnitalChunk`s; since they were so interconnected, I decided to list the issues I've noticed with the constructors and propose a new design that would better organize them. Please let me know what's good, what's bad, and what's missing!

_Note that I didn't address `ucuc` or `ucw` for this issue, since they are quite different and this issue is long enough._

# The Problems
## 1. Assuming the `Space` is `Real`

While a `UnitalChunk` is simply a `DefinedQuantityDict` with a `UnitDefn`, most `UnitalChunk` constructors create the `DefinedQuantityDict` internally instead of having it passed as a parameter. Of these, most of them assume the `Space` to be `Real`. 

https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Unital.hs#L30-L32

While (as far as I can tell) most quantities have a `Real` `Space`, I'm not sure if this should be assumed by the constructors for two main reasons.

1. It essentially doubles the number of required constructors: for each variant `UnitalChunk` constructor, it is likely that we will eventually need a variant that does NOT assume that the `Space` is `Real`. (This is actually the crux of https://github.com/JacquesCarette/Drasil/issues/3075, although I didn't quite understand what was actually happening.)
2. It is not consistently used: there are many instances where a constructor that assumes the `Space` is `Real` _could_ be used, but isn't. For example, the `UnitalChunk` constructor `ucs'` does not assume the `Space` to be `Real`, yet most of its calls use `Real` (all in `Data.Drasil.Quantities.Math`, 33/38 times in the unitals for GamePhysics, etc.) This is exacerbated when a `UnitalChunk` constructor that does not assume the `Space` is passed `Real` from a higher-level constructor.

@JacquesCarette also mentioned that assuming the `Space` to be `Real` is a hack in [this comment](https://github.com/JacquesCarette/Drasil/issues/1478#issuecomment-501735871), proposing we limit the scope of the `drasil-lang` constructors and have any "convenience" constructors be defined in `drasil-utils` to be "imported 'qualified' ... to indicate that ... assumptions are being used." For example, all quantities in `Data.Drasil.Quantities.Math` use the `Real` `Space`. I'm not sure if this should be used for cases like SWHS (which has exactly one `ConstrConcept` that does not have a `Space` of `Real`, shown below), but that seems like a different discussion.

https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs#L414-L419

## 2. Building from a `Concept c` or its parts
We have two constructors that build a `UnitalChunk` from a `Concept c`...
https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Unital.hs#L64-L67

and five that build from "a given 'UID', term, and definition".
https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Unital.hs#L74-L78

Since we have a similar number of instances of both approaches (by my count, 137 where the `Concept` is passed vs. 150 where the `Concept` is built), I'm not sure if we should remove the constructor(s) that build(s) the `Concept`. It might be good to force the user to create the `Concept` themselves to ensure it is done correctly; this would also allow us to remove separate `UnitalChunk` constructors for taking a `String` or a `Sentence` for the description (shown below). 
https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Unital.hs#L86-L96

Alternatively, we could also implement these extra constructors in `drasil-utils`. Since it is trivial to build a `Sentence` from a `String`, we can also just remove the `String`-based constructor.

This has some implications, like potentially making the code less readable, as discussed in the [wiki page for the `DefinedQuantityDict` chunk investigation](https://github.com/JacquesCarette/Drasil/wiki/DefinedQuantityDict#conceptchunks-made-only-to-be-used-in-a-dqd). Also, since `ucs` is used in `cuc'`, we would need to tweak it as well.

https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs#L125-L130

## 3. Missing constructors for `Symbol`s that depend on the `Stage`.

This is related to points 1 and 2 above, but we currently only have one `UnitalChunk` constructor for when the `Symbol` is dependent on the `Stage`. As shown below, this constructor is based on `uc'` and as such, assumes the `Space` is `Real` (point 1) and builds the `Concept` from the passed parameters (point 2).

https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Unital.hs#L80-L84

This is the core problem in https://github.com/JacquesCarette/Drasil/issues/3074, where again, I didn't really understand what was happening. I was seeing if there is was a way to split a `Concept` into its parts to be passed into this constructor; there was a bit of an afterthought of creating another constructor that took a `Concept`, but I didn't look into that too much.

## 4. Poor Naming
This is a common thread for all constructors. `uc` and `ucs'` seem to be paired up (both have `Concept`s being _passed_ where `ucs'` also takes a `Space`), as do `uc'` and `ucs` (both have `Concept`s being _built_ where `ucs` also takes a `Space`), which seems inconsistent to me. `makeUCWDS` is also the only constructor to not start with "`uc`", which is also inconsistent. `ucStaged` is also based on `uc'`, not on `uc` as one would expect.

# My Proposed Solution
The solution would address point 1 by not assuming the `Space` at all (although some helper functions that do could be created in `drasil-utils` if desired). The main two constructors would be `uc` and `ucStaged`. As per point 3, `ucStaged` is a variant to allow for `Symbol`s to depend on the `Stage`, and is based on `uc` to resolve point 4.
```haskell
-- | Used to create a 'UnitalChunk' from a 'Concept', 'Symbol', 'Space', and 'Unit'.
uc :: (Concept c, IsUnit u) => c -> Symbol -> Space -> u -> UnitalChunk
uc a sym space u = UC (dqd (cw a) sym space un) un
  where un = unitWrapper u

-- | Similar to 'uc', but 'Symbol' is dependent on the 'Stage'.
ucStaged :: (Concept c, IsUnit u) => c -> (Stage -> Symbol) -> Space -> u -> UnitalChunk
ucStaged a sym space u = UC (dqd' (cw a) sym space (Just un)) un
  where un = unitWrapper u
```

Depending on where we land on point 2, the following two constructors could also be included (either here or in `drasil-utils`). As mentioned above, these both only take a `Sentence` as the description (as opposed to a `String`), although this is also not set in stone. As before, neither assumes the `Space` to be `Real`, resolving point 1. `ucNoConcStaged` is a variant for a `Symbol` that depends on the `Stage`, as per point 3. Finally, they also have more descriptive names (a variant of `uc` that doesn't take a `Concept` -> "no `Concept`" -> `NoConc`) which resolves point 4.

```haskell
-- | Similar to 'uc', except it builds the 'Concept' portion of the 'UnitalChunk'
-- from a given 'UID', term, and definition (which are the first three arguments).
ucNoConc :: (IsUnit u) => String -> NP -> Sentence -> Symbol -> Space -> u -> UnitalChunk
ucNoConc i t d sym space u = UC (dqd (dccWDS i t d) sym space un) un
  where un = unitWrapper u

-- | Similar to 'ucNoConc', but 'Symbol' is dependent on the 'Stage'.
ucNoConcStaged :: (IsUnit u) => String -> NP -> Sentence -> (Stage -> Symbol) -> Space -> u -> UnitalChunk
ucNoConcStaged i t d sym space u = UC (dqd' (dccWDS i t d) sym space (Just un)) un
  where un = unitWrapper u
```

samm82:
This PR will close #3199 by redesigning the constructors for `UnitalChunk`s.

As part of this redesign, one instance of `dqd'` was replaced with `ucStaged` to test its behaviour; this is part of #2704.


samm82:
@JacquesCarette Just confirming that the conversion of all the `String` descriptions to `Sentence`s in [5e03d4a](https://github.com/JacquesCarette/Drasil/pull/3237/commits/5e03d4a5ad3a5bba77e4af52ce2b73e4dc0f918c) is OK? Should I be re-implementing these descriptions to use the relevant concepts and `Sentence` constructors or is that too much work for too little gain?

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
```diff
@@ -44,82 +44,82 @@ acceleration, angularAccel, angularDisplacement, angularVelocity, chgInVelocity,
...
-kEnergy              = uc CP.kEnergy  (Concat [cK, cE]) joule
-linearAccel          = uc CP.linAccel (Concat [vec lA, label "(", lT, label ")"]) accelU
-linearDisplacement   = uc CP.linDisp  (Concat [vec lU, label "(", lT, label ")"]) metre
-linearVelocity       = uc CP.linVelo  (Concat [vec lV, label "(", lT, label ")"]) velU
-momentOfInertia      = uc CP.momentOfInertia (vec cI) momtInertU
-chgMomentum          = uc CP.chgMomentum (Concat [cDelta,vec cP]) impulseU
-momentum             = uc CP.momentum (vec cP) impulseU
-moment               = uc CP.moment   (vec cM) torqueU
-moment2D             = uc CP.moment   cM       torqueU
+acceleration         = uc CP.acceleration (vec lA) (Vect Real) accelU
```
JacquesCarette, line 47: in general, when I have to edit these things, I then try hard to make things line up in columns. Makes the code nice for humans to read and see that it is really a "database"

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
```diff
@@ -44,82 +44,82 @@ acceleration, angularAccel, angularDisplacement, angularVelocity, chgInVelocity,
...
-scalarPos            = uc CP.scalarPos lP metre
-tension              = uc CP.tension (vec cT) newton
-time                 = uc CP.time lT second
-torque               = uc CP.torque (vec lTau) torqueU
-velocity             = ucs' CP.velocity (vec lV) (Vect Real) velU
-weight               = uc CP.weight cW newton
-fOfGravity           = uc CP.fOfGravity (sub (vec cF) (vec lG)) newton
+period               = uc CP.period cT Real second
+position             = uc CP.position (vec lP) Real metre
+positionVec          = uc CP.positionVec (vec lR) Real metre
```
JacquesCarette, line 76: So `positionVec` should probably be `Vec Real` and not `Real`! But doing that should be in a separate PR. Lots of things in here have `vec` for display and should probably have `Vec` type as well. But Drasil might not be ready for it, so it should be done incrementally.

JacquesCarette (changes requested):
This is a very good step forward. Only change I'd like to see before pulling it in would be the lining things up (and submission of a few issues). I'm quite sure this will keep evolving, and this goes in that direction.

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
samm82: Issue created: https://github.com/JacquesCarette/Drasil/issues/3246 - the fact that we could more easily see this seems to imply that not assuming the `Space` was a good idea!

samm82:
@JacquesCarette I lined things up in `Quantities.Physics` more nicely. Should reworking the `Sentence`s to use our constructors be a separate PR? I think doing so might overlap with https://github.com/JacquesCarette/Drasil/issues/3234, so it might be better to have a separate `sentenceReworking` branch for these changes

JacquesCarette (approved):

Issue #3159, opened 2022-11-30
Expanding generated README content


peter-michalski:
The following information is currently generated in each example's src README files:

- Example name
- Authors
- Making Examples
  - How to run and build the program
- Configuration Files
  - Lists required config files
- Version
- Blurb mentioning Drasil
- Copyright information and license link

Considering expanding what information is generated in these files, [Categorizing the Content of GitHub README Files](https://arxiv.org/pdf/1802.06997.pdf) suggested by @smiths lists some common README file information based on a study of GitHub repos. I have also reviewed some GitHub repos dedicated to sample README templates. I have combined the suggestions from the paper and the repos into one [file](https://gitlab.cas.mcmaster.ca/SEforSC/se4sc/-/blob/git-svn/GradStudents/Peter/Artifact-Data/Artifacts/README/README-content.xlsx). There is considerable overlap. The paper does break down frequency of sections and information quite nicely.

Some common README file information:

- What
  - introduction, project background, features, stack
- Why
  - advantages of the project, comparison with related work
- How 
  - getting started, how to run, installation, how update, configs, setup, reqs, depends, languages, platforms, demo, downloads, error and bugs, run tests
- When
  - project status, versions, project plans, road-map  
- Who 
  - project team, community, mailing lists, contact, acknowledgement, license, code of conduct
- References
  - API docs, get support, feedback, more info, translations, related projects
-  Contribution
  - contribution guidelines
- Other      
  - future features 
  - support project
  - FAQ

The remainder of this issue will add justification for why some of the information commonly found in README files should or  should not be generated with our examples. Once the updated README file content is decided, another issue can be created to investigate how to encode the information. Some of the information could already be specified in the examples.

smiths:
Thank you @peter-michalski.  Great summary.

During our last Drasil meeting (#3184) I mentioned that I was going through the generated code for the Projectile example.  I also noticed the absence of the "what" information.  In the [README file for Projectile](https://github.com/JacquesCarette/Drasil/blob/2e5c711c8aabe8c173b46c02b65f215494a5d5f1/code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/python/README.md) we never actually say what Projectile does.  

@peter-michalski, can you please look into what would be involved in adding "what" to our generated README files.  We have the "what" information in the SRS, but it is unclear if it is easily separated from the other information in the SRS.

[...]

PR #3248, opened 2023-01-28
Add scope, motivation, and software review information to generated software-READMEs

peter-michalski:
Contributes to #3159


peter-michalski:
A few outstanding items:

1.  `Goal` `Sentences` are capitalized and do not render nicely in the README. I'm wondering if `Sentences` in Drasil should instead be lowercase (in situations like `ConceptInstance` definitions) and a function could capitalize them near their final targets. This may make working with them easier.
2. I'm wondering if annotations like purpose, background, motivation, should become a chunk.

code/stable/dblpendulum/src/python/README.md
```diff
@@ -1,4 +1,6 @@
 # DblPendulum 
+> The program documented here is called DblPendulum.
```
smiths, line 2: This text seems strange in a `README` file.  The `README` file doesn't document DblPendulum.  I feel like the `README` file would just start with the program name, possibly as a title (like #DblPendulum)

smiths: You already have #DblPendulum.  Do we need the sentence at all?  If we need a sentence, I guess it might say something like "This program is named DblPendulum".  (Seems redundant though.)

smiths:
With respect to your potential issues @peter-michalski, my instinct is that sentences shouldn't be capitalized so that we can put them together in different combinations.  We can capitalize as needed.  This reminds me of the "extra the" issue (#3234).  We can reduce some problems by keeping the sentence information we codify simple.

smiths:
The change to the README file (to remove the part about "the program documented" here) looks good to me.

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -52,6 +52,8 @@ type CaseName = String
 type ExamplePurpose = String
 -- | Description of example
 type ExampleDescr = String
+-- | Motivation of example
+type ExampleMotivation = String
```
JacquesCarette, line 56: this information that is quite specific to the `README` has really gotten too large to fit in `Language.Drasil.Code.Imperative.GOOL.ClassInterface`. This should be split out into its own file.

It doesn't even belong to `Language.Drasil.Code.Imperative`, as it is quite independent of that. Finding a better home needs a design.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -49,6 +49,12 @@ data CodeSpec where
   purpose :: Purpose,
   -- | Example Background.
   background :: Background,
+  -- | Example Scope.
```
JacquesCarette, line 52: why are these all `Example X`? The data structure is a "Code Specification". It feels like there is a mismatch.

Futhermore, are we now requiring all this information for all codes?

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -104,6 +104,9 @@ si = SI {
   _authors = [naveen],
   _purpose = [],
   _background  = [],
+  _motivation  = [],
```
JacquesCarette, line 107: Reviewing multiple examples in a row that all set these 3 things to be empty. This really does start to feel like a hack. If we're going to implement a feature, shouldn't we be using it in all the examples? Just using it in Projectile does not convince me that this is a feature that is implemented at the right level.

JacquesCarette (changes requested):

samm82:
I wonder if this PR is related to #3430 at all (for example, in its capturing of motivation).

samm82:
As mentioned in the commit for the merge, there seems to be an issue rendering the purpose correctly; compare:
```markdown
> Purpose: efficiently and correctly predict whether a launched projectile hits its target

> Motivation: Projectile motion is a common problem in physics.
```
This is related to #3256

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
samm82: My guess is that these were meant to specify that it is the purpose, etc., of the example itself, but I think this is a bit redundant

JacquesCarette:
Yes, it is tangentially related to #3430 in that the larger topic is "what should be in an introduction?"

smiths:
I don't know if this is the best spot for this comment, but looking at the Purpose for Projectile I wonder if we should remove "efficiently and correctly"?  The purpose for Projectile is given as:

"efficiently and correctly predict whether a launched projectile hits its target"

Our SRS doesn't define what we mean by efficient.  Our code doesn't do anything (that I'm aware of) to be efficient.  Projectile is so simple that the notion of efficiency isn't really that relevant.  Even if it was a more complex problem, we really haven't been focusing on efficiency.  Our first goal for Drasil code has been correctness.

I don't think we need to specifically say our purpose is to "correctly predict".  It seems obvious that we wouldn't consider a prediction that was incorrect as valid.  Does it really need to be made explicit?

If we distill the purpose to its essential elements, we have:

"Predict whether a launched projectile hits its target"

JacquesCarette:
The question remains: are there useful things in this PR (not in the abstract - we know this goes in a positive direction - but in the concrete code) that ought to be adapted and merged in?

smiths:
I feel like the recent work that @hrzhuang has been doing with the "purpose" chunk of knowledge supersedes the work on this PR.  There are already conflicts with the README.md files, and I think the current (main branch) versions are closer to what we want.  Having said that, I'm just looking at stable.  There may be something in the Drasil code changes that is useful to us.  Someone else will have to comment on that.

JacquesCarette:
Indeed, given @hrzhuang 's recent work this might be moot. Ricky, could you take a serious look at the original issue and the PR and see?

hrzhuang:
I think this PR has a related goal, but is otherwise independent from my work with `purpose`.

Prior to my work, `purpose` already existed in `SystemInformation`, but it was not filled in for most of the examples. My work was extracting a purpose from the SRS, put it into the `SystemInformation`, then have the SRS pull that information from `SystemInformation`.

This PR introduces new fields for "motivation", "scope", and "software review" into `SystemInformation`, and they are mostly still blank except for the motivation of Projectile. So it is laying the groundwork similar to the introduction of `purpose` before I actually filled it in.

JacquesCarette:
So does it seem worth resurrecting, or is it not worth it in its current state (so we should close it)?

hrzhuang:
It doesn't seem too difficult to bring this up to date. I think it's really a matter of do we want the "motivation", "scope", and "software review" in `SystemInformation` and in the `README` files.

smiths:
I like the idea of having motivation and scope available for the README file, and possibly other artifacts besides the SRS.  Whether System Information is the best place for this information is a question for @JacquesCarette.  I'm not sure what "software review" refers to.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
balacij: We do have a form of this information that we write for the SRS documents, as we just recently did for #3777. It would be nice if we could share them with a chunk that defines what the purpose, background, and motivation of each case study, respectively, is.

This is also related to #3773. The specific purpose, background, and motivation of each case study are just instances of the generic concepts of purpose, background, and motivation. In other words, for our current design, these might just be unique `ConceptInstances` for each case study. Rather than having lists here, it might be that we just have a `UID` to the specific `ConceptInstance`.

balacij: Well, actually, would it be nice to share the same SRS discussion of the purp., back., and moti. in the generated READMEs? I'm not sure anymore. We might want a longer version of that in the SRS, and a shorter one here. @smiths do you have any thoughts on this? Do you think we can share them?

code/stable/projectile/projectile_u_p_l_b_wi_v_f/src/swift/README.md
```diff
@@ -1,7 +1,13 @@
...
 > Authors:  Samuel J. Crawford, Brooks MacLachlan, W. Spencer Smith
 
-> Purpose: efficiently and correctly predict whether a launched projectile hits its target.
+> Purpose: efficiently and correctly predict whether a launched projectile hits its target
+
+> Motivation: Projectile motion is a common problem in physics.
+
+------------------------------------------------------------
+## What 
+ Common examples of projectile motion include ballistics problems (missiles and bullets) and the flight of the balls in various sports (baseball, golf, football, etc.).
```
balacij, line 10: When I think of "motivation" related to software, I _think_ that I think (lol) strictly technical features:
1. What makes this software special?
2. What differentiates this software from other, similar, software?

However, this (the linked code) answers the question of "what is the motivation of the problem this software somehow solves?," which I think is more intended for a non-technical audience. So, I'm not quite sure who our audience is for this information (technical/non-technical/mix?).

If a software (assuming it were publicly available) is large/prominent enough such that it would have this kind of information, I imagine it would have a website, where this kind of marketing material would be found instead. Conversely, if the software isn't, then I imagine it would be somewhat self-evident to the audience of developers seeking out the software (they probably have a specific issue in mind, searching for a few specific keywords, and then looking at program/library documentation).

balacij:
In all honesty, I thought that "software **review**" was just a typo, and that "software revision" was intended.

Actually, it looks like it was intended to be "software review" and that "software review" means a comparison chart between a software (related to a README.md) and related projects. I think that "Comparison" is more appropriate, but I also don't really think this is something feasible for us to generate -- it would require an immense amount of knowledge of related popular software, and software qualities between our generated code and theirs (maintaining this does not sound feasible).

code/stable/projectile/projectile_u_p_l_b_wi_v_f/src/swift/README.md
balacij: Addendum to this after my latest comment below: the answers to (1) and (2) in a motivation section would be terse (e.g., X was too slow / A goes fast, Y is purely CLI-based / A creates a pretty GUI, Z is... / A has..., etc.), so it would not be as complete as the potential "Software Review"/Comparison Chart.

samm82: These could potentially be even more terse; 
> A solves various problems with the following alternatives: X is too slow, Y is purely CLI-based (while A creates a pretty GUI), .…

By formatting it this way, we have the flexibility to comment on A's functionality/qualities when necessary for completeness, but don't need to when it is redundant (although this completeness might be desired); a table may also be a nice way of formatting this data.

balacij: Yep, that's better. Though, I think the more complete, less marketing-like version, is reserved for the "Software Review" / Comparison Chart, but I agree with you. I think a good example of what we're talking about here is the Gitea project, where, in their [README.md](https://github.com/go-gitea/gitea), they give very quick highlights in the Purpose, and then have a dedicated page for [software comparison](https://docs.gitea.com/installation/comparison).

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
smiths: @balacij I think for every example we should be able (if we haven't already) to write a one sentence purpose.  This one sentence purpose would fit nicely in the README.  It should also be usable in the SRS. As @NoahCardoso recently did, we can reuse the purpose in the Introduction to the SRS.  Although we haven't done it yet, I could see us wanting to have a longer version of the purpose for the SRS.  For instance, our one sentence purpose could be something like: 

The software library provides a multi-domain Lattice Boltzmann solver.

(I'm working on a project with a colleague in mechanical engineering that could be described with that sentence.)  The problem with this one sentence purpose is that it introduces at least two concepts that not everyone will be familiar with.  This could be fine in a README file, but in the SRS, a description of what multi-domain means and what Lattice Boltzmann means would be in order.  Following #3773, we could say that the purpose is an idea definition that includes two other idea definitions.  In the README we might not generate the text for the definitions of the sub-ideas, but for the SRS, this would be a natural thing to do.

We could go through a similar conversation for Motivation and Scope.  I think Motivation would be fairly simple, but Scope does get more involved.  There are different views of Scope.  There is a simple "high level" view, and then there is the view that decomposes the Scope into Scope Decisions (things that are entirely left out) and Assumptions.  The Scope can also include typical values and examples.

code/stable/projectile/projectile_u_p_l_b_wi_v_f/src/swift/README.md
smiths: If the bar for the motivation is to do something better than the competition, than none of our case study projects have a motivation.  :smile:  Research and commercial projects should have the kind of motivation that distinguished the project from the "competition."  Although our software doesn't have this kind of motivation, since it doesn't do anything better than the competition, I still think we can motivate it.  Of course we have the motivation for our Drasil research project, but that isn't the motivation I'm thinking.  The motivation is for why have software to solve the kind of problems the software solves.  It isn't motivation for why the current software is the best, but rather the motivation for having software to solve the problem.  Projectile motion is an important physics problem.  A software tool to help with projectile motion will help physicists/engineers/scientists.

If we had enough examples, either real or brainstormed, we could identify the patterns in the Motivation statements and build a language for generating motivation statements.   For now, I think I'm fine with a non-parameterized statement that we make for each example.

balacij: You've convinced me :smile: 

> A software tool to help with projectile motion will help physicists/engineers/scientists.

Do you think we can replace the existing "What" section with your quote (or a variation of it)?

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
balacij: Maybe `CodeSpec` just needs to carry a reference to the `SystemInformation` it was constructed for? Right now, these fields are carried in both, and `CodeSpec`s are always a clone of `SystemInformation`s manually created ones.

code/stable/projectile/projectile_u_p_l_b_wi_v_f/src/swift/README.md
smiths: Yes, this does sound like a better "What".  I do think we also need examples, but probably not in the README file.  I think of the examples as more something that goes in the SRS to help define the scope.  In projectile motion if the scope is for velocities less than those for ballistics, we can use the example of the speed of the balls in various sports.  The example and the scope are related.  They are also related to assumptions and typical values.  I don't think this information belongs in the README; I just mention it as I sort out my thoughts on the matter.  :smile:

jackwyand:
@balacij the only thing I notice is that "software review" is mentioned here but not in #3950. However it seems like you ended up deciding not to implement this in the end.

Also some discussion about including this information in the SRS documents may want to be kept as it looks like it was something that you wanted to include, it might be worth making an issue about it?

Other than that the changes seem to align.

jackwyand:
>Also some discussion about including this information in the SRS documents may want to be kept as it looks like it was something that you wanted to include, it might be worth making an issue about it?

It looks like this was already implemented at some point, so with that this PR can be closed.

balacij:
Great! Can we delete the branch then?

PR #3250, opened 2023-01-31
Add flight duration to output values in Projectile example.

### LINKED ISSUE #3088: Should `t_flight` (of Projectile) be used "in code"?
Taking Java as an example, here is how the Projectile example uses `t_flight` (the theory):
https://github.com/JacquesCarette/Drasil/blob/b555167ce688040b766742b03442fbf7af750087/code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/java/Projectile/Projectile.java#L31-L56

`t_flight` is used as justification in the [derivation of `p_land`](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#IM:calOfLandingDist), but it just ends up being inlined rather than "used" in the code.

Is this the expected behaviour? If so, should `t_flight` still appear and be calculated in the code? As of right now, it seems to be calculated but the value is unused.

Of course, there is a greater discussion: we inlined it manually (I believe, and I think there is an open issue about this), but we also have created unused variables. Should we have not generated the code, but left it as an instance model that was just "inlined"?

balacij:
Closes #3088

I think this is all we need to do to have `t_flight` displayed.

It's a pretty nice `diff` to look at. One line of code "changed", or one output value designated, and all of this code is generated.


balacij:
Adding @smiths as reviewer because I'm not sure yet if this is how we should resolve #3088, or if we should just close 3088.

smiths (changes requested):
I really like how one change to the list of outputs had such a dramatic impact on the generated code.  Nice!  I do have a change to request though.  The generated code was changed (for the better), but now it doesn't match the requirements.  We should modify the functional requirement `Output-Values` in the SRS to include `t_flight`.  If our requirements were connected to our code generator, we could make this change in the requirements and see it propagate to the code, but right now generation of the SRS and of the code are independent.

samm82 (changes requested):
Another change to be made is to add this output to the Output Data Constraints table

Output Data Constraints

balacij:
So, I like both of these suggestions, thank you :smile: 

However, I feel that both of these things should be automatic. Reading the "FR:Output-*-Values" from the examples, they all are on the same mechanical wavelength. I don't think automating these sentences would be too bad since they look so similar. What do we think?

Regarding the physical constraint, would it just be that the flight duration would need to be a non-zero number? There's a few ways we can write out a constraint regarding this, but I'm not sure what it should be. Any suggestions? What exactly is a "physical constraint" here?

smiths:
Yes, the changes should be automatic.  There really is only one piece of information that triggers everything else.  Everything is a consequence of deciding that t_flight is an output.  The decision that t_flight is an output should be made in the requirements stage though, not as a list in `Body.hs`.  The list on `Body.hs` should come from the requirements.

I'm not sure how big a change it would be to more explicitly connect the SRS and the code generator.  Maybe we should close the current issue by manually fixing the SRS generator, and then create a separate issue to streamline how the decision to output a variable propagates?

With respect to physical constraints, those are constraints on variables that are a consequence of the physics.  In the current case, since time starts at 0, and time only goes forward in the simulation, the final time has to be greater than zero.  On your beam problem some example physical constraints are E> 0, I > 0, length of beam > 0, etc.

JacquesCarette:
I think we should merge this PR in regardless. But should open a new issue for getting the internal process to be better, i.e. for the output list to be driven by the requirements instead of being hand-written.

smiths (approved):

samm82:
@JacquesCarette Are you suggesting to fix these issues manually and then create an issue to automate that process, or create the issue to automate that process, and as a "test case" fix these issues automatically?

samm82:
Alrighty - @balacij you can re-request me for review once you've made our suggested changes manually. (Bonus points if you link to an issue for automating this in the future 😉)

samm82 (changes requested):
This output should also be reflected in the output requirement (which should likely be generated similarly to the input requirement):

output requirement of Projectile

samm82 (approved):
(Never mind; I remembered that this was going to be a separate task - #3259)

balacij:
I just realized I forgot to add the output value to the FR (manually), and added it just as you were reviewing, lol. The generated one will still be done in the future too, of course.

code/drasil-example/projectile/lib/Drasil/Projectile/Requirements.hs
```diff
@@ -39,8 +39,13 @@ calcValuesDesc = foldlSent [S "Calculate the following" +: plural value,
     ch offset    +:+ fromSource offsetIM,
     ch message   +:+ fromSource messageIM
   ]]
-outputValuesDesc = foldlSent [atStart output_, ch message,
-  fromSource messageIM `S.and_` ch offset, fromSource offsetIM]
+outputValuesDesc = atStart output_ +:+. outputs
+  where
+    outputs = foldlList Comma List $ map foldlSent_ [ 
+        [ch flightDur, fromSource messageIM],
```
samm82, line 45: `flightDur` doesn't come from `messageIM`

samm82 (changes requested):

samm82:
I just realized that I had been looking at the deployed version of Projectile and not the built one, so there's like two layers of coincidences here 😅

code/drasil-example/projectile/lib/Drasil/Projectile/Requirements.hs
balacij: :facepalm:  Thank you. I spent too much time fiddling with the sentence combinators that I rushed.

JacquesCarette:
Is this ready now? I won't be able to merge until @samm82 says that his requests have been done.

balacij:
I think everything is okay now. I addressed the previous review.

samm82 (approved):

JacquesCarette (approved):

Issue #2209, opened 2020-07-15
Switch to using GHC API to read Haskell files


Nathaniel-Hu:
The current version of the Data Table Generator (in /scripts/) uses a basic (albeit likely buggy) source code reader (SourceCodeReader.hs). This reader is used to extract the data, newtype and class types defined/used in all Haskell scripts stored in /code/. 

A more reliable option would be to switch over to using the GHC API to parse the Haskell scripts and extract the data, newtype and class types defined/used. This will most likely require some knowledge in using the [GHC API](https://hackage.haskell.org/package/ghc-8.8.3). Some additional resources include the modern toolkit [ghc-lib-parser (documentation)](http://hackage.haskell.org/package/ghc-lib-parser) and its [source code](https://github.com/digital-asset/ghc-lib).

Note: the current version of the Data Table Generator (in /scripts/) are currently contained in [this branch](https://github.com/JacquesCarette/Drasil/tree/Issue%23703_GenDataTable), and is awaiting merge approval in [this pull request](https://github.com/JacquesCarette/Drasil/pull/2208).

Ant13731:
Update for this issue: There doesn't really seem to be a lot of documentation on how to actually implement the ghc api, but this [stack overflow post](https://stackoverflow.com/questions/5396783/get-a-list-of-the-instances-in-a-type-class-in-haskell) might be a decent start. This [one](https://stackoverflow.com/questions/42727985/ghc-api-find-all-functions-and-their-types-in-scope) might be relevant too.

balacij:
I think @tingyuw is currently experiencing this issue with #3328.

While using GHC tooling to parse the data is feasible, there are other solutions too. For example, the scripts have been working thus far, so they could be properly analyzed for their issues. Alternatively, if/when we 'de-embed Drasil' (#3003), then this ticket _should_ be trivial to solve using our own native tooling.

balacij:
#2211 should also be completed as part of this work.

djarabek:
Another option: it should be possible to use the already generated `.hie` files combined with the [`hiedb`](https://hackage.haskell.org/package/hiedb) package instead. Using this approach, we also get fully resolved type data (the previous code operated without this).

PR #3328, opened 2023-04-03
Move CodeExpr to drasil-lang

### LINKED ISSUE #3307: CodeBlocks and package dependencies
In https://github.com/JacquesCarette/Drasil/commit/a6b2edef9c9bcd128796fd6848913bbafbf7bc4f#commitcomment-104797093 @tingyuw said:

> A new type CodeBlock is added in this commit. The code is not final, I'm testing but we are now able to generate code cells in the notebook after these changes (still need to address how the code will be printed). The reason why I use ModelExpr is because CodeExpr is defined in drasil-code-base, and due to package dependencies, we cannot use it in drasil-lang. What do you think would be the best approach to proceed this?

[This is better discussed in an issue than as a commit comment.]

As far as I can tell, the problem is that `drasil-code-base` contains two separate things
- `Language.Drasil.Chunk.*` which depends on drasil-database
- `Language.Drasil.Code.Expr` which does **not**.

At the moment, I think there are two routes forward that both make sense:
1. move `CodeExpr`'s defintion to `drasil-lang`
2. split `drasil-code-base` into two packages

Both of these would resolve the dependency problem. @balacij , since I think you're the last one to have committed code in `drasil-code-base`, opinions?

Personally, I think we should be looking hard at re-merging the 3 `Expr` types, so I think that all of them should be defined in the same spot. (I don't really mean 'merge', I mean to de-duplicate.)

tingyuw:
Closes #3307 

As discussed in #3307, this PR moves modules from `drasil-code-base` to `drasil-lang` and `drasil-code`.
@balacij Feel free to review the work if you'd like! Since you were the last person working on it, I thought you might want to take a look.


balacij:
I'm happy to see `drasil-code-base` go. The `-base` was really bothering me, haha. I think there is one more reference to `code-base` in the `Makefile` that you need to remove. Otherwise, it looks good to me :smile:

JacquesCarette (approved):
Sorry this took so long to review, it came in at an odd time, then got buried.

Good stuff.

JacquesCarette:
However, there is a build failure - can you fix please?

tingyuw:
I'm not sure how to fix it. @balacij didn't experience this issue when running it locally. It might be related to issue #2209.

balacij:
Are we able to add extra logging information around the `fromJust` to see where it's going awry?

image

JacquesCarette:
Not easily, no. Easiest way to deal with it is to never call `fromJust` but instead use `maybe` to force dealing with `Nothing` at the use site.

JacquesCarette (approved):

PR #3378, opened 2023-05-05
3209: Makes neglectCurv Assumption more Generic

### LINKED ISSUE #3209: Projectile: make neglectCurv assumption more generic
The neglectCurv assumption for Projectile explicitly mentions the Earth, but the physics of projectiles is the same for any celestial body.  Therefore, neglectCurv should be changed to:

"The distance is small enough that the curvature of the celestial body can be neglected."

harmanpreet-sagar:
fixes #3209 : Projectile: make neglectCurv assumption more generic. Changes have been made to stable code and Assumptions.hs of the Projectile example.


smiths (commented):
I've reviewed the changes to stable.  Looks good to me.

samm82 (approved):
Looks good to me!

JacquesCarette (approved):

Issue #3296, opened 2023-02-23
"Right-hand side" needs to be hyphenated


samm82:
The full version of "RHS" is "right-hand side", with a hyphen ([source](https://en.wikipedia.org/wiki/Sides_of_an_equation)), and likewise for "LHS" (if used).
![image](https://user-images.githubusercontent.com/35857611/220794982-7e0076d8-e8bf-4c0f-839d-966b76940d51.png)

Note that the capitalization should then be "Right-Hand Side", since "right" is a word on its own ([source](https://www.letpub.com/author_education_Title_Capitalization_and_You)).

Also note that this along with #3380 and #3381 are specific to SWHS but could come up in other examples as well.

smiths:
Thanks @samm82.  I like how picky you are!  Since our goal is to say everything once (and generate all other instances) we should make every effort to say things as best we can.  The acronym RHS is pretty common, so I think it is a good idea to use it.  Even if it isn't used much in SWHS, it will be used in other examples.

PR #3379, opened 2023-05-05
Hyphenate "right-hand side" and "left-hand side"

hrzhuang:
Addresses the first point in #3296


smiths:
Good point @hrzhuang.  We should still split the issue though, so that we can close the one that has been addressed, and keep the others open.

hrzhuang:
Right. I have created new issues #3380 and #3381. I also edited the original issue #3296 to be just the first point in the original list.

JacquesCarette (approved):

Issue #3189, opened 2022-12-19
Projectile Problem: Mix of 2D and 1D vectors in the same equation


smiths:
In the [general definitions for projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:GDs), we have an equation that mixes 2D and 1D vectors.  Specifically for GD:rectVel we have the equation:

![image](https://user-images.githubusercontent.com/1422000/208525026-de1df370-decf-4b3c-aadc-406942841f55.png)

On the left-hand side of the equation, we have a bold v, which suggests a 2D vector, but the right-hand side of the equation is for a 1D vector.  The equation shouldn't have a 2D vector, both sides of the equation should be 1D.  The equation is for a 1D problem (motion in a straight line).

@JacquesCarette I'll assign the issue to you since we (you, me and @balacij) were just talking about it, but you may want to reassign it to someone else.

JacquesCarette:
This is tricky, as I think what's going on is the the lhs uses the generic 'velocity' notion which is a vector, and the rhs implicitly assumes the situation is 1D. We might need to 'specialize' velocity (i.e. create a new chunk) that is 1D to get this to work properly.

Question for @smiths : should that be done locally in Projectile, or should 1D velocity be in our 'base' database of concepts?

smiths:
I vote for 1D velocity to be in our `base` database of concepts.  The concept of rectilinear motion comes up in many cases other than projectile motion.  Actually, any time we have a velocity vector, we still have 1D velocity, in the form of each of the components of the vector which are 1D velocities in the direction of the corresponding coordinate axis.

When I looked at what would be involved in modifying projectile to capture the [falling sack example](https://github.com/smiths/caseStudies/tree/master/CaseStudies/projectile/projectileLesson), I saw that the theories related to 1D rectilinear motion would be reused.  This further suggests that the concept should be in our base database of concepts.

If it is an issue with naming, we could call 1D velocity speed, instead of calling it velocity.  Speed acts in one dimension, with the sign specifying which direction (forward or reverse) in the dimension.  We usually think of the coordinate system as a Cartesian coordinate system, but this would work in a general curvilinear coordinate direction.

JacquesCarette:
That seems like a reasonable way forward (have `speed` as 1D `velocity` in our base concepts). We should have a way to know that it is a specialization, but we don't (yet).

smiths:
[...]

PR #3386, opened 2023-05-08
3211 show GD: rectPos depends on time

### LINKED ISSUE #3211: Projectile: GD:rectPos explicitly depends on time
In GD:rectPos the position should explicitly be shown to depend on time.  That is, the equation should show p(t), not just p.

harmanpreet-sagar:
Closes #3211: Projectile: GD:rectPos explicitly depends on time. Changes have been made to stable code, Body.hs and Unitals.hs of the Projectile example.


code/drasil-example/projectile/lib/Drasil/Projectile/Concepts.hs
```diff
@@ -31,15 +31,16 @@ projMotion = compoundNC projectile motion
...
 
-launcher, projectile, target, projSpeed :: ConceptChunk
+launcher, projectile, target, projSpeed, projPos :: ConceptChunk
 launcher   = dcc "launcher"   (nounPhraseSP "launcher")  ("where the projectile is launched from " ++
                                                           "and the device that does the launching")
 projectile = dcc "projectile" (nounPhraseSP "projectile") "the object to be launched at the target"
 target     = dcc "target"     (nounPhraseSP "target")     "where the projectile should be launched to"
 
 projSpeed  = dccWDS "projSpeed" (nounPhraseSP "1D speed") (S "1D speed under" +:+ phrase constant +:+ phrase acceleration)
+projPos  = dccWDS "projPos" (nounPhraseSP "1D position") (S "1D position under" +:+ phrase constant +:+ phrase speed)
```
samm82, line 43: For "database-like" entries like this, try to "align" the arguments/operators as best as possible (e.g., `projectile` and `target`).

samm82: Also, `getAcc oneD` should be used instead of `S "1D"` to make use of our concepts (including in `projSpeed`, which I know isn't part of this PR).

samm82 (changes requested):
In addition to my formatting comments, I'm a bit confused about whether these functions should be scalars or vectors. The "1D speed" function we have is formatted as a vector, even though its quantity (speed) is a scalar (as opposed to the velocity vector); is it a vector because it is a function? Or is this a mistake and it should be a scalar? @smiths 
v function

Depending on the answer to this, the 1D position function might need to be tweaked:
p function

smiths:
> In addition to my formatting comments, I'm a bit confused about whether these functions should be scalars or vectors. The "1D speed" function we have is formatted as a vector, even though its quantity (speed) is a scalar (as opposed to the velocity vector); is it a vector because it is a function? Or is this a mistake and it should be a scalar?

You are correct @samm82 -- the bold vector symbols are a mistake.  1D speed is a scalar, not a vector.  This problem has its own issue (#3189).  We shouldn't let it block the current fix.

smiths (commented):
I know there is some discussion about the implementation, but the actual generated changes to stable look fine.  I'm not sure why the order of the cross-references to GD:rectPos and GD:rectVel have swapped order, but that change is just cosmetic.  All the required information is still there.

code/drasil-example/projectile/lib/Drasil/Projectile/Concepts.hs
harmanpreet-sagar: Ok thanks! I will make those changes.

JacquesCarette (approved):
The code seems fine.

code/drasil-example/projectile/lib/Drasil/Projectile/Concepts.hs
```diff
@@ -31,15 +31,16 @@ projMotion = compoundNC projectile motion
...
-launcher, projectile, target, projSpeed :: ConceptChunk
+launcher, projectile, target, projSpeed, projPos :: ConceptChunk
 launcher   = dcc "launcher"   (nounPhraseSP "launcher")  ("where the projectile is launched from " ++
                                                           "and the device that does the launching")
 projectile = dcc "projectile" (nounPhraseSP "projectile") "the object to be launched at the target"
 target     = dcc "target"     (nounPhraseSP "target")     "where the projectile should be launched to"
 
-projSpeed  = dccWDS "projSpeed" (nounPhraseSP "1D speed") (S "1D speed under" +:+ phrase constant +:+ phrase acceleration)
+projSpeed  = dccWDS "projSpeed" (nounPhraseSP "1D speed")    (getAcc oneD +:+ S"speed under" +:+ phrase constant +:+ phrase acceleration)
+projPos    = dccWDS "projPos"   (nounPhraseSP "1D position") (getAcc oneD +:+ S"position under" +:+ phrase constant +:+ phrase speed)
```
samm82, line 43: A minor nitpick is that there should be a space after the `S` constructor. Other than that, this looks good!

samm82 (changes requested):

code/drasil-example/projectile/lib/Drasil/Projectile/Concepts.hs
samm82: Also, I just realized the concepts for `speed` and `position` should be used as well

harmanpreet-sagar:
Resolved conflicts and made formatting changes.

samm82 (approved):
Love to see it! ♥

JacquesCarette (approved):

PR #3388, opened 2023-05-08
Changed g to be scalar for ALL models

### LINKED ISSUE #3219: Projectile: Code: Change `g_vect` to `G`
The constant for the gravitational acceleration in the [Projectile code](https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/python/Projectile.py) is currently `g_vect`.  This name is a bit odd, since it is a symbolic constant that is not in ALL_CAPS and g is a component of acceleration, not a vector.  For this issue, we want to change the symbol from `g_vect` to `G`.  (`G` is also problematic since capital `G` is often used to represent Newton's gravitational constant, but this seems like a good compromise at the moment.)

Part of this issue is to record the steps required to make the change of the symbol.  The hope is that the code will only need to be touched in one spot.  This then will give us another simple example of the benefits of the Drasil approach.

A nice bonus would be if the units for `G` were generated as a comment in the code.

ManvendraJani:
closes #3219 
~~This changes g_vect to g everywhere it is used.~~
EDIT: This now implements a new scalar g


smiths:
Unfortunately this change fixes some things and breaks other things.  Is it possible to just update the symbol to g for the projectile example?  That would make things simpler.

In the double pendulum example, we sometimes mean the vector **g** and sometimes we mean the scalar g.  The vector **g** has two components, one for the acceleration in the x direction and one for the acceleration in the y direction.  This would usually be [0 -9.8].  We need this form in [Double pendulum](https://jacquescarette.github.io/Drasil/examples/dblpendulum/SRS/srs/DblPendulum_SRS.html) example for GD:yForce1 and GD:yForce2.  This is correct in the current version of Double Pendulum.  Accepting the PR will break this.  Unfortunately Double Pendulum is currently incorrect for IM:calOfAngularAcceleration1 where the acceleration vector is used, but the 1D scalar quantity should be used.  Accepting the PR will actually fix this mistake.

We need a way to distinguish between an acceleration vector and the 1D acceleration due to gravity.  We don't currently have this.  We have the same problem with other vectors, as discussed in #3189.

JacquesCarette:
FYI: I will wait until this is no longer a 'Draft' to review it.

ManvendraJani:
@smiths I introduced a new variable `gravitationalMagnitude`, this should fix both, the issue with projectiles and change it to be just the scalar quantity in IM:calOfAngularAcceleration1 and IM:calOfAngularAcceleration2. Does this work? 

Also, there may be an additional issue of better distinguishing between vectors and scalars within the documentation as it is not very clear especially in this case when you have both a vector and scalar represented by the same value. Instead of making them bold we could maybe have a line on top of them using something like /overline

JacquesCarette (approved):
The code-level changes seem fine to me. @smiths should approve the other changes before this gets merged in.

smiths:
@ManvendraJani I agree that the distinction between scalars and vectors needs to be clear.  When something is a vector we should say so in its type and in its description.  We don't really display types yet in Drasil, but the descriptions could be made to say vector.  With respect to overbars or arrows or underlines, that is a possible stylistic convention.  My preference is for bold, but eventually, the stylistic choice (typographic conventions) should be a variability that the user can select.  For now, let's leave vectors as bold.

smiths (approved):
Yes, the changes to stable look good.  We might in the future want to go further and more explicitly say when quantities are vectors, not just with bold face (or other typographic conventions), but in the description and with types.  These future changes are bigger picture stuff though.  The current changes fix the current issue.  Nice!

Issue #3391, opened 2023-05-09
Fill `purpose` fields of all `SystemInformation`s


hrzhuang:
The `purpose` field in `SystemInformation` is newly added and currently empty for all of our examples. #3390 has a discussion by @smiths explaining that the `purpose` information is mostly found elsewhere in the documents and that same information can be used to fill in the `purpose` information.

hrzhuang:
@smiths currently in our examples we have a "problem description." For example, for GlassBR this is "efficiently and correctly predict whether a glass slab can withstand a blast under given conditions." Should this be used as the `purpose`?

I'm also looking into #3218 where we want to use the problem description for the `\brief`. The problem here is that currently the problem description is only in the SRS and is not made available to the code generator, so it would be nice if it was included in `SystemInformation` somehow. Should this be a separate issue?

hrzhuang:
The problem description is a `Sentence`. I think it's called "problem description" because it's currently only used for the "problem description" section of the SRS.

smiths:
@hrzhuang would it be feasible to define the `purpose` in SystemInformation and then generate the "problem description" in the SRS from that?  As a first approximation, the `purpose` field could have identical text to the "problem description."  Unless the problem description example from GlassBR is atypical.  If the problem description is longer in the other examples, then we'll need to distill a purpose from the problem description.  The purpose should be one short phrase that captures what the software is trying to accomplish.  It is like a more abstract version of the goal statement(s).

hrzhuang:
Yes, I think that would work. All of the problem descriptions are short with the exception of the one for game physics.

Also, for future reference, what I've been referring to as "problem description" is more correctly the introduction to the "problem description" section of the SRS. The section contains more than just the short introduction.

smiths:
That sounds good @hrzhuang.  Let's start with the purpose statement being lifted from the problem description and then used to generate the problem description, and other things.  I think if we study the problem descriptions we'll find other nuggets of information that can be extracted and codified, and then used to generate the problem description.  That can be something we do on a subsequent pass though.  For now, we can focus on the purpose.

hrzhuang:
I submitted a pull request that lifts `purpose` from the existing introductions of the Problem Description sections. I mention this in the pull request but I will repeat here: Template and HGHC are excluded from this since they do not have a Problem Description section, and Game Physics is excluded since the introduction of its Problem Description section is too long. Should we create separate issues to fill in the `purpose` for these examples?

[...]

PR #3398, opened 2023-05-09
Fill `purpose` for examples other than Template, Game Physics, and HGHC

hrzhuang:
Contributes to #3391. `purpose` is lifted from the existing introductions of the Problem Description sections. Template and HGHC are excluded since they do not have such an introduction. Game Physics is excluded since its introduction is too long.

### DIFF
--- a/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
+++ b/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
@@ -123,7 +123,7 @@ si = SI {
 }
 
 purp :: Sentence
-purp = foldlSent_ [S "efficiently and correctly to predict the", phraseNP (motion `ofA` pendulum)]
+purp = foldlSent_ [S "efficiently and correctly predict the", phraseNP (motion `ofA` pendulum)]
 
 symbolsAll :: [QuantityDict]
 symbolsAll = symbols ++ scipyODESymbols ++ osloSymbols ++ apacheODESymbols ++ odeintSymbols 
--- a/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
+++ b/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
@@ -269,6 +269,7 @@ userCharacteristicsIntro = foldlSP
 -------------------------------
 -- 4.1 : System Constraints  --
 -------------------------------
+-- Introduction of the Problem Description section derived from purp
 
 ---------------------------------
 -- 4.1.1 Terminology and Definitions --
--- a/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
+++ b/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
@@ -75,7 +75,7 @@ mkSRS = [TableOfContents, -- This creates the Table of Contents
       SystCons [] []],                            
   SSDSec $ 
     SSDProg
-      [ SSDProblem $ PDProg prob []                --  This adds a is used to define the problem your system will solve
+      [ SSDProblem $ PDProg purp []                --  This adds a is used to define the problem your system will solve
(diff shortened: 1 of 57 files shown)


code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs
```diff
@@ -122,6 +122,9 @@ si = SI {
    refdb       = refDB
 }
 
+purp :: Sentence
+purp = foldlSent_ [S "efficiently and correctly to predict the", phraseNP (motion `ofA` pendulum)]
```
samm82, line 126: "to" should be removed here

```suggestion
purp = foldlSent_ [S "efficiently and correctly predict the", phraseNP (motion `ofA` pendulum)]
```

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -118,6 +117,11 @@ si = SI {
   _usedinfodb = usedDB,
    refdb = refDB}
 
+purp :: Sentence
+purp = foldlSent_ [S "provide a model of a", phrase pidC,
+         S "that can be used for the tuning of the gain constants before",
+         S "the deployment of the controller"]
```
samm82, line 123: Can any of these words be used from concepts (e.g., "constants", "controller")?

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -122,6 +122,10 @@ si = SI {
    refdb       = refDB
 }
 
+purp :: Sentence
+purp = foldlSent_ [S "efficiently" `S.and_` S "correctly predict whether a launched",
+  phrase projectile, S"hits its", phrase target]
```
samm82, line 127: There should be a space between `S` and `"`
```suggestion
  phrase projectile, S "hits its", phrase target]
```

samm82 (changes requested):
Other than these minor tweaks, this looks good to me!

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
hrzhuang: "Constant" and "controller" are not currently defined as concepts in `drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs` and are used all over the SRS simply in strings. Should this be an issue?

samm82: Potentially? I think there are a lot of concepts that we just use as strings, so a more thorough pass of all our examples to "pull out" these concepts could be beneficial, but is also pretty low on the priority because of the effort required for a relatively small amount of benefit. You could make a general issue (tagged as <kbd>[newcomers](https://github.com/JacquesCarette/Drasil/labels/newcomers)</kbd>
 and <kbd>[Low Priority](https://github.com/JacquesCarette/Drasil/labels/Low%20Priority)</kbd>) for doing this so we don't forget, but we would also need to decide on the right time to do this (e.g., when we confident that our current examples won't be extended to include more terms).

samm82 (approved):

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
hrzhuang: I see. I opened issue  #3402 for this. I will mark this conversation as resolved now.

JacquesCarette (approved):
FYI: it actually would have been better to have done one PR per example. Many small PRs are better than larger ones.

Main reason: one issue in one part can hold up a big PR from being merged. If it's many, then most will get merged right away, and only the problematic part will be held up.

JacquesCarette:
(Should I delete this branch, or will you re-use it for the other examples, or are you done for now?)

hrzhuang:
Noted, thanks for letting me know! I will delete this branch now and create separate branches for the outstanding examples.

PR #3401, opened 2023-05-10
Debug tool: dump all chunks (by UID)

balacij:
Builds on #3413 

Contributes to #2873 (after which we should upgrade this code) by allowing us to _start_ analyzing our chunk maps, for the 'conflicting' chunk UIDs.

With this PR, we create a `.drasil` folder for each of the examples, which we expect Drasil maintainers to use for.... maintaining Drasil! :smile: 

There are some changes to the `designLog.txt` files, but I have no idea why they are appearing. I was not touching anything related to the design of the generated code. I suppose this might be because I used a strict/eagerly loaded map, which _may have somehow affected the ChunkDB used afterwards (?)_. @JacquesCarette, do you have any thoughts on where this might have come up?


balacij:
After we have chunk generation via `State`, I think we can also dump the 'fruits.json' file which also includes the 'final' chunkdb.

Potential subsequent PR:
1. (almost certainly) Finding which UIDs appear more than once (listing those from which)
2. Finding the 'fruits.json' and some basic stats between the number of 'seeds' and the number of 'fruits'

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -5,28 +5,7 @@
...
-  termResolve, defResolve, symbResolve,
-  traceLookup, refbyLookup,
-  datadefnLookup, insmodelLookup, gendefLookup, theoryModelLookup,
-  conceptinsLookup, sectionLookup, labelledconLookup, refResolve,
-  -- ** Lenses
-  unitTable, traceTable, refbyTable,
-  dataDefnTable, insmodelTable, gendefTable, theoryModelTable,
-  conceptinsTable, sectionTable, labelledcontentTable, refTable
-  ) where
+module Database.Drasil.ChunkDB where
```
samm82, line 9: What's the rationale behind removing list of what is exported? If pretty much everything is exported and there's nothing that should be explicitly hidden, then this seems fine to me lol

code/drasil-example/projectile/app/Main.hs
```diff
@@ -12,11 +14,14 @@ import qualified Drasil.Projectile.Lesson.Body as PL (nb, printSetting)
...
   setLocaleEncoding utf8
+  dumpEverything fullSI ".drasil/"
   typeCheckSI fullSI
   gen (DocSpec (docChoices SRS [HTML, TeX, JSON]) "Projectile_SRS") srs printSetting
   gen (DocSpec (docChoices Jupyter [])      "Projectile Motion Lesson") PL.nb PL.printSetting
   genCodeWithChoices choiceCombos
   genDot fullSI
   genLog fullSI printSetting
+  -- if the chunkDB had a mutable state, then this would make more sense.
+  -- dumpChunkDB fullSI ".drasil/" "fruits.json"
```
samm82, line 25: Not sure what this means lol

code/drasil-gen/lib/Language/Drasil/Dump.hs
```diff
@@ -22,5 +22,5 @@ dumpChunkDB :: DB.ChunkDB -> String -> IO ()
 dumpChunkDB chks targetFile = do
     let dumpped = DB.dumpChunkDB chks
     trg <- openFile targetFile WriteMode
-    LB.hPutStrLn trg $ encode $ dumpped
+    LB.hPutStrLn trg $ encode dumpped
```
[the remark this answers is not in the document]
balacij: Typo :smile: Already fixed now.

code/drasil-example/projectile/app/Main.hs
balacij: I think of the user-inputted information as the `seeds`, the Drasil-provided features as the `soil`, and the final tree of things automatically generated from the `seeds` as the `fruits`. So, if we had more gradual generation, we can take snapshots (breakpoints) of the `ChunkDB` at specific spots during generation.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
JacquesCarette: They are indeed tedious to maintain, but I think they are a solid piece of documentation.

Very long lists of exports are a code smell (exception: drasil-data and utilities).

code/drasil-example/projectile/app/Main.hs
JacquesCarette: Nice!

code/stable/dblpendulum/.drasil/inverted_seeds.json
```diff
@@ -0,0 +1 @@
...references"],"verification":["terms"],"verifyInptVals":["references","concepts","conceptInstances"],"vertical":["terms"],"verticalMotion":["terms"],"verticalacceleration":["terms"],"verticalforce":["terms"],"verticalposition":["terms"],"verticalvelocity":["terms"],"video":["terms"],"videogame":["terms"],"volume":["terms"],"w_1":["terms","symbols"],"w_2":["terms","symbols"],"weight":["terms"],"xAxis":["terms"],"xComp":["terms"],"xConstAccel":["terms"],"xCoord":["terms"],"xDir":["terms"],"xDist":["terms"],"xForce1":["references","labelledContent","generalDefinitions"],"xForce2":["references","labelledContent","generalDefinitions"],"xPos":["terms"],"xScalAcc":["terms"],"xVel":["terms"],"x_numpy":["symbols"],"y0_apache":["symbols"],"yAxis":["terms"],"yAxisDir":["references","concepts","conceptInstances"],"yComp":["terms"],"yConstAccel":["terms"],"yCoord":["terms"],"yDir":["terms"],"yDist":["terms"],"yForce1":["references","labelledContent","generalDefinitions"],"yForce2":["references","labelledContent","generalDefinitions"],"yPos":["terms"],"yScalAcc":["terms"],"yVel":["terms"],"y_ode":["symbols"],"year":["terms"],"zAxis":["terms"],"zComp":["terms"],"zCoord":["terms"],"zDir":["terms"]}
```
JacquesCarette, line 1: Could you print these one-per-line instead of as 1 long line? It will make diffs way easier to read!

code/stable/glassbr/src/cpp/designLog.txt
```diff
@@ -19,6 +19,7 @@ Successfully matched Real with Double.
 Successfully matched Boolean with Boolean.
 Successfully matched Rational with Double.
 Successfully matched Integer with Integer.
+Successfully matched Function (String :| [Real,Real]) Real with Func [String] Double.
```
JacquesCarette, line 22: This seems like an odd match? These don't seem like comparable types.

code/stable/nopcm/src/cpp/designLog.txt
```diff
@@ -19,11 +19,9 @@ ODE Library package selected as odeint.
 Successfully matched String with String.
 Successfully matched Actor "Constants" with Object "Constants".
 Successfully matched Actor "InputParameters" with Object "InputParameters".
-Successfully matched Rational with Double.
```
JacquesCarette, line 22: Any idea why 2 of these lines disappeared?

code/stable/nopcm/src/csharp/designLog.txt
```diff
@@ -17,12 +17,10 @@ ODE Library package selected as OSLO.
 Successfully matched String with String.
 Successfully matched Actor "Constants" with Object "Constants".
 Successfully matched Actor "InputParameters" with Object "InputParameters".
-Successfully matched Rational with Double.
 Successfully matched Real with Double.
-Successfully matched Vect Rational with List Double.
+Successfully matched Vect Real with List Double.
```
JacquesCarette, line 21: This, on the other hand, is an improvement!

JacquesCarette (changes requested):

code/stable/glassbr/src/cpp/designLog.txt
balacij: #3411 was a previous issue, `make stabilize` inadvertently let us find it!

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
balacij: Reverted in 934dfdd

balacij:
There are 2 kinds of SIs for each project. The original one with manually gathered chunks, and the copy that is manually injected with some chunks. I chose to dump only the second kind for now because it will show more of the "problematic" chunks (unless there are more downstream in generation). Since I don't think we'll be building too much tooling around the existing CDB, I don't think we need to go for completeness. Is that okay?

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -4,12 +4,11 @@
 -- Changes to ChunkDB should be reflected in the 'Creating Your Project 
 -- in Drasil' tutorial found on the wiki:
 -- https://github.com/JacquesCarette/Drasil/wiki/Creating-Your-Project-in-Drasil
-
 module Database.Drasil.ChunkDB (
   -- * Types
   -- ** 'ChunkDB'
   -- | Main database type
-  ChunkDB(CDB, symbolTable, termTable, defTable),
```
JacquesCarette, line 12: I know these are a pain to maintain, but it is much better to be explicit about what is on the public interface of this module.

code/drasil-gen/lib/Language/Drasil/Dump.hs
```diff
@@ -0,0 +1,35 @@
...
+
+type Path = String
+type TargetFile = String
+
+dumpEverything :: SystemInformation -> Path -> IO ()
+dumpEverything si targetPath = do
+  createDirectoryIfMissing True targetPath
+  let chunkDump = DB.dumpChunkDB $ _sysinfodb si
+  let invertedChunkDump = invert chunkDump
+  let sharedUIDs = SM.filter (\x -> length x > 1) invertedChunkDump
```
JacquesCarette, line 24: This is actually going to force `invertedChunkDump` to be fully computed, and then lots thrown away, because of `length` being strict.

```haskell
longerThan1 :: [a] -> Bool
longerThan2 (_ : _ : _) = True
longerThan2 _ = False
```
will only force the first 2 elements (if they exist) and not the rest.

code/stable/dblpendulum/.drasil/inverted_seeds.json
```diff
@@ -0,0 +1,2050 @@
+{
```
JacquesCarette, line 1: Do we really want to save all this "debug" information in stable?

JacquesCarette (changes requested):

code/drasil-gen/lib/Language/Drasil/Dump.hs
balacij: Ah, good catch! Thank you 79dbe4b960924b852503ef4a972cc29526fc6bcb

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
balacij: No worries! I used them all in the `drasil-database` package, but you made me realize that I should've been using the lenses instead of the raw record accessors. I fixed this in 6a6a3532809465ff01cae91c6453522235d777c3

code/stable/dblpendulum/.drasil/inverted_seeds.json
balacij: I definitely wouldn't want to stuff every possible artifact in `stable/`, but I think this particular kind might be okay since all Drasil projects will share the same debugging mechanism and because it will allow us to debug changes to the background knowledge as the projects are changed, without looking at the Haskell syntax as often.

That being said, I don't hold a strong opinion about this, and I am definitely open to making it configurable at some level other than Haskell.

JacquesCarette: Let's experiment with it turned on and see how much churn that generates. I am concerned...

PR #3409, opened 2023-05-10
removes comma from Data Constraints descritption

### LINKED ISSUE #3087: Extra Comma in Projectile Data Constraints
The highlighted comma shown below should be removed, since the independent clause comes before the dependent clause (if they were reversed, the comma would be necessary).

image

ManvendraJani:
closes #3087
Does fix for all examples, as mentioned in issue.


code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -199,7 +199,7 @@ uncertSent = foldlSent [atStartNP (the uncertainty), phrase column,
 
 -- | General 'Sentence' that describes some conservative constraints on the model.
 conservConsSent :: Sentence
-conservConsSent = foldlSent [atStartNP' (the constraint) `S.are` S "conservative" `sC`
+conservConsSent = foldlSent [atStartNP' (the constraint) `S.are` S "conservative" +:+
```
samm82, line 202: A change in one place that fixes all other examples? That's the power of Drasil 😎

code/stable/glassbr/src/cpp/designLog.txt
```diff
@@ -19,6 +19,7 @@ Successfully matched Real with Double.
 Successfully matched Boolean with Boolean.
 Successfully matched Rational with Double.
 Successfully matched Integer with Integer.
+Successfully matched Function (String :| [Real,Real]) Real with Func [String] Double.
```
samm82, line 22: Why is this change happening here? Is something broken in `master`? @JacquesCarette @balacij

samm82 (changes requested):

code/stable/glassbr/src/cpp/designLog.txt
balacij: Yep, irrelevant to this work. It seems to have happened earlier but never caught because `diff` checking ignores `.txt` files, so `stabilize` was never executed on something that changed.

balacij: #3411

smiths (commented):

JacquesCarette (approved):

code/stable/glassbr/src/cpp/designLog.txt
JacquesCarette: So this can be ignored because it is being fixed elsewhere.

samm82: Oops! Forgot to re-review

samm82 (approved):

PR #3413, opened 2023-05-10
Add '*.txt' files back into diff checking

### LINKED ISSUE #3411: Difference between `make stabilize` and `make test`
`make test` ignores `.txt` files in the `diff` checks, which is why we now have (surprising) changes in current PRs.

To solve this, we either need to ignore `.txt` files in `make stabilize` or consider `.txt` files in `make test` (which I think we should do). If the latter, we should only need to remove the `-x '*.txt'` from the Makefile's defined `DIFF`.

balacij:
Closes #3411 

This should avoid future issues related to #3411 as `.txt` files will now be diff'd against `stable/`.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -103,7 +103,7 @@ ALL_EXPANDED_TARGETS = $(foreach P, $(PACKAGE_GEN_TARGET), $($(P)_PACKAGES)) $(f
 #-------------------------------#
 
 # make variables/configuration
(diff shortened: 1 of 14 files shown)


balacij:
The subsequent issues related to these changes will need to be investigated, but we shouldn't block other PRs in the meanwhile.

JacquesCarette (approved):

PR #3416, opened 2023-05-11
Added specification of 1D in Detail Derivation

### LINKED ISSUE #3212: Projectile: Clarification in Detailed Derivation for GD:rectVel and GD:rectPos
In the last sentence of the detailed derivations for GD:rectVel and GD:rectPos, TM:acceleration and TM:velocity are invoked.  The current derivation doesn't explain how we go from 3D equations to 1D equations.  We should explicitly mention this, by changing the text to:

"From TM: acceleration _in 1D_, and using the above symbols ..."

(The emphasis (italics) added above is to show the difference from the original.  The emphasis  won't be part of the new derivation.)

ManvendraJani:
closes #3212


code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs
```diff
@@ -120,7 +120,7 @@ rectDeriv c1 c2 motSent initc ctm = foldlSent_ [
   (atStartNP (the c1) `S.is` getScalar c1 `S.andThe` phrase c2 `S.is` getScalar c2 !.), motSent,
   atStartNP (the initc), sParen (S "at" +:+ eS (sy time $= exactDbl 0) `sC` S "from" +:+
   refS timeStartZero) `S.is` S "represented by" +:+. getScalar initc,
-  S "From", refS ctm `sC` S "using the above", plural symbol_ +: S "we have"]
+  S "From", refS ctm +:+ S "in 1D" `sC` S "and using the above", plural symbol_ +: S "we have"]
```
samm82, line 123: This should use the 1D concept we have (I think it's called `oneD`).

samm82 (changes requested):

code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs
```diff
@@ -120,7 +120,7 @@ rectDeriv c1 c2 motSent initc ctm = foldlSent_ [
   (atStartNP (the c1) `S.is` getScalar c1 `S.andThe` phrase c2 `S.is` getScalar c2 !.), motSent,
   atStartNP (the initc), sParen (S "at" +:+ eS (sy time $= exactDbl 0) `sC` S "from" +:+
   refS timeStartZero) `S.is` S "represented by" +:+. getScalar initc,
-  S "From", refS ctm `sC` S "using the above", plural symbol_ +: S "we have"]
+  S "From", refS ctm `S.in_` S (abrv oneD) `sC` S "and using the above", plural symbol_ +: S "we have"]
```
samm82, line 123: I believe `short` can replace `S . abrv`, but I'm not exactly sure offhand.

samm82 (changes requested):

PR #3432, opened 2023-05-17
Changed a v t to be time dependent

### LINKED ISSUE #3210: Projectile: Update TM:acceleration and TM:velocity - add explicit dependence on time and vector information
In the Projectile example, TM:acceleration and TM:velocity do not explicitly show that the acceleration and velocity depend on time.  The equations should be changed to include t:

**a**(t) = d**v**(t)/dt and **v**(t) = d**p**(t)/dt

The Description field of TM:acceleration and TM:velocity should also explicitly say that **a**, **v** and **p** are vectors.  For instance, instead of saying **a** is the acceleration, it should say that **a** is the acceleration vector.

ManvendraJani:
closes #3210 
Changed `acceleration`, `velocity`, and `position` to be time dependent.
Changed `linearAccel`, `linearVelocity`, and `linearDisplacement` to be scalars
Changed `projSpeed` to be a scalar


samm82 (approved):

balacij:
So, this solution relies on forcing “(t)” to appear in the symbol. There are 2 issues with this approach:
1. we're using "lT" instead of referencing the symbol of "time" -- what happens if we change "t" from "time" to "s"?
2. we're claiming that "(t)" is part of the symbol of these variables, but I don't think we should be doing that. I think an alternative solution would be to take thee derivative with respect to the function application expression to show the explicit connection.

What do you think @samm82 ?

smiths:
I think @balacij raises some good points.

samm82:
I agree, but I think that should be a separate issue, since this is how we have been dealing with functions in Drasil right now. I know we can represent functions using `Expr`, but I'm not sure if this is available at the symbol level. @balacij?

balacij:
This is oddly delicate. I'm not sure what the solution is. It would be helpful to have a complete understanding of where the symbols shown in the table of symbols should be coming from. I don't think that our current approach of "everything from the chunk database" is maintainable for larger projects. The issue here seems that the table of symbols ignores the information gathered from the instance models. Would it be good to limit the table of symbols to only those immediately relevant to the _solution_ schematic proposed by the `InstanceModel`s, and then have the other symbols locally defined as needed? This would allow us to (a) decrease the size of the table of symbols for larger projects, (b) tie them back to the `InstanceModel`s that involve them. Regarding (b), 
I vaguely imagine that `EquationalModel`s can be used to tie the function symbols nicely together. If it uses a locally defined symbol, then it might get a bit ugly with non-unique symbols or those strictly local to that definition, but perhaps we can deal with it by placing a restriction on `InstanceModel`s.

I think this is in a similar vein of thought as we've been having in CAS 741, @samm82 @smiths.

Also, I apologize for the late reply.

balacij:
Assuming what we need to do is exactly re-investigating where we're getting the symbols from (for the table of symbols in the SRS), then we should probably open a new ticket and continue discussion there -- I'll just need someone else to vet this idea.

The general idea might also apply to other things related to the `ChunkDB` -- should transformers _ever_ be accessing a _list of chunks of a particular type_? I completely understand it referencing _specific_ chunks by their UID, but it seems a bit dubious to ever have transformers that refer to a whole type of chunks (this would also mean that our effort in #3481 and #3482 might be for waste).

balacij:
Funny enough, I think this is one of those cases where an "easy" ticket shows that "easy" can be misleading.

To be clear, @ManvendraJani, I am not criticizing your work. As @samm82 mentioned, this PR is perfectly fine. We might just have stumbled upon a systemic issue.

smiths:
Does Drasil know that *a* is a function?  Does it know that the type of *a* is R -> R?  If it knows that, then we can give it the symbol that it should display for the independent variable.  In this case the symbol is *t*, but it could be something else.  If we know *a* is a function, then we don't even have to always display it as a(t).  The main thing is that we capture the knowledge.  How we display it is a style choice.

I agree with the discussion above from @balacij - the way we build our table of symbols is not sustainable.  I like the completeness of the approach that every symbol in the document is unique and that every symbol is displayed in the table of symbols, but as the projects get bigger it will be difficult to have a unique and idiomatic symbol for everything.  

Displaying symbols that are used very little, or used in the base theory, but not in the final theory (instance models) could lead to confusion.  As an illustrative example, we can think about the case of refining the fully 3D time dependent Temperature T, where T is a function of x, y, z and t; that is, T(x,y,z,t) appears in the base theory (theoretical model).  As we refine the theory we use assumptions that T does not depend on space, until we get to T(t).  Our problem only depends on time, but in the current approach of everything going to the table of symbols, we would have to define x, y and z in the global table of symbols.  Somebody reading the table could easily get confused.  

Another example is if we use a generic theory and apply it to a specific domain.  The generic theory uses generic symbols, but the specific domain uses domain specific symbols.  Will displaying the generic symbols in the T of S lead to confusion?  As an example, in the conservation of thermal energy equation we use Gauss's Divergence Theorem to convert a volume integral to a surface integral.  We don't currently document Gauss's Divergence Theorem, but if we did, we likely want to document it using generic symbols.  We'd use the kind of symbols you see in a math textbook.  We'd want the generic math theory to be phrased in a way that we can reuse it.  We wouldn't use temperature for the generic definition.  We would like to do a symbol transformation of the generic theory into the domain specific symbols.

I think that eventually we'll need more flexibility on the table of symbols.  The SRS document designer will specify the design choice that they would like to use.  There might be a global table, or maybe there would be several tables, with separate tables for the base theories and the final theories (instance models).

JacquesCarette:
The fundamental problem is that Drasil knows remarkably little about functions. Its knowledge of mathematics is very much about "ground expressions", i.e. values in places like the Reals. Even vectors are sub-optimal right now.

PR #3436, opened 2023-05-18
3427 Create SRS documents with empty IMs, TMs, and DDs

### LINKED ISSUE #3427: `drasil-docLang` should allow us to create SRS documents with empty TM, DD, and IM sections, similar to GDs
To make developing Drasil-based projects more immediately gratifying when starting from fresh, we should allow users to create empty SRS documents (at the documentation level, not the 'valid SRS' level). This should involve us removing some `error` cases in `drasil-docLang` and creating additional boilerplate text.

To make my beam bending project, [I hacked in a minimal viable solution for this](https://github.com/JacquesCarette/Drasil/compare/master...balacij:Drasil:beamBending?expand=1#diff-595ea8ad584d7c5d3bb228edd5ac885b9f7069b8abb0840bceb6e163668fb43b) but we should do this properly now too.

Contributes to #3326, #3356

harmanpreet-sagar:
Closes #3427: `drasil-docLang` should allow us to create SRS documents with empty TM, DD, and IM sections, similar to GDs. Changes made to SpecificSystemDescription.hs and DocumentLanguage.hs to create blank SRS sections with statements indicating no content instead of creating an error.


harmanpreet-sagar:
I was able to test out my functions for creating blank IM and TM sections in the present SRS document but was unsure of how to test the DD section. When I tried modifying the code for the various examples to display a blank DD despite having DDs, they break and stop generating the HTML and the other SRS docs right before the DD section. Nothing after the DD section gets generated if I try to test it out that way. I am unsure as to which other methods I could use to test this out.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -361,10 +361,10 @@ mkSolChSpec si (SCSProg l) =
     map (mkSubSCS si) l
   where
     mkSubSCS :: SystemInformation -> SCSSub -> Section
-    mkSubSCS _ (TMs _ _ [])      = error "There are no Theoretical Models"
+    mkSubSCS _ (TMs _ _ [])      = SSD.thModF Doc.section_ []
```
balacij, line 364: What is the empty list for in the `SSD.thModF` call?

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -119,13 +119,22 @@ assumpF otherContents = SRS.assumpt (assumpIntro : otherContents) []
 
 -- | Wrapper for 'thModelIntro'. Takes the program name and other 'Contents'.
 thModF :: (Idea a) => a -> [Contents] -> Section
-thModF progName otherContents = SRS.thModel (thModIntro progName : otherContents) []
+thModF progName otherContents = if not (null otherContents) then
```
balacij, line 122: You can use pattern matching on `otherContents` instead of this if-then-else.

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -166,6 +193,9 @@ inModelIntro r1 r2 r3 r4 = foldlSP [S "This", phrase section_,
   pluralNP $ symbol_ `inThePP` model, S "identified in", namedRef r3 (plural thModel) `S.and_`
   namedRef r4 (plural genDefn)]
 
+inModelIntroNoContent :: Section -> Section -> Section -> Section -> Contents
+inModelIntroNoContent _ _ _ _ = mkParagraph $ S "There are no instance models."
```
balacij, line 197: What are all the ignored function parameters for?

balacij (commented):
Nice!

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
harmanpreet-sagar: It is the empty list that would have otherwise included contents. I added that list because `SSD.thModF` takes in two parameters, an `Idea a` and a list of contents.

```
thModF :: (Idea a) => a -> [Contents] -> Section
thModF progName otherContents = if not (null otherContents) then
                                  SRS.thModel 
                                    (thModIntro progName : otherContents) []
                                else
                                  SRS.thModel 
                                    (thModIntroNoContent progName : 
                                    otherContents) []
```

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: Since `inModelF` has the type `Section -> Section -> Section -> Section -> [Contents] -> Section` as can be witnessed here:
```
inModelF :: Section -> Section -> Section -> Section -> [Contents] -> Section
inModelF probDes datDef theMod genDef otherContents = if not (null otherContents) then 
                                                        SRS.inModel 
                                                          (inModelIntro probDes 
                                                          datDef theMod genDef 
                                                          : otherContents) []
                                                      else
                                                        SRS.inModel 
                                                          (inModelIntroNoContent 
                                                          probDes datDef theMod 
                                                          genDef : 
                                                          otherContents) []
```
 I needed to have four section parameters for `inModelIntroNoContent` so that it can satisfy the type of the function`inModelF`.

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: Ok, I will implement that right now!

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
balacij: Sorry, what is the list of contents for? Is it ever used right now? It seems to be always empty, no?

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
balacij: I think you should be able to omit those parameters seeing as they're not used.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
harmanpreet-sagar: I believe it usually takes in the field names as its contents in the example to generate the sidebar titles in the table.
image

For example for `GamePhysics`, it takes in the following parameters to generate the TM:
https://github.com/JacquesCarette/Drasil/blob/2d90402363fb59a38ba062ec93c979ac84e83ff8/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L81

and the `stdFields` are defined further in the file as the following:
https://github.com/JacquesCarette/Drasil/blob/2d90402363fb59a38ba062ec93c979ac84e83ff8/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L134-L135

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
balacij: What happens if you remove the extra arguments in the function call?

smiths:
I'm used to checking stable for changes to the SRS generator, but I guess there are no changes because none of current examples have empty IMs, TMs, etc.  I wonder if we should create another stable example to test this functionality?  I wouldn't be opposed to changing the hghc example to have these empty sections.  If it slows progress on this PR, we can wait, but I like the idea of having a test case for as much of our functionality as possible.

harmanpreet-sagar:
@smiths Instead of creating a new example, could we just instead modify the template to show that the sections should be there as that would show that these sections would be generated regardless of whether there is content available? 

Also, since these sections are part of the SRS template which can be viewed [here](https://github.com/smiths/capTemplate/blob/main/docs/SRS/SRS.pdf), we can kind of ensure that anyone who is generating the SRS documents must have these sections so that they do follow the template strictly.

smiths:
@harmanpreet-sagar, yes adding the "empty" sections to the generated template example is a good idea.

I like the changes to the hghc example.

The changes to stable look fine to me.  I will let @balacij judge the other changes to determine whether we should accept this pull request.

harmanpreet-sagar:
Ok, I will commit the changes done in the template example then!

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -280,8 +297,9 @@ helperCI a c = over defn (\x -> foldlSent_ [x, refby $ helperRefs a c]) a
     refby sent   = sParen $ S "RefBy:" +:+. sent
 
 -- | Section stubs for implicit referencing of different models and definitions.
-tmStub, ddStub, imStub, pdStub :: Section
+tmStub, ddStub, imStub, pdStub, gdStub :: Section
```
samm82, line 300: I would reorder this to be consistent with the order of their definitions.
```suggestion
tmStub, ddStub, gdStub, imStub, pdStub :: Section
```

code/drasil-example/template/lib/Drasil/Template/Body.hs
```diff
@@ -9,6 +9,8 @@ import Language.Drasil
 import Drasil.SRSDocument
 import Theory.Drasil (DataDefinition, GenDefn, InstanceModel, TheoryModel)
 import qualified Language.Drasil.Sentence.Combinators as S
+import Data.Drasil.Concepts.Documentation(doccon, doccon')
```
samm82, line 12: I would add a space here for consistency.
```suggestion
import Data.Drasil.Concepts.Documentation (doccon, doccon')
```

code/drasil-example/template/lib/Drasil/Template/Body.hs
```diff
@@ -62,8 +71,8 @@ refDB :: ReferenceDB
 refDB = rdb [] []
 
 -- MOVE TO CONCEPTS
-example :: CI -- name of example
-example = commonIdeaWithDict "example" (pn "Template") "Template" []
+template :: CI
+template = commonIdeaWithDict "templateName" (pn "Template") "Template" []
```
samm82, line 75: I think the point of having it named "example" is so that changing its internal name (i.e., "Template") wouldn't require changing the variable name itself throughout the code (i.e., `example`), but I guess Haskell's type checking/compile time errors would help us make this change (case in point: you were able to do it!) - I seem to recall other examples using `example`, `progName`, or something similar. I would be curious to see what @smiths thinks.

samm82 (commented):
I have some super minor formatting gripes and a question about a change you made, but other than that, this looks good to me if @balacij is content with passing of the empty list discussion.

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: Ok, I will do that right now!

code/drasil-example/template/lib/Drasil/Template/Body.hs
harmanpreet-sagar: I will change that right now!

code/drasil-example/template/lib/Drasil/Template/Body.hs
smiths: What is the purpose of the concept `template`?  My instinct is that the previous `example` is a better concept to use, if the purpose of the template is to show the variation points that will be changed when the template is instantiated for a specific example.  We aren't currently looking at changing the template, so `templateName` seems odd to me.  I feel like our goal for the template is to show a generic template that highlights where a specific example will make changes.  The users will not be creating a new template, but rather a new example following the template.

However, I could very well be missing the point of the `template` CI.

harmanpreet-sagar: The reason I did it was that the title of the document changed from "Software Requirements Specification for Template" to "Software Requirements Specification for example". So in order to counter that I changed the way that it was defined.

smiths: Can we change the title to "Software Requirements Specification for ProgName".  @samm82 is correct that I often use ProgName as the generic thing that is being referred to.

harmanpreet-sagar: Changed!

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -119,13 +119,18 @@ assumpF otherContents = SRS.assumpt (assumpIntro : otherContents) []
...
+thModF progName otherContents = SRS.thModel (thModIntro progName : 
+                                              otherContents) []
 
 -- | Creates a eneralized Theoretical Model introduction given the program name.
 thModIntro :: (Idea a) => a -> Contents
 thModIntro progName = foldlSP [S "This", phrase section_, S "focuses on the",
   phrase general, plural equation `S.and_` S "laws that", short progName, S "is based on"]
 
+thModIntroNoContent :: Contents
+thModIntroNoContent = mkParagraph $ S "There are no theoretical models."
```
JacquesCarette, line 132: This is a 'pattern', i.e "There are no" followed by the name of the part you're looking at. There should be chunks for each (theoretical model, etc) from which you can pull the name.

We don't want to introduce multiple strings "theoretical model" in our code base.

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -141,21 +146,30 @@ generalDefinitionIntro _ = foldlSP [S "This", phrase section_,
...
+                                        endingSent : otherContents) []
 
 -- | Creates a general Data Definition introduction. Appends the given 'Sentence' to the end.
 dataDefinitionIntro :: Sentence -> Contents
 dataDefinitionIntro closingSent = mkParagraph (foldlSent [S "This", phrase section_, 
     S "collects and defines all the", plural datum, 
     S "needed to build the", plural inModel] +:+ closingSent)
 
+dataDefnIntroNoContent :: Contents
+dataDefnIntroNoContent = mkParagraph $ S "There are no data definitions."
```
JacquesCarette, line 160: Above comment applies here too (and below too, which I won't point out explicitly).

JacquesCarette (changes requested):

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: Ok, I will change it!

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: Ok, I will change it!

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -120,45 +120,62 @@ assumpF otherContents = SRS.assumpt (assumpIntro : otherContents) []
        atStartNP' (the assumption), S "refine the", phrase scope,
        S "by providing more detail"]
 
+emptySectSentence :: NamedIdea n => n -> Contents
+emptySectSentence var = foldlSP[ S "There are no" +:+ plural var]
```
samm82, line 124:
```suggestion
emptySectSentence var = foldlSP [S "There are no" +:+ plural var]
```

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -362,10 +362,10 @@ mkSolChSpec si (SCSProg l) =
...
     mkSubSCS _ (GDs _ _ [] _)    = SSD.genDefnF []
-    mkSubSCS _ (DDs _ _ [] _) = error "There are no Data Definitions"
-    mkSubSCS _ (IMs _ _ [] _)    = error "There are no Instance Models"
+    mkSubSCS _ (DDs _ _ [] _)    = SSD.dataDefnF EmptyS []
+    mkSubSCS _ (IMs _ _ [] _)    = SSD.inModelF SSD.pdStub SSD.ddStub SSD.tmStub SSD.gdStub []
```
samm82, line 368: Are these cases covered by the cases for non-empty lists below? For example,
```haskell
mkSubSCS _ (IMs _ _ [] _) = SSD.inModelF SSD.pdStub SSD.ddStub SSD.tmStub SSD.gdStub []
```
seems to just be a specific case of
```haskell
mkSubSCS si' (IMs intro fields ims ShowDerivation) =
      SSD.inModelF SSD.pdStub SSD.ddStub SSD.tmStub SSD.gdStub $ map mkParagraph intro ++
      concatMap (\x -> [LlC $ instanceModel fields si' x, derivation x]) ims
```
If so, we should remove the first case to improve readability (the cases were explicitly separated before since they had different functionality, but now that they have similar functionality, this is likely unnecessary).

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
```diff
@@ -120,45 +120,62 @@ assumpF otherContents = SRS.assumpt (assumpIntro : otherContents) []
...
+
 -- | Wrapper for 'thModelIntro'. Takes the program name and other 'Contents'.
 thModF :: (Idea a) => a -> [Contents] -> Section
-thModF progName otherContents = SRS.thModel (thModIntro progName : otherContents) []
+thModF _        []                   = SRS.thModel [thModIntroNoContent] []
```
samm82, line 128:
```suggestion
thModF _        []            = SRS.thModel [thModIntroNoContent] []
```

harmanpreet-sagar: I will make that change right now. Thanks for catching that!

code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs
harmanpreet-sagar: I will change that now!

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
harmanpreet-sagar: Yes, you are correct! The second case does cover the functionality of the first case. I will remove that now!

samm82 (commented):
I noticed a line length issue that I just fixed because it was so minor. Once again, if @balacij is happy with this, so am I!

balacij (approved):
Nice work!

samm82 (approved):

JacquesCarette (approved):

PR #3437, opened 2023-05-18
removed 'error' statement thrown when generating empty traceability matrix

### LINKED ISSUE #3428: `drasil-docLang` should allow us to have SRS documents with empty traceability matrices
Similar to #3427, in the name of decreasing the time to gratification (seeing tangible artifacts) when developing new projects, we should remove the restriction (from `drasil-docLang`) that says that each SRS document must have a non-empty traceability matrix.

Similar to #3427, I [hacked in a solution](https://github.com/JacquesCarette/Drasil/compare/master...balacij:Drasil:beamBending?expand=1#diff-404217f13ec9aff55d416f24e2ea76ad2f809455735b5d3f3b128d0f6e11440c). We should do this properly now, however (i.e., one that creates appropriate text and doesn't generate an empty table).

Contributes to #3326, #3356

ManvendraJani:
closes #3428

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -428,8 +428,8 @@ mkTraceabilitySec (TraceabilityProg progs) si@SI{_sys = sys} = TG.traceMGF trace
   where
     trace = map (\(TraceConfig u _ desc rows cols) -> TM.generateTraceTableView
       u desc rows cols si) fProgs
-    fProgs = filter (\(TraceConfig _ _ _ rows cols) -> eitherEmtpy 
-      (header (showUIDs rows sidb) si) (header (showUIDs cols sidb) si)) progs
+    fProgs = filter (\(TraceConfig _ _ _ rows cols) -> not $ null 
+      (header (showUIDs rows sidb) si) || null (header (showUIDs cols sidb) si)) progs
     sidb = _sysinfodb si
 
 -- | Helper to get UIDs
@@ -439,12 +439,6 @@ showUIDs a s e = filter (`elem` (Map.keys $ s ^. traceTable)) $ concatMap (\x ->
 -- | Helper to get headers of rows and columns
 header :: ([UID] -> [UID]) -> SystemInformation -> [Sentence]
 header f = TM.traceMHeader (f . nub . Map.keys . (^. refbyTable))
-
--- | Helper to check if either list is empty instead of using isEmpty
-eitherEmtpy :: [a] -> [a] -> Bool
-eitherEmtpy _ [] = False
-eitherEmtpy [] _ = False
-eitherEmtpy _ _  = True
 
 -- ** Off the Shelf Solutions
 
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -48,7 +48,7 @@ import qualified Drasil.Sections.SpecificSystemDescription as SSD (assumpF,
 import qualified Drasil.Sections.Stakeholders as Stk (stakeholderIntro,
   tClientF, tCustomerF)
 import qualified Drasil.DocumentLanguage.TraceabilityMatrix as TM (
-  generateTraceTableView)
+  generateTraceTableView, traceMHeader, TraceViewCat)
 import qualified Drasil.DocumentLanguage.TraceabilityGraph as TG (traceMGF)
 import Drasil.DocumentLanguage.TraceabilityGraph (traceyGraphGetRefs)
 import Drasil.Sections.TraceabilityMandGs (traceMatStandard)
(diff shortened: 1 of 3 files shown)


ManvendraJani:
@balacij it wasn't too clear what you wanted exactly but if this is not what you were after, please let me know.

balacij (approved):
This looks good to me, thank you!

@smiths @samm82 Do you have any idea what the alt text should be here?

code/drasil-docLang/lib/Drasil/DocumentLanguage/TraceabilityMatrix.hs
```diff
@@ -76,10 +76,9 @@ traceMColumns fc fr c = map ((\u -> filter (`elem` u) $ fc u) . flip traceLookup
 tableShows :: (Referable a, HasShortName a) => a -> Sentence -> Sentence
 tableShows r end = refS r +:+ S "shows the" +:+ plural dependency `S.of_` (end !.)
 
--- | Helper that makes sure the rows and columns of a traceability matrix have substance.
-ensureItems :: String -> [a] -> [a]
-ensureItems u [] = error $ "Expected non-empty matrix dimension for traceability matrix " ++ u
-ensureItems _ l = l
+-- | Helper that prints an error when traceability matrix is empty.
+emptyItems :: UID -> LabelledContent
+emptyItems u = llcc (makeTabRef' u) $ Paragraph $ S $ show u ++ " : is empty/incomplete"
```
samm82, line 81: Some notes about this message (in decreasing importance):
- [ ] The UID displayed here doesn't refer to a specific matrix dimension, but to the whole matrix. I would change this so that it is clear that at least one matrix dimension is empty
   - Maybe it could even determine what set of concepts is empty? This seems unnecessary to me, though.
- [ ] I think this should explicitly say "empty", since an incomplete list will still result in a table being generated correctly.
- [ ] I personally think that there shouldn't be a space before the colon (although others may disagree).

samm82 (changes requested):
I tried to generate a traceability matrix section for the Template example to get an idea for how this would look with empty sections (to avoid the pesky `CallStack` errors of `<concept> not found in ReferenceMap` from modifiying an existing example), and it doesn't look quite right:

image

(BTW: to do this, I redefined `mkSRS` to be `[TraceabilitySec $ TraceabilityProg $ traceMatStandard si]` and replaced the second argument of the `cdb` call for `symbMap` with `(nw example : nw mass : nw inValue : [nw errMsg, nw program] ++
    map nw doccon ++ map nw doccon' ++ map nw physicCon ++ map nw physicCon' ++
    map nw physicscon ++ map nw mathcon)`.)

code/drasil-docLang/lib/Drasil/DocumentLanguage/TraceabilityMatrix.hs
ManvendraJani: I implemented changes and fixed it so it now actually shows the messages.

samm82 (commented):
This output looks better, but I think it still can be improved.
image

I think that if a traceability matrix isn't rendered, its description (e.g., "Tab:TraceMatAllvsR shows the dependencies of on the data definitions, theoretical models, general definitions, and instance models.") should be removed. If none of them aren't rendered, I think it should say "There are no traceability tables." in a similar way to #3436. I'm not sure how useful it is to say that a given traceability table is empty instead of just omitting it. Others may disagree with me on this, however.

(This specific example of a description should be "...dependencies of the requirements on the...", which may need to be a separate issue, since I'm noticing other ways that the traceability table descriptions could be improved.)

samm82 (commented):
This looks good to me! @smiths I'm assuming we want to display a traceability matrix even if there are no connections between its components? This is what is currently done, but I'm wondering if there is another way we could handle this (e.g., "There are no connections between the assumptions and themselves.")

smiths:
@samm82 I don't feel super strongly about it, but if there are no connections between components your proposed text message is probably better than showing a traceability graph with no connections.  If I saw a traceability graph with no connections my first thought would be that information is missing from the graph, since no connections at all should be rare for a real project.  If the case of no connections comes up, it would likely help the reader to make this explicit.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -423,11 +423,28 @@ mkUCsSec (UCsProg c) = SRS.unlikeChg (intro : mkEnumSimpleD c) []
...
+showUIDs :: [TM.TraceViewCat] -> ChunkDB -> [UID] -> [UID]
+showUIDs a s e = filter (`elem` (Map.keys $ s ^. traceTable)) $ concatMap (\x -> x e s) a
+
+-- | Helper to get headers of rows and columns
+header :: ([UID] -> [UID]) -> SystemInformation -> [Sentence]
+header f = TM.traceMHeader (f . nub . Map.keys . (^. refbyTable))
+
+-- | Helper to check if either list is empty instead of using isEmpty
+eitherEmtpy :: [a] -> [a] -> Bool
+eitherEmtpy _ [] = False
```
balacij, line 445: `Emtpy` is a typo.

You can also use `null` to check if a list is empty instead of pattern matching.

balacij (changes requested):

balacij (approved):
Nice work!

PR #3439, opened 2023-05-23
3418 Fix Drasil image path in READMEs

### LINKED ISSUE #3418: Fix Drasil image path in READMEs
It seems like the path to the Drasil logo (?) from the READMEs is incorrectly generated; the path should have one extra "../" to go up one more directory.

## Relevant HTML code in README:
https://github.com/JacquesCarette/Drasil/blob/f003baf9c65a28c28a58b4aa5010b73cec672203/code/stable/projectile/projectile_c_p_nol_b_u_v_d/src/python/README.md?plain=1#L29-L31

## What this looks like:
image

## Clicking on the image link:
image

## What the path should be:
image

harmanpreet-sagar:
Closes #3418. Changes made to CreateMd.hs, WriteReadMe.hs and the projectile example in stable in order to update the broken image link in the Projectile READMEs.


code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,21 @@ license :: Doc -> Doc
...
 -- | Drasil Tree icon. Uses HTML directly to format image since normal markdown doesn't support it.
-drasilImage :: Doc
-drasilImage = alignImage "../../../../drasil-website/WebInfo/images/Icon.png"
+drasilImage :: String -> Doc
+drasilImage name = if name == "Projectile" then 
+                      alignImage 
+                        "../../../../../drasil-website/WebInfo/images/Icon.png"
+                   else
+                      alignImage 
+                        "../../../../drasil-website/WebInfo/images/Icon.png"
```
samm82, line 106: This should use pattern matching (or some other Haskell trick; can this be improved @balacij?)
```suggestion
drasilImage "Projectile" = alignImage "../../../../../drasil-website/WebInfo/images/Icon.png"
drasilImage _            = alignImage "../../../../drasil-website/WebInfo/images/Icon.png"
```

samm82 (changes requested):
Other than the Haskell syntax, this looks good!

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
JacquesCarette: I'm really uncomfortable with hard-coding any of our examples into the rest of Drasil. 

Much better would be some kind of other way to signal to the Markdown printer (from the example) the relative location. For example, maybe `drasilImage` could use some kind of (to be created) `buildPath` function which would take a natural number corresponding to the number of `..` to add. Such "configuration data" could then all be put into a record that could be passed around.

harmanpreet-sagar: I like that as well! I will look into creating the `buildPath` function now.

harmanpreet-sagar:
@samm82 @balacij 
I am trying to obtain the location of the ReadMes for each example so that I can count the number of `"/"` within the `FilePath` and then append the same number of `"../"` in front of the path of the image. However, I am unable to find the `FilePath` for the ReadMes.

The closest I have gotten is the following:
https://github.com/JacquesCarette/Drasil/blob/040ee6573bfcb84bf16514290c9f630960dd2a60/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs#L98-L113

Is there any function that I can use to obtain the `FilePath` from this code? Or is there any other place I should be looking to get the `FilePath` for the ReadMes?

samm82:
I wonder if simply adding an extra `../` if the generated code has design variabilities specified for generation (like Projectile does) is a good enough short-term solution. 🤔

balacij:
Sorry, I'm confused: aren't the READMEs supposed to be in the current working directory?

We might need to take a step back and figure out a proper (not String-based) design for working with external resources (as @JacquesCarette mentioned). If each SRS is meant to be standalone, then the resources needed should be copied for each individual generated artifact, or at least symbolically linked.

samm82:
@balacij Yes the READMEs are built in the current directory, but they reference an image that is stored in a different directory, so the relative file path has to go up the correct number of levels.

harmanpreet-sagar:
@ samm82 Could you expand on your earlier comment about design variabilities? I am not sure I entirely understand what you are trying to say.

samm82:
@harmanpreet-sagar Sure! The only reason (as far as I can tell) that Projectile has this extra directory is because we generate multiple "versions" of it based on design variabilities. For example, Projectile_U_P_NoL_U_WI_V_D is a version of Projectile with **U**nmodular code that is implemented as a **P**rogram, with **No L**ogging and an **U**nbundled input structure that stores constants **W**ith the **I**nputs as **V**ariables and represents real numbers as **D**oubles (see the Case Studies section of the [website](https://jacquescarette.github.io/Drasil/)). The Python version of this variability is found in the `python/projectile_u_p_nol_u_wi_v_d` directory, as opposed to examples _without_ variabilities, which would just be found in the `python` directory; this is where the need for the extra `../` comes from.

As far as I can tell, these choices are defined here:
https://github.com/JacquesCarette/Drasil/blob/017a0768ba6f6bd2a9e018e3c18cba9b2c47492d/code/drasil-example/projectile/lib/Drasil/Projectile/Choices.hs#L78-L103

and are actually used here:
https://github.com/JacquesCarette/Drasil/blob/017a0768ba6f6bd2a9e018e3c18cba9b2c47492d/code/drasil-example/projectile/app/Main.hs#L18

Where `genCodeWithChoices` generates code with these variabilities, as opposed to `genCode` which just generates one version.

https://github.com/JacquesCarette/Drasil/blob/017a0768ba6f6bd2a9e018e3c18cba9b2c47492d/code/drasil-example/glassbr/app/Main.hs#L13

I wonder if there is a way to only add the extra `../` if we actually generate these "design variabilities", maybe based on which function we use to generate code?

harmanpreet-sagar:
That makes so much sense now! Thank you for that clarification! That is a good idea. I will look into that to see whether it is feasible to do so or not!

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,34 @@ license :: Doc -> Doc
...
+numSlash :: [Char] -> Int
+numSlash [] = 4
+numSlash file = length (removeSlash file)
+
+-- Creates the prefix to be attached to the general drasil logo path
+buildPath :: [Char] -> [Char]
+buildPath file = [x | x <- pathPrefix, x /= ' ']
+  where
+    pathPrefix = unwords pathPrfx
+    pathPrfx = replicate (numSlash file) "../"
```
samm82, line 114: Combine these lines for readability:

```suggestion
    pathPrefix = unwords $ replicate (numSlash file) "../"
```

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,34 @@ license :: Doc -> Doc
...
-drasilImage :: Doc
-drasilImage = alignImage "../../../../drasil-website/WebInfo/images/Icon.png"
+-- Extracts the backslash from the filepath.
+removeSlash :: [Char] -> [Char]
+removeSlash st = [ x | x <- st, x `elem` "//"]
+
+-- Determines the number of slashes within the FilePath of the SampleInput file
+numSlash :: [Char] -> Int
+numSlash [] = 4
+numSlash file = length (removeSlash file)
```
samm82, line 107: Others may disagree with me, but if `removeSlash` and `numSlash` are only going to be used here, might as well add them to the locally defined functions in the `where` statement of `buildPath`.

samm82 (changes requested):
Some minor code formatting issues, but great job finding a fix for this! I'll wait for @JacquesCarette to review the design of this, but once you fix the formatting, it looks good to me.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
harmanpreet-sagar: Ok, I will change that right now!

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
harmanpreet-sagar: Ok, I will make that change!

samm82 (approved):
Just made some more minor tweaks (I felt bad asking you to do another when I didn't catch it the first time!) - `pathPrefix` makes sense to keep as a local variable, since it is more complex, but since the other function was just a list comprehension (I already forgot what it was called 😅), I just inlined it

harmanpreet-sagar:
Thank you!!!

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -66,5 +66,6 @@ data ReadMeInfo = ReadMeInfo {
   configFP :: [FilePath],
   caseName :: CaseName,
   examplePurpose :: ExamplePurpose,
-  exampleDescr :: ExampleDescr
+  exampleDescr :: ExampleDescr,
+  sampleFile :: FilePath
```
JacquesCarette, line 70: I'm not sure I like this: do all `README`s have a 'sample File' associated to them? If not, then we shouldn't require this information for all of them.

For optional information, it should be wrapped in a `Maybe`. We shouldn't use that `FilePath` is implemented as `List` to fake "not present" via an empty list. That is too fragile.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -276,10 +276,10 @@ instance RenderChoices AuxFile where
 
 -- | Gets the file path to a sample input data set from a 'Choices' structure, if 
 -- the user chose to generate a sample input file.
-getSampleData :: Choices -> Maybe FilePath
+getSampleData :: Choices -> FilePath
```
JacquesCarette, line 279: See review item below on `sampleFile` for more, but: I think this change is backwards. It uses a 'coding trick' instead of being explicit about the fact that `FilePath` is optional.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,27 @@ license :: Doc -> Doc
 license auth = text "Copyright (c) 2021," <+> auth <>
   text ". All rights reserved. Please see the [full license](https://github.com/JacquesCarette/Drasil/blob/4b9ad0a3016fecb3c7a2aa82ab142f9e805b5cc8/LICENSE) for more details."
 -- | Drasil Tree icon. Uses HTML directly to format image since normal markdown doesn't support it.
-drasilImage :: Doc
-drasilImage = alignImage "../../../../drasil-website/WebInfo/images/Icon.png"
+
+-- Creates the prefix to be attached to the general Drasil logo path
+buildPath :: [Char] -> [Char]
```
JacquesCarette, line 102: `[Char]` == `String`.

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
harmanpreet-sagar: The `sampleFile` is just referring to the sample input files that are present with all the `ReadMe`s in all examples. Thus, my code was more based on obtaining the `FilePath` from the input files and extracting all the information from those `FilePaths`s. 

I tried obtaining the `FilePath` for the `ReadMe`s but since all of those were obtained using the` IO ()` methods such as `getCurrentDirectory`, etc. I was unable to obtain the `String`/`FilePath` from them in a manner that would not include me using `IO ()` as one of the types throughout the entire fix.

Another potential way that I tried to solve this issue was to create a new choice within the `Choices `area which would take in an integer value which would have a default value of 4. This value essentially would obtain the number of folders that we would have had to go back to in order to get the image. We could then obtain any changes to that value for each example and then generate the image as well. That would I believe remove the modifications done to the Maybe Filepath as our code would no longer be based on the `FilePath` of the input file. 

Let me know if you would like me to modify it such that we create the new `Choice`.

code/drasil-code/lib/Language/Drasil/Choices.hs
harmanpreet-sagar: As I mentioned in the comment above, a way that we could counter this issue would be to create a new choice within the `Choices` area which would take in an integer value which would have a default value of 4. This value essentially would obtain the number of folders that we would have had to go back to in order to get the image. We could then receive any changes to that value for each example and generate the image. That would remove the modifications to the `Maybe Filepath` as our code would no longer be based on the FilePath of the input file.

Let me know if you would like me to modify it such that we create the new `Choice`.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
harmanpreet-sagar: I will change that right now!

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,27 @@ license :: Doc -> Doc
...
+buildPath file = [x | x <- pathPrefix, x /= ' ']
```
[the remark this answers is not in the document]
harmanpreet-sagar: That is a much better way! Thanks! I will change that right now.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -100,7 +100,7 @@ license auth = text "Copyright (c) 2021," <+> auth <>
 
 -- Creates the prefix to be attached to the general Drasil logo path
 buildPath :: String -> String
-buildPath file = [x | x <- pathPrefix, x /= ' ']
+buildPath file = filter (/=' ') pathPrefix
   where
     pathPrefix = unwords $ replicate (numSlash file) "../"
```
samm82: Now you could even pull `pathPrefix` out of the `where` statement, since the definition of `buildPath` is a lot simpler:

```suggestion
buildPath file = filter (/=' ') $ unwords $ replicate (numSlash file) "../"
  where
```

harmanpreet-sagar: Ok, I will make that change right now!

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
harmanpreet-sagar: Update: The method that I am suggesting in the above comment can be viewed [here ](https://github.com/JacquesCarette/Drasil/tree/3418-Fix-Drasil-Logo-Path-on-READMEs-alt)on a different branch.

code/drasil-code/lib/Language/Drasil/Choices.hs
harmanpreet-sagar: Update: The method that I am suggesting in the above comment can be viewed [here](https://github.com/JacquesCarette/Drasil/tree/3418-Fix-Drasil-Logo-Path-on-READMEs-alt) on a different branch.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -97,16 +97,27 @@ license :: Doc -> Doc
 license auth = text "Copyright (c) 2021," <+> auth <>
   text ". All rights reserved. Please see the [full license](https://github.com/JacquesCarette/Drasil/blob/4b9ad0a3016fecb3c7a2aa82ab142f9e805b5cc8/LICENSE) for more details."
 -- | Drasil Tree icon. Uses HTML directly to format image since normal markdown doesn't support it.
-drasilImage :: Doc
-drasilImage = alignImage "../../../../drasil-website/WebInfo/images/Icon.png"
+
+-- Creates the prefix to be attached to the general Drasil logo path
```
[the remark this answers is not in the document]
harmanpreet-sagar: I will change that right now!

harmanpreet-sagar:
I'm closing this PR to start a new one which would have the discussed changes as the new changes are on a different branch based on the master. The new PR is #3467.

PR #3447, opened 2023-05-26
Improved Traceability Matrix Description

### LINKED ISSUE #3438: Improve descriptions of traceability matrices
I noticed in #3437 that the description of the requirements vs all matrix wasn't being generated correctly: "Tab:TraceMatAllvsR shows the dependencies of on the data definitions, theoretical models, general definitions, and instance models." - there should be something between "dependencies of" and "on the". Looking to another example, I noticed that Projectile says "...dependencies of requirements, goal statements on the data definitions..." when it should say "...dependencies of **the** requirements **and** goal statements on the data definitions..."

Additionally, Tab:TraceMatRefvsRef's description says it "shows the dependencies of data definitions, ..., and instance models with each other." This wording of "each other" should be reused for Tab:TraceMatAvsA instead of saying it "shows the dependencies of assumptions on the assumptions."

# TODO
- [x] Change the list of `ConceptInstance`s in the document in the description of Tab:TraceMatAllvsR to be built using `foldlList Comma List`
- [x] Change description of Tab:TraceMatAvsA to "shows the dependencies of assumptions on each other.", possibly by reusing the method that Tab:TraceMatRefvsRef uses

ManvendraJani:
closes #3438


code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
```diff
@@ -87,10 +87,11 @@ traceMatRefinement = TraceConfig (mkUid "TraceMatRefvsRef") [plural Doc.dataDefn
 
 -- | Records other requirements. Converts the 'SystemInformation' into a 'TraceConfig'.
 traceMatOtherReq :: SystemInformation -> TraceConfig
-traceMatOtherReq si = TraceConfig (mkUid "TraceMatAllvsR") [x plural +:+ S "on the" +:+
-  plural Doc.dataDefn, plural Doc.thModel, plural Doc.genDefn, plural Doc.inModel]
-  (x titleize' +:+ S "and Other" +:+ titleize' item) [tvDataDefns, tvTheoryModels,
-  tvGenDefns, tvInsModels, tvReqs] [tvGoals, tvReqs] where
+traceMatOtherReq si = TraceConfig (mkUid "TraceMatAllvsR") [plural requirement +:+ 
+  S "and" +:+ pluralNP (goalStmt `onThePP` Doc.dataDefn), plural Doc.thModel, 
+  plural Doc.genDefn, plural Doc.inModel] (x titleize' +:+ S "and Other" +:+
```
samm82, line 92: It looks like this assumes that an example will have more than one instance of each model (e.g., instance model). How can we build this list from what is _actually_ present in the document, including (ideally) using the singular form of a model if there is only one instance of it? 

The number of DDs and IMs is present in `SystemInformation`:
https://github.com/JacquesCarette/Drasil/blob/017a0768ba6f6bd2a9e018e3c18cba9b2c47492d/code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs#L103-L123

and it looks like it _might_ be populated the rest of the way with `fillcdbSRS`, but I don't know enough about it to verify this:
https://github.com/JacquesCarette/Drasil/blob/017a0768ba6f6bd2a9e018e3c18cba9b2c47492d/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L80-L81

samm82 (changes requested):
While you're here, might as well fix the typo ("Traceabiliy" instead of "Traceability") in the comments of `TraceabilityMandGs.hs`.

Also, the first item of #3438 is not checked off, and it doesn't look like this PR addresses that, so I'm not sure that this PR actually closes the issue.

ManvendraJani:
@samm82 I don't see straightforward way to currently implement a check of plurality for all of the Common Ideas. I am looking especially at Goal Statements and Assumptions, as either is not part of SI directly.

samm82:
I think that we should make that a separate issue; it's a good thing to implement, but we don't want it holding up this PR.

samm82 (approved):
This looks good to me!

ManvendraJani:
> I think that we should make that a separate issue; it's a good thing to implement, but we don't want it holding up this PR.

Should I make an issue regarding this?

samm82:
Yes please! @ManvendraJani

Issue #3419, opened 2023-05-12
Fill `purpose` for Game Physics


hrzhuang:
Spun-off of #3391. Other examples (except for HGHC and Template which are meant to be bare bones) had a short introduction to the Problem Description section in the SRS which could be directly used as the `purpose` in `SystemInformation`. However, this introduction for Game Physics is too long to be used directly. A short summary of this introduction (and the Goal Statements) is needed.

smiths:
@hrzhuang this is a good issue.  I've assigned it to you, since I don't think it will take long and you know the context of what we are trying to accomplish.  You can put the proposed text as a comment first so that I can give you comments before we proceed.

smiths:
@hrzhuang I like your proposed purpose, but I think we should simplify it even further.  Our first goal is to get the functionality right.  I have less confidence that we'll achieve the nonfunctional requirements like simple, lightweight, etc. 
Also, I don't want to claim those things until we have an unambiguous way of specifying those qualities.  At some point, we will want to return to the conversation about the nonfunctional requirements, but for now, I think we should leave them out of the purpose.  Let's just say:

"simulate 2D rigid body physics for use in game development"

I thought about taking out the game development part, but I think we should keep it, since the scope of the project comes from an existing project built for game physics.

I agree with you that the introduction would be changed to something like "A simple, lightweight, fast, and portable 2D rigid body physics library allows ..."  I left the nonfunctional requirements in there to help remind us that this is something we should expand/clarify in the future.  @hrzhuang please either fix the introduction, or create an issue that we can assign to someone else for this.

hrzhuang:
I created a PR that adds this purpose and fixes the introduction. I changed the first sentence of the introduction to

> A system is needed to simulate 2D rigid body for use in game development, in a simple, lightweight, fast, and portable manner, which will allow for the production of higher quality products.

so it reuses the purpose instead of saying the same thing slightly differently. Also Drasil puts the phrase "A system is need to" at the beginning of every introduction (which is something we might want to relax as another issue).

PR #3450, opened 2023-05-29
Fix `\brief` for Projectile

### LINKED ISSUE #3218: Projectile: Code: Fix \brief for doxygen
The [`\brief` tag](https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/python/Projectile.py) in the generated `Projectile.py` file is currently a comment to the developers, as opposed to a comment to users of the code.  The current tag says "Contains the entire Projectile program."  We should fix this to provide the purpose of the program.  The purpose `\brief` should be "Efficiently and correctly predict whether a launched projectile hits its target."  Ideally the purpose here should come from the same source as for the README file (#3159) and the SRS.

hrzhuang:
Closes #3218

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
@@ -17,6 +17,7 @@ import Language.Drasil.Choices (ImplementationType(..), InputModule(..),
   Structure(..))
 import Language.Drasil.CodeSpec (CodeSpec(..))
 import Language.Drasil.Mod (Description)
+import Language.Drasil.Printers (Linearity(Linear), sentenceDoc)
 
 import Data.Map (member)
 import qualified Data.Map as Map (filter, lookup, null)
@@ -35,10 +36,11 @@ modDesc = fmap ((++) "Provides " . stringList)
 unmodularDesc :: GenState Description
 unmodularDesc = do
   g <- get
-  let n = pName $ codeSpec g
-      getDesc Library = "library"
-      getDesc Program = "program"
-  return $ "Contains the entire " ++ n ++ " " ++ getDesc (implType g)
+  let spec = codeSpec g
+      implTypeStr Program = "program"
+      implTypeStr Library = "library"
+  return $ show $ sentenceDoc (sysinfodb spec) Implementation Linear $ capSent $
+    foldlSent ([S "a", S (implTypeStr (implType g)), S "to"] ++ purpose spec)
 
 -- | Returns description of what is contained in the Input Parameters module.
 -- If user chooses the 'Bundled' input parameter, this module will include the structure for holding the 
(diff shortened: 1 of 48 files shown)


code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -154,7 +153,8 @@ concIns = assumptions ++ funcReqs ++ goals ++ nonfuncReqs
 -- Problem Description --
 -------------------------
 
--- Introduction of the Problem Description section derives from purp
+probDescIntro :: Sentence
+probDescIntro = S "efficiently" +:+ purp
```
samm82, line 157: Based on [Dr. Smith's comment](https://github.com/JacquesCarette/Drasil/pull/3248#issuecomment-1567400242) on #3248, I would remove this and just use `purp` instead of `probDescIntro`, since the "SRS doesn't define what we mean by efficient" and the "code doesn't do anything [specifically] to be efficient."

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
```diff
@@ -214,4 +216,4 @@ physAndSfwrCons = do
         (inputs $ codeSpec g)
   return $ stringList [
     if not (any isPhysC cns) then "" else "physical constraints",
-    if not (any isSfwrC cns) then "" else "software constraints"]
\ No newline at end of file
+    if not (any isSfwrC cns) then "" else "software constraints"]
```
samm82, line 219: A minor thing, but it is convention to end files with a trailing newline.
```suggestion
    if not (any isSfwrC cns) then "" else "software constraints"]
    
```

samm82 (changes requested):

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
hrzhuang: I think we should keep the "efficiently" based on Dr. Smith's comment on #3218 that "[he is] fine with keeping the non-functional requirements in the introduction." We also kept the non-functional requirements in the introduction for #3419.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
hrzhuang: It seems like there was already a trailing newline there and you added an additional one? Now there are two trailing newlines.

samm82: Hmm... my diff looks like there was a trailing line removed - I'll see what it looks like in the branch

samm82: Yes, you are right - my bad!

samm82 (approved):

smiths (approved):

PR #3455, opened 2023-05-31
removed instance of show for fields

### LINKED ISSUE #3393: Use concepts for field names
In #3389, it was noted that we hardcode a bunch of field names as strings instead of using their associated concepts, which should be fixed.

https://github.com/JacquesCarette/Drasil/blob/3725db2ce30e5eb47c0ba9f93f96549d4f61ea21/code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs#L216-L229

Note that simply mapping from a `Sentence` to a `String` is a code smell, so some other solution should be developed.

ManvendraJani:
closes #3393

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
@@ -95,16 +95,16 @@ tmDispExprs t = map express (t ^. defined_quant) ++ t ^. invariants
 
 -- | Create the fields for a model from a relation concept (used by 'tmodel').
 mkTMField :: TheoryModel -> SystemInformation -> Field -> ModRow -> ModRow
-mkTMField t _ l@Label fs  = (show l, [mkParagraph $ atStart t]) : fs
-mkTMField t _ l@DefiningEquation fs = (show l, map unlbldExpr $ tmDispExprs t) : fs
-mkTMField t m l@(Description v u) fs = (show l,
+mkTMField t _ Label fs  = ("Label", [mkParagraph $ atStart t]) : fs
+mkTMField t _ DefiningEquation fs = ("Equation", map unlbldExpr $ tmDispExprs t) : fs
+mkTMField t m (Description v u) fs = ("Description",
   foldr ((\x -> buildDescription v u x m) . express) [] $ tmDispExprs t) : fs
-mkTMField t m l@RefBy fs = (show l, [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
-mkTMField t _ l@Source fs = (show l, helperSources $ t ^. getDecRefs) : fs
-mkTMField t _ l@Notes fs =
-  nonEmpty fs (\ss -> (show l, map mkParagraph ss) : fs) (t ^. getNotes)
-mkTMField _ _ l _ = error $ "Label " ++ show l ++ " not supported " ++
-  "for theory models"
+mkTMField t m RefBy fs = ("RefBy", [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
+mkTMField t _ Source fs = ("Source", helperSources $ t ^. getDecRefs) : fs
+mkTMField t _ Notes fs =
+  nonEmpty fs (\ss -> ("Notes", map mkParagraph ss) : fs) (t ^. getNotes)
+mkTMField _ _ l fs = ("Label not supported for data definitions", []) : fs
+
 
 -- | Helper function to make a list of 'Sentence's from the current system information and something that has a 'UID'.
 helperRefs :: HasUID t => t -> SystemInformation -> Sentence
@@ -132,16 +132,15 @@ helperSources rs  = [mkParagraph $ foldlList Comma List $ map (\r -> Ref (r ^. u
 
 -- | Creates the fields for a definition from a 'QDefinition' (used by 'ddefn').
 mkDDField :: DataDefinition -> SystemInformation -> Field -> ModRow -> ModRow
-mkDDField d _ l@Label fs = (show l, [mkParagraph $ atStart d]) : fs
-mkDDField d _ l@Symbol fs = (show l, [mkParagraph . P $ eqSymb d]) : fs
-mkDDField d _ l@Units fs = (show l, [mkParagraph $ toSentenceUnitless d]) : fs
-mkDDField d _ l@DefiningEquation fs = (show l, [unlbldExpr $ express d]) : fs
-mkDDField d m l@(Description v u) fs = (show l, buildDDescription' v u d m) : fs
-mkDDField t m l@RefBy fs = (show l, [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
-mkDDField d _ l@Source fs = (show l, helperSources $ d ^. getDecRefs) : fs
-mkDDField d _ l@Notes fs = nonEmpty fs (\ss -> (show l, map mkParagraph ss) : fs) (d ^. getNotes)
-mkDDField _ _ l _ = error $ "Label " ++ show l ++ " not supported " ++
-  "for data definitions"
+mkDDField d _ Label fs = ("Label", [mkParagraph $ atStart d]) : fs
+mkDDField d _ Symbol fs = ("Symbol", [mkParagraph . P $ eqSymb d]) : fs
+mkDDField d _ Units fs = ("Units", [mkParagraph $ toSentenceUnitless d]) : fs
+mkDDField d _ DefiningEquation fs = ("Equation", [unlbldExpr $ express d]) : fs
+mkDDField d m (Description v u) fs = ("Description", buildDDescription' v u d m) : fs
+mkDDField t m RefBy fs = ("RefBy", [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
+mkDDField d _ Source fs = ("Source", helperSources $ d ^. getDecRefs) : fs
+mkDDField d _ Notes fs = nonEmpty fs (\ss -> ("Notes", map mkParagraph ss) : fs) (d ^. getNotes)
+mkDDField _ _ l fs = ("Label not supported for data definitions", []) : fs
 
 -- | Creates the description field for 'Contents' (if necessary) using the given verbosity and
 -- including or ignoring units for a model/general definition.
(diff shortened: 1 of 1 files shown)


samm82 (changes requested):
This is definitely a step in the right direction, but I don't think this is ready to be merged yet. Essentially, we are still using `String`s to represent these ideas, but we've moved them to where they are used, instead of having them all in one place, which is harder to maintain. I think merging this PR should wait until we grab these labels from concepts themselves (e.g., as a `Sentence`), which might require some design.

ManvendraJani:
> This is definitely a step in the right direction, but I don't think this is ready to be merged yet.

I felt that as well but I just wanted to know what direction to take as this is a change in how the code works. The type `ModRow` is a list of `(String, [Content])` tuples.  This would need to also be changed then, including how this is used for the generated output. 

This the direction I was thinking to go in, but let me know if you want something different?

ManvendraJani:
@samm82 the last commit breaks a few things but I feel this is very close to the final code, I was looking for ways to pass the information stored in ModRow without classifying it as a paragraph.

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -213,17 +211,16 @@ descPairs IncludeUnits =
...
+getSent Units             = S "Units"
+getSent RefBy             = S "RefBy"
+getSent Source            = S "Source"
+getSent Input             = S "Input"
+getSent Output            = S "Output"
+getSent InConstraints     = S "Input Constraints"
+getSent OutConstraints    = S "Output Constraints"
+getSent DefiningEquation  = S "Equation"
+getSent (Description _ _) = S "Description"
+getSent Notes             = S "Notes"
```
samm82, line 226: One of the next steps here would be passing these sentences from our previously defined objects (e.g., `atStart output_` or something similar) instead of hardcoding them like this (and add a trailing new line to this file - it doesn't look like there is one but I could be wrong 😅).

ManvendraJani: just had some issues with the generation of the JSON and tex files I have fixed those.

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -95,16 +98,16 @@ tmDispExprs t = map express (t ^. defined_quant) ++ t ^. invariants
...
-mkTMField t m l@RefBy fs = (show l, [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
-mkTMField t _ l@Source fs = (show l, helperSources $ t ^. getDecRefs) : fs
+mkTMField t m l@RefBy fs = (getSent l, [mkParagraph $ helperRefs t m]) : fs --FIXME: fill this in
+mkTMField t _ l@Source fs = (getSent l, helperSources $ t ^. getDecRefs) : fs
 mkTMField t _ l@Notes fs =
-  nonEmpty fs (\ss -> (show l, map mkParagraph ss) : fs) (t ^. getNotes)
-mkTMField _ _ l _ = error $ "Label " ++ show l ++ " not supported " ++
-  "for theory models"
+  nonEmpty fs (\ss -> (getSent l, map mkParagraph ss) : fs) (t ^. getNotes)
+mkTMField _ _ l fs = (getSent l +:+ S ":Label not supported for data definitions", []) : fs
```
samm82, line 109: a) This should still say "theory models" since it is the `mkTMField` constructor
b) I think the error message should stay as it is; if a pattern above isn't matched, then what value of `l` would be possible that would be caught by a case of `getSent`?

(This applies to the other instances as well)

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -162,42 +164,41 @@ buildDDescription' Verbose  u d m = [UlC . ulcc . Enumeration $ Definitions $
...
+mkGDField g m l@(Description v u) fs = (getSent l,
   buildDescription v u (express g) m []) : fs
-mkGDField g m l@RefBy fs = (show l, [mkParagraph $ helperRefs g m]) : fs --FIXME: fill this in
-mkGDField g _ l@Source fs = (show l, helperSources $ g ^. getDecRefs) : fs
-mkGDField g _ l@Notes fs = nonEmpty fs (\ss -> (show l, map mkParagraph ss) : fs) (g ^. getNotes)
-mkGDField _ _ l _ = error $ "Label " ++ show l ++ " not supported for gen defs"
+mkGDField g m l@RefBy fs = (getSent l, [mkParagraph $ helperRefs g m]) : fs --FIXME: fill this in
+mkGDField g _ l@Source fs = (getSent l, helperSources $ g ^. getDecRefs) : fs
+mkGDField g _ l@Notes fs = nonEmpty fs (\ss -> (getSent l, map mkParagraph ss) : fs) (g ^. getNotes)
+mkGDField _ _ l fs = (getSent l +:+ S ":Label not supported for data definitions", []) : fs
```
samm82, line 176: Instead of "gen defs", like was used before, I would put
```suggestion
mkGDField _ _ l fs = (getSent l +:+ S ":Label not supported for general definitions", []) : fs
```

samm82 (changes requested):
Oops - I made all these comments and forgot to submit my review! This PR is almost there!

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
ManvendraJani: in this specific case I was thinking something like input constraints, that is a field that could be passed but is not supported by mkTMField

samm82: Yes, you are right - my bad! I don't know why I just assumed that there would be a case for every label in these field functions 😔

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
ManvendraJani: I think for these I copy pasted the format and just forgot to adjust the error message, will push the fix in a minute

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -162,42 +164,41 @@ buildDDescription' Verbose  u d m = [UlC . ulcc . Enumeration $ Definitions $
...
+mkGDField g m l@(Description v u) fs = (getSent l,
   buildDescription v u (express g) m []) : fs
-mkGDField g m l@RefBy fs = (show l, [mkParagraph $ helperRefs g m]) : fs --FIXME: fill this in
-mkGDField g _ l@Source fs = (show l, helperSources $ g ^. getDecRefs) : fs
-mkGDField g _ l@Notes fs = nonEmpty fs (\ss -> (show l, map mkParagraph ss) : fs) (g ^. getNotes)
-mkGDField _ _ l _ = error $ "Label " ++ show l ++ " not supported for gen defs"
+mkGDField g m l@RefBy fs = (getSent l, [mkParagraph $ helperRefs g m]) : fs --FIXME: fill this in
+mkGDField g _ l@Source fs = (getSent l, helperSources $ g ^. getDecRefs) : fs
+mkGDField g _ l@Notes fs = nonEmpty fs (\ss -> (getSent l, map mkParagraph ss) : fs) (g ^. getNotes)
+mkGDField _ _ l fs = (getSent l +:+ S ":Label not supported for gen defs", []) : fs
```
samm82, line 176:
```suggestion
mkGDField _ _ l fs = (getSent l +:+ S ":Label not supported for general definitions", []) : fs
```

Also, what's the rationale for what the error message is? Is there supposed to be a space after the colon? Why change it from "Label `l` not supported..."?

ManvendraJani: I could change this back but, I don't think the exact message matters as long as we provide the necessary information and it is straightforward. I think the current message does what it needs to.

samm82 (approved):
If @smiths is OK with the error message, then this looks good to me!

smiths:
The error message ":Label not supported for general definitions" looks fine to me.

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
```diff
@@ -69,16 +69,13 @@ data RawContent =
   | EqnBlock ModelExpr                       -- ^ Block of Equations holds an expression.
   | DerivBlock Sentence [RawContent]         -- ^ Grants the ability to label a group of 'RawContent'.
   | Enumeration ListType                     -- ^ For enumerated lists.
-  | Defini DType [(Identifier, [Contents])]  -- ^ Defines something with a type, identifier, and 'Contents'.
+  | Defini DType [(Sentence, [Contents])]    -- ^ Defines something with a type, identifier, and 'Contents'.
```
JacquesCarette, line 72: Oh, I'm not so such about this particular change. While `Identifier` doesn't need to be `String`, I think there are still some non-trivial restrictions on them that `Sentence` does not obey.

I think we're going to need to introduce more types that are intermediate between `String` and `Sentence`.

JacquesCarette (changes requested):
A lot of this is great and makes sense, except for one tiny bit...

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
ManvendraJani: Is this something that you want done in this PR or are you talking more generally?

JacquesCarette: I don't want this PR to change this particular type to `Sentence`, that seems like too big a change.

ManvendraJani:
@samm82 is this good to be merged or should this be closed/reconsidered?

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
samm82: @ManvendraJani It seems like @JacquesCarette is suggesting still using the `Identifier` type, but defining it to be a subtype of `Sentence` (e.g., that enforces certain properties) and that this PR shouldn't be merged until that is taken care of

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
```diff
@@ -69,13 +69,16 @@ data RawContent =
...
-  | Defini DType [(Sentence, [Contents])]    -- ^ Defines something with a type, identifier, and 'Contents'.
+  | Defini DType [(Identifier, [Contents])]    -- ^ Defines something with a type, identifier, and 'Contents'.
   | Figure Lbl Filepath MaxWidthPercent      -- ^ For creating figures in a document. Should use relative file path.
   | Bib BibRef                               -- ^ Grants the ability to reference something.
   | Graph [(Sentence, Sentence)] (Maybe Width) (Maybe Height) Lbl -- ^ Contain a graph with coordinates ('Sentence's), maybe a width and height, and a label ('Sentence').
   | CodeBlock CodeExpr                       -- ^ Block for codes
                -- TODO: Fill this one in.
 
+-- |  An identifier is just a 'Sentence'.
+type Identifier = Sentence
```
JacquesCarette, line 80: This is not what I meant. You should be able to use complete `Sentence`s as `Indentifier`s.  Only a fairly restricted set of constructors should be allowed.

The only reasonable way to implement that is as a new type (via `data`) that has functions to transform it to a `Sentence` in latter passes.

The way forward will need to first figure out who creates `Defini` instances, and why are they (now) using a `Sentence` instead of something simpler.

Issue #3459, opened 2023-06-02
Make function for formatting ordinals (1st, 2nd, 3rd, ...) consistently


samm82:
_Note that this should probably be tackled in multiple PRs, but since the fixes are interconnected, I included them all in one issue._

I noticed that we don't really have a consistent way of dealing with ordinals, and our current methods are either really bad or inconsistent (see below for specific examples).

# TODO
- [ ] Improve GamePhysics's handling of quantities with ordinals in their definition to refer to the actual index $j$ (creating it if it doesn't exist)
    - This should involve creating a new function, that should be able to be reused for other instances of ordinals
- [ ] Use this function where applicable in SSP
- [ ] Introduce handling of ordinals from just straight numbers, either in a new function or adding functionality to the previously mentioned one
    - This will be interesting with Haskell's strong typing
- [ ] Decide on how to format ordinals: "_j_ th" vs. "_j_-th " vs. "_j_<sup>th</sup>" -  "_j_ th" shouldn't have a space, but Markdown doesn't needs it
    - This is the lowest priority and can be done at any time
    - My current vote is "_j_ th" since it seems the most reusable and not unnecessarily complicated; what do you think @smiths?

## Really Bad (GamePhysics)
![image](https://github.com/JacquesCarette/Drasil/assets/35857611/164da931-0cb9-48fa-9a7c-bbe9251a032d)
Example Implementation:

https://github.com/JacquesCarette/Drasil/blob/9153523ef3bed37e20808b26e0b4f2854d1b3ebf/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs#L185-L187

Affected Unitals: `massj`, `posj`, `accj`, `angAccj`, `velj`, `torquej`, `massIRigidBody`, `rRot`, and `forcej`
_There are also some comments that should either be removed in a separate PR or updated to use the new function:_
https://github.com/JacquesCarette/Drasil/blob/9153523ef3bed37e20808b26e0b4f2854d1b3ebf/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/IMods.hs#L154-L163

## Inconsistent (SSP)
![image](https://github.com/JacquesCarette/Drasil/assets/35857611/d1a7c606-b60a-4d8d-a28a-ae1900204b50)

[...]

PR #3461, opened 2023-06-02
Update `sufxer` to deal with mod 100

samm82:
While making #3459, I came across [this Stack Exchange post](https://codegolf.stackexchange.com/questions/4707/outputting-ordinal-numbers-1st-2nd-3rd) that pointed out that the pattern 11, 12, and 13 using "th" repeats every 100 (e.g., 11th, 112th, 498713th).

This PR implements this based on the work done in #3421


harmanpreet-sagar (approved):
Looks good to me! 😄

JacquesCarette (approved):

PR #3478, opened 2023-06-12
Removes duplicate 'the's

### LINKED ISSUE #3234: Duplicate "the" (coming from `the_ofThe`, etc.?)

ManvendraJani:
closes #3234 

Note: 
The 'term' and definition of `UnitalChunk` are being abused somewhere in the code, for example currently the 'term' for lenRod_1 is 
```Haskell
len `ofThe` firstRod
```
and the definition  is defined as
```Haskell
len `the_ofThe` firstRod
```
but use cases for both appear the same in the generated files. I believe this is because the definition is never actually being used. My question was if an issue should be raised about this or is this something we aim to clean up along with the [Chunk Investigation](https://github.com/JacquesCarette/Drasil/projects/10)


smiths (approved):
The changes to stable look good to me.  As usual, I haven't reviewed how the changes were made, but the generated files look good.  :smile:

samm82 (commented):
Skimming the changes to the Drasil code, these changes look good. However, the changes from #3455 seem to also be present here. The "proper" solution would be to revert those commits from this branch so that only the commits relevant to the "the" issue are present. An alternative would be to mark this as a draft until #3455 is merged in.

Each PR should be as small as possible to make it easy to see exactly what was changed. In future, make sure to be branching off of `master` instead of another branch (unless you specifically need changes from a branch to build off of them)!

code/drasil-data/lib/Data/Drasil/Concepts/Documentation.hs
```diff
@@ -173,6 +173,7 @@ module_         = nc "module"         (cn'    "module"             )
 model           = nc "model"          (cn'    "model"              )
 name_           = nc "name"           (cn'    "name"               )
 nonfunctional   = nc "non-functional" (cn'    "non-functional"     ) --FIXME: Adjective
+note_            = nc "note"           (cn'    "note"               )
```
JacquesCarette, line 176: nitpick: line it up properly!

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -83,7 +86,7 @@ makeDerivCons (E e) = EqnBlock e
 makeDerivCons s     = Paragraph s
 
 -- | Synonym for easy reading. Model rows are just 'String',['Contents'] pairs.
-type ModRow = [(String, [Contents])]
+type ModRow = [(Sentence, [Contents])]
```
JacquesCarette, line 89: I don't know if we want the first bit of a `ModRow` to be a full `Sentence`. Why is this change needed?

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -213,17 +214,16 @@ descPairs IncludeUnits =
...
-  show RefBy             = "RefBy"
-  show Source            = "Source"
-  show Input             = "Input"
-  show Output            = "Output"
-  show InConstraints     = "Input Constraints"
-  show OutConstraints    = "Output Constraints"
-  show DefiningEquation  = "Equation"
-  show (Description _ _) = "Description"
-  show Notes             = "Notes"
+getSent :: Field -> Sentence
```
JacquesCarette, line 217: I think the source of the problem is here: `getSent` goes immediately into "too big" a data-structure.

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
JacquesCarette: Also, these changes shouldn't be in this PR at all, they are about something other than duplicate 'the'.

code/stable/dblpendulum/SRS/HTML/DblPendulum_SRS.html
```diff
@@ -167,22 +167,22 @@ <h2>Table of Symbols</h2>
                 </tr>
                 <tr>
                   <td><em>a<sub>x1</sub></em></td>
-                  <td>The horizontal acceleration of the first object</td>
+                  <td>Horizontal acceleration of the first object</td>
```
JacquesCarette, line 170: So this change is by design? I like it, I prefer this without the extra leading `The`.

JacquesCarette (changes requested):

code/stable/dblpendulum/SRS/HTML/DblPendulum_SRS.html
ManvendraJani: This was the only change that was supposed to be committed, I accidentally branched off of my working branch for another issue.

ManvendraJani:
@samm82 is it ok to just merge this into the changes from #3455?

PR #3487, opened 2023-06-15
Added a 'How to Change Input' Section to README

### LINKED ISSUE #3214: Projectile: README add how to run instructions
Projectile reads the input from a file and writes the output to a file.  There is nothing wrong with the decision to use files, but this decision isn't documented anywhere.  The SRS is rightfully abstract, just talking about the generic ideas of input and output.  The decision to use files for input and output is a design decision that I don't believe we have explicitly made.  In the future, as we explore how we make and document the decisions that bridge us from requirements to design, we'll likely make the design decision explicit.  In the meantime, we should give the user information on how to use the software.  In general this is useful information to find in a README file (#3159).

To the projectile example README file, we should add additional text on running the Projectile code.  After the instructions on using make, we could add something like the following text:

"Projectile will take the inputs from input.txt and write the outputs to output.txt.  Inputs can be changed by editing input.txt."

ManvendraJani:
closes #3214

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -54,6 +54,10 @@ type ExamplePurpose = String
 type ExampleDescr = String
 -- | File contributors
 type Contributor = String
+-- | Input File
+type InFile = Maybe String
+-- | Output File
+type OutFile = Maybe String
 -- | Holds all information needed to create a README file.
 data ReadMeInfo = ReadMeInfo {
   langName :: LangAbbrev,
@@ -67,5 +71,6 @@ data ReadMeInfo = ReadMeInfo {
   caseName :: CaseName,
   examplePurpose :: ExamplePurpose,
   exampleDescr :: ExampleDescr,
-  folderNum :: Int
+  folderNum :: Int,
+  inputOutput :: (InFile, OutFile)
 }
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -55,9 +55,9 @@ type ExampleDescr = String
 -- | File contributors
 type Contributor = String
 -- | Input File
-type InFile = String
+type InFile = Maybe String
 -- | Output File
-type OutFile = String
+type OutFile = Maybe String
 -- | Holds all information needed to create a README file.
 data ReadMeInfo = ReadMeInfo {
   langName :: LangAbbrev,
@@ -72,5 +72,6 @@ data ReadMeInfo = ReadMeInfo {
   examplePurpose :: ExamplePurpose,
   exampleDescr :: ExampleDescr,
   folderNum :: Int,
-  inputOutput :: Maybe (InFile, OutFile)
+  inputf :: InFile,
+  outputf :: OutFile
 }
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -55,9 +55,9 @@ type ExampleDescr = String
 -- | File contributors
 type Contributor = String
 -- | Input File
-type InFile = Maybe String
+type InFile = String
 -- | Output File
-type OutFile = Maybe String
+type OutFile = String
 -- | Holds all information needed to create a README file.
 data ReadMeInfo = ReadMeInfo {
   langName :: LangAbbrev,
@@ -72,5 +72,5 @@ data ReadMeInfo = ReadMeInfo {
   examplePurpose :: ExamplePurpose,
   exampleDescr :: ExampleDescr,
   folderNum :: Int,
-  inputOutput :: (InFile, OutFile)
+  inputOutput :: Maybe (InFile, OutFile)
 }
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -55,7 +55,7 @@ type ExampleDescr = String
 -- | File contributors
 type Contributor = String
 -- | Input File
-type InFile = String -- There may not always be an Input/Output File
+type InFile = String -- TODO: There may not always be an Input/Output File
 -- | Output File
 type OutFile = String
 -- | Holds all information needed to create a README file.
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -55,9 +55,9 @@ type ExampleDescr = String
 -- | File contributors
 type Contributor = String
 -- | Input File
-type InFile = Maybe String
+type InFile = String -- There may not always be an Input/Output File
 -- | Output File
-type OutFile = Maybe String
+type OutFile = String
 -- | Holds all information needed to create a README file.
 data ReadMeInfo = ReadMeInfo {
   langName :: LangAbbrev,
@@ -72,6 +72,5 @@ data ReadMeInfo = ReadMeInfo {
   examplePurpose :: ExamplePurpose,
   exampleDescr :: ExampleDescr,
   folderNum :: Int,
-  inputf :: InFile,
-  outputf :: OutFile
+  inputOutput :: (InFile, OutFile)
 }
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer.hs
@@ -1,6 +1,6 @@
 module Language.Drasil.Code.Imperative.GOOL.LanguageRenderer (
   -- * Common Syntax
-  doxConfigName, makefileName, sampleInputName, sampleOutputName, readMeName
+  doxConfigName, makefileName, sampleInputName, readMeName
 ) where
 
 ----------------------------------------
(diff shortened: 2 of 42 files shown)


code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -44,13 +44,20 @@ commandLine = text $ "In your terminal command line, enter the same directory as
...
 -- | Helper for giving instructions on how to build the program.
 makeInstDoc :: Doc
 makeInstDoc = text "How to Build the Program:" <> contSep <> commandLine <> contSep <>
     bkQuote3 <> contSep <> text "make build" <> contSep <> bkQuote3
 
+-- | Helper for giving instuctions and Input and Output files.
+inOutFile :: String -> Doc
+inOutFile name = text "How to Change Input:" <> contSep <> text name <+> 
+    text "will take the inputs from input.txt and write the outputs to output.txt." 
+    <> contSep <> text "Inputs can be changed by editing input.txt."
```
samm82, line 59: Is there a way to pull "input.txt" and "output.txt" from the filenames themselves? Do we actually specify the file names anywhere or is this hardcoded when we actually generate them?

samm82 (commented):
Other than my note about where the filenames come from, this looks good to me! I also like how this change was meant for Projectile, but then was trivial to just automatically add it to all examples; perhaps this could be an example of a small change that propagates? @smiths

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
JacquesCarette: I agree that those should be parameters to `inOutFile` even if they end up being hard-coded constants elsewhere in the code.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -22,11 +22,11 @@ makeMd = vcat . punctuate secSep . filtEmp
...
 
--- | Instruction section, contains 3 paragraphs, Running, Building and Config Files.
+-- | Instruction section, contains 4 paragraphs, Running, Building, Input-Output and Config Files.
 -- The Config file section is only displayed if there are configuration files.
-instDoc :: [String] -> Doc
-instDoc cfp = regularSec (text "Making Examples") 
-    (runInstDoc <> doubleSep <> makeInstDoc) <> configSec cfp 
+instDoc :: [String] -> String -> Doc
+instDoc cfp name = regularSec (text "Making Examples") 
+    (runInstDoc <> doubleSep <> makeInstDoc <> doubleSep <> inOutFile name) <> configSec cfp
```
JacquesCarette, line 29: The instructions for changing input seem to be unconditional, i.e. input/output will always be done the same way (via files).

While this is true now, I don't think it will remain so. So, in a way, this code here makes an implicit assumption about the architecture of the generated code. And when we change that, it won't be obvious that this code needs to change as well.

So I'm unconfortable with the code as is.

JacquesCarette (changes requested):
I'm saying "request changes" but I think this points to a need for a deeper design, not just code changes.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -44,13 +44,25 @@ commandLine = text $ "In your terminal command line, enter the same directory as
...
 -- | Helper for giving instructions on how to build the program.
 makeInstDoc :: Doc
 makeInstDoc = text "How to Build the Program:" <> contSep <> commandLine <> contSep <>
     bkQuote3 <> contSep <> text "make build" <> contSep <> bkQuote3
 
+-- | Helper for giving instuctions and Input and Output files.
+inOutFile :: String -> (Maybe String, Maybe String) -> Doc
+inOutFile _    (Nothing, Nothing) = empty
+inOutFile _    (Nothing, _      ) = empty
+inOutFile _    (_      , Nothing) = empty
```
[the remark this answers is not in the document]
ManvendraJani: In light of the comment made by @JacquesCarette there may not be an input/output file in which case we would not generate that section.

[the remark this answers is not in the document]
ManvendraJani: although I realize that I may be misusing the Maybe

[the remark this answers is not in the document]
samm82: No, I understand now. Essentially, we don't care about what happens if there isn't an input file and/or an output file until we actually implement that possibility - makes sense to me!

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -53,15 +53,14 @@ makeInstDoc = text "How to Build the Program:" <> contSep <> commandLine <> cont
     bkQuote3 <> contSep <> text "make build" <> contSep <> bkQuote3
 
 -- | Helper for giving instuctions and Input and Output files.
-inOutFile :: String -> (Maybe String, Maybe String) -> Doc
-inOutFile _    (Nothing, Nothing) = empty
-inOutFile _    (Nothing, _      ) = empty
-inOutFile _    (_      , Nothing) = empty
-inOutFile name (Just inFile, Just outFile) = doubleSep <>
```
samm82: I liked this better; what if we have an input file but no output file? I'm not sure if I like them being wrapped in a tuple, though....

samm82: Also make sure to be giving commits descriptive messages; there is a lot more being done here than just "fixing generated text". 😅

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -22,11 +22,11 @@ makeMd = vcat . punctuate secSep . filtEmp
...
 
--- | Instruction section, contains 3 paragraphs, Running, Building and Config Files.
+-- | Instruction section, contains 4 paragraphs, Running, Building, Input-Output and Config Files.
 -- The Config file section is only displayed if there are configuration files.
-instDoc :: [String] -> Doc
-instDoc cfp = regularSec (text "Making Examples") 
-    (runInstDoc <> doubleSep <> makeInstDoc) <> configSec cfp 
+instDoc :: [String] -> String -> Maybe String ->  Maybe String -> Doc
+instDoc cfp name inf outf = regularSec (text "Making Examples") 
+    (runInstDoc <> doubleSep <> makeInstDoc) <> inOutFile name inf outf <> configSec cfp
```
samm82, line 29: If `inf == Nothing`,`runInstDoc` still includes the instructions `make run RUNARGS=input.txt`, even when there is no `input.txt`; this should be updated.

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
```diff
@@ -44,13 +44,25 @@ commandLine = text $ "In your terminal command line, enter the same directory as
...
+-- | Helper for giving instuctions and Input and Output files.
+inOutFile :: String -> Maybe String -> Maybe String -> Doc
+inOutFile _ Nothing _ = empty
+inOutFile _ _ Nothing = empty
+inOutFile name (Just inFile) (Just outFile) = doubleSep <>
+      text "How to Change Input:" <> contSep <> text name <+> 
+      text "will take the inputs from" <+> bkQuote <> text inFile <> bkQuote <+> 
+      text "and write the outputs to" <+> bkQuote <> text outFile <> bkQuote <> 
+      text "." <> contSep <> text "Inputs can be changed by editing" <+> bkQuote <> 
+      text inFile <> bkQuote <> text "."
```
samm82, line 64: Maybe add a comment to this function saying that it is to be expanded upon in the future, similar to the `-- This needs a more permanent solution` comment above.

samm82 (changes requested):

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
ManvendraJani: Currently, we have no alternative for this, should this simply be not generated in cases of input being empty for now?

samm82: Yes, or you can come up with an alternative message to be displayed

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -54,6 +54,10 @@ type ExamplePurpose = String
 type ExampleDescr = String
 -- | File contributors
 type Contributor = String
+-- | Input File
```
JacquesCarette, line 57: You should document, here, why there may not be an `Input File` (and same with output)

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer.hs
```diff
@@ -1,20 +1,22 @@
...
+  doxConfigName, makefileName, sampleInputName, sampleOutputName, readMeName
 ) where
 
 ----------------------------------------
 -- Syntax common to several renderers --
 ----------------------------------------
 
 -- | Common syntax for several renderers.
-doxConfigName, makefileName, sampleInputName, readMeName :: String
+doxConfigName, makefileName, sampleInputName, sampleOutputName, readMeName :: String
```
JacquesCarette, line 11: see issue #3490. Note that this shouldn't be fixed in this PR, but a subsequent one.

code/drasil-code/lib/Language/Drasil/Code/Imperative/WriteReadMe.hs
```diff
@@ -25,19 +25,21 @@ makeReadMe ReadMeInfo {
...
     verInfo progLang progLangVers,
     unsupOS unsupportedOSs,
     extLibSec extLibns extLibfp,
     endNote number auths] -- add date information to end note for license
 
 -- | Helper for encoding the type of program (either library or controller-based) in a README file.
-makeInstr :: ImplementationType -> [FilePath] -> String -> Doc
-makeInstr Library _ = const empty
-makeInstr Program cfp = instDoc cfp
+makeInstr :: ImplementationType -> [FilePath] -> String  -> Maybe String -> Maybe String -> Doc
```
JacquesCarette, line 40: 5 argument functions are a kind of 'design smell' that says that the wrong data is being passed-in. Maybe it's not worth unbundling some of it?

code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs
JacquesCarette: Personally, I think that this particular PR maybe should not have the `Maybe` for input or output, and simply focus on refactoring where the information comes from. A subsequent PR can implement a proper design for different architectures that may make some input/output optional.

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/WriteReadMe.hs
samm82: Sorry for my part in leading you astray! 😅😔

JacquesCarette (approved):

JacquesCarette:
@samm82 this is now blocked on your 'change requested'

samm82 (approved):
My bad!

PR #3499, opened 2023-06-23
3217 Projectile Improve Generated Makefiles

### LINKED ISSUE #3217: Projectile: Improve Generated Makefiles
The [Makefiles](https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/python/Makefile) could be improved by providing some header information that identifies the project.  A brief blurb about each rule in the Makefile would also be nice.

For Projectile, I propose that we provide Makefile comments with the name of the project (Projectile), and the purpose (Efficiently and correctly predict whether a launched projectile hits its target).  Since the makefile is generated, I don't think we want author information.  However, it would be nice to have a portion of the blurb from the README file that states that the Makefile is generated by Drasil.

We can decide in the future if we wish to have descriptions of the different makefile rules included as comments in the Makefile.

harmanpreet-sagar:
Closes #3217: Projectile: Improve Generated Makefiles. Changes made to `drasil-build`, `drasil-code`, `drasil-gool`, and stable for all examples.


smiths (changes requested):
I definitely like the addition of the program name and the purpose to the Makefiles.  This is a good example of reusing knowledge.  However, I don't like when the purpose says "To efficiently and correctly ...".  This is a fault in the original purpose knowledge that the use in the Makefiles highlights.  I don't like that we say "efficiently" because we don't actually do anything for efficiency.  We could eventually do more for efficiency, but even then I think the purpose should be the "functional purpose."  The purpose is what is to be calculated, without any modifiers about the qualities of the calculations.  I think we want the purpose to be simple.  I also think we should remove "correctly."  It is an empty statement.  We would never want to calculate an incorrect answer.  :smile:  Again, we could add a discussion about the accuracy of the solution, but that is another quality of the solution.  If we just focus on the functional purpose, we don't need to bring in other ideas about the quality of the solution.  (The qualities are still covered in the nonfunctional requirements; they are just left out of our purpose statement in the interest of making the purpose as simple and clear as possible.)

Please modify the examples that say "To efficiently and correctly ..." to "To ...".  We previously made this modification to Projectile.  Some of the examples have a different structure to the purpose.  They look fine to me.

harmanpreet-sagar:
I agree! I will make those changes right now!

smiths (commented):
The new purposes look great.

I have a new question though - why is the statement that the makefile is generated by Drasil in the README file?  I would think that the statement that the Makefile is generated would be in the Makefile itself.  I think there should be a comment at the end (beginning?) of the Makefile that says:

Generated by Drasil version <current version number>

I actually think all of the generated files should have a line like this.  The version number in particular seems helpful, although it will require a change to stable every time there is an update to Drasil.

I suggest you remove the changes to the README file and then create a new issue to modify all files to have the "Generated by Drasil version <x>" line.  We can then discuss that idea as a group, since I can't speak for everyone on whether that is a change in the desired direction.

harmanpreet-sagar:
The reason I put the statement "The makefile is generated by Drasil" in the READMEs is that I most probably misunderstood the original comment made in #3217 which can be viewed [here](https://github.com/JacquesCarette/Drasil/issues/3217#:~:text=However%2C%20it%20would%20be%20nice%20to%20have%20a%20portion%20of%20the%20blurb%20from%20the%20README%20file%20that%20states%20that%20the%20Makefile%20is%20generated%20by%20Drasil.).

I will remove the changes to the READMEs and will create a new issue for adding the "Generated by Drasil version " line in the Makefiles.

harmanpreet-sagar:
The new issue that has been created to add the "Generated by Drasil version" is #3501.

code/stable/dblpendulum/src/cpp/README.md
```diff
@@ -1,7 +1,7 @@
 # DblPendulum 
 > Author: Dong Chen
 
-> Purpose: efficiently and correctly predict the motion of a pendulum.
+> Purpose: predict the motion of a pendulum.
```
samm82, line 4: Should "predict" be capitalized? (This applies to the other READMEs as well)

harmanpreet-sagar: I believe that the reason that they are not is because READMEs obtain the purpose from the `purp` variable in the examples. The `purp` has the `Sentence` not capitalized because it is attached to a boilerplate text that results in a sentence such as the following: 

image

The boilerplate text is "A system is needed to".

samm82: Is there a way to capitalize a `Sentence` after it's been created? If not, there should be an issue created for that (if there isn't one already; I know we've started investigating something similar!)

harmanpreet-sagar: I was under the impression that the location at which the READMEs take the input for the purpose was a `Sentence`. However, it is a `String` and can be capitalized using the `capitalize` function. I am making that change right now and will commit the changes soon! Thanks for catching that!

samm82 (approved):

JacquesCarette (approved):

PR #3500, opened 2023-06-23
Added the GSD section to projectile

### LINKED ISSUE #3495: Add General System Description to Projectile
For Projectile, in the `Characteristics of Intended Reader` section, we reference the `User Characteristics` section.
image
This should be removed (only for projectile) as we do not generate the User Characteristics section for projectile.

ManvendraJani:
closes #3495


code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,69 @@ refDB = rdb citations concIns
 concIns :: [ConceptInstance]
 concIns = assumptions ++ funcReqs ++ goals ++ nonfuncReqs
 
--------------------------
+----------------------------------------
 -- Characteristics of Intended Reader --
--------------------------
+----------------------------------------
```
samm82, line 179: ❤️

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,69 @@ refDB = rdb citations concIns
...
+-----------------
+--SystemContext--
+-----------------
+
+sysCtxIntro :: Contents
+sysCtxIntro = foldlSP
+  [refS sysCtxFig1, S "shows the" +:+. phrase sysCont,
+   S "A circle represents an entity external to the", phrase software
+   `sC` phraseNP (the user), S "in this case. A rectangle represents the",
+   phrase softwareSys, S "itself", sParen (short projectileTitle) +:+. EmptyS,
```
samm82, line 194:
```suggestion
   phrase softwareSys, S "itself" +:+. sParen (short projectileTitle),
```

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,69 @@ refDB = rdb citations concIns
...
+  namedRef (SRS.assumpt ([]::[Contents]) ([]::[Section])) (phrase software +:+ plural assumption) +:+
+  S "are appropriate for any particular" +:+
+  phrase problem +:+ S "input to the" +:+. phrase software]
+
+sysCtxSysResp :: [Sentence]
+sysCtxSysResp = [S "Detect data type mismatch, such as a string of characters" +:+
+  phrase input_ +:+. S "instead of a floating point number",
+  S "Determine if the" +:+ plural input_ +:+ S "satisfy the required" +:+.
+  pluralNP (physical `and_` softwareConstraint),
+  S "Calculate the required" +:+. plural output_]
```
samm82, line 220: You could reduce the number of "+:+" used in `sysCtxUsrResp` and `sysCtxSysResp` by using `map foldlSent [[<Sentence 1 components>], [<Sentence 2 components], ... ]`

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,69 @@ refDB = rdb citations concIns
...
+
+sysCtxResp :: [Sentence]
+sysCtxResp = [titleize user +:+ S "Responsibilities",
+  short projectileTitle +:+ S "Responsibilities"]
```
samm82, line 224:
```suggestion
sysCtxResp = map (\x -> titleize x +:+ S "Responsibilities") [user, projectileTitle]
```
Of course, if `titleize projectileTitle` won't work for whatever reason, this will have to be modified.

code/drasil-example/projectile/lib/Drasil/Projectile/Figures.hs
```diff
@@ -1,13 +1,17 @@
...
 resourcePath :: String
 resourcePath = "../../../../datafiles/projectile/"
 
 figLaunch :: LabelledContent
 figLaunch = llcc (makeFigRef "Launch") $ figWithWidth (atStartNP (the physicalSystem))
   (resourcePath ++ "Launch.jpg") 70
+
+sysCtxFig1 :: LabelledContent
+sysCtxFig1 = llcc (makeFigRef "sysCtxDiag") $ fig (titleize sysCont) 
+  (resourcePath ++ "SystemContextFigure.png")
```
samm82, line 17: Was this image added in this branch? I can't find it

code/stable/projectile/SRS/HTML/Projectile_SRS.html
```diff
@@ -459,6 +467,71 @@ <h2>Organization of Document</h2>
...
+              <a href=#Figure:sysCtxDiag>Fig:sysCtxDiag</a> shows the system context. A circle represents an entity external to the software, the user in this case. A rectangle represents the software system itself (Projectile). Arrows are used to show the data flow between the system and its environment.
+            </p>
+            <div id="Figure:sysCtxDiag">
+              <figure>
+                <img src="../../../../datafiles/projectile/SystemContextFigure.png" alt="System Context" >
+                <figcaption>System Context</figcaption>
+              </figure>
+            </div>
+            <p class="paragraph">
+              The interaction between the product and the user is through an application programming interface. The responsibilities of the user and the system are as follows:
```
samm82, line 489: I don't think this is true; the user interacts with Projectile through an input file, unless that is what is meant by "application programming interface".

code/stable/projectile/SRS/HTML/Projectile_SRS.html
```diff
@@ -459,6 +467,71 @@ <h2>Organization of Document</h2>
...
+                  <li>
+                    Ensure required <a href=#Sec:Assumps>software assumptions</a> are appropriate for any particular problem input to the software.
+                  </li>
+                </ul>
+              </li>
+              <li>
+                Projectile Responsibilities
+                <ul class="list">
+                  <li>
+                    Detect data type mismatch, such as a string of characters input instead of a floating point number.
```
samm82, line 508: This will have interesting consequences for my testing project, since I don't think Projectile's code actually does this!

samm82 (changes requested):
Some comments are just notes, but there are some places where `Sentence` construction could be improved, in addition to a missing image file in `datafiles/` and a (possibly) incorrect description of user interaction with Projectile.

code/stable/projectile/SRS/HTML/Projectile_SRS.html
smiths: Although there is no code written to detect the type mismatch, Projectile does technically do this, since it will cause a run time error from the interpreter or compiler.  Having said that, if we had a language that wasn't typed, we could have a problem.  Even in languages with typing, it won't be a very graceful failure.

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
ManvendraJani: titleize will not work on projectileTitle, Is this ok to leave as is?

code/stable/projectile/SRS/HTML/Projectile_SRS.html
ManvendraJani: This is what is written accross other examples, We should create a different issue to fix this.

code/stable/projectile/SRS/HTML/Projectile_SRS.html
ManvendraJani: Should this line be removed? Because as said above, this is similar to what is said in other examples.

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
samm82: I would then change it to
```suggestion
sysCtxResp = map (\x -> x +:+ S "Responsibilities") [titleize user,
  short projectileTitle]
```
to reduce the amount of code duplication

code/stable/projectile/SRS/HTML/Projectile_SRS.html
samm82: Some other examples might _actually_ use an "application programming interface". Regardless, an issue is the best place to discuss this.

code/stable/projectile/SRS/HTML/Projectile_SRS.html
samm82: If other examples say it, my guess is that they don't actually implement this in the code, either. My guess is that this will come up during my project, so my vote is to leave this in for now and wait for it to be handled later

code/datafiles/projectile/SystemContextFigure.png
samm82, line 1: Minor nitpick, but there should be a comma after "velocity" to be consistent with how we format lists!

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,67 @@ refDB = rdb citations concIns
...
+
+sysCtxDesc :: Contents
+sysCtxDesc = foldlSPCol [S "The interaction between the", phraseNP (product_
+   `andThe` user), S "is through an application programming" +:+.
+   phrase interface, S "The responsibilities of the", phraseNP (user 
+   `andThe` system), S "are as follows"]
+
+sysCtxUsrResp :: [Sentence]
+sysCtxUsrResp = map foldlSent [[S "Provide initial", pluralNP (condition `ofThePS`
+  physical), S "state of the", phrase motion, S "and the", plural inDatum, S "related to the",
```
samm82, line 206:
```suggestion
  physical), S "state of the", phrase motion `andThe` plural inDatum, S "related to the",
```
Some other `Sentence` combinators could be used here, but since the string here only contains text that is captured in a combinator, it should really be used!

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -165,14 +174,67 @@ refDB = rdb citations concIns
...
+  phrase projectileTitle `sC` S "ensuring no errors in the", plural datum, S "entry"], 
+  [S "Ensure that consistent units are used for", pluralNP (combineNINI input_ Doc.variable)],
+  [S "Ensure required", namedRef (SRS.assumpt ([]::[Contents]) ([]::[Section])) 
+   (phrase software +:+ plural assumption), S "are appropriate for any particular",
+  phrase problem, S "input to the", phrase software]]
```
samm82, line 211: Since we have a concept of an `input_`, it (and the `toThe` combinator, if it exists!) should be used

samm82 (changes requested):
Some more nitpicks 😬

code/stable/projectile/SRS/HTML/Projectile_SRS.html
smiths: Yes, leave it in for now.  It is a good responsibility for the code to take on.  If we don't do it in an elegant way, we should address that (eventually), rather than change the requirements.

samm82 (approved):
As long as an issue for [this comment](https://github.com/JacquesCarette/Drasil/pull/3500#discussion_r1240175138) gets created, this PR looks good to me! Thanks for fixing all those little things 😅

smiths (approved):

PR #3502, opened 2023-06-23
Implement name generation in GOOL

hrzhuang:
## Scope of this PR

* All variables declared via methods of `DeclStatement` are marked as used
* Function parameters are marked as used
* `listSlice` and `stringListLists` generate variable names that are guaranteed not to conflict with the above

There are more places in GOOL that should mark variable names as used or take advantage of name generation, but I think this PR is a good starting point that provides a framework for future work.


JacquesCarette (approved):
Very nice! Really elegant changes throughout.

code/drasil-gool/lib/GOOL/Drasil/State.hs
```diff
@@ -548,7 +553,33 @@ getWordIndex = gets (^. (contentsIndices . _2))
...
 -- Helpers
 
 ifElemError :: (Eq a) => a -> [a] -> String -> [a]
-ifElemError e es err = if e `elem` es then error err else e : es
\ No newline at end of file
+ifElemError e es err = if e `elem` es then error err else e : es
+
+-- Split the longest numerical (0-9) suffix from the rest of the string
+splitVarName :: String -> (String, Maybe Int)
+splitVarName = over _2 readMaybe . over both reverse . swap . span isDigit . reverse
```
balacij, line 578: To avoid the 2 `reverse` calls, I think you can use `span (not . isDigit)` to immediately get the string component. Afterwards you can use `Bifunctor`s `second` to convert the remaining `String` into a number.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
```diff
@@ -1143,4 +1144,4 @@ typeDfltVal Char = litChar ' '
 typeDfltVal String = litString ""
 typeDfltVal (List t) = litList (convType t) []
 typeDfltVal (Array t) = litArray (convType t) []
-typeDfltVal _ = error "Attempt to get default value for type with none."
\ No newline at end of file
+typeDfltVal _ = error "Attempt to get default value for type with none."
```
balacij, line 1147: Underrated changes to each files. Drasils codebase has many files that are missing the trailing newline characters.

code/drasil-gool/lib/GOOL/Drasil/State.hs
```diff
@@ -548,7 +553,33 @@ getWordIndex = gets (^. (contentsIndices . _2))
 resetIndices :: MethodState -> MethodState
 resetIndices = set contentsIndices (0,0)
 
+useVarName :: String -> MethodState -> MethodState
+useVarName v = over (varNames . at prefix) (Just . max nextSuffix . fromMaybe 0)
+  where (prefix, nextSuffix) = over _2 (maybe 0 (+1)) $ splitVarName v
+
+genVarName :: [String] -> String -> MS String
```
balacij, line 560: It would be good to have some documentation about the sequence of variable names generated.

balacij (changes requested):

balacij:
Oops! I added some comments after you merged!

JacquesCarette:
No worries - I missed those. A new PR should be done to fix these. Good catch, thanks.

code/drasil-gool/lib/GOOL/Drasil/State.hs
hrzhuang: So the problem is that if I have `"ab12cb34"`, then `span (not . isDigit)` will give me `("ab", "12cb34")`, whereas I want `("ab12cb", "34")`.

bmaclach:
Great to see some nice improvements in GOOL 😀

smiths:
I haven't looked at how the improvements were made, but the changes to stable look like a move in the right direction.  Having loop indexes like i and j are much more idiomatic than what was previously in the code.

code/drasil-gool/lib/GOOL/Drasil/State.hs
hrzhuang: This is on me, I will document this better 😄

But basically `splitVarName` is used for both `useVarName` (to mark a user-chosen name as used) and `genVarName`. And `genVarName` generates a name based on a list of candidates chosen by the user (e.g. `i`, `j`, `k`). So `splitVarName` needs to handle any user-provided variable name.

PR #3504, opened 2023-06-27
Add vectors to GOOL

hrzhuang:
* Adds types and declaration statements for vectors
* Implements the vector operations add, scale, and dot product
* Vector operations result in "thunks," which can be assigned to variables


hrzhuang:
Since `listAccess` works on both lists and arrays, I assumed that this is true for all methods of the class `List`. I was mistaken. I am switching the underlying type for vectors from arrays to lists, since arrays currently do not implement all features needed for vectors, namely the ability to query the length of an array. I will create an issue for this, and possibly change vectors back to using arrays after it is resolved.

JacquesCarette (approved):

JacquesCarette:
There might be some code duplication here that could likely be abstracted out. But there probably is way more of that in GOOL already, so it doesn't feel worth holding up just now.

JacquesCarette:
(I didn't delete the branch quite yet, in case other people do catch things that ought to be fixed.)

PR #3509, opened 2023-06-28
Add GOOL examples to CI

hrzhuang:
This PR adds a step in the GitHub workflow to generate C++, Java, etc. from the GOOL examples under `drasil-code/test` and compile them to check for compile errors.


balacij:
Nice! This also reminds me that I need to clean up the CI. We should only compile artifacts if there was any changes to the artifacts themselves, to save CI time.

JacquesCarette (approved):
So few lines of code for so big / important a change!

JacquesCarette:
(will wait until build finishes though).

PR #3515, opened 2023-07-04
3501 add generated by drasil version text

### LINKED ISSUE #3501: Add "Generated by Drasil Version" text in all Drasil artifacts
Currently, the `Makefiles` do not have a statement that indicates that these `Makefiles` were generated by Drasil. As mentioned in the opening comment in #3217 and in this comment [here](https://github.com/JacquesCarette/Drasil/pull/3499#pullrequestreview-1495551883), a comment should be added at the beginning of the `Makefiles` that indicates the `Makefiles` were generated by Drasil.

A suggestion by @smiths was to add the following line "Generated by Drasil Version" where Version represents the version number of Drasil.

Adding the Version number in this line would require the [stable to be updated every time there is an update to Drasil](https://github.com/JacquesCarette/Drasil/pull/3499#:~:text=The%20version%20number%20in%20particular%20seems%20helpful%2C%20although%20it%20will%20require%20a%20change%20to%20stable%20every%20time%20there%20is%20an%20update%20to%20Drasil.). 

What are your thoughts on this @JacquesCarette @balacij @samm82 ?

harmanpreet-sagar:
Closes #3501. 2 new files added in `drasil-metadata`. Changes made to `drasil-gool`, `drasil-metadata`, `drasil-gen`, `drasil-code`, `drasil-build`, `drasil-example`, and stable


code/drasil-example/dblpendulum/app/Main.hs
```diff
@@ -13,6 +13,6 @@ main = do
   dumpEverything fullSI ".drasil/"
   typeCheckSI fullSI
   gen (DocSpec (docChoices SRS [HTML, TeX, JSON]) "DblPendulum_SRS") srs printSetting
-  genCode choices code
+  genCode choices code "../../../drasil-metadata/lib/Data/Drasil/DrasilConfiguration.json"
```
balacij, line 16: Maybe we can use template Haskell to inject the contents of the file inside a variable inside `drasil-metadata` instead of referring to the file URI and passing around a file URI + re-reading it for each project, when it's a static thing.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CSharpRenderer.hs
```diff
@@ -121,10 +121,10 @@ instance OOProg CSharpCode where
 
 instance ProgramSym CSharpCode where
   type Program CSharpCode = ProgData
-  prog n st files = do
+  prog n st v files = do
```
balacij, line 124: I'm getting the feeling that there should just be a record passed around for each typeclass as we can probably further add more to the record, but not everything in the record will be used everywhere. For example, license references. @JacquesCarette ?

code/stable/dblpendulum/src/cpp/Calculations.hpp
```diff
@@ -1,6 +1,7 @@
 /** \file Calculations.hpp
     \author Dong Chen
     \brief Provides functions for calculating the outputs
+    \Generated by Drasil Version: "v0.1-alpha"
```
balacij, line 4: Since you have the colon to separate the variable component of the text, you can probably remove the quotation marks around the version information. Though I might prefer "Generated by Drasil v0.1-alpha". Maybe format of the information message can be variable too?

balacij (changes requested):

code/drasil-build/lib/Build/Drasil/Make/Print.hs
```diff
@@ -39,3 +39,8 @@ printCmd (C c opts) = text $ (if IgnoreReturnCode `elem` opts then "-" else "")
 -- | Renders multiple commands.
 printCmds :: [Command] -> Doc
 printCmds = foldr (($+$) . (<>) tab . printCmd) empty
+
+noFirstElement :: [a] -> [a]
```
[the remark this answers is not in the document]
harmanpreet-sagar: I will change that right now!

code/drasil-build/lib/Build/Drasil/Make/Print.hs
```diff
@@ -25,8 +25,8 @@ printRule (R t d _ c) = printTarget t d $+$ printCmds c
 
 -- | Gathers all rules to abstract targets and tags them as phony.
 printPhony :: [Rule] -> Doc
-printPhony = (<+>) (text ".PHONY:") . hsep . tail . map (\(R t _ _ _) -> text $ renderMS t) . 
-  filter (\(R _ _ t _) -> t == Abstract)
+printPhony = (<+>) (text ".PHONY:") . hsep . noFirstElement . tail . map (\(R t _ _ _) -> text $
```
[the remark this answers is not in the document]
harmanpreet-sagar: The reason that there are two tail calls here is because otherwise, the last line of a makefile would read `PHONY: # Generated by Drasil Version build run doc` instead of `PHONY: build run doc`, which would make the remaining part of the line a comment

code/drasil-code/test/VectorTest.hs
```diff
@@ -7,7 +7,7 @@ import GOOL.Drasil (GSProgram, SVariable, SMethod, OOProg, ProgramSym(..),
   ModuleSym(..))
 
 vectorTest :: OOProg r => GSProgram r
-vectorTest = prog "VectorTest" "" [fileDoc $ buildModule "VectorTest" []
+vectorTest = prog "VectorTest" "" "" [fileDoc $ buildModule "VectorTest" []
```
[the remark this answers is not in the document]
harmanpreet-sagar: Since we are using the `prog` function to define the `VectorTest`, we can manually input the version number but not automatically. If you want I can manually input the version number into it.

code/drasil-example/dblpendulum/app/Main.hs
harmanpreet-sagar: Yes, that is a good idea. I will make that change right now.

code/stable/dblpendulum/src/cpp/Calculations.hpp
harmanpreet-sagar: Ok, I will change that right now!

code/drasil-code/package.yaml
```diff
@@ -2,7 +2,7 @@ name:                drasil-code
 version:             0.1.9.0
 author:              "Dan Szymczak, Steven Palmer, Jacques Carette, Spencer Smith"
 maintainer:          "Jacques Carette"
-synopsis:	     A framework for code and document generation for scientific software - Code SubPackage
+synopsis:	           A framework for code and document generation for scientific software - Code SubPackage
```
samm82, line 5: I don't know if this is even an issue, but what's the reason for this extra space?

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
```diff
@@ -0,0 +1,25 @@
...
+obtainVersion num = do
+  input <- B.readFile $ drasilConfigPath num
+  let dc = decode input :: Maybe DrasilConfig
+  case dc of
+    Nothing -> return "error parsing JSON"
+    Just d -> return ((show.version) d)
+
+-- The current Drasil version
+drasilVersion :: String
+drasilVersion = "v0.1-alpha"
```
samm82, line 25: If we're reading from the JSON, shouldn't this be removed?

code/stable/dblpendulum/src/cpp/Calculations.hpp
```diff
@@ -1,6 +1,7 @@
 /** \file Calculations.hpp
     \author Dong Chen
     \brief Provides functions for calculating the outputs
+    \"Generated by Drasil v0.1-alpha"
```
samm82, line 4: Do we want these quotation marks? I personally don't think so

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
harmanpreet-sagar: You're right! I will remove this right now. I missed removing that part.

code/stable/dblpendulum/src/cpp/Calculations.hpp
balacij: I was saying that we should remove the quotation marks in my message. I can see how I phrased my message awkwardly, sorry about that...

samm82: Yeah I figured the quotation marks in your message were just saying what should be displayed and shouldn't be included in the actual output

code/drasil-example/dblpendulum/app/Main.hs
balacij: While not exactly what I had in mind for a solution, it looks like dfdcbca33d6fe9c1b8ed8cff17b443948c49e441 does the trick. Since the "3" is a special number, though, we might as well have it hard-coded in the path too (and not passed around as an argument by things). Have you made sure that the path works fine for in the example subprojects too?

code/drasil-code/test/VectorTest.hs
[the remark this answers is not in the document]
balacij: Ah, gotcha. It's probably fine as is.

code/drasil-build/lib/Build/Drasil/Make/Print.hs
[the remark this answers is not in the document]
balacij: Oh, that sounds odd. The comment in the Makefile shouldn't be a `Rule` type, it should just be a comment. Looking at Drasils [definition of a `Makefile`](https://github.com/JacquesCarette/Drasil/blob/0241639a27891e96ebe1c22733a833a95ea1652f/code/drasil-build/lib/Build/Drasil/Make/AST.hs#L7-L11), it looks like it isn't quite sufficient. A `Makefile` isn't exactly a sequence of rule definitions, which is why we (well, you) had to use `Rule` to fix that. It would be good to have a base PR (for this PR) that adds just enough structure to the definition of `Makefile`s such that we can have comments interspersed between rule definitions. Afterwards, we can improve the definition of `Makefile`s to use the same 'statements' system that `Makefile`s use.

What do you think?

code/stable/dblpendulum/src/cpp/Calculations.hpp
harmanpreet-sagar: Ohhh ok! My bad. I will change that right now!

code/drasil-code/package.yaml
harmanpreet-sagar: So should I revert that change as the spacing may vary on each person's machine then?

code/drasil-example/dblpendulum/app/Main.hs
harmanpreet-sagar: When you say subprojects, you are referring to the projectile example, right? If we are talking about the projectile example, the only location where I am adding a number for the `genCode` is in the `Choices.hs` file. In that, I am putting in 4 instead of 3.

smiths (commented):
I reviewed the changes to stable and they look good to me.

code/drasil-example/dblpendulum/app/Main.hs
samm82: My guess is that this number will be added to some sort of record at some point (similarly to the number used for getting the file path of the Drasil logo)

harmanpreet-sagar: Currently, it is not added to any record as the `genCode` function is not a record and this number is only being used at a very early stage that it is not part of a `DrasilState` or `CodeSpec` or any of those major datasets. We do not have a record for this right now as this would be the only member of the record.  I am not adding it into the `DrasilConfig` record because that is created specifically for obtaining the data in the `DrasilConfiguration.json` file.

samm82: Agreed!

harmanpreet-sagar: Is what I have done in 95a5942d57b294c91f16e968f73a8a46d3f97829 what you were talking about or am I misunderstanding it @balacij?

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
```diff
@@ -10,10 +10,22 @@ import Utils.Drasil(drasilConfigPath)
...
+defaultFolderVal = 3
+
 instance FromJSON DrasilConfig
 
-obtainVersion :: Int -> IO String
-obtainVersion num = do
+-- Reads the DrasilConfiguration.json file and extracts the data from that file.
+obtainVersion :: Maybe Int -> IO String
+obtainVersion Nothing = do
+  input <- B.readFile $ drasilConfigPath defaultFolderVal
```
samm82, line 23: Instead of pattern matching, using [`fromMaybe`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Maybe.html#v:fromMaybe):

```suggestion
obtainVersion num = do
  input <- B.readFile $ drasilConfigPath $ fromMaybe defaultFolderVal num
```

will let you get rid of the `obtainVersion (Just num)` case below, reducing duplication

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
```diff
@@ -10,10 +10,22 @@ import Utils.Drasil(drasilConfigPath)
...
+  let dc = decode input :: Maybe DrasilConfig
+  case dc of
+    Nothing -> return "error parsing JSON"
+    Just d -> return ((show.version) d)
+
+obtainVersion (Just num) = do
```
samm82, line 29: This entire case (including the parts I can't select because they weren't changed in this PR) should be removed after using `fromMaybe` above

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
```diff
@@ -10,10 +10,22 @@ import Utils.Drasil(drasilConfigPath)
 
 newtype DrasilConfig = DrasilConfig {version :: String} deriving Generic
 
+-- The number of folders that needs to be exited in order to obtain the
```
samm82, line 13:
```suggestion
-- The number of folders that need to be exited in order to obtain the
```

Since it is the _folders_ that need to be exited, not the _number_ of folders, the verb should be plural.

samm82 (changes requested):

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
smiths: While we are looking at this comment, "in order to" should just be "to".  :smile:

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
harmanpreet-sagar: Thanks for that suggestion! I did not know that was a possibility!

code/drasil-metadata/lib/Data/Drasil/DrasilConfig.hs
samm82: See [this list of low-information content phrases](https://www.webpages.uidaho.edu/range357/extra-refs/empty-words.htm)

harmanpreet-sagar: This is a very good resource! Thank you for that!

code/drasil-build/lib/Build/Drasil/Make/Print.hs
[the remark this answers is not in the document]
harmanpreet-sagar: I agree! The `Makefile` should be able to have comments separate from the Rules. I like the idea of creating another PR  which provides a further structure to the way `Makefile`  as that would improve the creation of `Makefile`s.

samm82 (approved):
I'm approving this so my "request changes" review doesn't block this PR's eventual merge, but I think (at least) @balacij should approve it as well before it does get merged

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CSharpRenderer.hs
JacquesCarette: We should definitely be looking at bundling some data together whenever they meaningfully belong as a unit.

code/drasil-code/package.yaml
JacquesCarette: Definitely never use tabs.

code/drasil-gool/lib/GOOL/Drasil/AST.hs
```diff
@@ -107,13 +107,14 @@ updateParam :: (Doc -> Doc) -> ParamData -> ParamData
 updateParam f v = pd (paramVar v) ((f . paramDoc) v)
 
 -- Used as the underlying data type for Programs in all renderers
-data ProgData = ProgD {progName :: String, progPurp :: String, progMods :: [FileData]}
+data ProgData = ProgD {progName :: String, progPurp :: String, 
+  drasilVers :: String, progMods :: [FileData]}
 
-progD :: String -> String -> [FileData] -> ProgData
-progD n st fs = ProgD n st (filter (not . isEmpty . modDoc . fileMod) fs)
+progD :: String -> String -> String -> [FileData] -> ProgData
```
JacquesCarette, line 113: Here is where the unbundled arguments have gone crazy. Above the threshold that I can bear.

code/drasil-gool/lib/GOOL/Drasil/ClassInterface.hs
```diff
@@ -58,7 +58,7 @@ class (ProgramSym r, VectorType r, VectorDecl r, VectorThunk r,
 
 class (FileSym r) => ProgramSym r where
   type Program r
-  prog :: Label -> Label -> [SFile r] -> GSProgram r
+  prog :: Label -> Label -> Label -> [SFile r] -> GSProgram r
```
JacquesCarette, line 61: This change (and the one to `docMod` also seem to go too far.

JacquesCarette (changes requested):

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -117,21 +118,23 @@ genLog SI{_sys = sysName} pinfo = do
   setCurrentDirectory workingDir
 
 -- | Calls the code generator.
-genCode :: Choices -> CodeSpec -> IO ()
-genCode chs spec = do
+genCode :: Choices -> CodeSpec -> Maybe Int -> IO ()
```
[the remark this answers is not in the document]
samm82: If I remember correctly, I think this was just going to be introduced as a proof of concept and then we would later decide on a design that encapsulated it either with `Choices`, `CodeSpec`, or something else. It seems like this should be figured out in this PR 😅

code/drasil-example/dblpendulum/app/Main.hs
balacij: I was meaning that we could/should use [TemplateHaskell](http://wiki.haskell.org/Template_Haskell) to embed the file contents at Drasils compile-time. I made an [MVP demo](https://github.com/balacij/mvp-TemplateHaskellReadFile/blob/main/src/Lib.hs) that you can use to try it out, basically just wrapping a [stackoverflow answer](https://stackoverflow.com/a/44369564/16760741) in a Stack project.

I'm not sure if we want to go down this route because it's TemplateHaskell, but I think it's better than passing around a number related to the current working directory and the version json file.

Pinging @JacquesCarette for his thoughts here, since he was noted the `Maybe Int` in a later comment.

JacquesCarette: Yes, a more sophisticated implementation would use TH. We already use TH, so it's not a big deal to use it some more.

It does feel like maybe the first implementation could be simpler than using TH. Even with TH, we'd need to pass information around, and it shouldn't be a new argument, so the `Maybe Int` design issue is independent of that.

balacij: > Even with TH, we'd need to pass information around, and it shouldn't be a new argument, so the Maybe Int design issue is independent of that.

I think that if we computed the TH in `drasil-metadata`, and referenced `drasil-metadata` everywhere else, we _shouldn't_ (hopefully) need to play the "what is the folder depth?"/`Maybe Int` game.

JacquesCarette: Good point - I agree with that assessment.

So, maybe it should be done now? Can you help @harmanpreet-sagar with that?

harmanpreet-sagar: I have created a new PR #3566 to resolve #3501. In that while not explicitly taking in any input for the folder depth, I used the number that we have specified in the `Choices` regardless for the `README` generations and modified it slightly to generate the length as I determined a pattern for that. This way, we are not taking in any input for the specific folder depth nor do we have any `Maybe Int` issues. 

I unfortunately was unable to determine how to use TH in `drasil-metadata` so I used another method. Thus, if we would like to implement that, I might need a little bit of assistance in using TH.

harmanpreet-sagar:
This PR was a lot complicated in incorporating the changes and resolving any conflicts that arose from merging `master`. Hence, I created a new PR, #3566 to resolve issue #3501.

code/drasil-utils/lib/Utils/Drasil/Lists.hs
```diff
@@ -46,3 +46,9 @@ foldle1 f g (x:y:xs) = foldle f g (f x y) xs
 -- | Convert "row" of elements into "column" of elements.
 toColumn :: [a] -> [[a]]
 toColumn = map (: [])
+
+-- | Helper to remove the first element of a list if it exists.
+noFirstElement :: [a] -> [a]
+noFirstElement [] = []
+noFirstElement [_] = []
```
balacij, line 53: I think you can remove the 1-element case. A better name might be 'safeTail'.

code/drasil-code/test/VectorTest.hs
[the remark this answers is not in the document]
balacij: Since we're going to try out the `drasil-metadata` approach, I think this is possible now.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -47,8 +47,8 @@ import Text.PrettyPrint.HughesPJ (isEmpty, vcat)
 -- | Initializes the generator's 'DrasilState'.
 -- 'String' parameter is a string representing the date.
 -- \['Expr'\] parameter is the sample input values provided by the user.
-generator :: Lang -> String -> [Expr] -> Choices -> CodeSpec -> DrasilState
-generator l dt sd chs spec = DrasilState {
+generator :: Lang -> String -> [Expr] -> Choices -> CodeSpec -> String -> DrasilState
+generator l dt sd chs spec vers = DrasilState {
```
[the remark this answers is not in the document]
balacij: Since we're going to try out the TH approach, I don't think we need to add a parameter at all now. We can just directly refer to the version held in `drasil-metadata` wherever `vers` is currently used.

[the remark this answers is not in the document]
balacij: Ah, this comment applies to everything below too where the extra option is added.

code/drasil-example/dblpendulum/app/Main.hs
balacij: Let's try out the TH approach, I think it will be easiest to work with once we have it working. Did you have a look at the prototype I linked @harmanpreet-sagar ? I'll message you on Discord too.

JacquesCarette:
So should this PR be closed in favour of the later one?

harmanpreet-sagar:
PR closed in favour of #3566

PR #3516, opened 2023-07-04
Generate __init__.py files for ease of importing for testing

samm82:
To improve the imports of files while keeping the test files in a separate directory, I used `__init__.py` files to allow for a more consistent import structure (see https://github.com/samm82/Drasil/commit/ac36c574613a3acde75983b8bfdd141ae87d6c74 for an example of this). I figured I'd see how easy it would be to generate these files, and it was WAY easier than I expected, so I just opened a PR for it.

If this change should wait to be introduced alongside my generated testing code (when we get there), then this PR can be closed.


balacij:
Sometimes [HLint updates](https://github.com/ndmitchell/hlint/releases) and our CI breaks as a result, so we end up fixing those (newly found) issues in random PRs.

I like this idea. Following along the 'modularization' choice, we might also consider nesting the projects into packages.

JacquesCarette (approved):
So weird that adding an essentially empty file helps...

PR #3528, opened 2023-07-11
Remove unnecessary the and remove GamePhysics from table of abbrv.

### LINKED ISSUE #3491: Remove unnecessary 'the' from purpose of Documents
As mentioned in #3317. In the Purpose of Documents in the Introduction, the first sentence goes as follows:
```
The primary purpose of this document is to record the requirements of the GlassBR
```
where GlassBR is replaced by the respective title of the SRS.

The `the` in `requirements of the GlassBR` should be removed.

### LINKED ISSUE #3514: Remove Game Physics from table of Abbreviations
The `Abbreviations and Acronyms` table has `Game Physics` listed. Currently it is shown as: 
image
I believe this is unnecessary for 2 reasons; 
- as mentioned in #3491 the term 'game physics library' should not exist, as we don't specify that for other examples.
- `Game Physics` is neither and abbreviation or acronym.

ManvendraJani:
closes #3514 
closes #3491


code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Concepts.hs
```diff
@@ -14,9 +14,10 @@ import Control.Lens ((^.))
...
 
 acronyms :: [CI]
-acronyms = [assumption, centreMass, dataDefn, gamePhysics, genDefn, goalStmt, inModel,
+acronyms = [assumption, centreMass, dataDefn, genDefn, goalStmt, inModel,
   likelyChg, ode, requirement, refBy, refName, srs, thModel, threeD, twoD, typUnc, unlikelyChg]
 
 centreMass, gamePhysics :: CI
 centreMass  = commonIdeaWithDict "centreMass"  (ctrOfMass ^. term) "CM" [physics]
-gamePhysics = commonIdeaWithDict "gamePhysics" (pn "game physics library") "Game Physics" [physics]
+--This is a hack that is used accross the board, we should consider using something
```
samm82, line 22: I'm not fully sure what the purpose of this comment is; we should consider using something (like what?)

(Also, it should be "across", not "accross")

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -614,13 +610,13 @@ <h1>Introduction</h1>
           Due to the rising cost of developing video games, developers are looking for ways to save time and money for their projects. Using an open source physics library that is reliable and free will cut down development costs and lead to better quality products.
         </p>
         <p class="paragraph">
-          The following section provides an overview of the Software Requirements Specification (SRS) for Game Physics. This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
+          The following section provides an overview of the Software Requirements Specification (SRS) for GamePhysics. This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
```
samm82, line 613: As mentioned in our previous meeting, this would be a good place to introduce the abbreviation (e.g., "for the Game Physics library (GamePhysics)"), but since that is a more involved and lower priority change, this looks good to me for now. Could you add a <kbd>[Low Priority](https://github.com/JacquesCarette/Drasil/labels/Low%20Priority)</kbd> issue describing this change?

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
```diff
@@ -553,7 +553,7 @@ <h1>Introduction</h1>
           <div class="subsection">
             <h2>Purpose of Document</h2>
             <p class="paragraph">
-              The primary purpose of this document is to record the requirements of the Solar Water Heating System. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SWHS. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
+              The primary purpose of this document is to record the requirements of SWHS. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SWHS. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
```
samm82, line 556: This should say "NoPCM" instead of "SWHS", right?

code/stable/dblpendulum/SRS/JSON/DblPendulum_SRS.ipynb
```diff
@@ -155,7 +155,7 @@
     "## Purpose of Document\n",
     "<a id=\"Sec:DocPurpose\"></a>\n",
     "\n",
-    "The primary purpose of this document is to record the requirements of the Pendulum. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of DblPendulum. With the exception of [system constraints](#Sec:SysConstraints), this SRS will remain abstract, describing what problem is being solved, but not how to solve it.\n",
+    "The primary purpose of this document is to record the requirements of DblPendulum. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of DblPendulum. With the exception of [system constraints](#Sec:SysConstraints), this SRS will remain abstract, describing what problem is being solved, but not how to solve it.\n",
```
samm82, line 158: "DblPendulum" (and "SglPendulum") should be added to the table of abbreviations.

samm82 (changes requested):

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
ManvendraJani: See #3529

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
smiths: I agree with @samm82.

code/drasil-example/nopcm/lib/Drasil/NoPCM/Body.hs
```diff
@@ -194,14 +195,18 @@ si = SI {
    refdb       = refDB
 }
 
+progName :: CI
+progName = commonIdeaWithDict "swhsNoPCM"   (nounPhrase "solar water heating system"
+  "solar water heating systems") "NoPCM" [materialEng]
```
samm82, line 200: I think this should have a different "full" name than "solar water heating system"; maybe something like "solar water heating system with no phase change material" @smiths any thoughts?

code/stable/dblpendulum/SRS/HTML/DblPendulum_SRS.html
```diff
@@ -371,6 +371,10 @@ <h2>Abbreviations and Acronyms</h2>
                   <td>DD</td>
                   <td>Data Definition</td>
                 </tr>
+                <tr>
+                  <td>DblPend</td>
+                  <td>DblPendulum</td>
```
samm82, line 376: Shouldn't the full version of this be "Double Pendulum"?

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
```diff
@@ -547,13 +547,13 @@ <h1>Introduction</h1>
           Due to increasing costs, diminishing availability, and negative environmental impact of fossil fuels, the demand is high for renewable energy sources and energy storage technology. Solar water heating systems provide a novel way of storing energy.
         </p>
         <p class="paragraph">
-          The following section provides an overview of the Software Requirements Specification (SRS) for solar water heating systems. The developed program will be referred to as Solar Water Heating System (SWHS). This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
+          The following section provides an overview of the Software Requirements Specification (SRS) for solar water heating systems. The developed program will be referred to as Solar Water Heating System (NoPCM). This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
```
samm82, line 550: This is a good indicator that the "full name" of NoPCM should be different than that of SWHS (see my other comment)

samm82 (changes requested):

code/drasil-example/nopcm/lib/Drasil/NoPCM/Body.hs
smiths: Yes, I agree with @samm82.  The full name should distinguish the nopcm example from the with pcm example, just like the short name does.  I like the suggested full name from @samm82.

code/stable/dblpendulum/SRS/HTML/DblPendulum_SRS.html
ManvendraJani: In that case should the abbreviation be "DblPend" or "DblPendulum"

smiths: I like DblPend, since it is shorter.

code/datafiles/nopcm/SystemContextFigure.png
samm82, line 1: Would you be able to use Times New Roman and try to match the font size of "User" for "NoPCM" for consistency? Also, why did the resolution drop?

code/drasil-example/nopcm/lib/Drasil/NoPCM/Body.hs
```diff
@@ -257,14 +258,19 @@ orgDocEnd = foldlSent_ [atStartNP (the inModel),
...
 
 --TODO: If/when system constraints recieves any content, add s3_3_intro
 
 ------------------------------
 --Section 3.1 : SYSTEM CONTEXT
 ------------------------------
 
+sysCntxtFig :: LabelledContent
+sysCntxtFig = llcc (makeFigRef "SysCon") $ fig (foldlSent_
+  [refS sysCntxtFig +: EmptyS, titleize sysCont])
```
samm82, line 271:
```suggestion
sysCntxtFig = llcc (makeFigRef "SysCon") $ fig (refS sysCntxtFig +: titleize sysCont)
```
This should work and is cleaner. Nice job finding the use of "SWHS" in the non-searchable image!

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
```diff
@@ -544,10 +544,10 @@ <h2>Abbreviations and Acronyms</h2>
...
+          Due to increasing costs, diminishing availability, and negative environmental impact of fossil fuels, the demand is high for renewable energy sources and energy storage technology. Solar water heating systems with no phase change material provide a novel way of storing energy.
         </p>
         <p class="paragraph">
-          The following section provides an overview of the Software Requirements Specification (SRS) for solar water heating systems. The developed program will be referred to as Solar Water Heating System (NoPCM). This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
+          The following section provides an overview of the Software Requirements Specification (SRS) for solar water heating systems with no phase change material. The developed program will be referred to as Solar Water Heating System With No Phase Change Material (NoPCM). This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
```
samm82, line 550: I think @smiths usually likes only the letters used in an acronym to be capitalized. Do you think this is OK @smiths or would you prefer "solar water heating system with No Phase Change Material" (or even "...NO Phase Change Material" since the "o" is used, albeit lowercase)?

In either case, "with" shouldn't be capitalized. I believe we have a function for titleizing phrases while ignoring the smaller words that shouldn't be capitalized; can that be used here?

samm82 (changes requested):

code/drasil-example/nopcm/lib/Drasil/NoPCM/Body.hs
ManvendraJani: This results in the colon being after `sysCont` I don't believe we have a constructor for the colon in between

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
smiths: A generative approach really does make you think about every little decision!  That's a good thing, even if it is challenging.  :smile:  @samm82 is correct, the capitalization in the expanded acronym should tell the reader the acronym.  The acronym should have each of the capital letters.  

The name NoPCM is really for us to easily distinguish SWHS from NoPCM.  As a standalone example, it probably should be SWHS_NoPCM.  The NoPCM is like an option for the family of SWHS programs.  If it is easy to do, without too many unintended consequences, I think renaming NoPCM to SWHS_NoPCM would be nice.  That would mean for the specific capitalization question above, we just need to turn "With" into "with".

code/drasil-example/nopcm/lib/Drasil/NoPCM/Body.hs
samm82: Ah yes, good catch

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
ManvendraJani: We can wait for Dr. Smith's input but in the mean time I have added a temporary fix but it does not look right. I created an issue regarding how we can properly fix it, see #3535

ManvendraJani:
@samm82 using the abbreviation DblPend leads to issues with the tex file and code generation. To resolve this we would need to  change the name of the example to DblPend, is that ok to do? (This also applies to SglPend)

samm82:
Can you be more specific? I'm not sure what the issue is. Why would changing the name of the example to "DblPend" fix it? Because our file structure is set up to use the abbreviations of examples for their names when generating folders, etc.?

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
samm82: I'm not sure if #3535 really fixes this issue. It seems like we need a way to specify which letters are used to make the abbreviations of concepts, then capitalize, bold, etc. these letters when introducing the acronym (whether in the text or in the table). Capitalizing a concept (which would include capitalizing "with"; see the code snippet below) is different than specifying where its abbreviation comes from.

https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-lang/lib/Language/Drasil/NounPhrase.hs#L271-L274

ManvendraJani: I was referring to the capitalization in the case of #3535, you mentioned that the "with" should be lower case. As for the abbreviation, although we may create abbreviation in the printers in some cases this does not apply to the common ideas, we have just been pre-defining the abbreviation when that idea is created.

I will change the name to SWHS_NoPCM if it is ok to move on with changing the example names.

samm82:
Changing the whole example name seems like the best move here, and is probably OK to be done in this PR (even though it is getting a bit big, the work seems to be related, except for maybe the work done for #3491)

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
smiths: I think we might have to relax automation with turning a phrase into an abbreviation.  Not all abbreviations are ALLCAPS.

smiths:
I'm okay with changing the name of the example.  The default of having the program name and its abbreviation being the same makes sense.  I do think that we might eventually wish to allow for the possibility of the name and the abbreviation being different, but this isn't a priority.

JacquesCarette:
Changing the name of the examples seems to be the most coherent solution with respect to the discussion. (I'll reply to other items separately).

code/datafiles/swhs/SystemContextFigure.png
samm82, line 1: Why did the SWHS figure get removed; is it not needed for SWHS's SRS?

smiths: Unless I'm missing something, the system context figure shouldn't be removed for swhs.  Modifying the nopcm figure makes sense to reflect the change in program name.

ManvendraJani: I am sorry, this was deleted from the wrong file. i meant to delete it from NoPCM

code/stable/nopcm/SRS/HTML/NoPCM_SRS.html
ManvendraJani: Using `SWHS_NoPCM `as the name leads to issues in the .yaml files where an '_' (underscore) implies a dependency. We could get around this by using a name like `SWHSNoPCM `(a '-' leads to other issues within the haskell files). Let me know if this is ok or should it be changed to something else entirely?

smiths: I don't love the SWHSNoPCM name, but I'm fine with living with it for now in the interest of resolving this issue.  In the future it would help if we had options for the name.  The default option would be the name we prefer.  This name could have underscores, or dashes, accents, or even special symbols.  The other option would be the "conservative" name that isn't going to cause problems when it is used for a folder name, or in a yaml file etc.

hrzhuang:
The reason `make gool` was breaking is that it relies on symlinks such as `datafiles/dblpend/cpp` (which points to `datafiles/odelib/cpp`) in order to find the `odelib`. But these are UNIX symlinks, which are not recognized as symlinks on @ManvendraJani's computer, so when `dblpendulum` was renamed to `dblpend`, the symlinks turned into plain text files (containing the text `../odelib/cpp`, for example). I just pushed a fix for this.

This uncovers a bigger issue of UNIX symlinks in our repository. As it currently stands (even before the changes in this PR), `make gool` does not work on Windows under Git Bash because the symlinks do not work (and we haven't realized this because `make gool` usually just runs on the CI 😅). I'm not sure if it would work under Cygwin, but in any case I don't think we should be relying on UNIX symlinks in a repository that is supposed to be cross-platform.

samm82:
Sorry I didn't get a chance to look at this earlier, but good job for figuring it out! 🎉

samm82 (approved):
This looks good to me! (Still confused as to why the resolution of the SWHS figure dropped, but it seems like it is of high-enough quality so I don't know how big of a deal it is ¯\_(ツ)_/¯)

code/datafiles/swhs/SystemContextFigure.png
JacquesCarette, line 1: Why has the size of the picture been halved in both dimensions?

JacquesCarette (commented):
Just one question before I approve.

code/datafiles/swhs/SystemContextFigure.png
ManvendraJani: I am not sure why that happened but I just fixed it in the latest commit.

JacquesCarette (approved):
Approving now, but I'm going to wait a bit to see if others have comments, as this has been quite the storied PR!

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
```diff
@@ -422,16 +428,16 @@ <h2>Abbreviations and Acronyms</h2>
...
+          A pendulum consists of mass attached to the end of a rod and its moving curve is highly sensitive to initial conditions. Therefore, it is useful to have a program to simulate the motion of the pendulum to exhibit its chaotic characteristics. The program documented here is called Double Pendulum.
         </p>
         <p class="paragraph">
-          The following section provides an overview of the Software Requirements Specification (SRS) for Pendulum. This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
+          The following section provides an overview of the Software Requirements Specification (SRS) for Double Pendulum. This section explains the purpose of this document, the scope of the requirements, the characteristics of the intended reader, and the organization of the document.
         </p>
```
smiths, line 435: Up to this point we call the program Double Pendulum.  From this point forward we call it DblPend.  Unfortunately, it doesn't look like we introduce this short form (abbreviation).  We are treating Double Pendulum as the program name in the line above, but we call it DblPend below.  What is the program name?  I feel like the simple fix would be to make the program name DblPend, and never call it Double Pendulum.

code/stable/sglpend/SRS/HTML/SglPend_SRS.html
```diff
@@ -392,16 +398,16 @@ <h2>Abbreviations and Acronyms</h2>
...
         </p>
         <div id="Sec:DocPurpose">
           <div class="subsection">
             <h2>Purpose of Document</h2>
             <p class="paragraph">
-              The primary purpose of this document is to record the requirements of the Pendulum. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SglPendulum. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
+              The primary purpose of this document is to record the requirements of SglPend. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SglPend. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
```
smiths, line 410: Same comment as for DblPend.  If we called the program SglPend from the start, we wouldn't have the problem that we call it Single Pendulum twice and SglPend every other time (as far as I can tell).

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -767,7 +767,7 @@ <h1>Introduction</h1>
           <div class="subsection">
             <h2>Purpose of Document</h2>
             <p class="paragraph">
-              The primary purpose of this document is to record the requirements of the Solar Water Heating System. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SWHS. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
+              The primary purpose of this document is to record the requirements of SWHS. Goals, assumptions, theoretical models, definitions, and other model derivation information are specified, allowing the reader to fully understand and verify the purpose and scientific basis of SWHS. With the exception of <a href=#Sec:SysConstraints>system constraints</a>, this SRS will remain abstract, describing what problem is being solved, but not how to solve it.
```
smiths, line 770: I thought the plan was to have a different name for the NoPCM version of SWHS?  Wasn't the short form going to be SWHS_NoPCM?

smiths: or SWHSNoPCM, if the underscore is problematic.

smiths:
Given the comment from @JacquesCarette I reviewed again.  I commented in stable the parts that don't seem quite right to me.  Mainly the comments are around DblPend and SglPend.  It seems odd that we give the programs these names, but we also use the names Double Pendulum and Single Pendulum.  I guess we have the long-form version of the name and the abbreviation.  I'd be okay if we just called them by the short forms, or if we officially gave them the name of the long forms, but explicitly introduce that we will use the short form going forward (like we do with SWHS).

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
ManvendraJani: There is already an issue regarding this, see #3529

smiths: Okay.  We can handle this with issue #3529.

code/stable/sglpend/SRS/HTML/SglPend_SRS.html
smiths: Also addressed by #3529.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
ManvendraJani: I believe you are taking a look at the SWHS file, the SWHSNoPCM file does show the correct name. 
image

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
ManvendraJani: let me know if I should change the program name to be DblPend anyways

code/stable/swhs/SRS/HTML/SWHS_SRS.html
smiths: On second look, I was looking at the wrong file.  I thought it was the NoPCM version, but I see now that it is SWHS.  No problem here.

smiths (approved):
This PR looks good to me with respect to the changes to stable.

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
smiths: No its fine as it is.  This PR is confusing enough.  :smile:

JacquesCarette:
Sorry, I made more merge conflicts when I pulled in another PR...

Issue #3536, opened 2023-07-14
Get rid of `CommonConcept` altogether


JacquesCarette:
As far as I can tell, it is used 3 times in examples. Furthermore, it should be equivalent to a `ConceptChunk` with an empty concept domain.

PR #3543, opened 2023-07-18
Change `hghc`, `srsSWHS`, and `swhsPCM` to `CI`s

### LINKED ISSUE #3542: Change `hghc`, `srsSWHS`, and `swhsPCM` to `CI`s
Part of #3536. These hold the names of the examples (HGHC, NoPCM, and SWHS, respectively). They are currently `CommonConcept`s, but are only accessed via the type classes `CommonIdea` and `Idea`. All other example names are `CI`s (`gamePhysics`, `glassBR`, `pdControllerApp`, etc.).

hrzhuang:
Closes #3542.

A possibly controversial change is the change from `capSent (swhsPCM ^. defn)` to `capSent $ pluralNP $ progName ^. term, S "incorporating", phrase phsChgMtrl` (this is inside the list of `Setence`s passed to `foldlSent`) in `introStartSWHS`.

In the original code, `swhsPCM` is a `CommonConcept` whose term is "solar water heating systems incorporating PCM" and whose definition is "solar water heating systems incorporating phase change material" (the definition expands "PCM" to "phase change material"). This is not really a definition but rather a hack to get multiple levels of abbreviation, and since this "definition" is only used in one place, I don't think we are worse off if we just construct it in the spot where it is used, especially with respect to our goal of removing `CommonConcept` (#3536).

### DIFF
--- a/code/drasil-example/hghc/lib/Drasil/HGHC/HeatTransfer.hs
+++ b/code/drasil-example/hghc/lib/Drasil/HGHC/HeatTransfer.hs
@@ -65,8 +65,8 @@ htTransCladFuelEq = (exactDbl 2 `mulRe` sy cladCond `mulRe` sy gapFilmCond) $/ (
 
 ---
 
-hghc :: CommonConcept
-hghc = dcc' "hghc" (cn "HGHC") "HGHC program" "HGHC"
+hghc :: CI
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):
Not entirely sure that `CI` is hugely better, but that can be dealt with in a later pass. This still does seem like progress.

Issue #3525, opened 2023-07-10
Disambiguate the UIDs of models from those of concepts


hrzhuang:
Spun off of #3518. In the meeting (#3506) we decided that we should change the UIDs of models (i.e. data definitions, general definitions, instance models, and theory models) to be different from the UIDs of the underlying concepts. I think @JacquesCarette suggested that we prepend `theory:` to the UIDs of the models. Should we proceed with this change?

balacij:
#2904 started with `ModelKinds`. I think that starting with `InstanceModel`s is a good idea, I think a lot of them now inherit their `UID` from the quantity they define a solution for.

JacquesCarette:
Yes, we should, and indeed, implementation has started.

PR #3546, opened 2023-07-19
Implement namespaces for UIDs

hrzhuang:
The old UID type was just a `String` wrapped in a `newtype`. As a running example, the UID `energy` has string representation `"energy"` and JSON representation `"energy"`.

With this PR, a UID now have a `[String]` as its namespace. Currently all UIDs have the namespace `[]`, i.e. they are in the "global" or "root" namespace. So `energy` still has string representation `"energy"` but its JSON representation is now `["energy"]`. If we were to put the UID `velocityTM` under the `["theory"]` namespace (see #3525), then it would have string representation `"theory:velocityTM"` and JSON representation `["theory", "velocityTM"]`.


JacquesCarette (approved):
That is indeed the correct "raw starting point" for this. Then we'll see, by actually using it, what it is we **really** want.

PR #3553, opened 2023-07-21
Optionally build the chunk JSON files

### LINKED ISSUE #3552: json in stable brings a lot of noise to too many PRs - should remove
Sorry @balacij but this is getting tiresome. It's already sometimes quite a lot of work to review because of changes that affect the visuals in minor ways (i.e. to the .tex and .html) but the .json files changes way too often. Please remove, and let's find a different way to track this.

balacij:
Closes #3552 
Builds on #3554 in the name of testing #3554

I'm not really happy with this solution. I wanted to make it optional to dump the chunks. I thought the appropriate time to dump chunks was during 'debugging', but it seems our `make debug` command doesn't work! We'll need to fix that. For now, I've made this optional by allowing users to set a `DEBUG_ENV=true` environment variable if they want to generate `.drasil` folders of chunk JSON files.


smiths:
@balacij I like your `make debug` plan for the future.  It is more elegant than changing the environment variable.  For practical reasons though, you are right that changing the environment variable is good enough for now.  I verified that the new `Makefile` includes a comment that explains the use of the environment variable.

JacquesCarette (approved):

JacquesCarette:
Conflict to resolve.

balacij:
@hrzhuang Do you get weird error thrown when running `make debug`?

hrzhuang:
Do you mean errors like this?
```
*** Exception (reporting due to +RTS -xc): (THUNK_1_0), stack trace: 
  Language.Drasil.Generate.prntDoc',
  called from Language.Drasil.Generate.prntDoc,
  called from Language.Drasil.Generate.prnt,
  called from Language.Drasil.Generate.gen,
  called from Main.main
*** Exception (reporting due to +RTS -xc): (THUNK_1_0), stack trace: 
  Language.Drasil.Generate.prntDoc',
  called from Language.Drasil.Generate.prntDoc,
  called from Language.Drasil.Generate.prnt,
  called from Language.Drasil.Generate.gen,
  called from Main.main
*** Exception (reporting due to +RTS -xc): (THUNK_1_0), stack trace: 
  Language.Drasil.Generate.genLog,
  called from Main.main
*** Exception (reporting due to +RTS -xc): (THUNK_1_0), stack trace: 
  Language.Drasil.Generate.genLog,
  called from Main.main
```

balacij:
Yep! Do you know if these are expected? At a quick glance, they seem odd to me.

hrzhuang:
I think this is related to this [GHC issue](https://gitlab.haskell.org/ghc/ghc/-/issues/17761) where reading the end of a file is an exception that is thrown and caught, and the `-xc` flag logs this exception even though it is caught.

balacij:
It does look related, but I just have no idea where the actual error is originating from. I didn't find any calls to `read`. It looks like this issue is on `master` now too, so no point in delaying this merge at least, however.

JacquesCarette:
I wouldn't be at all surprised if `write` hard the same logging behaviour as `read` - and there are definitely calls to things that write!

samm82:
Just confirming @balacij, I have branches where changes were made to files in .drasil, so those changes are overriding their deletion in this PR; should we be deleting the .drasil/ folders in all branches?

balacij:
Yep, a good ole:
```
git pull origin master
git clean -fd
```
should do the trick.

balacij:
That's weird… Do you have a branch I could check out? Or was it this one?

samm82:
The branch it was happening on was `master`, although it was also happening on `genOutReq` for me

balacij:
Oh, this is odd, I can reproduce...

Looking at `master`, it doesn't look like my changes are there! Did someone accidentally revert these changes?

samm82:
Ooooh... maybe I did 😬 I remember I was reverting some changes on a branch and pulled in this branch to it, but when I reverted a commit, it also reverted the changes of this branch, so I just reverted it. I guess that revert commit did a lot of heavy lifting in the master branch as well.... 😅

PR #3554, opened 2023-07-21
Move HLint script to a separate workflow file

balacij:
Contributes to #2789 

Decreases build time by only running HLint when absolutely necessary (i.e., when a file in one of the `code/drasil-*` paths are edited) and concurrently alongside the "build & test" workflow.

Tested with #3553.


smiths:
Good idea!

JacquesCarette (approved):

Issue #3481, opened 2023-06-14
Remove `SystemInformation`s `_instModels` field


balacij:
I added the `_instModels` field to `SystemInformation` so that we can pass around information about which `InstanceModels` are relevant to SRS. Right now, the `ChunkDB`s of each problem contains only the `InstanceModel`s relevant to their own problem, and each one is presumably displayed in the SRS (and used). So, we should be able to naively remove this `_instModels` field from `SystemInformation`, replacing accessors to it with an accessor to the `ChunkDB` that gathers all the `InstanceModel`s automatically.

daijingz:
Can I have the file path? Let me take a look

balacij:
Hi @daijingz, `SystemInformation` is defined in `drasil-sysinfo`, it's one of the few files there. [grep](https://github.com/JacquesCarette/Drasil/wiki/Debugging-in-Drasil#grep-summary) will be helpful for searching where it is used too.

daijingz:
Let me find the corresponding code first.

daijingz:
I see the file at this path: C:\Users\david\Drasil\code\drasil-sysinfo\lib\SysInfo\Drasil\SystemInformation.hs. I hope it is right.
There are two places with the "_instModels" fields: one on line 13, and another on line 55. I think I should delete them from the code. This is the first step.

How to do the next step "replacement"? Seems like chunkDB is an external Haskell library. Can you give me more information?

daijingz:
Here is the modified code: 
![T SEFP MK44@P@FF SRGAN4](https://github.com/JacquesCarette/Drasil/assets/61481741/03c8b47a-b0d1-461a-ba41-b35e8762258c)
![~F4VCP{KA M04V)2${HJB0T](https://github.com/JacquesCarette/Drasil/assets/61481741/cab716f3-4307-4a61-b41e-9bff41e48929)
Other places I have not changed, and I guess this is the first way

JacquesCarette:
You need to make a PR for us to evaluate your changes.

daijingz:
> You need to make a PR for us to evaluate your changes.

Dr.Carette, I feel this is still an unsaved change. I have not submitted this change since I feel there are some other places that required to be changed, and I need some further assistance. Now I am still reading relevant codes. Give me more time please.

JacquesCarette:
Do all the tests pass?

daijingz:
I realize simple deletion is not working well, let me do it again.

[...]

PR #3559, opened 2023-07-23
Issue #3481 Change version #1

daijingz:
First version: deleting _instmodels objects in the "SystemInformation.hs" file


JacquesCarette:
As you can see, the build failed. You could have figured this out by doing `make test` on your own machine.

PR #3561, opened 2023-07-24
Fix `make debug`

### LINKED ISSUE #3560: Problem with `make debug`
By default, `make` only produces one level of stack trace for an error, i.e. indicate where the `error` functions is called and nothing else. I would like to get to get a full stack trace to track down the root cause of the error, so I'm looking into `make debug`. However, it gives me the following error.
```
cd "build/glassbr" && stack exec -- "glassbr" +RTS -xc -P
glassbr: the flag -xc requires the program to be built with -prof
glassbr: the flag -P requires the program to be built with -prof
```
I am running this after a `make clean`, so the programs should be built with the compiler flags that are needed.

hrzhuang:
Closes #3560. It seems that in addition to `stack build`, `stack exec` also requires the `--profile` flag (the `--profile` flag is equivalent to `--executable-profiling --library-profiling`). But `--profile` needs to be passed to `stack exec` before the `--`, so I've moved the `$(EXECARGS)` to before the `--` since the RTS flags that used to be passed after `--` can also be passed before `--`.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -146,8 +146,8 @@ WEBSITE_FOLDER = $(WEBSITE_FOLDER_NAME)/HTML/
 FULL ?= 0
 
 # GHC debug options
-PROFALL = --executable-profiling --library-profiling
-PROFEXEC = +RTS -xc -P
+PROFALL = --profile
+PROFEXEC = --profile --rts-options '-xc -P'
 
 # GHC build options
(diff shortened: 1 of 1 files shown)


balacij:
Ah, nice! I guess we'll need to fix up #3553 too now.

JacquesCarette (approved):

PR #3564, opened 2023-07-25
Add Stack-based shebangs to some of the `scripts/**.hs` files

### LINKED ISSUE #3369: `ClassInstDepGen` _implicitly_ relies on dependency `split`
We should make the dependency explicit somehow.

balacij:
Closes #3369 by adding direct traceability to the requirements of the scripts. I'm only adding it to the ones that are actual "scripts" and intended to be runnable. Adding the shebang to the local shared modules `SourceCodeReader...hs` wouldn't make sense since it is not intended to be run.


JacquesCarette (approved):
Nifty!

Issue #3534, opened 2023-07-14
Remove capitalization from concepts in PDController


samm82:
As seen in the example below, many concepts in the PDController example have capitalization baked in, instead of letting the sentence constructors handle it.

https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs#L17-L22

This has unwanted side effects; for example:

![image](https://github.com/JacquesCarette/Drasil/assets/35857611/cb5cd660-7dcf-431c-86c7-f9f685cdfb76)

daijingz:
Although I have chosen other issues, can I work on this as well?

balacij:
Hi @daijingz, that is fine since it doesn't look like anyone else is working on it. Try to submit a PR once you have something.

daijingz:
Can you give me some explanations about your mentioned concept of "capitalization"?
I feel this time I may over-simplify the given issue again. I feel "capitalization" is these 6 concepts concepts' parameters have capitalized letters, which should be changed to non-capitalized letters (Since sentence constructors have already worked on this)

daijingz:
This is my draft answer:

> --pdControllerApp = commonIdeaWithDict "pdControllerApp" (pn "PD Controller")                    "PD Controller" []
> pdControllerApp = commonIdeaWithDict "pdcontrollerapp" (pn "pd controller")                    "PD Controller" []
> --pdControllerCI  = commonIdeaWithDict "pdControllerCI"  (pn "Proportional Derivative")          "PD"            []
> pdControllerCI  = commonIdeaWithDict "pdcontrollerci"  (pn "proportional derivative")          "PD"            []
> --proportionalCI  = commonIdeaWithDict "proportionalCI"  (pn "Proportional")                     "P"             []
> proportionalCI  = commonIdeaWithDict "proportionalci"  (pn "proportional")                     "P"             []
> --derivativeCI    = commonIdeaWithDict "derivativeCI"    (pn "Derivative")                       "D"             []
> derivativeCI    = commonIdeaWithDict "derivativeci"    (pn "derivative")                       "D"             []
> --integralCI      = commonIdeaWithDict "integralCI"      (pn "Integral")                         "I"             []
> integralCI      = commonIdeaWithDict "integralci"      (pn "integral")                         "I"             []
[...]

PR #3568, opened 2023-07-28
Remove capitalization from concepts in PDController #3534 Draft

daijingz:
I only change the file "[Drasil/code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs](https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs#L17-L22)", while other files I have not changed.

It can work well with build and exec commands on my personal computer. I still have some confusion on this issue.


code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs
```diff
@@ -14,12 +14,18 @@ acronyms
...
                      pidCI :: CI
 
-pdControllerApp = commonIdeaWithDict "pdControllerApp" (pn "PD Controller")                    "PD Controller" []
-pdControllerCI  = commonIdeaWithDict "pdControllerCI"  (pn "Proportional Derivative")          "PD"            []
-proportionalCI  = commonIdeaWithDict "proportionalCI"  (pn "Proportional")                     "P"             []
-derivativeCI    = commonIdeaWithDict "derivativeCI"    (pn "Derivative")                       "D"             []
-integralCI      = commonIdeaWithDict "integralCI"      (pn "Integral")                         "I"             []
-pidCI           = commonIdeaWithDict "pidCI"           (pn "Proportional Integral Derivative") "PID"           []
+--pdControllerApp = commonIdeaWithDict "pdControllerApp" (pn "PD Controller")                    "PD Controller" []
+pdControllerApp = commonIdeaWithDict "pdcontrollerapp" (pn "pd controller")                    "PD Controller" []
```
samm82, line 18: Since "PD Controller" is the name of the program and is a proper noun, it should still be capitalized:
```suggestion
pdControllerApp = commonIdeaWithDict "pdControllerApp" (pn "PD Controller")                    "PD Controller" []
```
(Also see my more general note about UIDs

samm82 (changes requested):
Definitely on the right track; however, I have the following notes:

- [ ] The first argument to `commonIdeaWithDict` is what we call a "UID", or a "Unique IDentifier". All of these should be kept with the capitalization they have so make it easier to read (since they are one word).
- [ ] Instead of commenting out the previous versions of the code, the old version should just be removed. It is easy enough to see what changes have been made using git.
- [ ] If you compile the examples (using `make`), you should see some changes to the examples (summarized in the generated log files). If these changes are good (i.e., what you would expect from your changes to the code), then running `make stabilize` will update our stable version of the code to reflect these changes.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs
daijingz: Let me fix it quickly

daijingz: Because by using the command grep, I feel that the third parameter is always capitalized. Do I need to change the third parameter?

daijingz: I feel all second parameters are names of the mentioned programs and all of them are proper nouns, so I cannot change them. This time, I change all the third parameters, which has two effects:
1. Before the log file (C:\Users\david\Drasil\code\logs\pdcontroller_log.log) is empty, now it has 278 lines.
2. Build process still work well, although there are some "ignored errors"

daijingz:
Let me open a new pull request

code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs
samm82: The third parameter I would keep capitalized. However, `pdControllerApp` is the name of the PD Controller app (as the name suggests), while the others are just names for concepts used in its specification (like "integral"), so these should all be lowercase. 

In regards to the effects of your changes, that is good that the log has content! You made changes, and those changes mean that what is generated is different from what we "expect" to be generated. Now you should look at the generated files (likely a mix of looking at the log file and looking at each generated file individually) to verify that these changes are good. Then, running the `make stabilize` command will update our "expectations" to match what you now generate. (The "ignored errors" are these differences between the two versions and is to be expected.)

samm82:
Just for future, if you aren't quite done working on a PR, it can be turned into a "draft" PR. It seems like you could just continue working on this branch (in this PR) for this change, so closing it likely wasn't necessary

PR #3570, opened 2023-07-29
Remove capitalization from concepts in PDController #3534 Draft 2 #3568

daijingz:
Now you can see log file changes.
(Following the last pull request)

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/HTML/CSS.hs
+++ b/code/drasil-printers/lib/Language/Drasil/HTML/CSS.hs
@@ -66,9 +66,8 @@ makeCSS _ = vcat [
     text "  margin-left: auto;",
     text "  margin-right: auto;}"],
   text "th, td {border: 1px solid black; padding: 0.5em;}",
-  text ".tdefn, .ddefn {width: 75%; margin-top: 1%; margin-bottom: 1%;}",
-  text ".tdefn th {width: 15%;}",
-  text ".ddefn th {width: 15%;}",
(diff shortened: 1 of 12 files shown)


samm82 (changes requested):
I'm pretty sure that the issue here is with the second argument, not the third argument. The strings in the `pn` call (except for "PD Controller") should be made lowercase, and the strings after then (e.g., "PD") should still be capitalized (I think).

You mention in the description of this PR that the changes are visible in the log files. These log files aren't actually added to commits, since they change very frequently. They are just a way for the developers (us) to see what changes to our generated output have been made by our changes to the code. If all of the changes in the log files look good, then the `stable` directory should be updated so that the generated version (in the `build` directory) and the `stable` version match. When your PR is ready to merge, it should include all changes that you've made to `stable` and running `make` should result in no changes being stored in the log files.

daijingz:
> I'm pretty sure that the issue here is with the second argument, not the third argument. The strings in the `pn` call (except for "PD Controller") should be made lowercase, and the strings after then (e.g., "PD") should still be capitalized (I think).
> 
> You mention in the description of this PR that the changes are visible in the log files. These log files aren't actually added to commits, since they change very frequently. They are just a way for the developers (us) to see what changes to our generated output have been made by our changes to the code. If all of the changes in the log files look good, then the `stable` directory should be updated so that the generated version (in the `build` directory) and the `stable` version match. When your PR is ready to merge, it should include all changes that you've made to `stable` and running `make` should result in no changes being stored in the log files.

See what you means, Working on it now

JacquesCarette:
Sam is correct.

PR #3579, opened 2023-08-02
Stop using a lens for `HasSpace`

### LINKED ISSUE #3578: Does `HasSpace` have to use a lens?
Currently the method `typ` from `HasSpace` is a lens.
```haskell
class HasSpace c where
  typ      :: Lens' c Space
```
Is there any reason for this? Or can we make it a simple function as follows?
```haskell
class HasSpace c where
  typ      :: c -> Space
```

hrzhuang:
Closes #3578

### DIFF
--- a/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
+++ b/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
@@ -271,7 +271,7 @@ vecDepVar info = quantvar $ implVarUID (dv ^. uid) (dv ^. term) vecT
 -- be represented as one or else will hit type errors in GOOL.
 arrayVecDepVar :: ODEInfo -> CodeVarChunk
 arrayVecDepVar info = quantvar $ implVarUID (dv +++ "vec") (dv ^. term)
-  (dv ^. typ) (sub (symbol dv Implementation) (label "vec"))
+  (typ dv) (sub (symbol dv Implementation) (label "vec"))
   where dv = listToArray $ depVar info
 
 -- Apache Commons (Java)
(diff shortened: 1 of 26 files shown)


JacquesCarette (approved):
Actually, now I know what we lose: the pretty 'dot' syntax for accessors...

hrzhuang:
Would it be preferable to make it a `Getter` (from the `lens` package) instead of a simple function? This would allow us to keep the dot syntax.

JacquesCarette:
I think `Getter` is just an alias for a function? If not, maybe it is a good idea. There is something about the dot syntax which is appealing in what it communicates to the reader.

hrzhuang:
To my understanding, in the `lens` package a `Lens` is simply anything that is both a `Getter` and a `Traversal`. `Getter` provides the getting part of the `Lens`'s functionality, and `Traversal` provides the setting part. [This diagram](https://raw.githubusercontent.com/ekmett/lens/master/images/Hierarchy.png) in the `lens` package illustrates the relationships between the various constructs provided by the package.

JacquesCarette:
I understand that - I'm just wondering what `Getter` actually expands to.

hrzhuang:
From the `lens` package documentation,
```haskell
type Getter s a = forall f. (Contravariant f, Functor f) => (a -> f a) -> s -> f s 
```
The documentation says `Getter` is isomorphic to `s -> a`, but this type signature makes `Getter`s compatible with the "getting" part of a `Lens`'s functionality. I.e. `^.` works with `Getter`s and `Getter`s can be composed via `.` (and a `Lens` can be composed with a `Getter` to form a new `Getter`, because `Lens`s are also `Getter`s).

JacquesCarette:
Thanks, that's what I needed. I would be inclined to replace `Lens'` with `Getter` rather than going all the way down to function.

PR #3582, opened 2023-08-02
Stop producing an empty paragraph when there is no derivation

### LINKED ISSUE #3580: Empty paragraph produced when data definition has no derivation
In investigating #3576, I found that in the HTML SRS for GlassBR, there is spacing between data definition tables but not between instance model tables. As it turns out, the spacing between DD tables is put in on accident! It results from an empty paragraph inserted after each DD table, which is supposed to be the derivation of the DD but it is an empty paragraph when the DD has no derivation.

I think ideally we should really be outputting nothing as opposed to an empty paragraph when a DD has no derivation, and in this case the result actually looks different visually.

The most relevant piece of code is as follows.
```haskell
-- | Create a derivation from a chunk's attributes. This follows the TM, DD, GD,
-- or IM definition automatically (called automatically by 'SCSSub' program).
derivation :: (HasDerivation c, HasShortName c, Referable c) => c -> Contents
derivation c = maybe (mkParagraph EmptyS)
  (\(Derivation h d) -> LlC $ llcc (ref c) $ DerivBlock h $ map makeDerivCons d) $ c ^. derivations
```

hrzhuang:
Closes #3580


samm82 (commented):
I personally think we SHOULD be adding these spaces, since they break up the tables more obviously! Others may disagree with me, but I think I'd prefer the opposite change to be made: inserting these paragraphs between ALL models!

hrzhuang:
I agree there should be spacing between the tables, but I think that is the wrong way to do it.

I think we should keep "content" separate from "display", because the same content could be displayed in multiple ways (think the separation between HTML and CSS).

The empty paragraph in this case is inserted into the `Document`, which should just be content. Spacing between the tables is display, and should be added when we produce HTML, LaTeX, etc. from this `Document`.

samm82:
Good distinction! 😅

smiths:
I agree the space is a good thing, but it should be added the right way.

JacquesCarette (approved):

balacij:
It would probably be good to add a `theory` class to the generated DD, IM, GD, & TM HTML boxes, and then to add some CSS to add the spacing back, e.g.,
```css
div.theory + div.theory {
  margin-top: 1rem;
}
```

Issue #3259, opened 2023-02-02
Generate “Functional Requirement” for “Output Values” 


balacij:
# TODO

- [x] Update requirements to use existing commented-out code
   - [x] GamePhysics (#3609)
   - [ ] DblPend (#3533, #3610)
   - [x] SglPend (#3533)
   - [x] SWHS/NoPCM (see [this comment](https://github.com/JacquesCarette/Drasil/issues/3259#issuecomment-1636294793) and #3589)
   - [x] GlassBR (#3583)
- [ ] Add sources to tables (started in [`genOutReqAddSource`](https://github.com/JacquesCarette/Drasil/compare/genOutReqAddSource) by using the IMs/DDs themselves as the outputs rather than the quantities; waiting on DblPend's outputs to be resolved from above)
   - [x] HGHC (#3575)
- [x] Figure out case of one output (PDController) (see [this comment](https://github.com/JacquesCarette/Drasil/issues/3259#issuecomment-1636259007))
   - [ ] Make more globally usable (looks like there's some overlap with the following code - this should probably be a separate issue)
https://github.com/JacquesCarette/Drasil/blob/8d5e8f8088bb87cd351d07f4104faae7fd49111f/code/drasil-docLang/lib/Drasil/Sections/ReferenceMaterial.hs#L17-L24
- [x] Handle case of zero outputs (done in 15823ccc77)
   - [x] Create an issue for handling this at a higher level than requirement generation (#3652)
- [x] Reduce time complexity of `fullReqs` and `fullTables` (first attempt in 51cd19c010, but then overwritten by cb6ec6fb0a and 2ad9c309bf)
- [x] Make sure that the RefBy information of the IMs is preserved by the above change (i.e., each IM should be referenced by the output requirement) (done in 2794024)
- [x] Verify that the caption is generated in a way that allows for consistent obtaining of the requirement's reference
- [x] Document [`outReqDesc`](https://github.com/JacquesCarette/Drasil/commit/4424a3b3c81ee9cbcd0821b2d4d6ca79e7842d83#r121639809) (and any other related functions that need it)

### Background

[...]

Issue #3540, opened 2023-07-17
Generate FR for outputting input and known values


samm82:
As mentioned in [this comment](https://github.com/JacquesCarette/Drasil/commit/c3f57fd8365f36c625fe85ecb95de8a2bd0cb402#commitcomment-121759870), "it is a common ... practice in research software to output the inputs as part of the outputs ... for easy provenance/reproducibility.... Eventually I would see this as an option for users of Drasil. They could select to "output the inputs" if they want to."

This should be done similarly to the output FR generation in #3259, but with a choice to actually generate it or not (this choice would also affect code generation).

Issue #3523, opened 2023-07-08
Encode behaviour for constraint violations at a higher level


samm82:
_Note: this is mainly a place for me to collect my thoughts to see if this is something worth pursuing while facilitating conversation about it._

# Context

Currently, information about what to do when a constraint is violated is built by hand in a requirement:

https://github.com/JacquesCarette/Drasil/blob/afb6fb752b8364d2807ced7fc0c1dd6c6aba52b2/code/drasil-example/projectile/lib/Drasil/Projectile/Requirements.hs#L31-L34

This requirement is currently not actually met by the generated program; calculations continue even with constraint violations. (I noticed this while making [this manual test function for invalid inputs](https://github.com/samm82/Drasil/blob/c53404bfe8d2f44ac6d929bbe2aae031d8905c74/code/stable/projectile/projectile_c_p_nol_b_u_v_d/src/python/test/Control_test.py#L37-L53) for my project; note the "xfail" [or "expected failure"].)

I think this choice for the desired behaviour should be captured at a higher level (maybe as a `Choice`?) similarly to #3259 and the related work on generating input FRs (I can't find an issue number); this would allow us to generate the requirement text, the code, and (eventually!) the test cases consistently.

# (**VERY** Preliminary) Proposed Design

I'm thinking the options we could have might be something like:

```haskell
data ConsViolBehav = Error | Warn | Continue
```
where `Error` means "stop the program with an error to the user" (what Projectile _should_ do), `Warn` means "continue the program but warn the user" (what Projectile currently does), and `Continue` means "continue the program as normal". Note that the `Continue` option is essentially the same as not checking the input constraints at all; if we choose to implement this option, I think we can just not generate the requirement at all (and maybe not even display the constraints?) I also think the `Error` option should verify _all_ inputs, detect _all_ errors and output them to the user, _then_ stop execution so the user has the most information (avoiding a back-and-forth process of only correcting one thing at a time).

# Questions

1. Currently, this is what this requirement looks like for Projectile: "Verify-Input-Values: Check the entered input values to ensure that they do not exceed the [data constraints](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:DataConstraints). If any of the input values are out of bounds, an error message is displayed and the calculations stop." The second sentence is really what we're looking to generate here; should this be a separate requirement? If not, should we generate the first sentence as well (since it looks fairly stable) or should we allow the user some customization here?
[...]

PR #3583, opened 2023-08-02
Move GlassBR Output DDs to IMs

### LINKED ISSUE #3569: Move DDs that are outputs to be IMs in GlassBR
The dependencies (e.g., what should be moved first) are listed in parenthesis. The last two items are not output DDs, but are required by other ones.

## TODO:
- [x] AR (J, $\hat{q}_\text{tol}$)
- [x] B (Pb)
- [x] GTF ($\hat{q}$, LR)
- [x] h (B, NFL, $\hat{q}$, Jtol)
- [x] J (B)
- [x] Jtol ($\hat{q}_\text{tol}$)
- [x] NFL (LR)
- [x] $\hat{q}$ (J)
- [x] $\hat{q}_\text{tol}$ (NFL)
- [x] Pb (IM1)
- [x] LR (IM2)

### Miscellaneous
- [x] Add constraints for $P_\text{b}$ and $P_\text{tol}$
- [x] Create issue for getting inputs and their constraints automatically (#3584)
- [x] Clean up imports

samm82:
Closes #3569; originated from #3259.


JacquesCarette (approved):
The code changes seem fine.

However: there's a lot of changes to the generated code (various things are moved around) which doesn't seem as good. So this shouldn't be merged as is, not without some real discussion about why things shifted around so much [it might be fine!]

Also, maybe a master merge is needed too? There are some changes to .json files, but these should all be gone.

samm82:
As I made these changes, I tried my best to review the code changes; they seemed to be a logical consequence of the DD -> IM migration, but I figured I'd give @smiths a chance to review these changes before I went any deeper. I agree that a more in-depth investigation is in order, and will get around to it once I'm done the other preliminary work required to add the source information to #3259! 

The .json files are still present due to the pending merge of #3553

code/stable/glassbr/src/python/Calculations.py
```diff
@@ -34,121 +67,120 @@ def func_q(inParams):
...
+    print("  AR = ", end="", file=outfile)
+    print(AR, end="", file=outfile)
     print(", ", file=outfile)
     print("  q_hat = ", end="", file=outfile)
     print(q_hat, file=outfile)
     print("  }", file=outfile)
     outfile.close()
     
-    return Interpolation.interpZ("SDF.txt", inParams.AR, q_hat)
+    return Interpolation.interpZ("SDF.txt", AR, q_hat)
```
samm82, line 124: AR is no longer stored in `inParams`

code/stable/glassbr/src/python/InputConstraints.py
```diff
@@ -131,11 +122,3 @@ def input_constraints(inParams):
...
         print(".")
         raise Exception("InputError")
-    if (not(inParams.AR >= 1.0)) :
-        print("AR has value ", end="")
-        print(inParams.AR, end="")
-        print(", but is expected to be ", end="")
-        print("above ", end="")
-        print(1.0, end="")
-        print(".")
-        raise Exception("InputError")
```
samm82, line 141: AR constraint checking lost

samm82:
I looked into the changes to stable and noticed these groups of changes:

## Dependencies

Boiling down the dependencies of the IMs gives the following six groups, where the items in each group can be computed in any order. (Note that "X > _" means that the quantity only depends on values that are input or assumed, and "DD > _" means that the quantity is a DD, not an IM.) Looking at the side "Symbols" bar for [Calculations.py](https://github.com/JacquesCarette/Drasil/blob/moveGlassBrOutputDDsToIMs/code/stable/glassbr/src/python/Calculations.py) and [Control.py](https://github.com/JacquesCarette/Drasil/blob/moveGlassBrOutputDDsToIMs/code/stable/glassbr/src/python/Control.py) confirm that this ordering is preserved. (Side note: should we be ordering the IMs like this in the document? Follow up: if we figure out an appropriate ordering automatically when generating code, could this functionality be used to automatically sort the IMs?)

X > AR  
X > GTF  
X > h  
DD > LDF
DD > q

LDF, h > Jtol
q, h, GTF > ~q

AR, ~q > J
AR, Jtol > ~qtol

LDF, J, h, > B
~qtol > NFL

B > Pb
NFL, GTF > LR

Pb, Pbtol > isSafePb
LR, q > isSafeLR

## Input Parameters
A trivial change is that quantities that used to be DDs in the set values" step (see the [table of required assignments](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Table:ReqAssignments)) are no longer stored within `inParams` (see [here](https://github.com/JacquesCarette/Drasil/pull/3583/files#r1285146877)) for an example. This affects h, GTF, and AR.

### Input Constraints
Since AR is mainly used to calculate other values (i.e., J and $\hat{q}_\text{tol}$ and their dependencies), its constraints are treated as input constraints and can be found in the [relevant table](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Table:InDataConstraints). However, now that it is no longer treated like an input when checking the constraints, meaning that [its constraint checking code is lost](https://github.com/JacquesCarette/Drasil/pull/3583/files#r1285147148). Obviously losing this is not ideal, so I have some ideas:

1. **Verify it as an output as part of #3523.** Assuming that output verification happens after all the outputs are calculated, this means that every calculation will be performed even if AR is invalid, which is not ideal. This could potentially be made a future improvement if we go this route.
2. **Add an intermediate verification.** This is kind of what is done already, but instead of just having input and output verification, we could add a verification requirement for intermediate values. Hooking this up to code generation could potentially be more involved (and might be another distraction for me), so I'm not sure how practical this is.
3. **Revisit what we really want GlassBR to output.** When I initially started #3259, I just assumed that the outputs of GlassBR were "correct". Now, I'm realizing that may not be the case. In particular, AR shows up in the [Required Assignments table](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Table:ReqAssignments), meaning that it will already get output as part of the FR [Output-Values-and-Known-Values](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#outputValsAndKnownValues) (even if this is not actually done yet; see #3540). If enough quantities are not useful to be "final" outputs of the program (i.e., not part of Output-Values-and-Known-Values), then there may be no need to turn some of these DDs into IMs, including potentially AR! (Note that some DDs had to be moved to IMs so prevent cyclical imports.)

samm82:
As an update to [my previous comment](https://github.com/JacquesCarette/Drasil/pull/3583#issuecomment-1666698374), we decided to go with option 3 above: reverting the migration of DDs that would already be output by the FR [Output-Values-and-Known-Values](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#outputValsAndKnownValues): AR, GTF, and h. This undid most of the side effects listed above, although the order of LR and Pb swapped for some reason. Most importantly, the constraint checking on AR returned!

smiths:
I'm not done reviewing the changes, but I have found something that we should either fix in this PR, or create a separate issue for.  The problem actually predates the current changes, but looking at the changes highlights the previous mistake.  The change I'd like to see is that the TMs aren't actually used anywhere.  We don't have any consistency checks (that was the other potential project for you @samm82!), but if we did they would notice that the TMs aren't actually used for anything.  The TMs are generic statements about safety.  The corresponding IMs should refer to these TMs.  The IMs take the generic safety TMs and refine them into specific safety requirements for GlassBR.  The safety-related IMs should refer to their corresponding TMs.

samm82:
I think at some point the TMs were referenced, but in a place where really the IMs should have been referenced - I made that switch at some point, but didn't realize it meant that the TMs were no longer used. I think I'll make an issue for it so that it doesn't hold up (a) this PR or (b) my work on testing 😅 unless you think it's the right mix of important and straightforward where I should attempt it in this PR.

smiths:
@samm82, yes please create a separate issue about the safety-related TRs not being refined by the IMs.

I like the changes to stable.  I'll add my approval, although we should wait for approval from @JacquesCarette before merging.

smiths (approved):

JacquesCarette (approved):

PR #3587, opened 2023-08-03
Fix spacing between LaTeX tables

### LINKED ISSUE #3586: Lack of spacing between instance model tables in GlassBR PDF SRS
At a glance, this seems related to #3576, but as I will demonstrate I think this is a separate issue.

Since the instance models are so large, the two instance models in GlassBR are placed on separate pages, so this issue is not actually visible in the normally generated PDF. However, if I were to edit the `.tex` file and remove some of the rows from the tables, it becomes apparent.

Screenshot 2023-08-03 at 1 14 04 PM

This may not look terrible, but unlike in HTML (see #3576), in $\LaTeX$ we actually do ask for space between the two tables via the command `\vspace{\baselineskip}`, and that is way less space than we ask for! (`\baselineskip` is the distance between two lines of text in a normal paragraph)

The reason this happens is that `\vspace` needs to be followed by a newline, otherwise the table that follows gets placed on the same "line" as the `\vspace`. This is illustrated in the diagram below.

vspace

hrzhuang:
Closes #3586


samm82 (approved):

code/drasil-printers/lib/Language/Drasil/TeX/Helpers.hs
```diff
@@ -113,7 +113,7 @@ mkEnvArgSq nm args d =
 -- | Makes minipage environment.
 mkMinipage :: D -> D
 mkMinipage d = commandD "vspace" (command0 "baselineskip") $+$
-  command0 "noindent" $+$ mkEnvArgBr "minipage" "\\textwidth" d
+  command0 "noindent" $+$ mkEnvArgBr "minipage" "\\textwidth" d $+$ pure (text "")
```
samm82, line 116: Could `empty` or something like it be used here?

hrzhuang: No, I tried that but `Text.PrettyPrint` tries to be smart and omits the empty line.

code/stable/hghc/SRS/PDF/HGHC_SRS.tex
```diff
@@ -115,6 +115,7 @@ \subsubsection{Data Definitions}
 \\ \bottomrule
 \end{tabular}
 \end{minipage}
+
 \vspace{\baselineskip}
```
JacquesCarette, line 119: I'm additionally wondering if we should put some wiggle room (aka `glue` in TeX parlance)?

JacquesCarette (approved):

JacquesCarette:
There's a merge conflict to resolve before I can merge.

JacquesCarette:
I'm going to merge this (as it is an improvement) but then re-open #3586 as the place to keep going.

PR #3603, opened 2023-08-09
Work on clearing up some issues with the deploy stage target

balacij:
`make deploy_lite` seems to be largely broken, but with this, it's a bit less broken. 

Creating a PR because @harmanpreet-sagar was running into issues, and this is a decent POC to make it somewhat workable.

When running `make deploy`, there seems to be a dependency on `graphmod` which refers to a bad "dblpendulum" reference! So, it looks like there's some other issues too.

### DIFF
--- a/code/scripts/deploy_stage.sh
+++ b/code/scripts/deploy_stage.sh
@@ -74,12 +74,12 @@ copy_datafiles() {
   echo "FIXME: Drasil should copy needed images and resources to the appropriate output directory to avoid needing the entirety of datafiles (for HTML)."
   rm -rf datafiles
   mkdir -p datafiles
-  cp -r "$CUR_DIR"datafiles/. datafiles/
+  cp -r "$CUR_DIR/datafiles/" datafiles/
 }
 
 copy_graphs() {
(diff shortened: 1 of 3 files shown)


JacquesCarette (approved):

code/scripts/deploy_stage.sh
```diff
@@ -137,8 +139,7 @@ copy_images() {
     rm -rf "$CUR_DIR"deploy/images
   fi
   mkdir -p "$CUR_DIR"deploy/images
-  cp -r "$CUR_DIR"drasil-website/WebInfo/images/* "$CUR_DIR"deploy/images
-  
+  cp -r "$CUR_DIR/drasil-website/WebInfo/images/" "$CUR_DIR/deploy/images"
```
harmanpreet-sagar, line 142:
```suggestion
  cp -r "$CUR_DIR/drasil-website/WebInfo/images/" "$CUR_DIR/deploy"
```

Currently, a double images folder is being formed within deploy. 

image


Removing the `/images` part would prevent this from happening.

harmanpreet-sagar (approved):

PR #3605, opened 2023-08-09
Attempt 3: Remove capitalization from concepts in PDController

daijingz:
This time I follow the guidance, so I only change all of the pn parameters except the first one.
And mentioned as before, there is no content in the initial log files (files without change) After this change, log files have this output:
JDTN%$~E~{COL90 ZTSLEPA
2CT8@LO77~X`M@A(LV7@L`N
This result is consistent with my change, and the code works


samm82:
"These log files aren't actually added to commits, since they change very frequently. They are just a way for the developers (us) to see what changes to our generated output have been made by our changes to the code. If all of the changes in the log files look good, then the stable directory should be updated so that the generated version (in the build directory) and the stable version match. When your PR is ready to merge, it should include all changes that you've made to stable and running make should result in no changes being stored in the log files." ([from this comment](https://github.com/JacquesCarette/Drasil/pull/3570#pullrequestreview-1557629767))

Do you understand what the log files mean? Lines that start with "<" mean that that line has been removed from the generated file, and those with ">" mean that that line has been added. When the two occur together separated by "---", it means the first line has been replaced with the second.

These logs look good (although copying them would be better than pasting pictures of them), so "the stable directory should be updated so that the generated version (in the build directory) and the stable version match".

samm82:
(Also commit messages should be more descriptive then "add files" - what change(s) did you make in the commit and why?)

daijingz:
Let me upload more relevant information.

samm82:
Can you please stop closing your PRs? This change is on the right track; you just need to update `stable`, which can be done in this branch/PR.

Issue #3600, opened 2023-08-08
Allow `InstanceModel`s to have more than one output


samm82:
While working on #3589, I noticed in [this comment](https://github.com/JacquesCarette/Drasil/issues/3589#issuecomment-1664842523) that there are not one, but _three_ quantities "output" by a single IM: the actual function it defines, and two implicit values:
![SWHS's implicit values](https://github.com/JacquesCarette/Drasil/assets/35857611/3645d46f-c276-4aac-9259-f932f5b6d24e)

@smiths mentioned in #3593 that while most IMs have only one output, we shouldn't limit them to always have only one, since there are occasions like this where "coupled equations" exist.

_Note that when this is done, the above FRs should be removed from SWHS and the new outputs of the IM mentioned above should be added to its `outputs` in its `SystemInformation`; hopefully #3259 is done by then, and if the calculation requirement isn't automated yet, these outputs should also be added to it._

JacquesCarette:
Right - this is yet another reason why `HasOutput` is dangerous (@hrzhuang) to use a little too "blindly".

PR #3607, opened 2023-08-09
Reorganize SWHS FRs

### LINKED ISSUE #3589: Reorganize calculation and output FRs of SWHS
Currently, SWHS (and NoPCM) are the only two examples (that I've encountered) that group the calculation and output requirement by quantity (i.e., having one requirement for calculating **and** outputting **each** value) instead of by operation (i.e., having one requirement for calculating **all** values and one for outputting **all** values). Not only is this inconsistent and capture multiple functionalities in the same requirement, it also is a roadblock to #3259.

samm82:
Originated from #3259 and closes #3589.

I regrouped the requirements of SWHS (and NoPCM) so there is one* calculation requirement and one output requirement.

*As mentioned [here](https://github.com/JacquesCarette/Drasil/issues/3589#issuecomment-1670603265), two of the original requirements in SWHS will remain until #3600 is completed.


JacquesCarette (approved):
The code changes look good to me.

smiths (approved):
The changes to stable look good.  They are an improvement to what we had before.

Issue #3533, opened 2023-07-14
Confusion on DblPend's output (and maybe SglPend?)


samm82:
The output requirement of DblPend states that the outputs of the system are $\alpha_1$ and $\alpha_2$,
![image](https://github.com/JacquesCarette/Drasil/assets/35857611/d0a4c393-80d5-42d1-a59e-1411c3c0d174)

but the actual output defined in the code (passed to `SystemInformation`) is just $\theta$
https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-example/dblpendulum/lib/Drasil/DblPendulum/Unitals.hs#L35-L36

which shows up in the output table generated (as part of my work on #3259) as:

![image](https://github.com/JacquesCarette/Drasil/assets/35857611/00125789-2702-4da0-baf4-92e4d9c10dc6)

I'm assuming that the outputs of $\alpha_1$ and $\alpha_2$ are correct, since they are what the IMs output, but changing `outputs` to be these values (either just the unitals or the IMs themselves) results in the following error which prevents code from being generated:

```
dblpendulum.EXE: The following outputs cannot be computed: angularAccelerationIM1, angularAccelerationIM2
Unused definitions are: alpha_x1, alpha_y1, p_x1, p_y1, p_x2, p_y2, force
Known values are: l_1, l_2, m_1, m_2, gravitationalMagnitude, pendDisAngle
CallStack (from HasCallStack):
  error, called at lib\Language\Drasil\CodeSpec.hs:179:20 in drasil-code-0.1.9.0-HcWsGH034SJAsZRmMYbtvM:Language.Drasil.CodeSpec
make: *** [Makefile:270: dblpendulum_gen] Error 1
```

although the SRS output changes to match what is expected
![image](https://github.com/JacquesCarette/Drasil/assets/35857611/029ed6ed-7cab-4307-8d37-27514905bb54)

Is there something I'm missing here? How can we change the outputs of the system to be correct so they can actually be computed?

cd155:
In short, the `theta` is only for solving the higher-order ODE. It is a hack to make the ODE solver works. It does not belong to the SRS. More like to belong to a design document, and it is an artifact.

We have a discussion discuss it before, you can see more details in this link. https://github.com/JacquesCarette/Drasil/issues/2986#issuecomment-1152504793

smiths:
[...]

PR #3610, opened 2023-08-10
Update DblPend's IM outputs to be theta instead of alpha

samm82:
Based on the discussion on #3533, I updated the outputs of DblPend's IMs to output the thetas themselves, since that is what is currently desired from them.

image


samm82:
I also update the relevant FRs (note that the output one will soon be generated in table form):

image

After #3600, we could also output the alphas fairly trivially (depending on how it would affect the code generation).

JacquesCarette (approved):
Code changes make sense to me. Might be worth merging in master so that no .json leaks in.

smiths (approved):
The changes look good to me.  I'm assuming that these changes don't break the code generator, since there are no code changes for stable listed.

There are some odd things that I happened to notice in the generated code, like in the html code the symbol for omega is different between the equation and the list in the description.  The order of L_1 and L_2 are opposite to what would be expected in the description field.  These are things we can fix in the future.

smiths:
It looks like the json files have leaked in!

cd155 (approved):

JacquesCarette (approved):

PR #3614, opened 2023-08-10
Use `HasOutput` for `mkQRTuple`

### LINKED ISSUE #3613: `mkQRTuple` should use `HasOutput`
Currently `mkQRTuple` treats `InstanceModel`s and `DataDefinition`s as quantities, when in fact they should be treated as having quantities as outputs. See also #3575.

hrzhuang:
Closes #3613. Builds on #3612.

This resulted in a `stable` change to a functional requirements table in GlassBR, but I think it is actually a problem with the original table?


smiths (approved):
From the point of view of stable the changes are fine.  In fact, the wording that now shows in the new version of the required outputs table is better than the old wording.

JacquesCarette (approved):

Issue #3598, opened 2023-08-08
Add a stable for the main Drasil Website and change the manner in which `make deploy_lite` runs


harmanpreet-sagar:
The main Drasil website is currently generated by Drasil, however, it is not part of stable. Based on what was discussed, the best way to move forward on this would be to create a new stable folder specifically for the website. This would also allow us to change the way `make deploy_lite` works as we could use this issue to modify the conditions of the makefile to use artifacts currently present in stable to aid in the deployment of the main Drasil website.

What do you think? @smiths @JacquesCarette @balacij @samm82

smiths:
Sounds good to me.

JacquesCarette:
As we discussed in the meeting: yes, but not "merged" with the current stable. We should have a little bit of separate between our different use cases.

PR #3622, opened 2023-08-10
3532 generate description using purp

### LINKED ISSUE #3532: Generate description of Examples using `purp` in `Body.hs` of all examples on Website.
The description for most of the examples provided on the main Drasil website does not currently match the purposes specified for each of those examples in their `Body.hs` files. For example, the website's description for `GlassBR` reads: "predicts whether a given glass slab is likely to resist a specified blast." However, the purpose of `GlassBR` as evident below is slightly different from the description on the main website. 

https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs#L126-L128

The text on the Drasil website for all the examples is currently obtained from these lines of code here:

https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-website/lib/Drasil/Website/Example.hs#L158-L167

This is the case for all the various other examples as well. `PD Controller` currently does not have a description on the main website. The following image indicates the location on the website where the issue is prevalent. 

image

The location of the code where this description is being generated is as follows:

https://github.com/JacquesCarette/Drasil/blob/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-website/lib/Drasil/Website/Example.hs#L59-L88

We should obtain the description for each of these examples using the `purp` that is specified in the body files for each of these examples in order to remove the duplicity of these sentences. 

In addition, text files that contain just one line for most of these descriptions are also present in the folder linked [here](https://github.com/JacquesCarette/Drasil/tree/1acb8ba1553d16d1c949e3b55049fb2242dfe555/code/drasil-website/WebInfo/descriptions). I am wondering as to why these text files exist?

What do you think @JacquesCarette @smiths @balacij @samm82?

harmanpreet-sagar:
Closes #3532. Changes made to `drasil-example`, `drasil-website` and stable. Contributes to #3598.


code/drasil-website/lib/Drasil/Website/Body.hs
```diff
@@ -14,6 +14,17 @@ import Drasil.Website.Example (exampleSec, exampleRefs, allExampleSI)
 import Drasil.Website.Documentation (docsSec, docRefs)
 import Drasil.Website.Analysis (analysisSec, analysisRefs)
 import Drasil.Website.GettingStarted (gettingStartedSec)
+import Data.Drasil.Concepts.Physics (pendulum, motion, rigidBody)
```
JacquesCarette, line 17: Something is very wrong here: there is no way the website should depend on all these things added here!

code/drasil-website/lib/Drasil/Website/Body.hs
```diff
@@ -89,8 +100,12 @@ sections fl = [headerSec, introSec, gettingStartedSec quickStartWiki newWorkspac
 
 -- | Needed for references and terms to work.
 symbMap :: FolderLocation -> ChunkDB
-symbMap fl = cdb ([] :: [QuantityDict]) (map nw [webName, web] ++ map getSysName allExampleSI)
-  ([] :: [ConceptChunk]) ([] :: [UnitDefn]) [] [] [] [] [] [] [] $ allRefs fl
+symbMap fl = cdb ([] :: [QuantityDict]) (map nw [webName, web, phsChgMtrl] ++
```
JacquesCarette, line 103: ?????

JacquesCarette (changes requested):

code/drasil-website/lib/Drasil/Website/Body.hs
harmanpreet-sagar: The reason the website depends on all these things added here is that we are now using the purposes as defined in `Body.hs` of each of the examples. Since those purposes make use of these terms when writing the `Sentence`, the website now has to depend on these imports.

code/drasil-website/lib/Drasil/Website/Body.hs
harmanpreet-sagar: The reason I put `phsChgMtrl` there is because it is of the type `CI` whereas the following list that contains terms such as `pendulum`, `motion`, etc. are of type `ConceptChunk`. I tried putting phsChgMtrl with the others but it did not allow me to compile it due to type mismatch

harmanpreet-sagar:
The website would look like this after the latest commit:

image

JacquesCarette (approved):
This is sub-optimal, in the sense that the code embeds too much knowledge of things that lie outside the code (i.e. what examples we have). But this is still an improvement, so I'll merge it in. We can work on making things more dynamic later.

PR #3625, opened 2023-08-10
Replace `\vskip` with `\medskip` in model tables

### LINKED ISSUE #3586: Lack of spacing between instance model tables in GlassBR PDF SRS

hrzhuang:
Closes #3586.

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Helpers.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Helpers.hs
@@ -112,7 +112,7 @@ mkEnvArgSq nm args d =
 
 -- | Makes minipage environment.
 mkMinipage :: D -> D
-mkMinipage d = commandD "vspace" (command0 "baselineskip") $+$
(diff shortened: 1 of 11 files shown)


smiths (approved):

JacquesCarette (approved):

Issue #3584, opened 2023-08-02
Populate "Input" and "Input Constraints" fields of IMs automatically


samm82:
As mentioned in #3569, our IMs require their inputs (and the constraints on those inputs) to be entered manually. This results in duplication, causing a lack of traceability and a proneness to errors (again, see discussion on #3569). Since we can already "scrape" the quantities used in the "Equation" field to generate the "Description" field, it should be relatively straightforward to populate the "Input" field using these values, then getting all their constraints for the appropriate field.

One note about this is that the output is included in the list of quantities in the description. I'm tempted to assume that the left side of the equation can be used as the output, but with the examples involving DEs, I'm not sure this is the case. I think it makes sense to have the user specify the output, then remove this output from the list of quantities before it gets used in the "Inputs" field (while keeping it in the description). It may also be possible to automatically pull any constraints of this output to be used in the "Output Constraints", but this may have to be a separate issue if it is non-trivial.

smiths:
I'm definitely in favour of as much automation as possible.  If scraping the equation for the description field is working already, I think you are right that we can also find the inputs.  I'm not sure it will always work, but at least for our current case studies, I think it should.

@samm82 I agree that we can't just take the left-hand side of the equation as the output.  As you mention, we have ODEs where the left-hand side is a derivative, but the output is supposed to be the time history of the thing that is changing, not it's rate of change.  We have a similar problem with partial differential equations.  We also have equations like Ax = b for a linear system of equations.  The output for this equation is nearly always x, even though it isn't typically written as x = A^-1 b.  We also might define a quantity implicitly, where the thing that we want appears on both sides of the equation.  This comes up in some root-finding problems.  Even an example like finding the roots of a quadratic equation is often better expressed as finding the x values that make a x^2 + b x + c = 0, rather than giving the closed-form quadratic equation.  As the polynomials get higher order, we don't have access to such nice formulas as the quadratic formula.

PR #3627, opened 2023-08-14
Implement `DefinesQuantity` for `InstanceModel` and `DataDefinition`

### LINKED ISSUE #3621: Can `DataDefinition` and `InstanceModel` implement `DefinesQuantity`?
This is motivated by the `definedIn` family of functions.

```haskell
-- | Takes a 'HasSymbol' that is also 'Referable' and outputs as a 'Sentence': "@symbol@ is defined in @reference@."
definedIn :: (Referable r, HasShortName r, HasSymbol r) => r -> Sentence
definedIn q = ch q `S.is` S "defined in" +:+. refS q

-- | Same as 'definedIn', but allows for additional information to be appended to the 'Sentence'.
definedIn' :: (Referable r, HasShortName r, HasSymbol r) => r -> Sentence -> Sentence
definedIn' q info = ch q `S.is` S "defined" `S.in_` refS q +:+. info
```

Correct me if I'm wrong, but as far as I can see `InstanceModel` and `DataDefinition` are the only types that satisfy both `Referable` and `HasSymbol`. So those are currently the only possible arguments to the `definedIn` functions. As part of #3573 and #3617 (part of #3574), I would like to replace `HasSymbol` with `HasOutput` in the type signatures of the `definedIn` functions.

However, `HasOutput` is in `drasil-theory` and the `definedIn` functions are in `drasil-lang`, and we don't want `drasil-lang` to depend on `drasil-theory` (see also #3591 and #3597). Would it be appropriate then to implement `DefinesQuantity` for `InstanceModel` and `DataDefinition` so we can use it in the `definedIn` functions as follows?

```haskell
definedIn :: (Referable r, HasShortName r, DefinesQuantity r) => r -> Sentence
definedIn q = ch (q ^. defLhs) `S.is` S "defined in" +:+. refS q

definedIn' :: (Referable r, HasShortName r, DefinesQuantity r) => r -> Sentence -> Sentence
definedIn' q info = ch (q ^. defLhs) `S.is` S "defined" `S.in_` refS q +:+. info
```

hrzhuang:
Closes #3621.

### DIFF
--- a/code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
+++ b/code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
@@ -71,6 +71,9 @@ instance HasUID             DataDefinition where uid = ddQD uid uid
 instance NamedIdea          DataDefinition where term = ddQD term term
 -- | Finds the idea contained in the 'QDefinition' used to make the 'DataDefinition where'.
 instance Idea               DataDefinition where getA = either getA getA . qdFromDD
+-- | Finds the 'Quantity' defined by the 'DataDefinition'
+instance DefinesQuantity    DataDefinition where
+  defLhs = ddQDGetter defLhs defLhs
 -- | Finds the output variable of the 'DataDefinition'
 instance HasOutput          DataDefinition where
   output = ddQDGetter defLhs defLhs
--- a/code/drasil-theory/lib/Theory/Drasil/InstanceModel.hs
+++ b/code/drasil-theory/lib/Theory/Drasil/InstanceModel.hs
@@ -70,6 +70,9 @@ instance CommonIdea         InstanceModel where abrv _ = abrv inModel
 instance Referable          InstanceModel where
   refAdd      = ra
   renderRef l = RP (prepend $ abrv l) (refAdd l)
+-- | Finds the 'Quantity' of an 'InstanceModel'
+instance DefinesQuantity    InstanceModel where
+  defLhs = imOutput . _1
 -- | Finds the inputs of an 'InstanceModel'.
 instance HasInputs          InstanceModel where
   inputs          = imInputs


code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
```diff
@@ -71,6 +71,9 @@ instance HasUID             DataDefinition where uid = ddQD uid uid
 instance NamedIdea          DataDefinition where term = ddQD term term
 -- | Finds the idea contained in the 'QDefinition' used to make the 'DataDefinition where'.
 instance Idea               DataDefinition where getA = either getA getA . qdFromDD
+-- | Finds the 'Quantity' defined by the 'DataDefinition'
+instance DefinesQuantity    DataDefinition where
+  defLhs = ddQDGetter defLhs defLhs
```
samm82, line 76: What do we mean by "LHS"? Do we mean explicitly the LHS of the equation, or are we using it in the "lvalue" sense to literally mean the quantity that the model defines? (See a related discussion [here](https://github.com/JacquesCarette/Drasil/issues/3584#issuecomment-1675199708))

hrzhuang: The latter. I recently introduced `defLhs` in #3601, submission for a better name is open  😄.

samm82: It seems to me that this overlaps with `HasOutput`, just without the output constraints. Is there a case where something would be an instance of `DefinesQuantity` without being one of `HasOutput`? If not, is `DefinesQuantity` not already implicit in `HasOutput`? (Of course, we may want to keep the two separate in the case of constraining functions, etc.

Issue #3221, opened 2023-01-05
Generate PEP8 compliant Python code


smiths:
The generated Python code is not [PEP8](https://peps.python.org/pep-0008/) compliant.  There aren't too many problems though.  Running `flake8` finds errors related to the number `#` leading block comments, line length and whitespace.  For instance, the following output is produced when running `flake8` on Projectile:

```
./Calculations.py:1:1: E266 too many leading '#' for block comment
./Calculations.py:6:1: E266 too many leading '#' for block comment
./Calculations.py:10:1: E302 expected 2 blank lines, found 1
./Calculations.py:13:1: E266 too many leading '#' for block comment
./Calculations.py:13:80: E501 line too long (113 > 79 characters)
./Calculations.py:16:80: E501 line too long (102 > 79 characters)
./Calculations.py:17:1: E302 expected 2 blank lines, found 1
./Calculations.py:18:80: E501 line too long (104 > 79 characters)
./Calculations.py:20:1: E266 too many leading '#' for block comment
./Calculations.py:20:80: E501 line too long (151 > 79 characters)
./Calculations.py:22:80: E501 line too long (108 > 79 characters)
./Calculations.py:23:80: E501 line too long (140 > 79 characters)
./Calculations.py:24:1: E302 expected 2 blank lines, found 1
./Calculations.py:27:1: E266 too many leading '#' for block comment
./Calculations.py:30:80: E501 line too long (148 > 79 characters)
./Calculations.py:32:1: E302 expected 2 blank lines, found 1
./Calculations.py:33:59: E203 whitespace before ':'
./Calculations.py:35:26: E203 whitespace before ':'
./Calculations.py:37:9: E203 whitespace before ':'
./Control.py:1:1: E266 too many leading '#' for block comment
./InputParameters.py:1:1: E266 too many leading '#' for block comment
./InputParameters.py:3:80: E501 line too long (158 > 79 characters)
./InputParameters.py:6:1: E266 too many leading '#' for block comment
./InputParameters.py:7:1: E302 expected 2 blank lines, found 1
./InputParameters.py:8:5: E266 too many leading '#' for block comment
./InputParameters.py:8:80: E501 line too long (103 > 79 characters)
./InputParameters.py:13:1: W293 blank line contains whitespace
./InputParameters.py:14:5: E266 too many leading '#' for block comment
./InputParameters.py:25:1: W293 blank line contains whitespace
./InputParameters.py:26:5: E266 too many leading '#' for block comment
[...]

PR #3640, opened 2023-08-15
Remove space before colon in Python files

harmanpreet-sagar:
Contributes to #3221. Removes spaces before the colon in Python files, making it more PEP8 compliant. Changes have been made to `drasil-gool` and stable.


smiths (approved):
The changes to stable look good to me.  I seem to remember a previous attempt to fix this that had an unwanted consequence on the generation of the other programming languages, but there are no changes in stable to any of the other languages, so there are no problems on that front.

JacquesCarette:
Strictly speaking, I'm not super keen on this method of fixing the problem. It's a hack. I think that Python should choose to have its own renderer for `if` statements rather than over-engineering the language-polymorphic version with things that only Python cares about.

smiths:
Makes sense @JacquesCarette.   Should we create an issue to add the Python rendered for `if` statements?

JacquesCarette:
I'm thinking it's not worth bothering - we may as well wait until we have someone that needs to do a substantial refactor.

PR #3643, opened 2023-08-18
3221 remove colon part 2

harmanpreet-sagar:
Contirbutes to #3221. While writing an update to the issue, I realized that I missed a few areas in the python files that still have the whitespace before colon, so here is another PR to fix that. Changes made to `drasil-gool` and stable. This should now fix all the `E203 whitespace before ':'` errors.


smiths (approved):
I checked a few instances of the generated code to verify that all instances of E203 have been removed.  The changes to stable look good to me.

JacquesCarette (approved):

PR #3644, opened 2023-08-18
3221 Create flake8 config file.

harmanpreet-sagar:
Contributes to #3221. I added this file directly in the code folder because running `flake8` in the code folder checks all the Python files, whether they are in stable, build or in scripts. This config would apply to all of them.


harmanpreet-sagar:
The ignore statement in this file should allow us to ignore `E266 too many leading '#' for block comment` as we need the multiple '#' to tell doxygen to parse a given piece of code. For more info view [this comment](https://github.com/JacquesCarette/Drasil/issues/3221#issuecomment-1397619288).

smiths:
@harmanpreet-sagar, I verified that this works.  However, did you branch from right branch?  The error E203 was reintroduced for a couple of cases in the projectile example I tests.  (My guess is that this will all get sorted when everything gets merged.)

Putting the `.flake8` file in the `code` folder is a minimal solution, and it has a certain elegance, but I wonder if the `.flake8` file should be produced in the folder of each python project?  For what we are doing now, your solution is great.  However, in the future, we'll generate code for one project at a time.  We won't be generating a bunch of case studies; we'll be generating the code that someone can take and use.  If someone copies the contents of a Python folder, they won't get `.flake8` and the PEP8 errors will come back.

I'm not actually sure which approach is better:

1) Putting `.flake8` in the top folder (`code`) using the fact that all projects will implicitly inherit this configuration.
2) Putting `.flake8` in every python project explicitly.

The disadvantage of the second approach is repeating the same file in multiple places.  The advantage is that each project is self-contained, without an implicitly included file.  Another advantage is that this would let us have variability in the `.flake8` configuration.  Right now we are assuming all projects have the same configuration, but in the future we may want more flexibility.

My vote is for option 2, but maybe @balacij or @JacquesCarette have a different opinion?

harmanpreet-sagar:
This was a separate branch and hence the E203 error was showing up. I have now merged that branch as well into this PR. The main goal of my putting it in the code was to prevent file duplication. I tried going into separate folders and this config seemed to work even when running flake8.

smiths:
@harmanpreet-sagar, yes your `code` folder solution will work.  The question is whether it is a better solution in the longterm compared to duplicating the file in each python folder.  There are pros and cons to each approach.

balacij:
Regarding @smiths comment, I think having the config files be per-project is best so that we can bring it closer to each project. For example, we might want some generated code to be compliant with Python Black standards rather than just PEP8.

Speaking of which, do we want these configuration files to do the linting and formatting, or do we want them to only show that the generated code already obeys these relevant standards?

smiths:
@balacij I was just thinking of the configuration file as configuration for the `flake8` check for `PEP8` compliance.  The generator is responsible for generating PEP8-compliant code.  In the cases where we wish to generate code that isn't `PEP8` compliant, like when we generate doxygen code, we configure `flake8` to ignore these warnings.

If there is something clever that we can potentially do with linting and formatting the generated code, I'd be happy to learn more about it.  :smile:

JacquesCarette:
@balacij since @harmanpreet-sagar 's internship is over, would you be able to take over this PR to conclusion?

balacij:
I don't mind. I might split his PR into 2 separate ones. The `flake8` configuration discussion is a bit of a different solution than proposed in this PR, but this PR does seemingly fix a bug with spacing Python expressions.

balacij:
Merged in stable, and now the diff is truly just the 1 file, so we can close this PR until we have the right solution.

PR #3647, opened 2023-08-21
Remove capitalization from concepts in PDController #3534

daijingz:
Now I submitted both code/stable/ and the modified file
Let me take a break and left the field blank (I am exhausted, wait a moment)


balacij:
The size of the changes are quite worrying. It looks like the encoding of the files were changed. I don't think the diffs that GitHub shows us your PR is making should happen. Do you think you can look into this @daijingz?

smiths:
I agree with @balacij.  I'm glad you are making progress @daijingz, but it isn't possible to review the diffs since they show the entire files deleted and then replaced.

daijingz:
I am confused with this as well because I just change one file in the folder.
However, the repository /code/stable/ has 60 files, I have to upload them as well
I do not know solve this problem

balacij:
I don't use Windows, so I'm unfortunately of little help here, but have you set your locale encoding in your terminal to UTF-8? https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#unicode-support-important

daijingz:
I do not know how can we see the setting UTF-8. I guess it is already the default setting

balacij:
Did you have a look at the paragraph I linked to from the 'new workspace' wiki page? Was it any help? Note that after trying things, you should try to `make stabilize` again to make sure that the change was positive.

daijingz:
This time I have already used the command Make stabilize, so I think I just run the program again with the mentioned command, and then I just upload all the files from my desktop.
(Before I just change my modified files in the GitHub folder, then I make a submission)

JacquesCarette:
What do you mean by 'upload'? I'm afraid you might be mis-using `git`, which certainly could also lead into odd problems.

Issue #3648, opened 2023-08-22
Fix outputs of SSP


samm82:
Currently, there are many problems with SSP (the following is quoted from Dr. Smith in #3630):

- [ ] IM:crtSlpId should be phrased as a minimization problem. The output should be the minimum factor of safety and the critical slip surface [x_cs and y_cs]. It is found by minimizing Fs (as defined in IM:fctSfty) with respect to the coordinates of the slip surface.... We probably also need an F_s^c to represent the critical safety factor.
   - [Currently,] IM:crtSlpId is poorly written and incorrect. I believe the equation for IM:crtSlpId must have been inadvertently changed at some time. The equation x_slope = y_slope is not correct. It defines a straight line with a slope of 1 and an intercept of 0. That isn't what we are looking for. 
- [ ] The output[s of SSP] should be the factor of safety and the critical slip surface.

JacquesCarette:
Some things got worse in SSP back in the day when we only had `Expr`. Now that `Expr` has been split into 3, it is possible to re-add some things into the model/display expressions that are not meant to be internally understood.

PR #3649, opened 2023-08-22
Remove (x,y) from outputs of SSP (and update a table caption)

### LINKED ISSUE #3630: Confusion on SSP's output
From my WIP on #3259, these are the current outputs of SSP:
image

F<sub>S</sub> is unambiguous:
image
but where does (x,y) come from? First of all, what does it represent? It looks like it is just a generic representation of any point (here is the most information I could find about it in the document:)
image

Based on the rest of the FRs (shown below), it looks like this should actually represent the coordinates of the slip slope surface, in which case it looks like it is calculated by the IM below:
image
image

From its description, it doesn't look like F<sub>S</sub><sup>min</sup> is actually "outputted" by its IM, rather, it looks to be a sequence of x and y values (which could either be two separate outputs - #3600 - or a single sequence of (x,y) values). Is my analysis correct (or at least on the right track)? Thoughts?

_There are also other values in the FRs that look like they should be defined as outputs but aren't currently; I'm wondering if that should just be something we tackle later._

samm82:
This, along with the creation of #3648, closes #3630


smiths (approved):
I'm good with the changes to stable.

samm82:
I also updated the caption of the following table to reference its relevant FR to be consistent with the rest of the FR tables (and since it made testing the code from 279402458fc6bbcc66e28a358cb0095bfd419aad more challenging).

I also capitalized "Returned" and "Output" because I thought they should be; was this correct or should I revert this?

image

smiths:
The changes sound good to me.

JacquesCarette (approved):

Issue #3581, opened 2023-08-02
`HasDerivation` should be called `MayHaveDerivation`


hrzhuang:
The class `HasDerivation` actually only gives you a `Maybe Derivation`.
```haskell
class HasDerivation c where
  derivations :: Lens' c (Maybe Derivation)
```

JacquesCarette:
Good remark. This would be an easy thing to just knock off and get a sense of accomplishment!

daijingz:
After using the grep command, I see these occurrences:
![U~ Z(FQ_JUUAX@H71SSX4CA](https://github.com/JacquesCarette/Drasil/assets/61481741/c04a5a2c-8da8-4d8a-ad7b-5d3d7df09c6e)
I am going to change these in the next step

daijingz:
After changing all these places, there is no obvious errors
Here is the output:
![1QE2I5WWVT( L3825CT$EV](https://github.com/JacquesCarette/Drasil/assets/61481741/e3d59b1f-34b8-47f8-b23e-185961a8f23b)
I am going to build a new request soon

PR #3653, opened 2023-08-29
HasDerivation should be called MayHaveDerivation #3581

daijingz:
My Pull request for issue #3581

I feel there are not too many things that need to be mentioned here. What kind of information should I include here?

Assume all mentioned "HasDerivation" places are relevant to the target class "HasDerivation" so that those instances are required to be modified

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.11.0
+        uses: styfle/cancel-workflow-action@0.12.0
         with:
           access_token: ${{ github.token }}
   build:


code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
```diff
@@ -87,7 +87,7 @@ instance HasDecRef          DataDefinition where getDecRefs = ddPkt pktDR
 -- | Equal if 'UID's are equal.
 instance Eq                 DataDefinition where a == b = (a ^. uid) == (b ^. uid)
 -- | Finds the derivation of the 'DataDefinition where'. May contain Nothing.
-instance HasDerivation      DataDefinition where derivations = ddPkt pktMD
+instance MayHaveDerivation      DataDefinition where derivations = ddPkt pktMD
```
balacij, line 90: Minor code style issue: we align the type arguments of lists of instances.

balacij (changes requested):

code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
daijingz: Do not understand what should I change, do you mean this? @balacij (I use the above example)
instance HasDerivation               DataDefinition where derivations = ddPkt pktMD
instance MayHaveDerivation      DataDefinition where derivations = ddPkt pktMD <-- now there argument types are aligned

balacij: Just so you know, you can wrap code with 3 backticks to make a code block. Using the code block will use a monospaced font, so seeing the spacing is easier :smile: 

I _think_ that looks right. The point is that the `DataDefinition`s should be vertically aligned there.

daijingz: So what should I do is just aligning all ```DataDefinition```
Here is an example (as stated beyond):
```
instance HasDerivation               DataDefinition where derivations = ddPkt pktMD
instance MayHaveDerivation           DataDefinition where derivations = ddPkt pktMD
```
Now it is aligned

balacij: Yes, but that looks like you added too much space. Align the `DataDefinition where...` onwards with the others listed in the files, please.

daijingz: Sorry for the delay. (Working on other things) In this modification, I changed three files: 
```code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs```
```code/drasil-theory/lib/Theory/Drasil/GenDefn.hs```
```code/drasil-theory/lib/Theory/Drasil/InstanceModel.hs```

They have similar alignment problems, but now all these issues have been solved. As you can see: https://github.com/JacquesCarette/Drasil/pull/3653/files#diff-bab435096d39e524c93900a45bea3ad6e1b439001c0a54ea1f27d94d28bd824e

balacij (approved):
Thanks, @daijingz!

PR #3662, opened 2023-09-15
Fix website deploy script by copying files to expected areas

### LINKED ISSUE #3646: Website links to the HTML versions of our examples are broken
While the URLs seems to be correct, the HTML file can't be found - was there a change to how we deploy the website that may have broken this?

image

balacij:
Closes #3646 

Builds on #3657

I'm leaving this as a draft because I want to make sure folder links are plugged in properly. The structure of the `gh-pages` branch is very confusing, we might need to re-evaluate that at some point.

To test this branch:
1. `make clean`,
2. `make deploy`,
3. get a coffee while waiting for (2) to finish -- it takes a crazy amount of time building the PDFs (which it shouldn't even be needing to do!),
4. open `deploy/HTML/index.html` in your local web browser,
5. look at the "HTML" and "PDF" links related to each example, and make sure they don't lead to 404s

### DIFF
--- a/code/scripts/deploy_stage.sh
+++ b/code/scripts/deploy_stage.sh
@@ -90,7 +90,7 @@ copy_examples() {
     example_name=$(basename "$example")
     # Only copy actual examples
     if [[ "$EXAMPLE_DIRS" == *"$example_name"* ]]; then
-      target_srs_dir="./HTML/$EXAMPLE_DEST$example_name/$SRS_DEST"
+      target_srs_dir="./$EXAMPLE_DEST$example_name/$SRS_DEST"
       mkdir -p "$target_srs_dir"
       if [ -d "$example/SRS/PDF" ]; then
         cp "$example/SRS/PDF/"*.pdf "$target_srs_dir"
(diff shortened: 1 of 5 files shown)


code/scripts/deploy_stage.sh
```diff
@@ -90,19 +90,22 @@ copy_examples() {
...
       fi
-      if [ -d "$example/"src ]; then
+      if [ -d "$example/src" ]; then
         mkdir -p "$EXAMPLE_DEST$example_name/$DOX_DEST"
-        for lang in "$example/"src/*; do
+        for lang in "$example/src/"*; do
           lang_name=$(basename "$lang")
-          mkdir -p "$EXAMPLE_DEST$example_name/$DOX_DEST$lang_name"
-          cp -r "$lang/" "$EXAMPLE_DEST$example_name/$DOX_DEST$lang_name/"
+          if [ "$lang_name" != "swift" ]; then
```
balacij, line 105: Having "swift" hard-coded here is odd because it's based on implicit knowledge, but I think this whole file has the same issue. It would be nice if we had "generation receipts" made by each example, containing an itemized (structured/nested) list of things it generated, that we could use in 2 ways: (i) in `drasil-website` to know what things to "post" on the website, and (ii) with a small script that does the necessary copying needed.

JacquesCarette (approved):

PR #3663, opened 2023-09-16
Hide debug chunk tables behind the `make debug` target

balacij:
There was a leftover chunk table dump generated (in the `debug/` directory) with each example run, slowing them down. So, I reconciled the chunk table dumps with the existing `make debug` target.


JacquesCarette (changes requested):
I debated between just pulling this in, and pointing out that you've introduced a huge amount of duplication... and duplication won.

I'll let you analyze what I think is duplicated as an exercise that we can discuss.

balacij:
Hmm, are you referring to `code/drasil-printers/lib/Language/Drasil/Debug/Print.hs`? That file is just a formatted replacement for the (existing) `code/drasil-printers/lib/Language/Drasil/Log/Print.hs` file, which I thought deserved a better folder name. However, if you are, that whole file will likely be replaced entirely with #2873 with a less-duplicated variant.

If you're not, then the only other duplication issue I can see is the `app/Main.hs` files which could be rolled up into a single `smithEtAlProg` 'program' that processes a `SystemInformation`+`PrintingInformation` with the the typical SRS+optional code generation.

JacquesCarette (approved):

JacquesCarette:
There are some conflicts to resolve @balacij

balacij:
Thanks! Resolved them now.

PR #3664, opened 2023-09-16
Shellcheck scripts in Lint workflow

### LINKED ISSUE #2936: Follow `shellcheck` recommendations on our `sh` scripts
[`shellcheck`](https://www.shellcheck.net/) is a nice tool we can use to audit our `scripts/*.sh` scripts. It is very similar to HLint, in that it audits the scripts and suggests improvements.

balacij:
Leaving this PR as a draft until we've checked the shell scripts and fixed any issues, or else our CI will constantly fail due to CI issues.

Closes #2936


JacquesCarette:
It would be nice to get this resolved one way or another?

balacij:
Yeah, it has been sitting here for a while...

I don't think any of the messages were bad -- @BilalM04 Would you be interested in this, since you're going to be working with the scripts soon anyway?

BilalM04:
@balacij yes I can take a look at it!

BilalM04:
#### Commit https://github.com/JacquesCarette/Drasil/pull/3664/commits/422e81e1a2ab39e88eb99d0f7cea33eb70ad8729

- Changed `dorny/paths-filter` to latest version.
- Existing process to run ShellCheck didn't work, changed  `scandir` value to `'./code/scripts'`.
- Hlint will only be installed, set-up, and run if there is a change to a `.hs` or `.lhs` file.  

#### Commit https://github.com/JacquesCarette/Drasil/pull/3664/commits/b6c1f06d99442ac8409f772fc59ed402d22130d2

- Implemented all the suggestions from ShellCheck.
- Suppressed two instances of the same warning: [here](https://github.com/JacquesCarette/Drasil/pull/3664/commits/b6c1f06d99442ac8409f772fc59ed402d22130d2#diff-ca9625c5433000422b98ce4718d298763a0bea13dc445c6b55007d4fd94edddeR21) and [here](https://github.com/JacquesCarette/Drasil/pull/3664/commits/b6c1f06d99442ac8409f772fc59ed402d22130d2#diff-3d953ffbcd3a46fab24264df5061ffaf9e5280dfcb6d99f660b5ae381222a3a9R25).
  - ShellCheck recommends to put quotes around the `$TARGET`; however, the script breaks as it is a makefile target.

#### Commit https://github.com/JacquesCarette/Drasil/pull/3664/commits/ef555e809eaf5bd71e86a4b0649f74e037448ee7

- Added `make shellcheck` target to the makefile to run ShellCheck locally.
- Added ShellCheck installation instructions to the Wiki.

code/scripts/code_build.sh
```diff
@@ -15,29 +15,30 @@ fi
 
 build_code () {
   if [ -d "$EXAMPLE_CODE_SUBFOLDER" ]; then
-    cd "$EXAMPLE_CODE_SUBFOLDER"
+    cd "$EXAMPLE_CODE_SUBFOLDER" || exit
```
balacij, line 18: `exit`s that we can ignore are okay to have a 0 return (i.e., 'success' code, but I _think_ we'll want the majority of these to return a 1.

code/scripts/tex_build.sh
```diff
@@ -24,17 +24,17 @@ fi
...
 if [ "$SUMMARIZE_TEX" = "yes" ]; then
-  echo "\n\n\033[0;33m$EDIR TeX Summary\033[0m:"
+  printf "\n\n\033[0;33m%s TeX Summary\033[0m:" "$EDIR"
   if [ "$RET" -eq 0 ]; then
     # Approximate error gathering from TeX logs.
-    cat "$EDIR$GEN_NAME_SUFFIX".log | grep -E "erfull|Warning"
-    cat "$EDIR$GEN_NAME_SUFFIX".blg | grep -B3 -E "Error"
-    BIBERRS=$(cat "$EDIR$GEN_NAME_SUFFIX".blg | grep -E "Error" | wc -l)
+    grep -E "erfull|Warning" < "$EDIR$GEN_NAME_SUFFIX".log
+    grep -B3 -E "Error" < "$EDIR$GEN_NAME_SUFFIX".blg
```
balacij, line 36: I think we can also remove the `<` on these lines.

code/scripts/update_default_stackage.sh
```diff
@@ -7,15 +7,17 @@ update_files() {
...
     done
 }
 
 fetch_hlint_version() {
-    local redirect_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://www.stackage.org/lts-$1/package/hlint")
-    local hlint_version=$(echo "$redirect_url" | grep -oE "hlint-[0-9]+\.[0-9]+\.[0-9]+" | sed 's/hlint-//')
+    local redirect_url
+    redirect_url=$(curl -Ls -o /dev/null -w "%{url_effective}" "https://www.stackage.org/lts-$1/package/hlint")
+    local hlint_version
+    hlint_version=$(echo "$redirect_url" | grep -oE "hlint-[0-9]+\.[0-9]+\.[0-9]+" | sed 's/hlint-//')
```
balacij, line 20: I'm a little bit confused by these changes. What was shellcheck complaining about here?

Multi-line is less readable, and the quotation around a literal string with no special characters is unnecessary.

wiki/New-Workspace-Setup.md
```diff
@@ -496,5 +497,52 @@ The [Doxygen download page](https://www.doxygen.nl/download.html) has a disk ima
...
+```
+brew install shellcheck
+```
+Or, you can use [MacPorts](https://www.macports.org/):
+```
+sudo port install shellcheck
+```
+
+</details>
+
```
balacij, line 546: Good idea!

balacij (commented):
Thank you for updating the scripts @BilalM04 !

code/scripts/tex_build.sh
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3664/commits/b1f11dde00fc5e9e77c04442f4b29ce1395404d4.

code/scripts/code_build.sh
BilalM04: Yes, you're right, these should return 1. Fixed in https://github.com/JacquesCarette/Drasil/pull/3664/commits/b1f11dde00fc5e9e77c04442f4b29ce1395404d4.

code/scripts/update_default_stackage.sh
BilalM04: It was made multi-line because of [SC2155](https://www.shellcheck.net/wiki/SC2155), which says to declare and assign separately.

The quotations were added in because of [SC1083](https://www.shellcheck.net/wiki/SC1083).

balacij: If we wanted to handle SC2155, we can -- we would just want to add error code checking, or else the change wouldn't make any effective difference. Regarding SC1083, I see -- I can get behind that viewpoint. I write my scripts by writing out the commands myself and then copy-pasting, and editing, so any script I would write is biased towards that. Let's keep the change from SC1083, but unless we're going to change anything in regard to error checking with SC2155, we can ignore it, or, I suppose, at least slightly tweak how it's shown, i.e.,

```bash
local X
local y

x=$(...)
y=$(...)
```

code/scripts/code_build.sh
balacij: Nice!

code/scripts/update_default_stackage.sh
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3664/commits/fc3246a49571ae47c5876e7e72952adfe43335e8.

balacij: Thanks!

balacij:
Note: I can't review my own PR, but I agree with all of @BilalM04's changes. So, I think this one is ready for @JacquesCarette for final review.

Thank you again @BilalM04 :slightly_smiling_face:

JacquesCarette (approved):

PR #3550, opened 2023-07-20
3545 change abrv to program name

harmanpreet-sagar:
Closes #3545. No changes made to stable. Changes mad to `drasil-website`, `drasil-example` (Projectile),  and `drasil-gen`,

Issue #3545, opened 2023-07-19
Change `abrv` to `programName` for usages that do not cause a stable change


harmanpreet-sagar:
Based on the findings of #3472, `abrv` needs to be changed to `programName` in areas that do not cause any stable changes. 

Background: Most uses of `abrv` come along with `sys`, which essentially holds the name of the examples. There is no difference between `programName` and `abrv` in all examples other than `Game Physics` and `PD Controller`. In the cases of these two examples, the only difference that is created based on the usage of `programName` instead of `abrv` would be the addition of an underscore `(_)` between the two words in its name. More information on this investigation can be found in this Excel sheet [here ](https://mcmasteru365-my.sharepoint.com/:x:/g/personal/sagarh_mcmaster_ca/EYuyx9tXvwhBiRhrO-Ft3UIBOTqIKkipPtl3HGFTG-OOZg?e=i8mQco) or in this comment [here](https://github.com/JacquesCarette/Drasil/issues/3472#issuecomment-1638883175).

harmanpreet-sagar:
@smiths @JacquesCarette Should I have the underscore here or not as in this mention, the name of the example is part of an ongoing sentence.

![image](https://github.com/JacquesCarette/Drasil/assets/110783820/ad34a1a9-f674-49ac-81c0-5d52f3a4b763)

JacquesCarette:
The issue of `_` is completely different and should be discussed in its proper context rather than here. (Which is probably why @smiths has a hard time answering here.)

harmanpreet-sagar:
@smiths I believe it is referring to the Game Physics Library as a program. I can create a new issue to explore this further.

smiths:
@harmanpreet-sagar if the intention is to define what the Game_Physics library does, then the sentence should be reworded.  Something like "Game_Physics is a library for modelling and simulating 2D rigid body physics for use in video games."  In this case, we need the underscore, since the library really has to have a name that is different from the phrase "game physics".  People will have trouble reading the sentence if it is read as the concept of game physics, rather than the library called game physics.  I think we really should come up with a better name for this library.  I don't want to cause additional headaches, so that's something we can talk about in the future.  :smile:

smiths:
Game_Physics is really one of our more neglected examples.  It really requires a thorough review, but I've been waiting until we get our refined theories version in order.  It also would help to find someone that is enthusiastic about physics, since there is some neat physics even in 2D rigid body mechanics.  :smile:

Issue #3557, opened 2023-07-21
Investigate simplifying/rewriting `Makefile`


balacij:
The `Makefile` has some issues, for example:
1. the Stack version check if superfluous since Stack already does this,
2. it constantly rebuilds artifacts for various build targets (e.g., see #3548),
3. the target names are confusing (e.g., analysis, convertAnalyzed, graphs, tracegraphs, deploy_code_path), some can be folded into each other via environment variables (e.g., hlint vs hot_hlint, debug vs *, deploy vs deploy_lite) to simplify the scripts, and some just aren't helpful (e.g., install),
4. the file is difficult to read and maintain,
5. the targets with `X` parameters aren't extensible,
6. it has an incomplete system integrity check script (i.e., checking system dependencies), and
7. the targets are very finicky at times

<details>
<summary>Current output of `make help`</summary>

```
λ ~/Programming/Drasil/code/ logNone make help
HLint:
  hlint               Run HLint through the drasil packages. Uses a local HLint installation.
  hot_hlint           Run HLint through the drasil packages. Uses the latest HLint version, downloading the binary each time.
                      
Documentation:
  docs                Create Haddock documentation of all Drasil modules. Set FULL=1 if you want to additionally test Haddock generation for website deployment.
                      
GOOL:
  gool                Generate code from examples and test each one.
  doxygen             Generate doxygen documentation for all examples.
  deploy_code_path    Find all code file paths for all examples.
                      
Checks:
  check_stack         Check that we are using the right stack version.
  check_dot           Download the most recent version of graphviz.
  graphmod            Check that we can generate dot graphs.
                      
Help:
  help                Show this help.
  help_examples       Lists all examples.
  help_packages       Lists all packages.
                      
Deploy:
  website             First builds all Drasil packages, and then executes the drasil-website.
  deploy_lite         Deploy the Drasil website without regenerating artifacts. For local development.
  deploy              Generates all artifacts and the Drasil website locally.
                      
Analysis:
  graphs              Generate all module dependency graphs.
  analysis            Generate a table and some graphs to analyze Drasil's class, datatype, and instance structures.
  convertAnalyzed     Convert analyzed dot graphs into SVGs.
                      
General:
  pr_ready            Check if your current work is ready to for a PR via `all` and `hot_hlint`.
                      
Examples:
  all                 Run examples and test against stable.
  install             Install all example project binaries into your local binary path (see $(stack path) for local-bin-path).
  debug               Run test target with better debugging tools.
[...]

PR #3665, opened 2023-09-17
Reconcile _some_ conflicting/implicit knowledge about 'program' names during generation of various artifacts

### LINKED ISSUE #3565: PD Controller example is labelled as "PD_Controller" instead of "PDController" on website
I noticed this before but waited in case #3521 fixed this. The PD Controller example is called exactly that, space included:

https://github.com/JacquesCarette/Drasil/blob/8d5e8f8088bb87cd351d07f4104faae7fd49111f/code/drasil-example/pdcontroller/lib/Drasil/PDController/Concepts.hs#L17

However, it seems that on the website, this space is replaced with an underscore instead of just being omitted:
image
A consequence of this is that the links generated are incorrect:
image
Removing these underscores from the URL leads to the correct page:
image

Since this is the only example name with a space, I think there is likely a function that gets called at some point that replaces spaces with underscores, when they should just be omitted (at least in the use case of the website).

balacij:
Closes #3565
Builds on #3667

Instead of writing a description, I will annotate the code through a review. This PR is largely a follow-up to PR #3550 / Issue #3545.


code/Makefile
```diff
@@ -179,9 +178,9 @@ BATCHTERM=dumb
 #--- Actual targets ---#
 ########################
 
-# Default `make` will just run examples and test them against stable.
-# See `test` target for more details.
-all: test ##@Examples Run examples and test against stable.
+# Default `make` will just run examples and test them against stable. See `test`
+# target for more details.
+all: test test_website ##@Examples Run examples and test against stable.
```
balacij, line 183: `test_website` is a good target to add here because it makes `make all` and `make pr_ready` test the website against the stabilized website artifacts as well.

code/Makefile
```diff
@@ -239,7 +238,7 @@ test: $(TEST_EXAMPLES) ##@Examples Run all examples and compare against the stab
 
 # First builds the website and then tests the generated contenst with those in
 # the stable-website folder.
-test_website: $(TEST_WEBSITE)
+test_website: $(TEST_WEBSITE) ##@Website Test freshly generated website against last known 'stable' copy.
```
balacij, line 241: Adding a comment about the `test_website` target, for `make help`.

code/Makefile
```diff
@@ -254,7 +253,7 @@ stabilize: $(STABILIZE_EXAMPLES) ##@Examples Overwrites the stable folder with u
 
 # First build the website, and then overwrite the contents in the stable-website
 # folder with currently generated ones.
-website_stabilize: $(STABILIZE_WEBSITE)
+website_stabilize: $(STABILIZE_WEBSITE) ##@Website Stabilize a freshly generated copy of the website artifacts.
```
balacij, line 256: Again, adding a comment for another target, for `make help`.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -179,7 +178,7 @@ fillReferences dd si@SI{_sys = sys} = si2
     -- set new reference table in the chunk database
     chkdb2 = set refTable (idMap $ nub $ refsFromSRS ++ inRefs
       ++ map (ref.makeTabRef'.getTraceConfigUID) (traceMatStandard si) ++ secRefs -- secRefs can be removed once #946 is complete
-      ++ traceyGraphGetRefs (filter (not.isSpace) $ abrv sys) ++ map ref cites
+      ++ traceyGraphGetRefs (programName sys) ++ map ref cites
```
balacij, line 181: The first of many similar changes. `filter (not.isSpace)` is often applied to the names of our various 'programs' (case study abbreviations) to remove spaces from the name. The goal of the filter was to make the program names friendly to code and file-systems. So, instead, I made it default to the calculated program names, which now remove spaces instead of replacing them with underscores.

code/drasil-docLang/lib/Drasil/DocumentLanguage/TraceabilityGraph.hs
```diff
@@ -187,10 +187,10 @@ traceyGraphPath :: String -> String -> String
 
 traceGFiles = ["avsa", "avsall", "refvsref", "allvsr", "allvsall"]
 traceGUIDs = map mkUid ["TraceGraphAvsA", "TraceGraphAvsAll", "TraceGraphRefvsRef", "TraceGraphAllvsR", "TraceGraphAllvsAll"]
-traceyGraphPaths ex = map (\x -> resourcePath ++ map toLower (filter (not.isSpace) ex) ++ "/" ++ x ++ ".svg") traceGFiles
-traceyGraphGetRefs ex = map makeFigRef' traceGUIDs ++ zipWith (\x y -> Reference (x +++. "Link") (URI y) (shortname' $ S $ show x)) traceGUIDs (traceyGraphPaths $ map toLower $ filter (not.isSpace) ex)
+traceyGraphPaths ex = map (\x -> resourcePath ++ map toLower ex ++ "/" ++ x ++ ".svg") traceGFiles
+traceyGraphGetRefs ex = map makeFigRef' traceGUIDs ++ zipWith (\x y -> Reference (x +++. "Link") (URI y) (shortname' $ S $ show x)) traceGUIDs (traceyGraphPaths $ map toLower ex)
 -- for actual use in creating the graph figures
-traceyGraphPath ex f = resourcePath ++ map toLower (filter (not.isSpace) ex) ++ "/" ++ f ++ ".svg"
```
balacij, line 193: Note: `filter (not.isSpace)` on abbreviations is only 1 issue of impromptu formatting of case study names. The `map toLower` is another issue, which occurs as often.

code/drasil-website/lib/Drasil/Website/CaseStudy.hs
```diff
@@ -59,7 +59,7 @@ data CaseStudy = CS {
 -- so we take the naming scheme from there.
 mkCaseStudy :: Example -> [CaseStudy]
 mkCaseStudy E{choicesE = []} = []
-mkCaseStudy E{sysInfoE = si@SI{_sys = sys}, choicesE = [x]} = [CS{sysInfoCS = si, progName = S $ programName sys, choicesCS = x}]
+mkCaseStudy E{sysInfoE = si@SI{_sys = sys}, choicesE = [x]} = [CS{sysInfoCS = si, progName = S $ abrv sys, choicesCS = x}]
```
balacij, line 62: This was one of the main issues with the website display names. 

Previously, `programName` replaced spaces with underscores. Now it removes them altogether. Hence, keeping `programName` (as done in @3550) here wouldn't work. So, since we're already displaying the human-friendly abbreviation for the case study, we should be free to include our spaces, so we revert to using `abrv`.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -79,16 +79,16 @@ fullExList codePth srsDoxPth = Enumeration $ Bullet $ map (, Nothing) (allExampl
...
 -- | Display the points for generated documents and call 'versionList' to display the code.
 individualExList :: Example -> [ItemType]
 -- No choices mean no generated code, so we do not need to display generated code and thus do not call versionList.
 individualExList E{sysInfoE = SI{_sys = sys}, choicesE = [], codePath = srsP} = 
-  [Flat $ S (programName sys ++ "_SRS") +:+ namedRef (getSRSRef srsP "html" $ programName sys) (S "[HTML]") +:+ namedRef (getSRSRef srsP "pdf" $ programName sys) (S "[PDF]")]
+  [Flat $ S "SRS:" +:+ namedRef (getSRSRef srsP "html" $ programName sys) (S "[HTML]") +:+ namedRef (getSRSRef srsP "pdf" $ programName sys) (S "[PDF]")]
 -- Anything else means we need to display program information, so use versionList.
 individualExList ex@E{sysInfoE = SI{_sys = sys}, codePath = srsP} = 
-  [Flat $ S (programName sys ++ "_SRS") +:+ namedRef (getSRSRef srsP "html" $ programName sys) (S "[HTML]") +:+ namedRef (getSRSRef srsP "pdf" $ programName sys) (S "[PDF]"),
+  [Flat $ S "SRS:" +:+ namedRef (getSRSRef srsP "html" $ programName sys) (S "[HTML]") +:+ namedRef (getSRSRef srsP "pdf" $ programName sys) (S "[PDF]"),
```
balacij, line 91: I was going to replace this with an `abrv sys ++ "SRS"`, resulting in `PD Controller_SRS [HTML] [PDF]` on the website, but I found it odd to keep the `_`, so I removed that. Afterwards, I found the name repetition too frequent in such a small area, so I just opted to remove the abbreviation altogether.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -208,15 +208,18 @@ getSRSRef path sufx ex = makeURI refUID (getSRSPath path (map toLower sufx) ex)
 -- | Get the paths of where each reference exist for SRS files. Some example abbreviations have spaces,
 -- so we just filter those out. The suffix should only be either html or pdf.
 getSRSPath :: FilePath -> String -> String -> FilePath
-getSRSPath path sufx ex = path ++ map toLower (filter (not.isSpace) ex)
-  ++ "/SRS/srs/" ++ filter (not.isSpace) ex ++ "_SRS." ++ map toLower sufx
+getSRSPath path sufx ex = 
+  path
+  ++ map toLower ex -- FIXME: The majority of these `map toLower`s are implicit knowledge!!!
```
balacij, line 213: The `map toLower`s happen quite often!

code/stable-website/index.html
```diff
@@ -236,24 +236,24 @@ <h1>Generated Examples</h1>
...
               </li>
               <li>
                 Generated Code:
                 <ul class="list">
                   <li>
-                    PD_Controller <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pd_controller/src/python">[Python]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pd_controller/src/csharp">[C Sharp]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pd_controller/src/java">[Java]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pd_controller/src/cpp">[C++]</a>
+                    PDController <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pdcontroller/src/python">[Python]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pdcontroller/src/csharp">[C Sharp]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pdcontroller/src/java">[Java]</a> <a href="https://github.com/JacquesCarette/Drasil/tree/master/code/stable/pdcontroller/src/cpp">[C++]</a>
```
balacij, line 248: Here is the resultant 'actual' fix. Note that the name shown here doesn't align with the name shown on line 239!

Why? Well, the top-level `<li>` nodes show the main abbreviation, but this is the name of the final 'executable program/project/case study, taking into consideration the choices'. Following in line with how the other projects (including the multi-src Projectile case study in mind), we use the name of the 'executable program', left uppercase. 

If we wanted to make this name lowercase, we can, similar to how this name is done everywhere else. If we wanted to avoid putting the name of the case study here again, we can (I would prefer this, but I thought I was pushing the boundary of this PR too much now). If we want to include the space, we can. Ultimately, I didn't quite know what name we wanted here, but finally, it notes the 'real' issue here:

We have impromptu naming and calculation of things everywhere. This PR is specifically about the name of `PD Controller` and how we propagate it, but us having the freedom of Haskell is a bit of a double-edged sword here, whereby we can have impromptu temporary code placed anywhere that breaks uniformity and actual knowledge captured for a specific 'fix'.

code/stable/pdcontroller/src/cpp/Makefile
```diff
@@ -12,18 +12,18 @@ endif
 
 # Generated by Drasil v0.1-alpha
 
-# Project Name: PD_Controller
+# Project Name: PDController
```
balacij, line 15: Again, similar to the above comment, using the same uniform naming scheme that we use in URIs, etc.

balacij:
Ultimately, the solution isn't complex, but it does rely on the abbreviations of our case studies and how we want to translate them to 'code' artifacts.

code/drasil-website/lib/Drasil/Website/CaseStudy.hs
balacij: I reverted this one to keep continuity with the other names in the case studies table containing the final case studies project folder name, but the blurb of discussion is still good.

code/drasil-website/lib/Drasil/Website/Example.hs
samm82: Do these show up as a redundancy to enforce them being lowercase, or is there something that is being made lowercase in multiple places (instead of just being made lowercase at its definition)?

balacij: Ah, that's a good catch. I completely forgot to address why they appear.

It does not appear to be a redundancy to enforce them being lowercase. I don't really think we should _need_ nearly any sort of redundancy, since we generate through staging.

These `map toLower`s applied to our case study names is to relate to the things that were named after our project folder names in the `code/drasil-example` folder -- they are all lowercase. We should probably figure out a better way of handling external resources internally, preferrably reading them in externally and having an external dictionary that's passed to/compiled into Drasil.

samm82: I feel like this could be related to how we've been handling paths to Projectile-related files (since it uses design variabilities that introduce another directory level) and figuring out how to handle this in the long term 🤔

balacij: Yep, it definitely is. We might need to do another pass at cleaning this up.

balacij: Also related to #3557

code/drasil-gen/lib/Language/Drasil/Dump.hs
```diff
@@ -16,22 +16,25 @@ import Database.Drasil (traceTable, refbyTable)
...
+
 type Path = String
 type TargetFile = String
 
 -- | For debugging purposes, if the system has a `DEBUG_ENV` environment
 --   variable set to anything, we can dump the chunk maps in a system to the
 --   host system.
-dumpEverything :: SystemInformation -> Path -> IO ()
-dumpEverything si p = do
+dumpEverything :: SystemInformation -> PrintingInformation -> Path -> IO ()
```
JacquesCarette, line 28: Why is this PR dependent on the one that deals with dumping debug information? These two should be separate. Especially as I've got qualms about that one, which will now delay this one!

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -79,16 +79,16 @@ fullExList codePth srsDoxPth = Enumeration $ Bullet $ map (, Nothing) (allExampl
 allExampleList :: [Example] -> [ItemType]
 allExampleList = map (\x -> Nested (nameAndDesc x) $ Bullet $ map (, Nothing) (individualExList x))
   where
-    nameAndDesc E{sysInfoE = SI{_sys = sys, _purpose = purp}} = S (programName sys) +:+ S " - To" +:+. head purp
+    nameAndDesc E{sysInfoE = SI{_sys = sys, _purpose = purp}} = S (abrv sys) +:+ S " - To" +:+. head purp
```
JacquesCarette, line 82: Isn't this a regression? We want to use `abrv` less, and `programName` was a step in that direction.

JacquesCarette (changes requested):
Please disentangle these two PRs.

code/drasil-gen/lib/Language/Drasil/Dump.hs
balacij: Rebased!

code/drasil-website/lib/Drasil/Website/Example.hs
balacij: Isn't this reverting a regression instead? `abrv` carries the actual abbreviation of the `PD Controller` project. `programName` calculates what a 'valid program name' would be given the abbreviation (replacing a space with an underscore).

JacquesCarette: Ok, for this one,  I think you're right.

JacquesCarette (approved):

code/drasil-website/lib/Drasil/Website/Example.hs
samm82: One thing I just noticed: should this line (and any other related lines) be using our abbreviations for "SRS", "HTML", etc.?

balacij: Aha! They definitely shouldn't be using the raw `String` versions, so I think that's a great big "yes" to reduce duplication :smile:

JacquesCarette (approved):

PR #3670, opened 2023-09-18
Remove unnecessary deployment scripts for the website and clean up (a bit)

### LINKED ISSUE #2834: `scripts/deploy_wrapper.bash` & `scripts/deploy.sh` can be merged and cleaned up a little bit
These scripts were repurposed to just prepare the deployable folder for GitHub Actions, they no longer do the actual deployment themselves. As such, we should rename them. However, `deploy_wrapper.bash` seems redundant, I think we can remove it entirely/merge it into `deploy.sh`.

balacij:
Closes #2834
Contributes to #3557

Leaving this as a draft until I've built the website locally (it takes a while because of the TeX artifacts) and made sure things work as expected, at least locally.

EDIT: It looks good, but I want to double-check again. I'm testing on my fork with the same actions script that would run here.


JacquesCarette (commented):
I'll have to trust you on this - I see nothing wrong, but I'm not sure I would catch anything subtle.

balacij:
After some toying around with my fork's settings (which apparently changed without me changing them), my forked [website](https://balacij.github.io/Drasil/) is working, so I think we should be fine to merge.

JacquesCarette (approved):

Issue #3660, opened 2023-09-13
Doxygen warns about “bad” comments: `\Generated`


balacij:
#3566 introduced small “Generated by Drasil...” watermarks to each file's Doxygen comments. The watermarks appear as Doxygen annotations `\Generated by Drasil ...`. Unfortunately, `\Generated` is not a [valid annotation](https://www.doxygen.nl/manual/commands.html#cmdnote) for Doxygen, and so `make doxygen` throws a bunch of extra warnings. I think we should change it to `\note Generated ...`.

smiths:
Yes, I agree with this proposed change.  Nice to have another newcomer issue identified too.  :smile:  Maybe @daijingz will be interested?

daijingz:
> Yes, I agree with this proposed change. Nice to have another newcomer issue identified too. 😄 Maybe @daijingz will be interested?

Let me work on it, because of lots of things, I just arrive today.

daijingz:
This issue I use a month not because I feel hard to do because I spend a lot of time reviewing Doxygen. Now you can see my pull request.

daijingz:
Write it there. At first, let me find how many project folders have `\Generated` problem

After using the command `grep "Generated" -r * --include "*.*"`, I found these 5 segmental projects need to be contributed:
`dblpend`, `glassbr`, `pdcontroller`, `projectile`, `swhsnopcm`

PR #3673, opened 2023-11-07
Doxygen warns about “bad” comments: \Generated #3660 (Update 1)

daijingz:
After I had a correct running result (both make and stack exec work well), I submitted my changes.
I am worried about something:
1. The last issue #3653 changes has not been saved last time, and I do not know why. This time, after detecting the differences, I have to submit both changes this time.
2. I do not know why these modified files did not show my changes, it only says "Some generated files are not rendered by default. Learn more about [how customized files appear on GitHub](https://docs.github.com/github/administering-a-repository/customizing-how-changed-files-appear-on-github)." I only change a few lines on these files, and they are well displayed at my local GitHub desktop. (I have to say, it is very strange)
3. I did not change any files in this address: code/SRS/HTML/. These files are just automatically generated.

@balacij I am not sure whether these confusions are good. I changed over 200 files this time, and I checked each file one by one.


daijingz:
I do not understand this, and I find it has not occur in my testing. What should I change here

balacij:
It looks like the new stable was accidentally unpacked into the `code/` directory? The CI is failing because the old stable is still in `code/stable/`. That needs to be updated.

balacij:
I was a bit surprised with the other PR being suddenly closed. If you're going to work on different things (for any project), you should use branches to split your work by goal. That way you don't need to worry about manipulating a single branch, because you can have different branches track different change sets.

daijingz:
> I was a bit surprised with the other PR being suddenly closed. If you're going to work on different things (for any project), you should use branches to split your work by goal. That way you don't need to worry about manipulating a single branch, because you can have different branches track different change sets.

I did not close the previous pull request, and the previous pull request can only be determined by all the reviewers. I do not know why this issue was closed. (I am pretty sure I did not do anything, but when I saw the request few days ago, I found it already closed)

balacij:
I think your previous PR was automatically closed because you had a force push on your `master` branch -- #3653. Anyway, I think you should split this PR in two. This PR contains code that resolves the "MayHaveDerivation" issue (good), but then I think something went awry, and you didn't actually commit the Haskell changes needed for prepending `\note` (or at least I don't see it in the diffs). I think you had that code completed, however, because you updated stable, but you posted unfolded the contents of stable into the `code/` folder directly instead of in the nested `stable/` folder, which is why this PR has a huge diff size.

PR #3566, opened 2023-07-25
3501 add Generated by Drasil Version try 2

harmanpreet-sagar:
Closes #3501. 2 new files were added in `drasil-metadata`. Changes made to `drasil-gool`, `drasil-metadata`, `drasil-gen`, `drasil-code`, and stable.

I have made this new PR because it would be a lot more work to merge `master` and change the previous PR. I also tried to incorporate all the feedback I received on the previous PR into this new PR. I tried using `template-haskell`, however, I was unsuccessful in determining how to use it correctly for this program.

PR #3677, opened 2023-11-13
#3660 Doxygen warns about “bad” comments: \Generated

daijingz:
The second update is now.
All other generated files have been removed, and the git command system has been learned. (Feel afraid of applying now)

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -91,7 +91,7 @@ jobs:
         run: make deploy_lite
       - name: "Deploy 🚀"
         if: ${{ fromJSON(env.is_deployment) }}
-        uses: JamesIves/github-pages-deploy-action@v4.4.3
+        uses: JamesIves/github-pages-deploy-action@v4.5.0
         with:
           git-config-name: ${{ secrets.BOT_NAME }}
           git-config-email: ${{ secrets.BOT_EMAIL }}


daijingz:
@JacquesCarette Dr.Carette, I do not understand why I have this error. I only change stable directory files, and the only changeable place is stable directory.

balacij:
@daijingz It looks like you forgot to post the Haskell code changes. So `stable/` does not match a fresh `build/`. You can check the CI for the exact step and diffs that cause the CI to fail.

daijingz:
Let me do it.

balacij:
Yes, exactly. We need the Haskell code to be able to generate a fresh `build/` folder that is identical to `stable/`. Here, it looks like you've updated `stable/` but didn't update the Haskell code.

daijingz:
This time I use the command make stabilize

JacquesCarette:
I don't understand the current state of this PR. It has a single file change, but it is in `stable/` yet it looks like it is passing? Very odd.

daijingz:
> I don't understand the current state of this PR. It has a single file change, but it is in `stable/` yet it looks like it is passing? Very odd.

Dr.Carette, I am sorry to say, most of updates has been ignored after running command `make stabilize`.

balacij:
@JacquesCarette In all honesty, I didn't even realize that GOOL was in stable at all... The diff tests (and by extension, the CI) do not test the `gooltest` suite against `stable/` right now. I'll make an issue.

@daijingz The related issue links to a PR that shows how those `\Generated...` notes are generated. Did you modify that related _Haskell_ code to now make it generate `\note Generated...`? If you did, then the reason why the CI was failing with all the diffs you had was because you didn't push the Haskell changes. If you didn't, then I believe there is a misunderstanding -- you should have modified the Haskell code to generate the desired `stable/` artifacts you had manually edited.

daijingz:
@JacquesCarette @balacij Here are all my original steps: (This time it is a simplified version because I am changing three files)

**Step 1** -- `grep "Generated" -r * --include "*.*"`
This command is looking for all file lines with the string "Generated". It does not change the file.

**Step 2** -- Change three files with the Doxygen (Below are results after using the command `git diff`)
MAXOY6A(BY(7 70 2((@9A
Command `git status` has a consistent result in this area.

**Step 3** -- `make` -> This step has a normal result

**Step 4** -- `make clean` This is because there are a lot of generated files that we should not include.
(Below are results after using the command `make clean` and `git diff`) -> I think it is the same as **Step 2**
~6AY%~V)WA(44~)V$_MWHED

**Step 5** -- `make stabilize`
(Below are results after using the command `make stabilize` and both mentioned commands)
2TX}NRA36CXT$P 6`BOK}PB

**IMPORTANT** -> I tried to swap step 4 and step 5, however, the final result is still the same.

daijingz:
> @JacquesCarette In all honesty, I didn't even realize that GOOL was in stable at all... The diff tests (and by extension, the CI) do not test the `gooltest` suite against `stable/` right now. I'll make an issue.
> 
> @daijingz The related issue links to a PR that shows how those `\Generated...` notes are generated. Did you modify that related _Haskell_ code to now make it generate `\note Generated...`? If you did, then the reason why the CI was failing with all the diffs you had was because you didn't push the Haskell changes. If you didn't, then I believe there is a misunderstanding -- you should have modified the Haskell code to generate the desired `stable/` artifacts you had manually edited.

@balacij Jason, I did not. I do not know which part of the code generates this comment. I think your thinking is consistent with my suspicions. 

Then the problem occurs: where are these files? I know we should only change the `stable/` folder, but now I scanned through all the files in it, and I did not find any Haskell code. Can you give me some hints about the corresponding Haskell code address?

balacij:
Gotcha, thank you for explaining that @daijingz. The code that generates `build/` and `stable/` isn't actually in those folders at all. Rather, it's the projects in the same directory, for example `code/drasil-examples/projectile` is responsible for building the code that goes into `code/build/projectile/` and if permitted, `code/stable/projectile`.

In the issue, I linked to the original PR that generates the `\Generated...` comments.

See:

https://github.com/JacquesCarette/Drasil/issues/3660
> #3566 introduced small “Generated by Drasil...” watermarks to each file's Doxygen comments. The watermarks appear as Doxygen annotations `\Generated by Drasil ...`. Unfortunately, `\Generated` is not a [valid annotation](https://www.doxygen.nl/manual/commands.html#cmdnote) for Doxygen, and so `make doxygen` throws a bunch of extra warnings. I think we should change it to `\note Generated ...`.

Then open the referenced #3566. The diff in that PR is quite large, so it would be good to open the [files tab](https://github.com/JacquesCarette/Drasil/pull/3566/files) and search for ".hs" so that you only see the Haskell files that were edited. You can study those files to understad what needs to be changed such that we generate the appropriately changed artifacts :)

daijingz:
> Gotcha, thank you for explaining that @daijingz. The code that generates `build/` and `stable/` isn't actually in those folders at all. Rather, it's the projects in the same directory, for example `code/drasil-examples/projectile` is responsible for building the code that goes into `code/build/projectile/` and if permitted, `code/stable/projectile`.
> 
> In the issue, I linked to the original PR that generates the `\Generated...` comments.
> 
> See:
> 
> #3660
> 
> > #3566 introduced small “Generated by Drasil...” watermarks to each file's Doxygen comments. The watermarks appear as Doxygen annotations `\Generated by Drasil ...`. Unfortunately, `\Generated` is not a [valid annotation](https://www.doxygen.nl/manual/commands.html#cmdnote) for Doxygen, and so `make doxygen` throws a bunch of extra warnings. I think we should change it to `\note Generated ...`.
> 
> Then open the referenced #3566. The diff in that PR is quite large, so it would be good to open the [files tab](https://github.com/JacquesCarette/Drasil/pull/3566/files) and search for ".hs" so that you only see the Haskell files that were edited. You can study those files to understad what needs to be changed such that we generate the appropriately changed artifacts :)

I have a confusion -> In this example, there is no file in the /detail-example/ folder.

balacij:
There are plenty of files in [that folder](https://github.com/JacquesCarette/Drasil/tree/master/code/drasil-example/projectile). The `app` folder contains the code that generates the content in the `build/projectile` folder (on-demand) and that we copy over to the `stable/projectile` folder.

The tools it uses to do that generation are shared across the other case studies, and by improving those tools, we can improve the quality of the produced artifacts. For example, the PR I linked (on generating the `\Generated` messages) in the original issue is highly relevant to the work you're trying to do. It would be good to study that PR to figure out how it generated the comments and how to update what it outputs.

PR #3687, opened 2024-01-22
Use Swift Markup language for writing watermarks, remove 'note' from watermark text and Makefile

### LINKED ISSUE #3660: Doxygen warns about “bad” comments: `\Generated`

balacij:
Closes #3660 
Follow-up to #3681

Swift's documentation [Markup](https://swiftrocks.com/swifts-documentation-markup) is neat.


smiths (commented):
I looked at the changes to `stable` and everything looks good to me.  I also spot-checked in the new version of stable that `make doc` does not generate the warning about `\Generated`.

JacquesCarette (approved):

PR #3695, opened 2024-02-21
Switch `compsy` usage to an instance of `Ord` for `Symbol`

balacij:
Removes an export, re-using `Ord` instead.

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -21,7 +21,6 @@ import Drasil.ExtractDocDesc (getDocDesc, egetDocDesc)
 import Drasil.TraceTable (generateTraceMap)
 
 import Language.Drasil hiding (kind)
-import Language.Drasil.Display (compsy)
 
 import Database.Drasil hiding (cdb)
 import SysInfo.Drasil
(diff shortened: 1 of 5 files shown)


JacquesCarette (commented):
So I am really unsure about this change. Code-wise, it's 100% fine. It's design-wise that I am not sure.

More precisely: long ago, there was an `Ord` instead for `Symbol` and it was removed. The point is that it was used implicitly in places (as instances are wont to be, that's part of their design) where we really were not sure we actually wanted that. So we remove the `Ord`instance to make it clear, now and in the future, where we dependent on symbols being ordered.

So while this makes sense code-wise, I'm not sure we should do it?

balacij:
Hmm, I wasn't aware of that. The commit removing it is 603afc693d2b4e845de19badf715aea019b41529. I guess the relevant parts are:

image

and

image

Is the issue that we're ignoring `Equational` vs `Implementation` `Stage` information?

JacquesCarette:
The point is that we *don't* want to use symbols to 'order' things except for purely display-oriented things (like tables of symbols).

An `Ord` instance makes it super tempting to use it for other purposes. And because it is global and 'silent', it becomes really hard to see that that is what is going on.

So I am not convinced that this is a bad change.

PS: often a good idea to use the Discord to discuss if a design change like this is worth doing. I'm not sure where exactly to record this design decision so that no one else in the future attempts to do the same (obvious!) refactor.

JacquesCarette:
Note: I didn't delete the branch as there might be some comment edits worth keeping.

balacij:
If the comment changes are worth keeping, I can make a PR porting them, and writing down the design decision.

Sounds good, will switch there then.

JacquesCarette:
Comment changes made sense to me, may as well not lose that work.

Issue #3700, opened 2024-03-13
Add more static analysis to our Haskell tool chain?


JacquesCarette:
[Stan](https://github.com/kowainik/stan/) is another tool, which does things differently than hlint. Maybe we should use it too?

This issue is basically for us to come up with a yes/no decision; if yes, then this should become a (small) project instead of a long-lived issue.

balacij:
`weeder` is another one that might be useful too.

Are there any conflicts between Stan and hlint?

In either case, I'm a vote for 'yes.'

smiths:
I looked at a sample [report](https://kowainik.github.io/projects/stan/report) that Stan generated for another project.  I can't see any reason, other than the time it will take to do it, to not be interested in a report on Drasil.  We might not need to integrate Stan into our CI process.  Maybe it is just something we do on an ad hoc basis?  The decision will depend on how useful the output is from Stan.  

I looked up [weeder](https://hackage.haskell.org/package/weeder) as well.  The purpose of weeder is to detect dead code.  I'd certainly be interested in knowing if we have dead code.  :smile:  Again, I don't know if we need it as part of our CI process.  I think we should run it on our code base first to see how useful the output is.

balacij:
I used `stan` on `drasil-lang` and was very pleased with the feedback. It output a clear report, providing concrete solutions.

Notes:
* I upgraded the project GHC version to 9.4.8 and used Stan v0.1.1.0. The GHC version change is to include a bugfix for my system, completely irrelevant for everyone else, but will show up in the report.
* `stan` and `weeder` work using [HIE](https://www.haskell.org/ghc/blog/20190626-HIEFiles.html) files, which is great, because they will rely on information from GHC directly.
* `stan` does not appear to enjoy lazy data, so I enabled the `StrictData` language extension on the project, otherwise there is a ton more warnings (at least 1 for each data definition, but can also be an extra 1 for each argument in each constructor!). I don't have any issue with this. @JacquesCarette any thoughts on this one in particular?

Unfortunately, GitHub does not allow uploading HTML files in comments. So, here is [stan.txt](https://github.com/JacquesCarette/Drasil/files/14643081/stan.txt), the same HTML-encoded file except in a .txt file extension. Renaming is all you'll need to do :smile:  

Highlights:
* `foldl` use should be avoided for `foldl'` (agreed)
* finds inefficiencies (`nub`)
* warns about partial functions (`head`, `tail`)
* warns about infix operators defined without a fixity
[...]

PR #3708, opened 2024-04-02
Write hie info

balacij:
Contributes to #3700 by writing HIE information necessary for both `stan` and `weeder`

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -492,16 +492,17 @@ deploy: analysis graphs convertAnalyzed packagedeps docs tex doxygen ##@Deploy G
 
 # Removes all generated folders.
 $(filter $(CLEAN_GF_PREFIX)%, $(CLEAN_FOLDERS)): $(CLEAN_GF_PREFIX)%:
-	- rm -rf "./$*/"
+	rm -rf "./$*/"
 
 clean_artifacts: $(CLEAN_FOLDERS) ##@Cleaning Remove generated artifacts & folders. (alt.: cleanArtifacts)
-	- rm -f "$(CACHED_MSV_FILE)"
+	rm -f "./$(CACHED_MSV_FILE)"
 
 # TODO: HELP_FUN's pattern match ignores targets with aliases, this is a temporary exterior alias until we fix it.
 cleanArtifacts: clean_artifacts
 
 clean: clean_artifacts ##@Cleaning Fully clean all generated builds, artifacts, & folders.
-	- stack clean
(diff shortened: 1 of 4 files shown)


balacij:
Force-pushed to remove a commit I didn't intend to write to this branch...

code/Makefile
```diff
@@ -492,16 +492,17 @@ deploy: analysis graphs convertAnalyzed packagedeps docs tex doxygen ##@Deploy G
 
 # Removes all generated folders.
 $(filter $(CLEAN_GF_PREFIX)%, $(CLEAN_FOLDERS)): $(CLEAN_GF_PREFIX)%:
-	- rm -rf "./$*/"
+	rm -rf "./$*/"
```
JacquesCarette, line 495: why this change?

balacij: It's _almost completely_ redundant. Passing `-f` `rm` already suppresses the majority of errors, which is what `-` would do for that Makefile line. The errors that `rm -f` return all have to do with a system that we somehow were trying to edit, but which we were stopped by the OS for some reason (e.g., we accidentally tried to delete System32 or any other read-only file, the local working folders became read-only [permission loss], etc.). I think we would want to see those errors, because it would mean something has gone very seriously wrong.

balacij:
Yep, [`.hie` files](https://gitlab.haskell.org/ghc/ghc/-/wikis/hie-files) (and the directory) contain compiler-related information that was found while GHC was working with it. From what I've understood, its ultimate goal is to enable external tools to 'search/analyze/manipulate the AST' as if it were built into the compiler itself, so it has an easier time dealing with the AST //instead of needing to work with the CST. The [`hiedb` package](https://hackage.haskell.org/package/hiedb) allows us to write Haskell code that handles the `.hie` information (hence enumerating what information we can readily see from the GHC output), and it also has a nifty ['Query'](https://hackage.haskell.org/package/hiedb-0.6.0.0/docs/HieDb-Query.html) that we might be able to use as well! For example, we have those two scripts in the `scripts/` folder -- I forget what they do completely, but I know they scan the source code -- that could probably be simplified by relying on the `.hie` information instead of a string search. Maybe we make that a work task actually? I'm not sure how robust those scripts are.

balacij:
@JacquesCarette I see on #3739, you mentioned that you wanted more information about this -- did the above comment clarify everything?

JacquesCarette:
Upon re-reading: yes.

Issue #3244, opened 2023-01-24
Is it redundant to have "Chunk" in the suffix of a chunk type name?


balacij:
In #3017, we established that we will continue to use the name "Chunk" to refer to information Drasil works with. If we think of "chunks" as instances of our Haskell types (which is how we build them in Drasil), then we also have "chunk types" (or categories/classifications/etc). I believe we also have an assumption that all data types we build are chunks too. So, it feels redundant when we have chunks (such as those listed below) with "Chunk" in their type's name. Should we rename these chunks?

```console
rg "data [a-zA-Z]+Chunk "

drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs
40:data NamedChunk = NC {

drasil-lang/lib/Language/Drasil/Chunk/Unital.hs
30:data UnitalChunk = UC { _defq' :: DefinedQuantityDict

drasil-lang/lib/Language/Drasil/Chunk/UncertainQuantity.hs
33:data UncertainChunk  = UCh { _conc :: ConstrainedChunk , _unc' :: Uncertainty }

drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs
36:data ConstrainedChunk = ConstrainedChunk { _qd     :: QuantityDict

drasil-lang/lib/Language/Drasil/Chunk/Unitary.hs
29:data UnitaryChunk = UC { _quant :: QuantityDict

drasil-lang/lib/Language/Drasil/Chunk/Concept/Core.hs
32:data ConceptChunk = ConDict { _idea :: IdeaDict -- ^ Contains the idea of the concept.

drasil-code-base/lib/Language/Drasil/Chunk/CodeBase.hs
43:data CodeChunk = CodeC { _qc  :: QuantityDict
67:data CodeVarChunk = CodeVC {_ccv :: CodeChunk,

drasil-code/lib/Language/Drasil/Chunk/Parameter.hs
15:data ParameterChunk = PC {_pcc :: CodeChunk
```

(Not assigning anyone, but pinging @smiths and @JacquesCarette in particular.)

balacij:
Readability might be an issue for some of them, but they might just need to be renamed to something else that better describes them (“NamedChunk” in particular).

Keeping “Chunks” in the suffix seems odd, it feels akin to adding “Type” to the end of each type name in any typed language.

peter-michalski:
> they might just need to be renamed to something else that better describes them

I agree with this.

JacquesCarette:
[...]

PR #3726, opened 2024-05-08
Renamed UncertainChunk to Uncertain

NoahCardoso:
I renamed the UncertainChunk datatype to just Uncertain as part of the renaming chunks named in the form "XChunk" as just "X".
Contributes to #3244

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
@@ -36,7 +36,7 @@ constrained = map cnstrw inputDataConstraints ++
   [cnstrw probBr, cnstrw probFail, cnstrw stressDistFac] 
 
 plateLen, plateWidth, aspectRatio, charWeight, standOffDist :: UncertQ
-pbTol, tNT :: UncertainChunk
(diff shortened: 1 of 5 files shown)


samm82:
Based on @JacquesCarette's [comment](https://github.com/JacquesCarette/Drasil/issues/3244#issuecomment-1404049326), since "Uncertain" is also an adjective, we may want to think of another name for these chunks as well. Perhaps investigating the difference between `UncertainChunk` and `UncertQ` would help shed some light on this!

NoahCardoso:
@samm82 would changing UncertainChunk to Uncertainty be better? Also do you want all of the Chunks to be nouns?

JacquesCarette:
Now is our opportunity to think a bit deeper: `UncertainChunk` is the encoding of a `ConstrainedChunk` that *add*s an uncertainty field to it. It isn't "an uncertainty".

The problem with coming up with a good name for it is that it is likely to require coming up with a new better name for `ConstrainedChunk` first!

So there's a bigger issue (which you should probably get @samm82 and @balacij help you draft) about naming all things derived from Quantity.

NoahCardoso:
This pr is no longer relevant and should be deleted

PR #3727, opened 2024-05-08
Website: add links to the source code for each generated example #3668

### LINKED ISSUE #3668: Website: case studies list: add a link to the source code too
See the below image. In our list of “Generated Examples”/case studies, we show the final fruits of the projects, but not the inputs. We should somehow link to the relevant source code folder too. For example, for `DblPend`, we should add a bullet item: "Source Code" that is clickable to `https://github.com/JacquesCarette/Drasil/tree/master/code/drasil-example/dblpend` (at the relevant commit tree however, similar to the stable code linked), or we could make the shown names clickable themselves to the same link.

image

BilalM04:
Added links to the source code for each generated example on the website. Also updated the stable website to reflect those changes.

NOTE: I followed the existing code to add the new references. However, a problem, which @balacij pointed out, is that each chunk is being created twice. This may require a redesign of the code.

image

Closes #3668

### DIFF
--- a/code/drasil-website/lib/Drasil/Website/Example.hs
+++ b/code/drasil-website/lib/Drasil/Website/Example.hs
@@ -185,16 +185,16 @@ getCodeRef ex@E{sysInfoE=SI{_sys = sys}, choicesE = chcs} l verName =
     -- Program language converted for use in file folder navigation.
     programLang = convertLang l
 
--- | Similar to 'getCodeRef', but builds the source code references 
--- and uses 'getExampleSrcCodePath' instead.
+-- | Similar to 'getCodeRef', but builds the source code references
 buildDrasilExSrcRef :: Example -> Reference
 buildDrasilExSrcRef ex@E{sysInfoE=SI{_sys = sys}} = 
   makeURI refUID refURI refShortNm
   where
     refUID = "srcCodeRef" ++ sysName
-    refURI = getExampleSrcCodePath (codePath ex) sysName
+    refURI = path ++ "code/drasil-example/" ++ sysName
     refShortNm = shortname' $ S refUID
     sysName = map toLower $ programName sys
+    path = codePath ex
 
 -- | Similar to 'getCodeRef', but gets the doxygen references and uses 'getDoxRef' instead.
 getDoxRef :: Example -> Lang -> String -> Reference
@@ -233,11 +233,6 @@ getCodePath, getDoxPath :: FilePath -> String -> String -> FilePath
 getCodePath path ex programLang = path ++ "code/stable/" ++ map toLower ex ++ "/src/" ++ programLang -- need repoCommit path
 -- | Uses 'exRt' path (srsDoxPath in this module).
 getDoxPath path ex programLang = path ++ map toLower ex ++ "/doxygen/" ++ programLang ++ "/index.html" -- need example path
-
--- | Get the paths for the source code located in the code/drasil-example directory.
-getExampleSrcCodePath :: FilePath -> String -> FilePath
--- | Uses 'repoRt' path (codePath in this module).
-getExampleSrcCodePath path ex = path ++ "code/drasil-example/" ++ map toLower ex
 
 -- | Gather all references used in making the Examples section.
 exampleRefs :: FilePath -> FilePath -> [Reference]
--- a/code/drasil-website/lib/Drasil/Website/Example.hs
+++ b/code/drasil-website/lib/Drasil/Website/Example.hs
@@ -85,11 +85,11 @@ allExampleList = map (\x -> Nested (nameAndDesc x) $ Bullet $ map (, Nothing) (i
 individualExList :: Example -> [ItemType]
 -- No choices mean no generated code, so we do not need to display generated code and thus do not call versionList.
 individualExList ex@E{sysInfoE = SI{_sys = sys}, choicesE = [], codePath = srsP} = 
-  [Flat $ namedRef (getSourceCodeRef ex) (S "Source Code"),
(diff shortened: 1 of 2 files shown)


smiths:
Can we change the link name from "Source Code" to "Drasil Source Code"?  It isn't that big a deal, but when I first read Source Code, it made me think of the generated code.  If we add in the word Drasil, then the reader will know right away that this is something different.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -221,11 +234,18 @@ getCodePath path ex programLang = path ++ "code/stable/" ++ map toLower ex ++ "/
 -- | Uses 'exRt' path (srsDoxPath in this module).
 getDoxPath path ex programLang = path ++ map toLower ex ++ "/doxygen/" ++ programLang ++ "/index.html" -- need example path
 
+-- | Get the paths for the source code locations
```
JacquesCarette, line 237: The name of the function and the comment don't give me any clue that this is related to `drasil-example`, and it should. Also, we should try to not hard-code so much, so we should be constructing "code/drasil-example/" from other piecees.

JacquesCarette (changes requested):
The changes to the web site are great. Minor clean up in the source code needed first.

BilalM04:
@JacquesCarette I updated the function name and comments. Please let me know if it is still not what you are looking for. Here is the updated code:

https://github.com/JacquesCarette/Drasil/blob/111168e78117ca0ec3c523d669a81618e794ddd9/code/drasil-website/lib/Drasil/Website/Example.hs#L236-L241

As for the portion of the path being hardcoded, I followed the existing code used to build the other paths: 

https://github.com/JacquesCarette/Drasil/blob/111168e78117ca0ec3c523d669a81618e794ddd9/code/drasil-website/lib/Drasil/Website/Example.hs#L229-L236

Here, portions of the existing code for `getCodePath` (building the generated code path) and `getDoxPath` (building the doxygen paths) are hard-coded. I could not find any existing pieces I can use to replace the hard-coded parts and I'm not too sure where or how to add new pieces. One possible solution is to add another path root for the drasil-example directory in the following code; however, the `repoCommitRoot` path has an identical root, apart from the hard-coded values.

https://github.com/JacquesCarette/Drasil/blob/111168e78117ca0ec3c523d669a81618e794ddd9/code/drasil-website/app/Main.hs#L36-L51

I would appreciate your guidance and recommendation on the approach I should follow to resolve this issue.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -221,11 +234,19 @@ getCodePath path ex programLang = path ++ "code/stable/" ++ map toLower ex ++ "/
 -- | Uses 'exRt' path (srsDoxPath in this module).
 getDoxPath path ex programLang = path ++ map toLower ex ++ "/doxygen/" ++ programLang ++ "/index.html" -- need example path
 
+-- | Get the paths for the source code located in the code/drasil-example directory.
+getExampleSrcCodePath :: FilePath -> String -> FilePath
```
balacij, line 238: Is there any chance you can change this change the type signature to `:: FilePath -> Example -> FilePath`? Alternatively, since we only use this function in `getSourceCodeRef`, we can just inline the function body. I don't think this is a very reusable function (since it should only be called once for each example and packaged in a chunk anyway).

balacij: I think the latter option is better, actually.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -183,6 +185,17 @@ getCodeRef ex@E{sysInfoE=SI{_sys = sys}, choicesE = chcs} l verName =
     -- Program language converted for use in file folder navigation.
     programLang = convertLang l
 
+-- | Similar to 'getCodeRef', but gets the source code references 
+-- and uses 'getSourceCodePath' instead.
+getSourceCodeRef :: Example -> Reference
+getSourceCodeRef ex@E{sysInfoE=SI{_sys = sys}} =
```
balacij, line 191: `buildDrasilSrcRef` might be a _slightly_ better name. `get Source Code Ref`  tells me that we're going to be getting some sort of information _from_ `Example` rather than building something using it.

balacij: Alt.: `buildDrasilExSrcRef` or `buildDrasilExampleSrcRef`.

balacij (approved):
Very minor notes that might help address @JacquesCarette's concern.

@JacquesCarette Fixing the issue of "code/drasil-example/" appearing here might be out of scope for this PR. That issue is common in `drasil-website`, so I think we should create a follow-up issue and figure out how to fix that issue separately because it's not immediately clear to me how we should be fixing that.

We should also open up a separate issue (once this PR is merged) about the created chunks not being adequately saved and used that we discussed (and you mentioned in the OP), @BilalM04.

code/drasil-website/lib/Drasil/Website/Example.hs
BilalM04: Moved it inline here:

https://github.com/JacquesCarette/Drasil/blob/ac9f9b53fcb6dc5c3185ca2753713ac3a4567d53/code/drasil-website/lib/Drasil/Website/Example.hs#L187-L198

balacij:
@JacquesCarette Have you had a chance to look at the changes and comments since you last reviewed?

JacquesCarette (approved):

Issue #3288, opened 2023-02-14
Nonfunctional requirement(s) for maintainability should refer to the code, not the SRS


samm82:
As mentioned by @smiths in [his review of the ChemCode SRS](https://github.com/JacquesCarette/Drasil/blob/chemCode/People/Sam/ChemCode/SRS/SRS_Fdbck.pdf), an NFR for maintainability should refer to code's maintainability, not the SRS's.

Currently, a common NFR for maintainability is something along the lines of "Maintainable: The traceability between requirements, assumptions, theoretical models, general definitions, data definitions, instance models, likely changes, unlikely changes, and modules is completely recorded in traceability matrices in the SRS and module guide."

This change should be implemented in:

- [x] [GamePhysics](https://jacquescarette.github.io/Drasil/examples/gamephysics/SRS/srs/GamePhysics_SRS.html#Sec:NFRs) (this says "Maintainability: The development time for any of the likely changes should not exceed 10% percent of the original development time," but the value of 10% should be defined as a symbolic parameter)
- [x] [GlassBR](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Sec:NFRs)
- [x] [NoPCM](https://jacquescarette.github.io/Drasil/examples/nopcm/SRS/srs/NoPCM_SRS.html#Sec:NFRs)
- [x] [PDController](https://jacquescarette.github.io/Drasil/examples/pdcontroller/SRS/srs/PDController_SRS.html#Sec:NFRs)
- [x] [Projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:NFRs)
- [x] [SSP](https://jacquescarette.github.io/Drasil/examples/ssp/SRS/srs/SSP_SRS.html#Sec:NFRs)
- [x] [SWHS](https://jacquescarette.github.io/Drasil/examples/swhs/SRS/srs/SWHS_SRS.html#Sec:NFRs)

smiths:
[...]

PR #3735, opened 2024-05-13
Updated NFR for maintainability to refer to code, not SRS: #3288

BilalM04:
Updated the NFR for maintainability to the following description:

> If a likely change is made to the finished software, it will take 10% of the original development time, assuming the same development resources are available.

Change applied to the following examples:

- [GamePhysics](https://jacquescarette.github.io/Drasil/examples/gamephysics/SRS/srs/GamePhysics_SRS.html#Sec:NFRs)
- [GlassBR](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Sec:NFRs)
- [NoPCM](https://jacquescarette.github.io/Drasil/examples/nopcm/SRS/srs/NoPCM_SRS.html#Sec:NFRs)
- [PDController](https://jacquescarette.github.io/Drasil/examples/pdcontroller/SRS/srs/PDController_SRS.html#Sec:NFRs)
- [Projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:NFRs)
- [SSP](https://jacquescarette.github.io/Drasil/examples/ssp/SRS/srs/SSP_SRS.html#Sec:NFRs)
- [SWHS](https://jacquescarette.github.io/Drasil/examples/swhs/SRS/srs/SWHS_SRS.html#Sec:NFRs)

TO DO: Instead of hard-coding the "10%", it should be a parameter than can vary at the time of SRS generation.

Contributes to #3288

### DIFF
--- a/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
+++ b/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
@@ -106,7 +106,7 @@ nfReqIntro _  = mkParagraph $ reqIntroStart +:+. nfrReqIntroBody
 mkMaintainableNFR :: String -> Integer -> String -> ConceptInstance
 mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
   S "If a likely change is made" `S.toThe` 
-  S "finished software, it will take", addPercent percent, EmptyS `S.ofThe`
+  S "finished software, it will take", addPercent percent `S.ofThe`
   S "original development time,",
   S "assuming the same development resources are available"
   ]) lbl nonFuncReqDom
--- a/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
+++ b/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
@@ -8,7 +8,7 @@ module Drasil.Sections.Requirements (
   fullReqs, fullTables, inReq, inTable,
   mkInputPropsTable, mkQRTuple, mkQRTupleRef, mkValsSourceTable,
   -- * Non-functional Requirements
-  nfReqF
+  nfReqF, mkMaintainableNFR
   ) where
 
 import Language.Drasil
@@ -17,7 +17,7 @@ import qualified Language.Drasil.Sentence.Combinators as S
 import Drasil.Sections.ReferenceMaterial(emptySectSentPlu)
 import Theory.Drasil (HasOutput(output))
 
-import Data.Drasil.Concepts.Documentation (description, funcReqDom,
+import Data.Drasil.Concepts.Documentation (description, funcReqDom, nonFuncReqDom,
   functionalRequirement, input_, nonfunctionalRequirement, {-output_,-} section_,
   software, symbol_, value, reqInput)
 import Data.Drasil.Concepts.Math (unit_)
@@ -99,6 +99,17 @@ fReqIntro _  = mkParagraph $ reqIntroStart +:+. frReqIntroBody
 nfReqIntro :: [Contents] -> Contents
 nfReqIntro [] = mkParagraph $ emptySectSentPlu [nonfunctionalRequirement]
 nfReqIntro _  = mkParagraph $ reqIntroStart +:+. nfrReqIntroBody
+
+-- | Common Non-Functional Requirement for Maintainability.
+-- Takes in a Reference Address ('String'), a percent value ('Integer'), 
+-- and a label ('String').
+mkMaintainableNFR :: String -> Integer -> String -> ConceptInstance
+mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
+  S "If a likely change is made" `S.toThe` 
+  S "finished software, it will take", addPercent percent, EmptyS `S.ofThe`
+  S "original development time,",
+  S "assuming the same development resources are available"
+  ]) lbl nonFuncReqDom
 
 -- | Creates an Input Data Table for use in the Functional Requirments section. Takes a list of wrapped variables and something that is 'Referable'.
 mkInputPropsTable :: (Quantity i, MayHaveUnit i, HasShortName r, Referable r) => 
--- a/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
+++ b/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
@@ -106,7 +106,7 @@ nfReqIntro _  = mkParagraph $ reqIntroStart +:+. nfrReqIntroBody
 mkMaintainableNFR :: String -> Integer -> String -> ConceptInstance
 mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
   S "If a likely change is made" `S.toThe` 
-  S "finished software, it will take", addPercent percent `S.ofThe`
+  S "finished software, it will take at most", addPercent percent `S.ofThe`
   S "original development time,",
   S "assuming the same development resources are available"
   ]) lbl nonFuncReqDom
--- a/code/drasil-docLang/package.yaml
+++ b/code/drasil-docLang/package.yaml
@@ -37,7 +37,6 @@ library:
   - Drasil.DocumentLanguage.Units
   - Drasil.DocumentLanguage.TraceabilityGraph
   - Drasil.SRSDocument
-  - Drasil.Sections.Requirements
   when:
   - condition: false
     other-modules: Paths_drasil_docLang
--- a/code/drasil-docLang/package.yaml
+++ b/code/drasil-docLang/package.yaml
@@ -37,6 +37,7 @@ library:
   - Drasil.DocumentLanguage.Units
   - Drasil.DocumentLanguage.TraceabilityGraph
   - Drasil.SRSDocument
+  - Drasil.Sections.Requirements
   when:
   - condition: false
     other-modules: Paths_drasil_docLang
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
@@ -132,7 +132,8 @@ understandability = cic "understandability" (foldlSent [
 
 maintainability :: ConceptInstance
 maintainability = cic "maintainability" (foldlSent [
-  S "development time for any" `S.the_ofTheC` S "likely changes should not exceed", 
-  addPercent (10 :: Integer), S "percent of the original development time"
+  S "If a likely change is made" `S.toThe` S "finished software, it will take", 
+  addPercent (10 :: Integer), S "percent" `S.ofThe` S "original development time,",
+  S "assuming the same development resources are available"
   ]) "Maintainability" nonFuncReqDom
 
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
@@ -6,6 +6,7 @@ import Language.Drasil.Chunk.Concept.NamedCombinators
 import qualified Language.Drasil.Sentence.Combinators as S
 
 import qualified Drasil.DocLang.SRS as SRS (solCharSpec)
+import Drasil.Sections.Requirements (mkMaintainableNFR)
 import Data.Drasil.Concepts.Documentation as Doc (body, funcReqDom, input_, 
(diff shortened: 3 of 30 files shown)


code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -2651,7 +2651,7 @@ <h2>Non-Functional Requirements</h2>
               </p>
               <p>
                 <div id="maintainability">
-                  Maintainability: The development time for any of the likely changes should not exceed 10<em>%</em> percent of the original development time.
+                  Maintainability: If a likely change is made to the finished software, it will take 10<em>%</em> percent of the original development time, assuming the same development resources are available.
```
samm82, line 2654: "10% percent" is redundant; it should probably just be "10%"

Also, is there a reason that the percent sign is in italics? I know this isn't part of this PR, so this might be pulled out to a new issue to at least investigate (but I think it shouldn't be in italics).

code/drasil-example/swhs/lib/Drasil/SWHS/Requirements.hs
```diff
@@ -160,10 +157,10 @@ reusable = cic "reusable" (foldlSent [
...
 maintainable :: ConceptInstance
 maintainable = cic "maintainable" (foldlSent [
-  S "The traceability between", foldlList Comma List [plural requirement,
-  plural assumption, plural thModel, plural genDefn, plural dataDefn, plural inModel,
-  plural likelyChg, plural unlikelyChg, plural module_], S "is completely recorded in",
-  plural traceyMatrix, S "in the", getAcc srs `S.and_` phrase mg]) "Maintainable" nonFuncReqDom
+  S "If a likely change is made" `S.toThe` S "finished software, it will take", 
+  addPercent (10 :: Integer), S "percent" `S.ofThe` S "original development time,",
+  S "assuming the same development resources are available"
+  ]) "Maintainable" nonFuncReqDom
```
samm82, line 163: This requirement is duplicated a bunch; do we have some place for storing "common" requirements where this could be defined once and reused?

samm82 (changes requested):

smiths (commented):
I agree with the feedback from @samm82, including that the word percent shouldn't be there and the percentage symbol should not be in italics.  Those changes will make the change fine from the point of view of stable.  How the changes are done should also be investigated.  Specifically, we don't want the same Drasil code repeated.

BilalM04:
@samm82 @smiths 

It seems the "%" symbol is always generated between an `<em></em>` tag in HTML, which by default renders the contained text in italics. The issue is currently present in the existing SRS documents, for example in the [Input Data Constraints](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Table:InDataConstraints:~:text=1000%20m-,10%25,-vlaunch) table in the SRS for Projectile.

In regards to the code duplication, I am unsure where to define the requirement so it can be accessed by various examples. Currently, it appears that requirements are defined for each example, with no place for "common" requirements. One possible file I was looking at is [code/drasil-docLang/lib/Drasil/Sections/Requirements.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs) which has the following description in its README: 

> Contains the functions to generate the Requirements section

smiths:
> @samm82 @smiths
> 
> It seems the "%" symbol is always generated between an `\<em\>\</em\>` tag in HTML, which by default renders the contained text in italics. The issue is currently present in the existing SRS documents, for example in the [Input Data Constraints](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Table:InDataConstraints:~:text=1000%20m-,10%25,-vlaunch) table in the SRS for Projectile.
> 

The percent symbol isn't a priority.  We can leave it as it is for now.

> In regards to the code duplication, I am unsure where to define the requirement so it can be accessed by various examples. Currently, it appears that requirements are defined for each example, with no place for "common" requirements. One possible module I was looking at is [code/drasil-docLang/lib/Drasil/Sections/Requirements.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs) which has the following description in its [README](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-docLang/lib/Drasil/Sections/README.md):
> 
> > Contains the functions to generate the Requirements section

@samm82 or @balacij can likely point you to the best way to add common requirements.  Much of the boilerplate in the requirements documents are shared between examples.  @samm82 was working on a way to standardize common requirements, like the input and output requirements.  Ideally the structure of the requirements will be the same between examples, but they will be parametrized by the percentage of time needed for the change, so that different examples can have different percentages.  There really isn't a compelling reason for a specific value for any of our examples, but in general not every future maintainability requirement is going to be identical.

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
```diff
@@ -100,6 +100,17 @@ nfReqIntro :: [Contents] -> Contents
 nfReqIntro [] = mkParagraph $ emptySectSentPlu [nonfunctionalRequirement]
 nfReqIntro _  = mkParagraph $ reqIntroStart +:+. nfrReqIntroBody
 
+-- | Common Non-Functional Requirement for Maintainability.
+-- Takes in a Reference Address ('String'), a percent value ('Integer'), 
+-- and a label ('String').
+mkMaintainableNFR :: String -> Integer -> String -> ConceptInstance
+mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
+  S "If a likely change is made" `S.toThe` 
+  S "finished software, it will take", addPercent percent, EmptyS `S.ofThe`
```
samm82, line 109: Is this `EmptyS` needed? I feel like it can be removed, but maybe not

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
```diff
@@ -131,9 +132,4 @@ understandability = cic "understandability" (foldlSent [
   ]) "Understandability" nonFuncReqDom
 
 maintainability :: ConceptInstance
-maintainability = cic "maintainability" (foldlSent [
-  S "If a likely change is made" `S.toThe` S "finished software, it will take", 
-  addPercent (10 :: Integer), S "percent" `S.ofThe` S "original development time,",
-  S "assuming the same development resources are available"
-  ]) "Maintainability" nonFuncReqDom
-
+maintainability = mkMaintainableNFR "maintainability" (10 :: Integer) "Maintainability"
```
samm82, line 135: Do we need to specify `:: Integer` for each of these calls, or can we just pass in `10`; either way, great job! I like this code reuse!

samm82 (changes requested):

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
samm82: Looks like this has been moved to #3748

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
BilalM04: Yep, you're right!

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
BilalM04: Before, it was necessary; however, since the function `mkMaintainableNFR` has an explicit type declaration for that parameter, it is no longer needed. Good catch!

samm82 (approved):

code/drasil-docLang/package.yaml
```diff
@@ -37,6 +37,7 @@ library:
   - Drasil.DocumentLanguage.Units
   - Drasil.DocumentLanguage.TraceabilityGraph
   - Drasil.SRSDocument
+  - Drasil.Sections.Requirements
```
balacij, line 40: I _think_ that whatever we're interested in importing from `Drasil.Sections.Requirements` should instead be imported from `Drasil.SRSDocument` (I see that at the bottom of this file, there is a list of `Drasil.Sections.X` re-exports).

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
```diff
@@ -100,6 +100,17 @@ nfReqIntro :: [Contents] -> Contents
 nfReqIntro [] = mkParagraph $ emptySectSentPlu [nonfunctionalRequirement]
 nfReqIntro _  = mkParagraph $ reqIntroStart +:+. nfrReqIntroBody
 
+-- | Common Non-Functional Requirement for Maintainability.
+-- Takes in a Reference Address ('String'), a percent value ('Integer'), 
+-- and a label ('String').
+mkMaintainableNFR :: String -> Integer -> String -> ConceptInstance
+mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
+  S "If a likely change is made" `S.toThe` 
+  S "finished software, it will take", addPercent percent `S.ofThe`
```
balacij, line 109: What do you think of "it will take _at most_ X%" instead of an exact %?

balacij (changes requested):
Certainly cleans up the code nicely :smile:

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
smiths: I think we should make the change suggested by @balacij.  The 10% is a the bound on the time.  The developer can do the redevelopment (change) in less than 10% of the time.

code/drasil-docLang/package.yaml
BilalM04: The header for `Drasil.SRSDocument` says the following:

https://github.com/JacquesCarette/Drasil/blob/21ff3588595f34c3135af6e6db538e16d5318767/code/drasil-docLang/lib/Drasil/SRSDocument.hs#L1-L4

Since this function would not be common to *all* SRS examples, would it still make sense to export from here? Additionally, in `Drasil.SRSDocument` the functions from `Drasil.Sections.X` are being imported from [`Drasil.DocLang`](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-docLang/lib/Drasil/DocLang.hs).

Looking more intro `Drasil.DocLang` it exports functions from `Drasil.Sections.Requirements`, as well as the other sections:

https://github.com/JacquesCarette/Drasil/blob/21ff3588595f34c3135af6e6db538e16d5318767/code/drasil-docLang/lib/Drasil/DocLang.hs#L57-L58

What are your thoughts on exporting the function just from `Drasil.DocLang`, which is already an exposed module. Then importing from `Drasil.DocLang` wherever it is needed.

balacij: That makes sense to me. Let's go with your suggestion :smile:

samm82 (approved):

JacquesCarette (approved):

JacquesCarette:
Up to @balacij to verify that the changes he'd asked for have been done.

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
```diff
@@ -131,8 +132,4 @@ understandability = cic "understandability" (foldlSent [
   ]) "Understandability" nonFuncReqDom
 
 maintainability :: ConceptInstance
-maintainability = cic "maintainability" (foldlSent [
-  S "development time for any" `S.the_ofTheC` S "likely changes should not exceed", 
-  addPercent (10 :: Integer), S "percent of the original development time"
-  ]) "Maintainability" nonFuncReqDom
-
+maintainability = mkMaintainableNFR "maintainability" 10 "Maintainability"
```
balacij, line 135: Sorry to be the nitpick... can this standard maintainability be moved to `drasil-docLang` now too?

BilalM04: I'm not sure what you mean. The code you linked is a call to the function exported from `drasil-docLang`. How would I go about moving it to `drasil-docLang`? Additionally, the parameters are not consistent across all the examples.

balacij: Discussed in person. We'll just move this to a new ticket.

Issue #3710, opened 2024-04-03
Add `stan` and `weeder` to our dev. workflow


balacij:
Motivation: #3700

Objective: Adding more static analysis tools (here, `stan` and `weeder`) to our development workflow.

`stan` and `weeder` require [HIE](https://www.haskell.org/ghc/blog/20190626-HIEFiles.html) files post-compilation of a project before they can be used.

Approximate steps:
1. *Tell GHC to print out the HIE information that `stan` and `weeder` need.* - #3708
2. *Install `stan` locally.* Careful! `stan` must be compiled with the same GHC version we intend to use for compiling Drasil. As of #3707, we're going to use GHC 9.2.8, but I could not get `stan` installed locally using GHC 9.2.X, so I had to update our project further to the GHC 9.4.X series by switching our targeted stack release to lts-21.25 (you can use the script found in #3707 to quickly update this too). Afterwards, you will need to add [some dependencies](https://github.com/kowainik/stan/blob/a610d74fcb4a852a9e6b0926580c055bb3dddc23/stack-ghc-9.4.8.yaml#L3-L11) to our main `stack.yaml` file in the `code` folder and run `stack install stan`. That's what worked for me, but YMMV.
3. *Address (or suppress) issues reported by `stan`.* After installing `stan`, you should run it on all of our `drasil-*` projects and the example projects. To get you started, you can run `find . -type d -name 'drasil-*' -exec sh -c 'cd {} && stan > stan.log' \;` to run `stan` on all `drasil-` libraries, and then run `cat drasil-X/stan.log` to read the console output for each respective library. It would be helpful to first go through each _category_ of error and fix the reports by _category_ rather than report. Two of our notable reports as of Apr. 2nd, 2024:
    1. Lazy data causing memory leaks. `stan` suggests adding the `StrictData` language extension to make all of our data types strict, but I found that Drasil runs into an infinite loop! This will require some debugging.
    2. Infix operators lacking fixity. We define many infix operators, many lacking a related fixity (i.e., what binds tighter? + or * in the expression `a+b*c`?). We will need to go through each area that related infix operators are used and come up with a binding order.
[...]

PR #3707, opened 2024-04-02
Update stackage release: 20.20 (GHC 9.2.7) -> 20.26 (GHC 9.2.8)

balacij:
And create a script for automating future updates.

PR #3739, opened 2024-05-16
Added `stan` dependencies to stack.yaml #3710

NoahCardoso:
Contributes to #3710. Added `stan` dependencies to stack.yaml. I was able to install `stan` and get it working on Drasil. Still need to address (or suppress) issues reported by `stan` as mentioned in #3710.


balacij:
Does this install `stan` as something we can run? I'm not quite sure how this works. Can you explain a bit more?

NoahCardoso:
>2. Install stan locally. Careful! stan must be compiled with the same GHC version we intend to use for compiling Drasil. As of https://github.com/JacquesCarette/Drasil/pull/3707, we're going to use GHC 9.2.8, but I could not get stan installed locally using GHC 9.2.X, so I had to update our project further to the GHC 9.4.X series by switching our targeted stack release to lts-21.25 (you can use the script found in https://github.com/JacquesCarette/Drasil/pull/3707 to quickly update this too). Afterwards, you will need to add [some dependencies](https://github.com/kowainik/stan/blob/a610d74fcb4a852a9e6b0926580c055bb3dddc23/stack-ghc-9.4.8.yaml#L3-L11) to our main stack.yaml file in the code folder and run stack install stan. That's what worked for me, but YMMV. 

I completed this step. I added the dependencies and was able to run `stan` on the Drasil libraries. You still need to run `stack install stan`.

code/stack.yaml
```diff
@@ -69,6 +69,24 @@ extra-deps:
...
+- clay-0.14.0@sha256:a50ba73137a39c55e89f24a7792107ec40ba07320b2c5ff7932049845c50ffc9,2204
+- dir-traverse-0.2.3.0@sha256:adcc128f201ff95131b15ffe41365dc99c50dc3fa3a910f021521dc734013bfa,2137
+- extensions-0.1.0.0@sha256:b8105dc43a57b0b3b54879e8dbb905676dfee3e8b59301fefbf2409a0fe95710,4447
+- tomland-1.3.3.2@sha256:887dc39a8c9819deb8fcb6fde72e87dad4c94108b1736a5bf7215ccf3117bd0f,9474
+- trial-0.0.0.0@sha256:ebd93f3485dd7f0ce8426fa46b500f26edbea285c2890150123e1b0f6a92c7db,4410
+- trial-optparse-applicative-0.0.0.0@sha256:ee878dbf412849d4fb52d7bca93bec45a6392a93e03a46bd320109b9a0ff2a38,2540
+- trial-tomland-0.0.0.0@sha256:c10c5c3f94e3dcfcad4a66cf70682ace77d5381596aff91fa2030636d07de977,2541
+- validation-selective-0.2.0.0@sha256:e1ab5482dede8bf676d729a09109c7c5f798363b9d458e4197a27afb8b48fdd3,3907
+- colourista-0.1.0.2@sha256:87b6c096563c3dc5afb1161f06891235370862fc3406406f1f10f1864e55e0a1,3364
+- slist-0.2.1.0@sha256:679013d3a03ed1f09fc46cc6ef87f94cc36a1320b2bf4fe4f91d102f85277528,3513
```
balacij, line 81: Can you please add a note mentioning that these are here specifically for `stan`?

balacij: And specifically, tested against GHC 9.4.8.

code/stack.yaml
```diff
@@ -69,6 +69,24 @@ extra-deps:
...
+
+# This configuration option allows the use of newer versions of dependencies than what is specified
+allow-newer: true
+
+# This configuration option specifies a list of dependencies for which newer versions are allowed.
+# In this case, it's allowing newer versions of the 'stan' package.
+allow-newer-deps:
+- stan
```
balacij, line 91: Do we only need this for Stackage's LTS 21.25 (GHC 9.4.8) list? If we update our LTS to, say, 22.22, do we still need this? If we only need it for our existing stackage list used, we should probably also just document it.

I'm not sure if we still want to do it, but I think that in the future we'll be wanting to compile against a few versions of GHC, so just having that information around is helpful.

balacij (approved):
Gotcha! Ok, that makes more sense now, just two minor notes, but otherwise, this lets me install `stan` locally :slightly_smiling_face:

code/stack.yaml
NoahCardoso: Actually, I don't think that the code is necessary. I just deleted it and everything still worked but I don't know if that is just because I already have stan installed. 

Also, why was #3708 never merged? do we need to generated the .hie files as in #3710 it says that we need them?
>Tell GHC to print out the HIE information that stan and weeder need. - https://github.com/JacquesCarette/Drasil/pull/3708

>ghc-options:
  "$everything": -fwrite-ide-info -hiedir=.hie

code/stack.yaml
NoahCardoso: Yes I will add that

JacquesCarette: Do they really all need to be pinned that hard?

code/stack.yaml
JacquesCarette: #3708 not merged because I didn't feel I quite understood the need for it and what it really did. And @balacij got busy... so it's kind of on his plate to convince me!

code/stack.yaml
NoahCardoso: No, they do not have to be this should also work
- clay>=0.14.0
- dir-traverse>=0.2.3.0
- extensions>=0.1.0.0
- tomland>=1.3.3.2
- trial>=0.0.0.0
- trial-optparse-applicative>=0.0.0.0
- trial-tomland>=0.0.0.0
- validation-selective>=0.2.0.0
- colourista>=0.1.0.2
- slist>=0.2.1.0

JacquesCarette: It's usually a good habit to give upper bounds too, since major version bumps tend to break things.  So
- clay < 1.0.0
- tomland < 2.0.0.0
and so on.

NoahCardoso: Like this
- clay>=0.14.0 && <0.15.0
- dir-traverse>=0.2.3.0 && <0.3.0.0
- extensions>=0.1.0.0 && <0.1.0.2
- tomland>=1.3.3.2 && <1.4.0.0
- trial>=0.0.0.0 && <0.1.0.0
- trial-optparse-applicative>=0.0.0.0 && <0.1.0.0
- trial-tomland>=0.0.0.0 && <0.1.0.0
- validation-selective>=0.2.0.0 && <0.3.0.0
- colourista>=0.1.0.2 && <0.2.0.0
- slist>=0.2.1.0 && <0.3.0.0

JacquesCarette: Well, that might be too tight, but we can certainly try it and see.

NoahCardoso: Actually, they have to be hard-pinned otherwise it doesn't work

code/stack.yaml
balacij: @NoahCardoso I believe it is required. I just tried it without the referenced code, and it didn't work.

balacij:
We force-pushed IRL to (1) rebase on `master` to avoid having the `master` commits shown in the PR [might be worth adding to the contributor's guide if it isn't already there], and (2) to remove accidentally committed `.hie` files. The third one was because we did the order in 2->1->2 and after the second one, the `.hie` files re-appeared, so we had to erase them again.

balacij:
@NoahCardoso Can you please move that last commit to a new branch+pr? Realistically, I know why `foldl'` is an improvement over `foldl` but it should be reflected in the PRs description, but it is good practice to keep our PRs separate.

NoahCardoso:
I used a force push to undo my last commit

balacij:
Thanks @NoahCardoso!

JacquesCarette (approved):

PR #3740, opened 2024-05-16
Renderer template generator

B-rando1:
I created a Python script that parses an existing renderer file, and creates a new file with unimplemented instances of each of the typeclasses, with each of the methods set to `undefined`.  This is useful because it lets you test the code as you implement it, without Haskell's typechecker getting in the way.


JacquesCarette (approved):
Nice.

In theory, one could even do this in Haskell (as there tools for dealing with Haskell code in hackage), which would tend to be less fragile. There are also tools like [tree-sitter](https://tree-sitter.github.io/tree-sitter/) that might be useful.

But that might all be overkill.

PR #3743, opened 2024-05-17
Abstracted a couple common functions between Julia and Swift

B-rando1:
First, Julia and Swift both call booleans `Bool`, so I moved that naming into `CommonPseudoOO`.

Second, Julia and Swift are both not supported by Doxygen, so I followed @bmaclach's convention from Swift, using point form for the doc comments instead of backslashes, which are required for Doxygen.  Since this is common between Julia and Swift, I extracted this functionality out of Swift and put it into `CommonPseudoOO` as well.

Also just an FYI, my work on the Julia renderer is on its own branch at https://github.com/JacquesCarette/Drasil/tree/JuliaRenderer.  It's still got a long way to go, but it successfully generates an empty Julia file with a doc comment.


code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
```diff
@@ -442,3 +443,34 @@ inherit n = toCode $ maybe empty ((colon <+>) . text) n
...
+
+-- | Generates a doc comment for a language that isn't supported by Doxygen.
+-- Since Doxygen doesn't support the language, it uses dashes instead of backslashes.
+modDoc' :: ModuleDocRenderer
+modDoc' desc as date m = m : [desc | not (null desc)] ++
+      [docField authorDoc (stringList as) | not (null as)] ++
+      [docField dateDoc date | not (null date)] ++
+      [docField noteDoc watermark]
+
+-- | Creates an arbitrary DocC Markup field for documentation.
```
bmaclach, line 465: DocC is a swift-specific term (see https://www.swift.org/documentation/docc/). It's based on Markdown, so it may still be a good candidate as a language-agnostic alternative to doxygen, but this comment specifically referencing DocC should probably be updated now that this function is being used for more than just swift.

If Julia has its own standard syntax for documentation generation, it would be good for GOOL to generate that syntax (perhaps a future improvement?)

bmaclach (approved):
The changes seem logical and correct to me, and I don't believe my previous comment is something that should block this from being merged, so I'm approving.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
B-rando1: Oh, thanks for pointing that out @bmaclach!  I'll make sure to edit the comment about DocC, and see if there are any better ways to treat doc comments in Julia.

B-rando1: I looked at documentation for Julia.  It has a [built-in documentation system](https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation) where you put a string with markdown-formatted documentation above an object, and it automatically assigns that information to that object.  Since it uses markdown, it should work well to use the same format as the DocC format for Julia.

JacquesCarette (approved):

PR #3758, opened 2024-05-29
STAN-0202 foldl' #3710

NoahCardoso:
In large Haskell projects, using `foldl'` can help improve performance and reduce memory usage, making your code more efficient.  In large data structures, using `foldl` can lead to space leaks and stack overflows because it builds up thunks that represent the unevaluated computations of each step. `foldl'` avoids this problem by immediately evaluating the accumulator at each step, which can significantly reduce memory usage for large data structures.

### DIFF
--- a/code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
+++ b/code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
@@ -22,7 +22,7 @@ import SysInfo.Drasil
 import Language.Drasil
 import Language.Drasil.Chunk.Concept.NamedCombinators as NC
 import Language.Drasil.Sentence.Combinators as S
-import Data.Foldable(foldl')
+import Data.Foldable (foldl')
 
 -- | Makes a Traceability Table/Matrix that contains Items of Different Sections.
 generateTraceTable :: SystemInformation -> LabelledContent
--- a/code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
+++ b/code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
@@ -22,6 +22,7 @@ import SysInfo.Drasil
 import Language.Drasil
 import Language.Drasil.Chunk.Concept.NamedCombinators as NC
 import Language.Drasil.Sentence.Combinators as S
+import Data.Foldable(foldl')
 
 -- | Makes a Traceability Table/Matrix that contains Items of Different Sections.
 generateTraceTable :: SystemInformation -> LabelledContent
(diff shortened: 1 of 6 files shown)


code/stack.yaml
```diff
@@ -73,6 +73,26 @@ extra-deps:
...
+- slist-0.2.1.0@sha256:679013d3a03ed1f09fc46cc6ef87f94cc36a1320b2bf4fe4f91d102f85277528,3513
+
+# This configuration option allows the use of newer versions of dependencies than what is specified
+allow-newer: true
+
+# This configuration option specifies a list of dependencies for which newer versions are allowed.
+# In this case, it's allowing newer versions of the 'stan' package.
+allow-newer-deps:
+- stan
+
```
samm82, line 95: This code looks like a carryover from #3739, which I think is OK (as long as that gets merged first lol)

code/drasil-docLang/lib/Drasil/Sections/TraceabilityMandGs.hs
```diff
@@ -22,6 +22,7 @@ import SysInfo.Drasil
 import Language.Drasil
 import Language.Drasil.Chunk.Concept.NamedCombinators as NC
 import Language.Drasil.Sentence.Combinators as S
+import Data.Foldable(foldl')
```
samm82, line 25:
```suggestion
import Data.Foldable (foldl')
```

Super minor formatting thing (I'd do this across these import statements), but I don't think this should be blocking

samm82 (commented):
Just some (probably) non-blocking improvements that could be made. Note that it looks like there are [some instances](https://wiki.haskell.org/Foldr_Foldl_Foldl'#Conclusion) where `foldl` would be better than `foldl'`, but "`foldl` is rarely the right choice", so this looks good to me (since the CI passes)

balacij (approved):

code/stack.yaml
JacquesCarette: It would still be better to decouple these changes.

NoahCardoso:
I used force push to decouple the changes from my stack.yaml pr.

balacij:
I think this one is ready for review again, @samm82 @JacquesCarette.

code/drasil-printers/lib/Language/Drasil/Debug/Print.hs
```diff
@@ -9,6 +9,7 @@ import           Utils.Drasil (stringList)
 import qualified Data.Map as Map
 import           Control.Lens ((^.), view)
 import           Data.List (nub, sort, sortBy)
+import           Data.Foldable(foldl')
```
samm82, line 12:
```suggestion
import           Data.Foldable (foldl')
```

samm82 (commented):
Still looks good to me (again with one minor, non-blocking spacing discrepancy lol)

samm82 (approved):
💖

JacquesCarette (approved):

PR #3761, opened 2024-05-30
Switch some lists to Set from containers

NoahCardoso:
Nub is slow and suggestion 209 pops up a lot. When working with lists that have `Ord` we can use toList and fromList from Data.Set. Nub has quadratic complexity because it checks each element of the list against every other element of the list. If you convert your list to a Set, you can avoid this quadratic complexity because Set does not allow duplicate elements.
image


JacquesCarette (changes requested):

code/drasil-lang/lib/Language/Drasil/CodeExpr/Extract.hs
```diff
@@ -90,7 +91,7 @@ eNamesRI' (UpFrom il)     = eNames' (snd il)
 
 -- | Get dependencies from an equation.
 eDep :: CodeExpr -> [UID]
-eDep = nub . eNames
+eDep = toList . fromList . eNames
```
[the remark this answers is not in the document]
balacij: I'm fairly certain that `toList . fromList` would also sort the list, since Haskell's internal `Set` is really an ordered binary tree (like we discussed yesterday, @NoahCardoso). That is to say, I'm a bit skeptical that `eDep` uses care for the order of the `UID`s. Have you tried switching the type here from a list to a set instead? Our `eDep`-like (UID-dependency searching) tools have a lot of `nub` usage, so switching to a set type might make sense.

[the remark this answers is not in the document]
NoahCardoso: @balacij should I change every list of UIDs to Set of UIDs?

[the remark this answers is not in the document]
balacij: I think there will be many, but I'm not sure if all should be changed. For example, if we have a list of UIDs in a chunk, we will need to check if order matters/makes sense.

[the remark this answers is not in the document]
JacquesCarette: It's likely a lot more subtle than that: the question is when does order matter, and when it doesn't. And sometimes order matters only for display purposes.

We do over-use lists, but that's quite typical of most Haskell projects which make lists too easy to use!

code/drasil-lang/lib/Language/Drasil/Expr/Extract.hs
```diff
@@ -76,4 +76,4 @@ eNamesRI' (UpFrom il)     = eNames' (snd il)
 
 -- | Get dependencies from an equation.  
 eDep :: Expr -> [UID]
-eDep = nub . eNames
+eDep = toList . fromList . eNames
```
balacij, line 79: @NoahCardoso If you're struggling with figuring out which things should be Sets and which should be lists, I think there are two ways you can go about this:
1. Research first.
2. Make a hypothesis and test it.

In other words,
1. Traverse over all things that use `eDep`, and then all things that use those things, and figure out if order is somehow important and why it is (or why it shouldn't be...). Thankfully, there isn't too much usage of `eDep`, this might be easier and better, logically.
2. Semi-confidently assume that this function should return a Set rather than a list. Our hypothesis of sorts -- the comment just says "Get dependencies from an equation," but it doesn't mention any sort of ordering of the dependencies, it just lists what a specific expression depends on (i.e., variables and functions). You can try this, and see what breaks.

Similar to `eDep`, I would imagine this extends to `sdep`, `shortdep`, `eNames`, but you can figure it out as you go along.

<details>

<summary>I took a quick peek at `eDep` usage. It doesn't like there's too much to traverse</summary>

```console
λ ~/Programming/Drasil/ glassbrIntro rg "[^a-z]eDep" -ths
code/drasil-code/lib/Language/Drasil/Chunk/CodeBase.hs
20:codevars e m = map (varResolve m) $ eDep e
24:codevars' e m = map (varResolve m) $ nub $ eDep' e

code/drasil-lang/lib/Language/Drasil/Expr/Development.hs
9:  , eDep, eNames, eNames', eNamesRI
15:import Language.Drasil.Expr.Extract (eDep, eNames, eNames', eNamesRI)

code/drasil-lang/lib/Language/Drasil/CodeExpr/Extract.hs
2:    eDep, eDep',

code/drasil-lang/lib/Language/Drasil/CodeExpr/Development.hs
11:    eDep, eDep',
25:import Language.Drasil.CodeExpr.Extract (eDep, eDep', eNamesRI, eNamesRI')
```

</details>

balacij:
Since this is a WIP branch now, I changed the status to a Draft PR.

code/drasil-lang/lib/Language/Drasil/Expr/Extract.hs
NoahCardoso: eDep is only used in codevars. I have eDep as a Set currently. This is roughly the order of the other functions that relate to eDep. eDep -> codevars -> fstvars -> fstdecl -> genFunc -> genModDef, genModFuncs ... -> ... generates things. Once things start to get generated in gool I assume that order matters. Codervars, fstvars and fstdecl might be able to be Sets however codevars is a list of CodeChunks which are not `ord`

NoahCardoso: Also `meDep` is used for printing. Should it be? I thought it was only for getting dependencies and its order shouldn't matter.
https://github.com/JacquesCarette/Drasil/blob/864e1385e94f900d8035f605a0b170b9c439dc61/code/drasil-lang/lib/Language/Drasil/ModelExpr/Extract.hs#L87-L89

code/drasil-gool/lib/GOOL/Drasil/State.hs
```diff
@@ -468,7 +469,7 @@ callMapTransClosure = over callMap tClosure
         traceCalls :: Map QualifiedName [QualifiedName] -> [QualifiedName] -> 
           [QualifiedName]
         traceCalls _ [] = []
-        traceCalls cm (c:cs) = nub $ c : traceCalls cm (nub $ cs ++ 
+        traceCalls cm (c:cs) = toList . fromList $ c : traceCalls cm ( cs ++
```
balacij, line 472: I think that if you remove the existing `toList . fromList`s, this PR will be in a mergeable state, no? I think we should do that, since you're now going through the `nub`d lists noted by `stan` and figuring out which ones really need a `nub` and which ones really want to be a different type.

balacij (changes requested):

code/drasil-gool/lib/GOOL/Drasil/State.hs
NoahCardoso: For this example, you can but for nubSort `toList . fromList` is used to remove all duplicates from the list
nubSort :: Ord a => [a] -> [a]
nubSort = toList . fromList . sort

balacij: Yep, we can just revert that one to `nub` as it was before this PR. This way, we minimize the changes in the PR and put it in an immediately merge-able state with just the few changes from lists to `Set`s.

After this PR, we can figure out which `nub`s involve a list that should be a `Set` or not, and go from there. However, since `toList . fromList` sorts lists, I think that is also a clue that we can also change this area to using a `Set` (since a `Set` in Haskell is ordered already).

NoahCardoso:
The order of `eNames` does not matter however the order of `eNames'` does in drasil-lang/lib/Language/Drasil/CodeExpr/Extract.hs. I believe the only difference between the two is that the order of `eNames'` matters for printing. Is this important? I think that unless there is a specific reason that the order is important we should try to replace `eNames'` with `eNames`.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -192,4 +192,4 @@ getConstraints cm cs = concat $ mapMaybe (\c -> Map.lookup (c ^. uid) cm) cs
 -- | Get a list of 'CodeChunk's from a constraint.
 constraintvars :: ConstraintCE -> ChunkDB -> [CodeChunk]
 constraintvars (Range _ ri) m =
-  map (codeChunk . varResolve m) $ nub $ eNamesRI ri
+  map (codeChunk . varResolve m) $ nub $ toList $ eNamesRI ri
```
JacquesCarette, line 195: since `eNamesRI` now returns a `Set`, isn't the `nub` redundant?

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
NoahCardoso: Yes it has been removed

balacij:
Sorry, why was this closed? Should we keep the branch or delete it? Was there anything we could salvage @NoahCardoso ?

PR #3764, opened 2024-05-31
Split representation of constructors and methods.

### LINKED ISSUE #3763: Splitting up Constructors and Methods in GOOL's Representation
Currently, in GOOL's representation of classes it stores the constructor in the same list as the other methods.  This is fine in most cases, but there are a few benefits to storing them separately:
- Currently, GOOL prints all methods (including the constructor) in the order they appear in its list.  Constructors should appear at the top of a class's definition, and right now we rely on `drasil-code` to put the constructor at the top of the list of methods.
- With Julia, 'methods' are really just functions that need to be outside the definition of the struct, but constructors are part of the struct, and in some cases must be within the struct definition.  Currently when generating a Julia struct library from a GOOL class, the only way to know that a method is the constructor is to check if its name is the same as the object's name.  Keeping them separate from the beginning would make this possible in a much cleaner way.

I started work on making this change, and it seems pretty straightforward to do.  I'll submit a PR when it's ready, but until then, does this sound like a reasonable change to make?

B-rando1:
Closes #3763
I noted the motivations for this change in the issue.

One thing to note: I used the `unzip` function, and hlint suggested I used `Data.Functor.unzip` instead, but that doesn't seem to be available on our version of Haskell.  I asked @balacij about it, and it sounds like we'll either need to upgrade our Haskell version to match hlint, or achieve the functionality a different way.


balacij:
An alternative way to get HLint running is for us to switch over to https://github.com/haskell-actions/hlint-run and install a specific version of HLint, rather than always using the newest version.

balacij:
You _should_ be able to pull in the `useCorrectHlint` branch (marking this PR as dependent on #3803 if you do so) to make the Lint workflow pass.

code/drasil-code/lib/Language/Drasil/Code/ExtLibImport.hs
```diff
@@ -211,11 +211,18 @@ genArguments _ _ = error argumentMismatch
...
   State ExtLibState (Class, [String])
-genClassInfo o c n desc svs ci cif = let (mis, mifs, f) = genCI ci cif in
-  if length mis /= length mifs then error methodInfoNumberMismatch else do
-    ms <- zipWithM (genMethodInfo o c) mis mifs
-    modify (if any isConstructor mis then id else addDef (new c []) o)
-    return (f desc svs (map fst ms), concatMap snd ms)
+genClassInfo o c n desc svs ci cif = let 
+  (mis, mifs, f) = genCI ci cif 
+  zMs = zip mis mifs
+  (zCtrs, zMths) = partition (\(mi, _) -> isConstructor mi) zMs
```
balacij, line 217: I _vaguely_ remember us talking about this in person, but... is there any way we can split up the constructors from the methods without first having them all considered methods first and sorting them post-facto? IIRC, `isConstructor` checked the name of the method in the context (checking if it is named as the same as the class), and the determination of whether a method is a constructor or not is based on the name of the "Method" being identical or not to the class. That is not robust -- for example, Dart permits named constructors.

B-rando1: That's a good point, and I hadn't thought of that Dart example.  I'll try to take a look soon.  I haven't looked at that code in a while, but I'll see if I can find a better way of keeping them separate from the source.

B-rando1: Ah, after looking at it I remember what `isConstructor` does.  It's defined like so:
https://github.com/JacquesCarette/Drasil/blob/e0a0658f73f026aa5ae875fb9309a542ef444e99/code/drasil-code/lib/Language/Drasil/Code/ExtLibImport.hs#L274-L277

And here's the definition of `MethodInfo`:
https://github.com/JacquesCarette/Drasil/blob/e0a0658f73f026aa5ae875fb9309a542ef444e99/code/drasil-code/lib/Language/Drasil/Code/ExternalLibrary.hs#L84-L87

So while constructors and methods are stored in the same list, they are separated by the datatype holding them.  I think what you might be remembering is a conversation we had before I discovered the `isConstructor` function and was trying to separate them by name.

Now with respect to what you said about allowing constructors to have different names (as Dart allows this), it looks like `drasil-code` and `drasil-gool` each assume that a constructor name can be derived from the class name.
- I'm less sure about `drasil-code`, but that comment above `MethodInfo` seems to suggest that constructors are tied directly to their class rather than to a `CodeFuncChunk`, which I assume gives methods their name.
- In `drasil-gool`, the `constructor` function doesn't take in a name, and iirc it gets the class name from the `buildClass` that it's passed to.

I'm not sure if it's worth adding that functionality now, as none of our current targets support constructors with different names, but it's good to keep in mind.

balacij: Ah! I see what you're getting at. Yes, and Dart is the only language that I know of which _natively_ supports named constructors.

Tangent (and, I don't really know where I'm going with this): We can fake named constructors in the other languages too. That would involve creating `public static` methods that return an instance of the carrying class.

For example, in Java:
```java
public class Point {
    private double x, y;

    private Point(double x, double y) {
        this.x = x;
        this.y = y;
    }

    public static Point cartesian(double x, double y) {
        return new Point(x, y);
    }

    public static Point polar(double radius, double angle) {
        return new Point(radius * Math.cos(angle), radius * Math.sin(angle));
    }
}
```

Now this is 'faked' because instantiation of `Point` no longer involves the `new` keyword (e.g., we would write `Point.polar(x,y)` instead of `new Point.polar(x,y)`). This can also be extended to some of our other languages, like C++ and Swift, relatively straight forwardly.

balacij (dismissed):
Other than some comments needing to be updated (or created, as necessary), this all LGTM!

B-rando1:
I updated and added comments relating to the changes @balacij.

B-rando1:
This PR has been sitting a while.  It's not currently blocking anything, but it would still be good to do something with it.

Are there any other changes we need here?  I know there was some discussion about multiple constructors, but that is covered in #3765.  Other than that, I don't think there are any requested changes that we're waiting on.

As far as I know this is still something we want to do.  It is less necessary due to the decisions we made in #3792, but it still is likely to come in handy for #3765, as well as other projects down the road.

balacij (approved):
Everything looks good to me!

JacquesCarette (approved):

JacquesCarette:
Can you fix the conflicts @B-rando1 please?

JacquesCarette (approved):

PR #3766, opened 2024-06-03
Update Haskell installation script source in GH Workflow

balacij:
`haskell/actions/setup` is deprecated, but there is a maintained fork: `haskell-actions/setup`


balacij:
Remaking this, so that I can test #3769 with my own repo.

PR #3767, opened 2024-06-03
Adding Strict Data

NoahCardoso:
This is a WIP for adding StrictData to Drasil. Contributes #3710


balacij:
If you set the default-extensions field at the top-level (instead of being a field under the `library` section), it will also propagate to the executables -- I'm not sure how much this is needed, but I don't think it would hurt.

PR #3769, opened 2024-06-03
Merge wiki into main repo, and create deployment script

balacij:
As per the title.

I used `git log --format='%aN <%aE>' | sort -u > contributors.txt ` to get and alphabetically sort the list of contributors by name for the commit message.


balacij:
I would say that's a pretty unique commit -- "24 people contributed"!

I just have to make sure that the main CI does not run.

PR #3771, opened 2024-06-04
Deploy wiki from repo

balacij:
As per the title (sorry for the spam), this PR:
* Deploys our public wiki from the repo.
* Moves the wiki contents to a local `./wiki/` folder.
* Retains the git history of our existing wiki.
* Writes commits to `.wiki` repo using SHAs from our git log.
* Overwrites manually-made changes to the wiki with content from the repo -- we'll need to be careful, I suppose. I'm not sure how we can reasonably watch for (manual) changes to the wiki and PR them back in to the main repo. I think this will have to be a caveat to this change.

PR Tests:
* Sanity-checking PRs (good case): https://github.com/balacij/Drasil/actions/runs/9358769164
* Sanity-checking PRs (bad case -- subdirectories of `./wiki/` are not allowed): https://github.com/balacij/Drasil/actions/runs/9358784521
* Deploying: https://github.com/balacij/Drasil/actions/runs/9358631564/job/25760829296

Originally, I was trying to use one of the pre-made actions scripts I linked from #3769, but it seemed to (1) erase git history of the `.wiki.git` repo and (2) not use a 'user' account to commit (I used our corresponding 'bot' accounts to do it for us). So, I just wrote the git commands myself. It took a bit longer than I was hoping for, but whatever, I think it's worth it.

I added everyone who has edited the wiki in any shape or form as a co-author of this PR through 8deb7caf8442b90a1fa56f1926e2d44a378230a4. I used `git log --format='%aN <%aE>' | sort -u > contributors.txt ` to get and alphabetically sort the list of contributors by name for the commit message.

Requirements: it uses the same secret variables as `Build.yaml`, so it should work out of the box, without any extra configuration on your end @JacquesCarette, but if something goes awry with variables, we need to set BOT_NAME, BOT_EMAIL, and BOT_TOKEN.


JacquesCarette (approved):

smiths:
I really this idea @balacij.  This is a much better way to track changes to the wiki.  I agree that we need to let users know to not edit the wiki the old way, since there changes will be lost.  Would the contributor's guide be a good place to say this?  I believe this summer's team read the guide as part of their onboarding.

balacij:
Well, I like that our first change to the wiki will be for this! That sounds like a good idea to me. It might also be good for us to add a small textual warning at the top (or maybe bottom?) of each file, just to remind folks that they should only edit the wiki through the main repo and PR system? Unfortunately, I haven't found any good way of restricting edit access to the wiki.

smiths:
I did wonder whether we could restrict access to editing the wiki.  Glad you looked into it.  I like the idea of updating the contributor's guide as our first test.  A small textual warning in each file sounds like a good idea.  We point out that other files are generated, so we could possibly add a note about which version of Drasil generated the wiki files?

JacquesCarette:
Small warning at the top of each page seems like a good idea. The wiki is as restricted as can be, i.e. only collaborators can edit. Starting with the contributor's guide is indeed a good test.

PR #3772, opened 2024-06-04
Line under module doc comment

### LINKED ISSUE #3732: Generated Python header comment missing empty line beneath
I noticed it in the `ReadTable` file from GlassBR, there's no empty line between the header comment block and the comment block for the first function:

https://github.com/JacquesCarette/Drasil/blob/f34609c5fee391de84fbb112b6bbaf15b0dfffe9/code/stable/glassbr/src/python/ReadTable.py#L1-L10

My guess is that this happens whenever there are no import statements.  I took a quick look at the other targets, and it seems to only happen with the Python target.  As far as I know this is not intended.

B-rando1:
We had discussed that there should be a blank line beneath the module-level doc comment at the top of files.  It also fixes an issue I was having with the generated Julia code, which comes in handy.

This changes stable (175 files in stable, to be precise 😅).  I didn't check every file, but I checked about half of them and they all have exactly one blank space under the module doc comment now.

The one thing I just wanted to double-check was, are there any cases where that extra line would break Doxygen (or whatever documentation generator the target uses)?  I ran `make doxygen` and it ran fine, but I'm not too familiar with Doxygen so I thought it would be smart to ask.  @bmaclach if you get a chance would you be able to weigh-in on DocC?  Thanks.

Closes #3732


JacquesCarette (approved):

smiths:
Yes, I agree with @samm82.  We'll be looking for examples like this in the future and it would be nice to be able to find them.  There aren't any labels for pull requests, so we can't just label the PR.  We could label the issue, but the connection between an issue and a PR isn't always one to one.  My suggestion is that we create a wiki page where we list our examples.  The list can start with the current example, but there will be more.  :smile:  With the wiki we can have a link directly to the PR.  We can also have an explanation of why it is a good example.  @BilalM04 can you please create a wiki page?  With the changes that @balacij has made you shouldn't edit the wiki directly.  You will need to edit the files in the wiki folder in a branch and do a pull request.

samm82:
@smiths We can label PRs (I just labelled this one <kbd>[enhancement](https://github.com/JacquesCarette/Drasil/labels/enhancement)</kbd>)! Although I think a wiki with more details would also be a good idea

smiths:
Thank you @samm82.  I didn't look hard enough to find the label field.  :smile:  We should definitely do that, and enhancement is a good label.  I think the wiki is still a good idea to record the explanation of why the enhancement is so efficient, since that isn't always going to be highlighted in the PR notes.

bmaclach:
@B-rando1 I don't have any insight about whether the blank line would break DocC. I never did use DocC outside of the context of GOOL, so I don't actually know much more about it than you do, at this point.

B-rando1:
No worries at all @bmaclach, thanks for getting back to us 🙂!

I'll take a quick look around tomorrow, and if everything looks good we should be able to merge after that.

BilalM04:
@smiths which section in the [sidebar](https://github.com/JacquesCarette/Drasil/blob/master/wiki/_Sidebar.md) should I put the wiki page for "Power of Drasil" under?

B-rando1:
I think we should be ready to go ahead with the merge.  I generated the Doxygen files again, and I was able to find the information from the header at the top, which means that Doxygen was able to find it.  With Swift I can't be as confident, as none of the information online gives examples of code.  I asked Copilot about it though, and it seems pretty confident that a blank line is ok.

@smiths do you think we're good to merge?

smiths:
> Yes, I agree with @samm82. We'll be looking for examples like this in the future and it would be nice to be able to find them. There aren't any labels for pull requests, so we can't just label the PR. We could label the issue, but the connection between an issue and a PR isn't always one to one. My suggestion is that we create a wiki page where we list our examples. The list can start with the current example, but there will be more. 😄 With the wiki we can have a link directly to the PR. We can also have an explanation of why it is a good example. @BilalM04 can you please create a wiki page? With the changes that @balacij has made you shouldn't edit the wiki directly. You will need to edit the files in the wiki folder in a branch and do a pull request.

I inadvertently asked @BilalM04 to create the wiki entry when @B-rando1 has been the one working on this issue.  The autocomplete (and my lack of proof reading) is to blame for this mistake.  Maybe @BilalM04 and @B-rando1 can work together on this?  Maybe @BilalM04 could create the page in the wiki and @B-rando1 could add the content describing this specific example.  The main thing is to start this important piece of infrastructure so that when we are writing we can easily find great examples.  :smile:

smiths:
> sidebar

We are going to have to redesign the wiki, but for right now, you can put this topic under Misc.

smiths:
Yes, we are good to merge.  @BilalM04 and @B-rando1 if we are at risk of not updating the wiki, please create an issue to remind us.  :smile:

B-rando1:
Sounds good.  I'll create an issue so that we make sure to remember.

BilalM04:
Sounds good! I've made the additions in this [branch](https://github.com/JacquesCarette/Drasil/tree/wikiDrasilPower) and was just about to create a PR. I'll link the new issue to the PR when @B-rando1 makes it.

B-rando1:
I created the issue!  @BilalM04 it's #3778.

PR #3777, opened 2024-06-05
Updated Glassbr Introduction #2299

### LINKED ISSUE #2299: Update GlassBR SRS Introduction
The GlassBR SRS introduction reads poorly, so it would be nice to clean it up to read like a proper introduction. It assumes the reader already knows what the system is supposed to do / why it is there.

NoahCardoso:
Closes #2299 

>The GlassBR SRS introduction reads poorly, so it would be nice to clean it up to read like a proper introduction. It assumes the reader already knows what the system is supposed to do / why it is there.

I made some improvements to the glassbr introduction from #2299
image

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
@@ -197,12 +197,15 @@ priorityNFReqs = [correctness, verifiability, understandability,
 
 startIntro :: (NamedIdea n) => n -> Sentence -> CI -> Sentence
 startIntro prgm _ progName = foldlSent [
-  atStart' explosion, S "in downtown areas are dangerous from the", phrase blast, S "itself" `S.and_`
-  S "also potentially from the secondary effect of falling glass. Therefore" `sC` phrase prgm, 
-  S "is needed to predict whether a",phrase glaSlab, S "can withstand a", phrase blast, S "under given conditions. For",
-  S "example" `sC` S "we might wish to know whether a pane of", phrase glass, S "from a gas main", phrase explosion,
-  S "from a small fertilizer truck bomb. The program documented here is called", 
-  short progName]
+  atStart' explosion, S "in downtown areas are dangerous from the", 
+  phrase blast +:+ S "itself" `S.and_` S "also potentially from the secondary" 
+  +:+ S "effect of falling glass. Therefore" `sC` phrase prgm, S "is needed to" 
+  +:+ S "predict whether a", phrase glaSlab, S "can withstand a", phrase blast, 
+  S "under given conditions.", S "For example" `sC` 
+  S "we might wish to know whether a pane of", phrase glass, S "from a gas main",
+  phrase explosion, S "from a small fertilizer truck bomb." +:+ 
+  S"The program documented here is called", short progName]
+
 undIR, appStanddIR :: [Sentence]
 undIR = [phrase scndYrCalculus, phrase structuralMechanics, phrase glBreakage,
   phrase blastRisk, pluralNP (computerApp `in_PS` Edu.civilEng)]
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
@@ -196,13 +196,13 @@ priorityNFReqs = [correctness, verifiability, understandability,
 {--INTRODUCTION--}
 
 startIntro :: (NamedIdea n) => n -> Sentence -> CI -> Sentence
-startIntro prgm sfwrPredicts progName = foldlSent [
-  atStart prgm, S "is helpful to efficiently" `S.and_` S "correctly predict the"
-  +:+. sfwrPredicts, underConsidertn blast,
-  atStartNP (the prgm) `sC` S "herein called", short progName `sC`
-  S "aims to predict the", sfwrPredicts, S "using an intuitive",
-  phrase interface]
-
+startIntro prgm _ progName = foldlSent [
+  atStart' explosion, S "in downtown areas are dangerous from the", phrase blast, S "itself" `S.and_`
+  S "also potentially from the secondary effect of falling glass. Therefore" `sC` phrase prgm, 
+  S "is needed to predict whether a",phrase glaSlab, S "can withstand a", phrase blast, S "under given conditions. For",
+  S "example" `sC` S "we might wish to know whether a pane of", phrase glass, S "from a gas main", phrase explosion,
+  S "from a small fertilizer truck bomb. The program documented here is called", 
+  short progName]
 undIR, appStanddIR :: [Sentence]
 undIR = [phrase scndYrCalculus, phrase structuralMechanics, phrase glBreakage,
   phrase blastRisk, pluralNP (computerApp `in_PS` Edu.civilEng)]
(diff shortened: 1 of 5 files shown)


code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
```diff
@@ -196,13 +196,13 @@ priorityNFReqs = [correctness, verifiability, understandability,
...
-  +:+. sfwrPredicts, underConsidertn blast,
-  atStartNP (the prgm) `sC` S "herein called", short progName `sC`
-  S "aims to predict the", sfwrPredicts, S "using an intuitive",
-  phrase interface]
-
+startIntro prgm _ progName = foldlSent [
+  atStart' explosion, S "in downtown areas are dangerous from the", phrase blast, S "itself" `S.and_`
+  S "also potentially from the secondary effect of falling glass. Therefore" `sC` phrase prgm, 
+  S "is needed to predict whether a",phrase glaSlab, S "can withstand a", phrase blast, S "under given conditions. For",
+  S "example" `sC` S "we might wish to know whether a pane of", phrase glass, S "from a gas main", phrase explosion,
```
samm82, line 203:
```suggestion
  S "is needed to predict whether a", phrase glaSlab, S "can withstand a", phrase blast +:+. S "under given conditions"
  S "For example" `sC` S "we might wish to know whether a pane of", phrase glass, S "from a gas main", phrase explosion,
```

It'd also be nice to keep the line length under 80, if possible (no idea what it actually is; it might already be good!)

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
```diff
@@ -552,7 +552,7 @@ <h2>Abbreviations and Acronyms</h2>
       <div class="section">
         <h1>Introduction</h1>
         <p class="paragraph">
-          Software is helpful to efficiently and correctly predict the blast risk involved with the glass slab. The blast under consideration is any kind of man-made explosion. The software, herein called GlassBR, aims to predict the blast risk involved with the glass slab using an intuitive interface.
+          Explosion in downtown areas are dangerous from the blast itself and also potentially from the secondary effect of falling glass. Therefore, software is needed to predict whether a glass slab can withstand a blast under given conditions. For example, we might wish to know whether a pane of glass from a gas main explosion from a small fertilizer truck bomb. The program documented here is called GlassBR.
```
samm82, line 555: Should be "Explosion**s**"; not sure what variant of `atStart` would do this (if any), so you might need to look into `plural` as well.

Also should be "…whether a pane of glass **fails** from…"

Is the gas main explosion caused by the small fertilizer truck bomb here, or are they two separate examples (and thus should be separated by "or")? @smiths

samm82 (changes requested):
Some minor things, but overall looks good!

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
JacquesCarette: It does seem to make it uppercase - maybe `explosion` is done using a phrase constructor that prevents pluralization?

JacquesCarette (approved):
This is a "conditional approval" in the sense that I too want `Explosions`, but that's already covered, and there are no other changes that I want to see.

smiths (changes requested):
Before merging we need to sort out the comment on explosion from @samm82.  I'd also like to see us use the Purpose information, or if that is too big a change, create a new issue (or discussion if that is more appropriate) to look at doing this.

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
balacij: @NoahCardoso and I looked at it together just now. It looks like the majority of the "words" defined in GlassBR's `Unitals.hs` use `nounPhraseSP` (i.e., "noun phrase that is self-plural"). Our first idea was to use `nounPhrase` to manually add the `s` ourselves, but that doesn't seem to be too commonly used, and when it is used, it's used for a complex phrase. The same issue occurs with other examples, such as Projectile and PD Controller. It's not quite clear how we should proceed -- is there a 'right' way to go about this? Is `nounPhrase` just what we want to use then?

balacij: RE: the local README.md: it looks to be largely out of date and duplicates information that should exist in the Haddock docs. Shall we remove the out of date information, without replacement?

balacij: Ah, it looks like definition of `explosion` should use `cn'` to make an `NP`. A smidge bit hidden.

NoahCardoso:
>In re-reading my https://github.com/JacquesCarette/Drasil/issues/2299#issuecomment-2146534772, I see that some of what I wrote was lost because I put it in angle brackets. I didn't really think about it, but now I think I was telling markdown that these were links. The text that didn't appear was the following proposed pattern for introductions:

@smiths I am not really sure what you mean by this.

>I know we don't have all the fields of information I mentioned when I https://github.com/JacquesCarette/Drasil/issues/2299#issuecomment-2146534772, but I tried to use the Purpose that we already have (I believe) for GlassBR. Can we use the Purpose in the introduction? This would let us use the same text in multiple places. The purpose really should be the same everywhere it appears.

In your comment, I believe the purpose statement is "Therefore, software is needed to predict whether a glass slab can withstand a blast under given conditions." I am not sure how this can be done by reusing the purpose text. Also I am not 100% sure what the purpose I am to reuse is.

image
When you mention to reuse the purpose are you referring to this?

samm82 (approved):
"Explosions" 😂

samm82:
@NoahCardoso You can find the "purpose" statement for GlassBR in the code permalinked below; hopefully it's trivial to reuse this in the introduction!

https://github.com/JacquesCarette/Drasil/blob/ef35006748d969442cdfaaf37f731d0d5f4e8bba/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs#L126-L128

(My approval is likewise a conditional one, based on this either being done, or an issue created for doing it in the future if it's not trivial.)

smiths:
Thank you @samm82 that is exactly the purpose I was thinking about.  @NoahCardoso hopefully  it is relatively easy to reuse the existing purpose data.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
```diff
@@ -200,10 +200,9 @@ startIntro prgm _ progName = foldlSent [
   atStart' explosion, S "in downtown areas are dangerous from the", 
   phrase blast +:+ S "itself" `S.and_` S "also potentially from the secondary" 
   +:+ S "effect of falling glass. Therefore" `sC` phrase prgm, S "is needed to" 
-  +:+ S "predict whether a", phrase glaSlab, S "can withstand a", phrase blast, 
-  S "under given conditions.", S "For example" `sC` 
-  S "we might wish to know whether a pane of", phrase glass, S "fails from a gas main",
-  phrase explosion `S.or_` S "from a small fertilizer truck bomb." +:+ 
+  +:+ purp :+: S ". For example" `sC` S "we might wish to know whether a pane of",
```
samm82, line 203: This _should_ work

```suggestion
  +:+. purp, S "For example" `sC` S "we might wish to know whether a pane of",
```

code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
```diff
@@ -200,10 +200,9 @@ startIntro prgm _ progName = foldlSent [
...
+  phrase glass, S "fails from a gas main", phrase explosion `S.or_` 
+  S "from a small fertilizer truck bomb." +:+
   S"The program documented here is called", short progName]
```
samm82, line 206:
```suggestion
  S "The program documented here is called", short progName]
```

samm82 (changes requested):

smiths (approved):

smiths:
The changes look good to me, as long as @samm82 is satisfied.

samm82 (approved):
Awesome! 🥳🎉

PR #3786, opened 2024-06-07
Added documentation for the `objMethodCall` functions.

### LINKED ISSUE #3785: Odd behaviour when printing results of methods
I was working on implementing method calls in Julia when I noticed some odd behaviour.

I wrote the following GOOL code:
```haskell
objTest :: (OOProg r) => MSBlock r
objTest = block [
  varDecDef s newSimpleClass,
  printLn $ valueOf s,                                                        -- Print s directly
  varDecDef x $ objMethodCallNoParams simpleClassType (valueOf s) "getX",     -- Store s.getX() in x
  printLn $ valueOf x,                                                        -- Print x
  valStmt $ objMethodCall simpleClassType (valueOf s) "setX" [litInt 2],
  printLn $ objMethodCallNoParams simpleClassType (valueOf s) "getX",
  valStmt $ objMethodCallNoParams simpleClassType (valueOf s) "printXMethod",
  printLn $ objMethodCallNoParams simpleClassType (valueOf s) "getX"]         -- Print s.getX()
```

This translates to the following Python code:
```python
s = SimpleData.SimpleData()
print("Instance of SimpleData object") # Print s directly - this is expected behaviour
x = s.getX()                           # Store s.getX() in x - this is expected behaviour
print(x)                               # Print x - this is expected behaviour
s.setX(2)
print("Instance of SimpleData object")
s.printXMethod()
print("Instance of SimpleData object") # print s.getX() - this is weird
```

You can see that on the last line, instead of printing the return result of `s.getX()`, it prints `Instance of SimpleData object`, which is only supposed to happen when you're printing objects.

This happens regardless of target.  Tracing the problem, the value passed to `printLn` is eventually passed to the `print` function from LanguagePolymorphic:
https://github.com/JacquesCarette/Drasil/blob/85cf74daaccb74131373df9d52cc2bc751fd48e9/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/LanguagePolymorphic.hs#L355-L365
On line `357` we get the type of `v`, and on line `360`, if the type is `Object n` then we use the `printObj` function, which gives the standard string description of the object.

From all this, it seems that the bit on line `357` where it calls `zoom lensMStoVS v >>= print' . getType . valueType` seems to be where it goes wrong.  For some reason, either `valueType` or `getType` see the type of `s.getX()` as `Object "SimpleData"` rather than `Int`.  

These functions both read fields from records defined in `AST.hs`.  At this point it's not super clear to me where and how these fields are set.  I'll keep looking, but if anyone else knows, that would be helpful!

B-rando1:
I also corrected a comment for the `call` function from `LanguagePolymorphic`

Lastly, I added the rest of the `objMethodCall` functions to the export list of `Drasil.hs` (previously only `objCallMixedArgs` was exported).

Closes #3785


code/drasil-gool/lib/GOOL/Drasil/ClassInterface.hs
```diff
@@ -300,21 +300,31 @@ exists :: (ValueExpression r) => SValue r -> SValue r
...
   -> NamedArgs r -> SValue r
 objMethodCallNamedArgs t o f = objMethodCallMixedArgs' f t o []
 
+-- | Function for calling an object method with a mix of positional
+--   and named arguments.
 objMethodCallMixedArgs :: (InternalValueExp r) => VSType r -> SValue r -> Label 
   -> [SValue r] -> NamedArgs r -> SValue r
 objMethodCallMixedArgs t o f = objMethodCallMixedArgs' f t o
 
+-- | Function for calling an object method with no parameters.
```
balacij, line 327: I think the filename indicates that much of its content will be "utilities"-like already, so I think you can replace "Function for X" with just "X".

Is the "object" part of this needed? I think methods are always tied to objects. I _think_ Java's `public static` methods are more like functions than methods -- do we want to access them from these functions too?

B-rando1: Ah, good point.  I think I noticed somewhere that there's a different function for calling static methods, so it is only for 'true' methods.

bmaclach (approved):

balacij (approved):
Yep, LGTM!

PR #3705, opened 2024-03-26
Add Well-Understood paper to front page description

daijingz:
@samm82 Wait a moment for the new updates. 
Inherited from previous PR #3686, but with some changes. Please wait for some further checks.

Closes #3360 

Newest inspection: Coding style requirements.
- [x] Double spaces of indent levels.
- [x] Camel case variable names
- [x] Correct meaningful variable names.
- [x] Single blank lines between top-level objects.

PR #3793, opened 2024-06-11
Wiki Workflow: Ensure all files have windows-friendly names

balacij:
@daijingz is exhibiting an issue with #3705 regarding a "bad path name". He is using Windows and Git Desktop. I believe the "bad path name" is a result of Windows' file system having [more restrictions](https://stackoverflow.com/a/31976060/16760741) on what valid file names are in comparison to Linux and Darwin. In other words, the current repo might be unusable for Windows users who _don't_ use Linux on Windows (i.e., use Windows-native-build tooling).

This PR tries to:
1. Fix existing issues (just one file name), and
2. Restrict us from at least few common edge cases involving bad file names.

I don't think it is necessary to go all out and make sure that all file paths and names are valid in the git repo -- this PR does just the bare minimum.

@daijingz if you could please try out this branch and report if it works, that would be appreciated!


balacij:
Ok, @B-rando1 and @samm82 tested this for me (testing breakage + fix) -- thank you guys again! So this fixes the path issue related to Windows machines cloning our repo.

JacquesCarette (approved):

JacquesCarette:
There's a merge conflict to resolve first.

balacij:
Good to go now!

Issue #3781, opened 2024-06-06
Adding to `Choices` for new input patterns


B-rando1:
As discussed in #3752, we want to add new input patterns to Drasil.  From looking at drasil-code it seems that we need to change `Choices` in order to allow users to express the input patterns they would like.  I have a design, but it has flaws so I would like some advice.

## Changes to `Choices`
`Choices` already has a couple of relevant properties:
- `architecture :: Architecture` contains information about whether the parsing and validation functions are in the same class or in separate classes.
- `dataInfo :: DataInfo` contains information about whether the input is bundled in a class or contained in separate variables.

I created a new datatype `data LogicLoc = InCons | InMthd | InFunc`.  I then added two fields of this type to `Choices`: `parseLoc` and `validateLoc`.  (I'm not too sure about any of these names, feel free to suggest better ones 😄).  The idea is that we can specify whether parsing and validation are done directly in the constructor, in a method to be called by the constructor, or in an external function.

## Notes
- We still don't have the ability to specify "I want parsing and validation in the same function/method, with each 'piece' of validation being performed as early in the parsing process as possible.
- If input parameters are `Unbundled`, then only the `InFunc` option for parsing and validation makes sense.
- So there are some patterns that we want to express but can't, and some patterns that we can express but shouldn't be able to.

It's possible I'm overanalyzing this, but it just seems like a better design of `Choices` would lead to fewer edge-cases to deal with.

B-rando1:
@smiths had a good idea in related discussion #3780.  Rather than giving the user full control over every aspect of the code, we could create a curated list of decisions that fit well together, and just offer those.  This would simplify the structure of `Choices`, and would also remove the chance of having a lot of edge cases.

B-rando1:
One change to `Choices` that I think makes sense is getting rid of the part of `Architecture` that deals with separate vs combined input methods.  @smiths mentioned in #3752 that parsing and validation were originally separated in order to separate the 'secrets' of input format and input constraints.  He went on to say that they later decided that since parsing and validation are only ever done together, they might as well be in the same module with the meta-secret of "how to provide valid input parameters".  I agree with this, and I also think there's an argument that Parnas' idea of a "Module" doesn't have to refer to files - it could refer to functions or methods that hold a secret.

The other reason is that the currently available design choices that involve `Separate` (e.g. GlassBR) generally have more issues with them.

We should probably wait to make changes until we have an overall direction, but I just thought I should mention this piece here.

[...]

PR #3642, opened 2023-08-17
Extract duplicate strings into `InternalConcept`s

hrzhuang:
This is a continuation of #2966. I brought it up to date with `master`, but do the changes look good to merge in? If not, what are the outstanding issues?

Closes #2843 
Closes #2844 

Builds on #3795

PR #3795, opened 2024-06-11
Removed `InputModule` from `Choices`

B-rando1:
We had discussed a few times that there isn't much need to put the different pieces related to input (parsing, validation, and storage) in different modules.  Since the `Modular Separated` pattern is not very useful, and in the case of `Bundled` inputs uses some patterns that are not good form (creating an empty object, then populating its members externally), I suggested in #3781 that we remove `InputModule` altogether and assume that all the input functionality is in the same module.

I don't believe we ever got around to discussing this, so we might have to make sure we want to do this.  I talked with @balacij and he agrees that this is a good change to be making.

Notable changes:
- Removed `InputModule` from `Choices` in [Choices.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-code/lib/Language/Drasil/Choices.hs)
- Renamed `genInputModCombined` to `genInputMod` in [Modules.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs)
- Changed label of modularity from `'C'` for `Combined` and `'S'` for `Separated` to `'M'` for `Modular` in a few places, e.g. in [Choices.hs](https://github.com/JacquesCarette/Drasil/blob/master/code/drasil-example/projectile/lib/Drasil/Projectile/Choices.hs) from Projectile.
    - This last change caused a large number of changes in `stable`.
- Changed any of the examples which previously used `Modular Separated` to use `Modular`.

Because of the large number of things this change affects I think it's good that we be thorough in double-checking the changes, but if we can reach a consensus soon on whether these are good changes to be making, that would help me with the rest of the changes to `Choices` going forward.

Contributes to #3781


code/drasil-code/lib/Language/Drasil/Code/Imperative/Parameters.hs
```diff
@@ -45,17 +44,12 @@ getInConstructorParams = do
...
 getInputFormatIns :: GenState [CodeVarChunk]
 getInputFormatIns = do
-  g <- get
-  let getIns :: Structure -> InputModule -> [CodeVarChunk]
-      getIns Bundled Separated = [quantvar inParams]
-      getIns _ _ = []
-  getParams "get_input" In $ quantvar inFileName : getIns (inStruct g) (inMod g)
+  let getIns :: [CodeVarChunk]
+      getIns = []
+  getParams "get_input" In $ quantvar inFileName : getIns
```
samm82, line 52: Is there a reason we can't do this? Or this is "scaffolding" for a later change?
```suggestion
  getParams "get_input" In $ quantvar inFileName : []
```

B-rando1: Oh true, that's from the fix I made a minute ago.  I tried simplifying that more, but Haskell couldn't figure out its type.  There might be a better way to do it than where I left it, though.  I'll take a look.

samm82: I think you can do `([] :: [CodeVarChunk])` to specify the type of an empty list

B-rando1: Turns out I was just messing up the syntax, and we don't even need the `[]`!

JacquesCarette (approved):

balacij (approved):

smiths (approved):
The changes make sense to me.  I looked through the generated code in stable.  I didn't run the code to verify that it works.  It is too bad that we can't currently run test cases on our generated code.  We've talked about doing this before.  @samm82 is getting us in the right direction to do this in the future.

JacquesCarette:
Now there's a conflict that needs fixed before it can be merged.

B-rando1:
This branch was merged as a part of #3642.

PR #3810, opened 2024-06-18
Add support for 1-indexed languages

### LINKED ISSUE #3746: Adding Support for 1-Indexed Arrays
Arrays in both Julia and MATLAB are 1-indexed.  We'll need some way to account for this when translating GOOL to these languages.  As far as I can tell, we have to assume that GOOL's arrays are 0-indexed, since the line of GOOL code 
```haskell
varDecDef (var "e" double) (listAccess (valueOf $ var "a"
    (listType double)) (litInt 1))
```
currently translates to the following line of Python code:
```python
e = a[1]
```
which means that in Julia it should translate to
```julia
e = a[2]
```

If we are assuming that GOOL's arrays are 0-indexed, we'll need to convert indices to 1-indexed (by adding or subtracting 1 in various cases) at some point when generating code for these targets.  I thought of various ways that we could do this, but the only feasible way I could find is to assume that all indices in the generated code are 0-indexed, until they are used with the List API, and only then do we convert.  For example, GOOL code producing the Python code
```python
l = ["a", "b", "c"]
i = 2
print(l[i])
```
would translate to the following Julia code:
```julia
l = ["a", "b", "c"]
i = 2
println(l[i + 1])
```
This isn't always the most idiomatic way to do it (e.g. above setting `i = 3` would have been better), but making the conversion at any other point leads to difficulties because we'd have to keep track of whether or not a variable is storing an index, etc.

I've started implementing it this way for Julia, and I'm trying to think how to do it in a way that is general enough that adding new methods to the List API, as well as adding new languages that are 1-indexed, won't require as much care with when to add or subtract `1`.  Here's a design I've thought of so far:
- Add two new functions to the `InternalListFunc` typeclass: `varToIndexFunc :: SValue r -> SValue r` and `indexToVarFunc :: SValue r -> SValue r`, which would add and subtract `1`, respectively, to the index given for 1-indexed targets, and return the index given for 0-indexed targets.
- Call these two functions within some general functions that are either implemented in every target language, or at least most of them.  One way to do this would be to call them on the index given for the functions defined in `LanguagePolymorphic.hs`, such as `ListAccess` in [LanguagePolymorphic.hs](https://github.com/JacquesCarette/Drasil/blob/21ff3588595f34c3135af6e6db538e16d5318767/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/LanguagePolymorphic.hs#L280-L288).

What are your thoughts?  Does this  seem like a good direction, or do we need to think about this some more?

B-rando1:
I added functions to the `List` type class for converting between integers and indices, the idea being that integers are always 0-indexed, but indices match the language's semantics.  I then integrated these functions into the default functions supplied by `LanguagePolymorphic.hs`, `CommonPseudoOO.hs`, and `Macros.hs`.  This makes it very easy for most if not all of list indexing to automatically switch for a language just by changing which function it uses for `intToIndex` and `indexToInt`.  Most languages have one or two functions that they implement on their own, which I did not update to use these new functions, but anything that is shared between languages supports them.

I know in the [issue](https://github.com/JacquesCarette/Drasil/issues/3746#issuecomment-2127853787) @JacquesCarette had said it might be better to have the functions outside the type classes, and give the type classes an `IndexingScheme` to give this function.  To be honest I tried that, and had a lot of trouble with Haskell's types 😅 .  I'm not sure if this is a big deal, but if it is I can definitely give it another shot, asking for help this time.

The one major benefit of @JacquesCarette's implementation, is that it allows us to shift the index in other ways.  The concrete example is list slicing.  Currently with some languages a list slice involves something like:
```java
for (int i = start; i < end; i++) {
```
Which, if converted to 1-indexed lists, would look like:
```java
for (int i = start + 1; i < end + 1; i++) {
```
When this would be more idiomatic:
```java
for (int i = start + 1; i <= end; i++) {
```
It's not a big deal, and there are likely not many similar cases, but it does show one way that we  could benefit from the other implementation.

Builds on #3808
Closes #3746

P.S. I know we had talked about constant-folding in the list API, so I'm not sure if I should say 'Closes' or 'Contributes to'.  I'd rather move on to other things for now, and reopen the issue if when we get the Julia renderer working, we see that constant-folding would be beneficial for the examples we're generating.

### DIFF
--- a/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
+++ b/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
@@ -1,15 +1,15 @@
 -- | Implementations defined here are valid in some, but not all, language renderers
-module GOOL.Drasil.LanguageRenderer.CommonPseudoOO (int, constructor, doxFunc, 
-  doxClass, doxMod, docMod', extVar, classVar, objVarSelf, indexOf, listAddFunc, 
-  discardFileLine, intClass, funcType, buildModule, arrayType, pi, printSt, 
-  arrayDec, arrayDecDef, openFileA, forEach, docMain, mainFunction, 
-  buildModule', call', listSizeFunc, listAccessFunc', string, constDecDef, 
-  docInOutFunc, bindingError, extFuncAppMixedArgs, notNull, listDecDef, 
-  destructorError, stateVarDef, constVar, litArray, listSetFunc, extraClass, 
-  listAccessFunc, doubleRender, double, openFileR, openFileW, stateVar, self, 
-  multiAssign, multiReturn, listDec, funcDecDef, inOutCall, forLoopError, 
-  mainBody, inOutFunc, docInOutFunc', boolRender, bool, floatRender, float, stringRender', 
-  string', inherit, implements
+module GOOL.Drasil.LanguageRenderer.CommonPseudoOO (int, constructor, doxFunc,
+  doxClass, doxMod, docMod', functionDoc, extVar, classVar, objVarSelf, indexOf, listAddFunc,
+  discardFileLine, intClass, funcType, buildModule, arrayType, pi, printSt,
+  arrayDec, arrayDecDef, openFileA, forEach, docMain, mainFunction,
+  buildModule', call', listSizeFunc, listAccessFunc', string, constDecDef,
+  docInOutFunc, bindingError, extFuncAppMixedArgs, notNull, listDecDef,
+  destructorError, stateVarDef, constVar, litArray, listSetFunc, extraClass,
+  listAccessFunc, doubleRender, double, openFileR, openFileW, stateVar, self,
+  multiAssign, multiReturn, listDec, funcDecDef, inOutCall, forLoopError,
+  mainBody, inOutFunc, docInOutFunc', boolRender, bool, floatRender, float, stringRender',
+  string', inherit, implements, listSize, listAdd, listAppend
 ) where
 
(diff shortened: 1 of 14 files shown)


code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
```diff
@@ -474,4 +485,27 @@ docCommandInit = "- "
...
+listSize :: (RenderSym r) => SValue r -> SValue r
+listSize l = do
+  f <- S.listSizeFunc l
+  mkVal (RC.functionType f) (RC.function f)
+
+-- Julia and MATLAB --
+listAdd :: (RenderSym r) => SValue r -> SValue r -> SValue r -> SValue r
+listAdd l i v = do
+  f <- S.listAddFunc l (S.intToIndex i) v
+  mkVal (RC.functionType f) (RC.function f)
```
balacij, line 500: Continuing the theme of your comments, it would be great to also add to comments to these ones :smile:

balacij (approved):
LGTM! Though, I'm probably not the best person to be reviewing this one :slightly_smiling_face:

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
B-rando1: Good point 😄 I'll take a look and add comments where I can

JacquesCarette (approved):
I was rather hoping that some output of one test case would change because of this. Ah well -- as far as I can tell, all this code is great.

Issue #3815, opened 2024-06-22
Drasil Research Group Meeting, Wednesday July 3, 11am, ITB/225 + online


JacquesCarette:
Thanks for your weekly updates @BilalM04 @B-rando1 @NoahCardoso . Efficiently done. If you could do the same on Friday 28th (here), that would we useful. I'll create an agenda when that becomes clearer.

smiths:
@JacquesCarette you might want to add a discussion of creating a GitHub Organization that Drasil can be a member of.  From what @BilalM04 said in #3811 this would allow contributors to create their own GitHub projects.

B-rando1:
For my weekly summary:

#### Summary
I spent most of the week working on the Julia renderer.  Progress is going well, and we can now generate Julia code for Projectile with most `Unbundled` options.  As we discussed in #3811, I am hoping to have it working with `Unmodular Bundled` by Wednesday, so that we can take another look at classes in Julia and come to a better consensus about whether it's worthwhile.
- I also ran into some issues involving scope, and created an issue about it in #3821.

I also worked on #3809.  That has been going well, and has been a fun side-project when I need a break from the main projects.  I don't think there's anything there that we need to discuss.

#### Blockers/Discussion Points
- I'd like to discuss #3792 some more.  The team seems a bit divided over this at the moment, so I think it would be helpful to talk it out.  Everyone is in favour of eventually adding `struct` support to GOOL and using those to generate the 'best' Julia code, but we don't have a consensus about classes.  To summarize where I believe everyone is at (if I misrepresent you, feel free to correct me):
    - @balacij and @smiths have expressed that it would be nice to be able to generate as many current examples as possible in Julia, even if the code isn't perfectly idiomatic.  They seem optimistic that converting classes to structs should be possible, and that the wrinkles we'll have to work out will be worth it.
    - @JacquesCarette has said that while the conversion from classes to structs seems logical, issues will continue coming up and we will end up needing too many hacks in order to make it work.
    - I definitely see both sides.  I like the idea of translating classes to structs, as well as making the Julia renderer as compatible with existing Drasil as possible. I also wouldn't say I've run into anything so far that's a complete dealbreaker for making it work.  At the same time, I have seen a good number of non-trivial issues spring up as I tried to make it work, such as #3752, #3789, and #3824.  I'm good to keep going if that's what we want, but I also understand if we'd rather cut our losses.

[...]

Issue #2655, opened 2021-07-05
Add back `OneOf` Value constraints


balacij:
Related to: https://github.com/JacquesCarette/Drasil/issues/2646#issuecomment-871564379 , https://github.com/JacquesCarette/Drasil/issues/2646#issuecomment-873109412 , and #2118 

> The fact that removing those constraints does not alter stable feels like a bug. That information should be used, but apparently it's not.
> So I think the pragmatic thing to do is to remove `Enumerated` but create two high-priority issues to bring the information thus cut out back. Note that a first hack of bringing the information back could be display-only.  #2118 is hard because it tries to deal with code generation, and that design is highly non-trivial. So the high-priority would be to get things back at least for display.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/issues/2646#issuecomment-873109412_

balacij:
As a reminder, this ticket is about building constraints on quantities, specifically 'this quantity must be one of these values' constraints.

@NoahCardoso To re-iterate our intended work:
1. You have prototyped (are prototyping?) what the 'generated' code _should_ look like for constraints that bind variables to one of many possible values. After this work, we can dissect what is going on and what we really want to be saying (and having generated for us) when we define these constraints.

As of right now, we decided that this should involve sets in the generated code, but we can come back to it later to make sure.

As such,
2. We have a critical piece missing: sets in `*Expr`. We will need to bridge this gap in all `Expr`s and in `drasil-gool` (@B-rando1 can assist with GOOL :slightly_smiling_face:).
3. We can finally define the constraint as something that 'means' `x \in {... , values, ...}`.

NoahCardoso:
Here are the prototyped examples in Python and Java and how I think they should look like
![image](https://github.com/JacquesCarette/Drasil/assets/106560551/3a7553dc-b885-4a01-950f-034c72c1f63f)
![image](https://github.com/JacquesCarette/Drasil/assets/106560551/fb27e477-84ce-44c1-870b-13d5cd038067)

JacquesCarette:
It all looks good - except that enumerated floats are a bad idea. Once in a while, things will go wrong. I would much rather use integers (i.e. multiply everything by 10) for the internal storage of these numbers.

NoahCardoso:
[...]

PR #3818, opened 2024-06-24
Making Glassbr compatible with Strict Data

NoahCardoso:
In `glassbr.unitals` the ConstrainedChunk `nomThick` is self-referential creating an infinite loop when used with strict data. To prevent this, I created a new function called `nomThick'` that gives the `displayDblConstrntsAsSet` function in `nomThick` the information it needs to properly display the nominal thickness constraints as a set.

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
@@ -104,15 +104,9 @@ standOffDist = uq (constrained' (uc sD (variable "SD") Real metre)
   [ gtZeroConstr,
     sfwrc $ Bounded (Inc, sy sdMin) (Inc, sy sdMax)] (exactDbl 45)) defaultUncrt
 
-nomThick' :: ConstrainedChunk
-nomThick' = cuc "nomThick" 
-  (nounPhraseSent $ S "nominal thickness")
-  lT millimetre Rational 
-  [] $ exactDbl 0
-
 nomThick = cuc "nomThick" 
   (nounPhraseSent $ S "nominal thickness" +:+ displayDblConstrntsAsSet 
-    nomThick' nominalThicknesses)
+    (mkQuant "nomThick" (nounPhraseSent $ S "nominal thickness") lT Rational Nothing Nothing) nominalThicknesses)
   lT millimetre {-Discrete nominalThicknesses, but not implemented-} Rational 
   [{- TODO: add back constraint: enumc nominalThicknesses -}] $ exactDbl 8
 
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
@@ -104,9 +104,15 @@ standOffDist = uq (constrained' (uc sD (variable "SD") Real metre)
   [ gtZeroConstr,
     sfwrc $ Bounded (Inc, sy sdMin) (Inc, sy sdMax)] (exactDbl 45)) defaultUncrt
 
+nomThick' :: ConstrainedChunk
+nomThick' = cuc "nomThick" 
+  (nounPhraseSent $ S "nominal thickness")
+  lT millimetre Rational 
+  [] $ exactDbl 0
+
 nomThick = cuc "nomThick" 
   (nounPhraseSent $ S "nominal thickness" +:+ displayDblConstrntsAsSet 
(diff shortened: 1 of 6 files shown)


samm82:
Is there a specific reason this was the approach taken? `displayDblConstrntsAsSet` only needs a `Quantity`, so we shouldn't have to make a whole new `ConstrainedChunk`. How else does the `Quantity` portion of `nomThick` get used? Does it make sense to have a separate `Quantity`, or should this "helper" object be inlined?

NoahCardoso:
I argee I think it looks better when inlined. I changed it to a QuantityDict since making it a ConstrainedChunk adds unnecessary information. 
nomThick = cuc "nomThick" 
  (nounPhraseSent $ S "nominal thickness" +:+ displayDblConstrntsAsSet 
    `(mkQuant "nomThick" (nounPhraseSent $ S "nominal thickness") lT Rational Nothing Nothing)` nominalThicknesses)
  lT millimetre {-Discrete nominalThicknesses, but not implemented-} Rational 
  [{- TODO: add back constraint: enumc nominalThicknesses -}] $ exactDbl 8

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -106,7 +106,7 @@ standOffDist = uq (constrained' (uc sD (variable "SD") Real metre)
 
 nomThick = cuc "nomThick" 
   (nounPhraseSent $ S "nominal thickness" +:+ displayDblConstrntsAsSet 
-    nomThick nominalThicknesses)
+    (mkQuant "nomThick" (nounPhraseSent $ S "nominal thickness") lT Rational Nothing Nothing) nominalThicknesses)
```
balacij, line 109: Seeing this added here, I would imagine it would be removed from somewhere else, too, no? Does this result in code duplication?

NoahCardoso: I'm not sure what you are referring to, could you be more specific?

samm82: I _think_ the underlying question is "is this information written anywhere else in our code base"?

A similar thought I had: the duplication of `S "nominal thickness"` and the use of the `Nothing`s seems to suggest that this chunk could be built with a different constructor than `cuc` that builds it more "logically", but I'm not sure what other constructors are available.…

NoahCardoso: Why would we not use cuc? The `Nothing`s and the `S "nominal thickness"` are used in the `mkQuant` constructor. The `displayDblConstrntsAsSet` function uses this to make a sentence for the `cuc` constructor.

samm82: What I mean by this is that we are duplicating the arguments `"nomThick"`, `nounPhraseSent $ S "nominal thickness" +:+ x`, `lT`, and `Rational`, which seems redundant. Are these defined anywhere else in the code that we could borrow from instead of redefining it (there may not be)? Looking at the constructors for `ConstrainedChunk`, it looks like they all require some "base" "chunk" to build off of (which is one argument for moving away from our "nested" infrastructure), so we might be stuck using `cuc`.

While investigating this, I noticed that the reason we have this "circular definition" is because we seem to be attempting to define it in terms of itself, instead of just actually giving it a definition, which prompted [https://github.com/JacquesCarette/Drasil/issues/3815#issuecomment-2197788525](a potential agenda item for our next meeting). 👀

JacquesCarette: I agree that this "smells bad". Feels like the wrong constructor is being used.

JacquesCarette (changes requested):

balacij:
@NoahCardoso and I had a discussion _just_ before 5pm on Friday, so the discussion didn't hit here yet, but we're going to look into removing the problematic part of the description entirely. Right now, the description of "t / nominal thickness" is "nominal thickness t \in {....}", which carries information about a _constraint_ (i.e., #2655). So, it looks like we need to...
1. Add Set support to *Expr and Space
2. Add basic Set support to GOOL (@B-rando1 mentioned he can help us with this :slightly_smiling_face: )
3. Add back the constraint
4. Remove the constraint from the description

Then, we end up completely avoiding the issue.

balacij:
Closing in favour of (new) work towards #2655.

Issue #3809, opened 2024-06-18
Negative Step in `listSlice` (GOOL)


B-rando1:
I noticed that in the `listSlice` function that GOOL uses, a negative step value results in different behaviour between different targets.  For example I'll be looking at the following pseudocode:
```python
l = [1, 2, 3, 4]
s = l.slice(start=2, end=0, step=-1)
print(s)
```
In Python this translates to the nicest (in my opinion) functionality: it prints `[3, 2]` - that is, it created a slice from index `2` to index `1`, reversing the list because of the negative step.

In other languages, this prints `[]`, either because those languages don't support negative steps or because we aren't taking advantage of them.

It also gets even worse when we leave out the start or end values.  GOOL allows us to leave either of those out, which has the effect of automatically starting at the start or ending at the end.  In Python when these are combined with a negative step, it simply reverses the list.  In other languages though, it throws out-of-bounds errors.

It would probably be wise to create some consistency between targets.  Our two options are to either assume that the step will never be negative (and ideally enforce this constraint), or bring the other languages up to speed with Python.  The second option is probably better, but there are some tricky edge cases we would need to think about.

For example, what does the pseudocode
```python
s = l.slice(, b, c)
```
translate to?  Depending on whether the runtime value of `c` is positive or negative, we have to start either at `0` or at `b`, and have to keep going either until the index is greater than `b` or less than `0`.  It's not impossible to get this right, but it would result in some messy code.

What are your thoughts?  At this point I'm wondering if it might be better to assume that the step will always be positive, and add a `listReverse` function later if we find we need a way of reversing a list.

JacquesCarette:
The semantics of negative list slice is supposed to be Python's. If it doesn't work "directly" in other languages, then proper code should be generated instead.

We definitely want to allow to state, at a higher level, what we want, and have the low-level generator 'do the right thing'. GOOL is not supposed to represent *code* but rather code-like *intent*.

B-rando1:
I did some work on a fix for this yesterday, focusing on Swift.  Swift mostly worked how it was supposed to, except that if `beg` or `end` wasn't given, they would default to the start or end of the list respectively, instead of checking the sign of `step`.  I added some logic to check step and assign a value to each of `beg` or `end` at runtime if it isn't given.  This fixes the issue, and has the same behaviour as Python for all 10 of the test cases I created.

The one thing I don't like is that some of these checks don't look smart.  For example, code generated for one test:
```swift
var endIdx0: Int
[...]

PR #3820, opened 2024-06-25
Fixed List Slicing in Swift

B-rando1:
### Main Fix
In Swift, the only issue was that the `beg` and `end` bounds would automatically be set to `0` and `list.length` respectively if they weren't given.  I updated the code to check if `step` is negative, and if so set `beg` and `end` to `list.length - 1` and `-1`, respectively, to bring it in line with Python's semantics.  If `step` is a variable, it generates an `if`-statement to do this at runtime, and if `step` is a `litInt`, then it does it at generation-time.

### Changes to `ValData`
In order to get the value of a `litInt`, I added an extra field to `valInt` to `ValData`, so its new definition is:
https://github.com/JacquesCarette/Drasil/blob/282c204c56e627466b903f0e37b1076366193068/code/drasil-gool/lib/GOOL/Drasil/AST.hs#L136

This feels somewhat suboptimal, as `valInt` is `Nothing` unless the value is `litInt`.  Because of smart constructors it's not the end of the world, but I wonder if it would be better to have a new datatype, e.g.
```haskell
data ValDataValue = LitInt Integer | LitBool Bool | ... | ValExpr
```

I do think that this functionality is worth the extra complexity, at least for integers.  For example, we could use it in #3746 to add constant-folding of the `+1`'s that need to take place.

### Test Cases
I switched out the single existing test case for `listSlice` with ten new ones, in order to target every edge case I could think of.  This makes it really easy to see if a language has these edge cases covered, as we can simply run the code and see if the output is the same as the Python output.

### Future Changes
The fix will be similar for Java, C#, and C++.  I might get to them soon, or I might wait  for feedback here about changes to `ValData`.

Contributes to #3809


smiths:
The change sounds good to me @B-rando1, but I'm not an expert.  :smile:  I do have a thought about your test cases.  It looks like you just print the results and leave it to the human to decide if the tests pass.  There is some cognitive load for this because the human will have to think what does correct output look like.  Is there any way you can automate the check of whether the test cases pass?

B-rando1:
Yes, you are correct @smiths.  Currently the test cases only print their results.  I agree that it would be better to have automated testing.

I'm not very familiar with testing methods.  One way I can think of is to have the makefile run the generated code, and pipe the results into a testing script.  I don't know if there are better options, though.  @balacij or @samm82, do you have any ideas for this?

smiths:
If "proper" testing isn't an option, there are some other potential options, as follows:

- have a "golden" version of the generated code from the list slicing tests in stable and make running the test case generation code part of our CI.  This wouldn't test the answers to running the code, but instead it would test the generated code matches our golden version, as we do for the other artifacts.  @balacij will have to comment on whether this is feasible/desirable
- record in comments the expected output, so at least future developers don't have to reinvent what the answers should be

B-rando1:
@samm82 yeah, the tests in `drasil-code/test` all contain GOOL code that is used to generate code in each of the targets.  In theory, I believe we want that running any of the generated files should produce the same result, and for the test cases I added the Python code is the standard.  I wasn't quite sure what your question was, does that help?

As a side note, I just noticed that `stable/gooltest` seems to be less supported than the rest of stable.  This PR passed CI without me updating stable, and there doesn't seem to be a command to stabilize the GOOL tests either.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
```diff
@@ -942,52 +942,83 @@ swiftIndexFunc l v' = do
...
+  Maybe (SValue r) -> Maybe (SValue r) -> SValue r -> MSBlock r
+swiftListSlice vn vo beg' end' step = do
+  stepV <- zoom lensMStoVS step
+  let needIf = (isNothing beg' || isNothing end') && isNothing (valueInt stepV)
+  bName <- if isJust beg' || not needIf then pure ""
+           else genVarName [] "begIdx"
+  eName <- if isJust end' || not needIf then pure ""
+           else genVarName [] "endIdx"
+  let -- If beg or end aren't given and step isn't a litInt, calculate them at
+      --  and store in a variable.
```
bmaclach, line 997: Something seems to be missing from this comment, "calculate them at ... and store in a variable"

bmaclach:
> As a side note, I just noticed that `stable/gooltest` seems to be less supported than the rest of stable. This PR passed CI without me updating stable, and there doesn't seem to be a command to stabilize the GOOL tests either.
I guess this is why some of the changes to stable seem unrelated to your changes? stable hadn't been updated in a while?

bmaclach (dismissed):
The changes to GOOL look good to me. I agree that your `ValDataValue` idea would be an improvement, but this seems valuable on its own!

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
B-rando1: Good catch!  It's supposed to say "at runtime".  I'll add that back in.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
```diff
@@ -942,52 +942,80 @@ swiftIndexFunc l v' = do
...
+      computedBeg = valueInt stepV >>= (\s -> if s > 0 then pure $ litInt 0
+                                              else pure $ listSize vo #- litInt 1)
+      computedEnd = valueInt stepV >>= (\s -> if s > 0 then pure $ listSize vo
+                                                       else pure $ litInt (-1))
+      -- Choose the values to use for beg and end based on given information
+      begVal = fromMaybe (fromMaybe (valueOf begVar) computedBeg) beg'
+      endVal = fromMaybe (fromMaybe (valueOf endVar) computedEnd) end'
+      i = var "i" int
+  block $ findBounds ++ [vn &= swiftMapFunc
+    (swiftStrideFunc begVal endVal step) (lambda [i] (listAccess vo (valueOf i)))]
```
balacij, line 1016: After we spoke about this in person, I jumped down the rabbit hole... it can definitely still be cleaned up (better variable names and comments would go a long way in making it more readable), but I think I've de-duplicated all that I can now.

```haskell
swiftListSlice vn vo beg' end' step = do
  stepV <- zoom lensMStoVS step

  let (setBeg, begVal) = makeSetterVal "begIdx" (valueInt stepV) beg' (litInt 0)    (listSize vo #- litInt 1)
      (setEnd, endVal) = makeSetterVal "endIdx" (valueInt stepV) end' (listSize vo) (litInt (-1))
      i = var "i" int

      setToSlice = vn &= swiftMapFunc (swiftStrideFunc begVal endVal step) (lambda [i] (listAccess vo (valueOf i)))
  block [
      setBeg,
      setEnd,
      setToSlice
    ]
  where
    makeSetterVal _     _       (Just v) _  _  = (emptyStmt, v)
    makeSetterVal vName Nothing _        lb rb = 
      let theVar = var vName int
          theSetter = varDecDef theVar $ inlineIf (step ?> litInt 0) lb rb
      in (theSetter, valueOf theVar)
    makeSetterVal vName (Just s) Nothing  lb rb = 
      let theVar = var vName int
          computed = pure $ if s > 0 then lb else rb
      in (emptyStmt, fromMaybe (valueOf theVar) computed)
```

Here were my steps:
1. Bunch up the begX and endX code by beg and end.
2. Abstract those bunches of code into a local function.
3. Realized a small redundancy in how `needIf` is used, so I inlined it and simplified all expressions.

What do you think?

balacij: Forgot to inline the last `makeSetterVal` case-set...

```haskell
swiftListSlice vn vo beg' end' step = do
  stepV <- zoom lensMStoVS step

  let (setBeg, begVal) = makeSetterVal "begIdx" (valueInt stepV) beg' (litInt 0)    (listSize vo #- litInt 1)
      (setEnd, endVal) = makeSetterVal "endIdx" (valueInt stepV) end' (listSize vo) (litInt (-1))
      i = var "i" int

      setToSlice = vn &= swiftMapFunc (swiftStrideFunc begVal endVal step) (lambda [i] (listAccess vo (valueOf i)))
  block [
      setBeg,
      setEnd,
      setToSlice
    ]
  where
    makeSetterVal _     _       (Just v) _  _  = (emptyStmt, v)
    makeSetterVal vName Nothing _        lb rb = 
      let theVar = var vName int
          theSetter = varDecDef theVar $ inlineIf (step ?> litInt 0) lb rb
      in (theSetter, valueOf theVar)
    makeSetterVal _    (Just s) Nothing  lb rb = (emptyStmt, if s > 0 then lb else rb)
```

balacij: Ok, final time:

```haskell
swiftListSlice vn vo beg end step = do
  stepV <- zoom lensMStoVS step

  let mbStepV = valueInt stepV
      (setBeg, begVal) = makeSetterVal "begIdx" mbStepV beg (litInt 0)    (listSize vo #- litInt 1)
      (setEnd, endVal) = makeSetterVal "endIdx" mbStepV end (listSize vo) (litInt (-1))
      
      i = var "i" int
      setToSlice = vn &= swiftMapFunc (swiftStrideFunc begVal endVal step) (lambda [i] (listAccess vo (valueOf i)))
  block [
      setBeg,
      setEnd,
      setToSlice
    ]
  where
    makeSetterVal _     _       (Just v) _  _  = (emptyStmt, v)
    makeSetterVal _    (Just s) _        lb rb = (emptyStmt, if s > 0 then lb else rb)
    makeSetterVal vName _       _        lb rb = 
      let theVar = var vName int
          theSetter = varDecDef theVar $ inlineIf (step ?> litInt 0) lb rb
      in (theSetter, valueOf theVar)
```

code/stable/gooltest/python/PatternTest/Makefile
```diff
@@ -1,6 +1,10 @@
+# Generated by Drasil v0.1-alpha
+
+# Project Name: PatternTest
+
 build:
 
 run: build
-	python PatternTest.py $(RUNARGS)
+	python3 PatternTest.py $(RUNARGS)
```
balacij, line 8: I guess these GOOL diff tests weren't plugged into the CI?

balacij (dismissed):
Everything else looks good to me! Great work!

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
B-rando1: That looks great to me!  It'll also be handy soon when we make the fix for the `listSlice` in [Macros.hs](https://github.com/JacquesCarette/Drasil/blob/ab8043b24dbd6e5e4664013cf4a52d77d227956c/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/Macros.hs#L61-L78), that Java, C#, and C++ use.  We can move `makeSetterVal` to [CommonPseudoOO.hs](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs), and then both functions can use it!

B-rando1: I tested it, and it passed all the test cases!  Would you like to push the commit, or should I?

JacquesCarette (approved):

JacquesCarette:
Looks good. Yes, your suggested refactor of `ValData` would be a good follow-up to this.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/SwiftRenderer.hs
balacij: (spoke in person)

balacij (approved):
Great work!

PR #3814, opened 2024-06-20
Generating an mdBook-based variant of the SRS

BilalM04:
*Build failure is due to Markdown SRS not being added to stable yet.*

This branch is outdated, additions have been made in #3823 

This draft PR is meant to present my work contributing to #3722 and to receive feedback on my implementation. Progress is being tracked via the checklist in #3722. I have successfully generated a complete [single-page SRS](https://github.com/BilalM04/projectileSRS-mdbook/blob/main/src/generated-example.md?plain=1), rendering all the layout objects in pure Markdown, resorting to HTML syntax only when absolutely necessary.

The implementation reuses code from the HTML and LaTeX printers. I was careful to avoid duplication and ensured this by  importing functions from these printers.

However, there is code duplication between the Markdown and JSON printers. This is due to the fact that JSON currently uses a combination of Markdown and HTML (ideally, it should be pure Markdown). Once this mdBook format SRS is completed, the JSON format should be revisited to leverage the Markdown functions and eliminate the hacky HTML. This will not only resolve the code duplication issue but also address several existing problems in the JSON SRS.

Additionally, I was able to separate the different SRS sections into separate files, consistent with the [manually ported example](https://github.com/BilalM04/projectileSRS-mdbook). 

## Next Steps

- Further investigate how I separate the sections, weigh pros and cons, and consider alternatives
- Generate the title page file and the `SUMMARY.md` for the side panel.
- Figure out how to modify the references and anchors to be able to link across files https://github.com/JacquesCarette/Drasil/discussions/3796#discussioncomment-9824376
- Generate a `book.toml`
- Implement the feedback received from this draft :)

Issue #3722, opened 2024-05-08
Experiment with generating an mdBook-based variant of the SRS


balacij:
# Project Proposal

[mdBook](https://rust-lang.github.io/mdBook/) is used to build the [Rust Book](https://doc.rust-lang.org/book/), and it looks pretty good to me. For reference, mdBook...
* is heavily Markdown based,
* uses a file-based structure, and
* should support our needs nicely (code [see above example] and [math](https://rust-lang.github.io/mdBook/format/mathjax.html)).

This project should be fun. Approximately, the steps should be...
1. Create a demo of the SRS with mdBook. Manually porting over one of the examples.
2. Create a Markdown definition DSL in Drasil.
3. Create an encoding of an mdBook-based project.
4. Create a renderer for the SRS into the mdBook-based project encoding. 

Recalling that Markdown is really just a language over HTML, we can (1) reuse our special HTML components in particular. Furthermore, since we have MathJaX support in mdBook, we can (2) reuse our LaTeX renderer.

# Project Progress

Development is ongoing by @BilalM04 under the [mdBookPrinter](https://github.com/JacquesCarette/Drasil/tree/mdBookPrinter) and [mdBookPrinterTest](https://github.com/JacquesCarette/Drasil/tree/mdBookPrinterTest) branches.

TODO:
- [x] Create a demo of the SRS with mdBook. Found [here](https://github.com/JacquesCarette/Drasil/tree/master/People/Mohd-Bilal/projectileSRS-mdBook).
- [x] Generate a single-page SRS in Markdown
  - [x] Render the layout objects
    - [x] Header
    - [x] Paragraph
    - [x] Table (raw markdown should be visually readable)
    - [x] Definition Table (raw markdown should be visually readable)
    - [x] List
    - [x] Figure
    - [x] Bib
  - [x] Render the Mathjax expressions
  - [x] Render captions using HTML
  - [x] Render reference anchors
  - [x] Render reference link syntax
- [x] Generate the different SRS sections in separate Markdown files
- [x] Fix reference links to jump across files
- [ ] Figure out how to render assets (Most recent update: assets must be in src folder)
- [x] Generate a `SUMMARY.md`
- [x] Generate a `Book.toml`

smiths:
[...]

PR #3823, opened 2024-06-27
Generating an mdBook-based variant of the SRS - UPDATED

### LINKED ISSUE #3827: Add mdBook installation instructions to the New Workspace Setup Wiki
@samm82 mentioned in https://github.com/JacquesCarette/Drasil/pull/3823#issuecomment-2197367526 to add the mdBook [installation instructions](https://rust-lang.github.io/mdBook/guide/installation.html) to the [New Workspace Setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup) so contributors can run it locally.

BilalM04:
This PR is an extension of #3814. Please read https://github.com/JacquesCarette/Drasil/pull/3814#issue-2365220779 before reading this PR.

Contributes to #3722 
Closes #3827 

## Updates / New Additions

- Refactored how the sections are separated (more info below).
- Generated a `SUMMARY.md` file, which is used for the side panel in mdBook.
- Generated a `book.toml` config file, similar to how the CSS is currently generated.
- Generated a title page file (`title.md`)
- Separated into two new SRS formats: Markdown and mdBook (more info below).
- Got the cross-file references working for the mdBook format.
- Fixed minor issues in expression printing.
- Generated SRS for all examples and updated stable. I roughly scanned over and ran all of them. They seem to all be working and rendering as desired.
- Added mdBook installation instructions to the New Workspace Setup wiki.
- Added makefile targets to build and run server of mdBook projects.
  - To build all mdBook projects run `make mdbook_build`. 
  - To build a specific example, run `make X_mdbook_build`. 
  - To launch the local server for a specific example, run `make X_mdbook_server`.

## Refactoring the Separation of Sections

Please read https://github.com/JacquesCarette/Drasil/discussions/3796#discussion-6809343 to understand how SRS generation currently works in Drasil. To summarize, here is the main info:

> The format specific gen functions receive a single Document, which contains all the SRS details. The Document it receives is in the format of the data type defined in [Language.Drasil.Document](https://github.com/JacquesCarette/Drasil/blob/bfcf92f09b07bd883ef2af438f300499d35305e0/code/drasil-lang/lib/Language/Drasil/Document.hs#L61-L64), where the SRS contents is represented through [Section]. The format specific gen functions, using the [makeDocument](https://github.com/JacquesCarette/Drasil/blob/bfcf92f09b07bd883ef2af438f300499d35305e0/code/drasil-printers/lib/Language/Drasil/Printing/Import/Document.hs#L22-L27) function, then convert it to the Document data type defined in [Language.Drasil.Printing.LayoutObj](https://github.com/JacquesCarette/Drasil/blob/bfcf92f09b07bd883ef2af438f300499d35305e0/code/drasil-printers/lib/Language/Drasil/Printing/LayoutObj.hs#L9-L10), where the SRS content is represented through [LayoutObj]. The SRS printers then render these LayoutObj's into a single Doc.

In #3814, the SRS was separated into separate files once it became in the format of the `Document` data type defined in [`Language.Drasil.Printing.LayoutObj`](https://github.com/JacquesCarette/Drasil/blob/bfcf92f09b07bd883ef2af438f300499d35305e0/code/drasil-printers/lib/Language/Drasil/Printing/LayoutObj.hs#L9-L10). I found this to be quite a late stage to be separating, as it was hard to distinguish where each `LayoutObj` belonged. Additionally, the `Document` type is implicitly assuming a single file is being generated. Instead I created a new data type `Project` and `File`:

https://github.com/JacquesCarette/Drasil/blob/5f9ba9280a05c8eca11af14c06ef3957e1fdc64f/code/drasil-printers/lib/Language/Drasil/Printing/LayoutObj.hs#L14-L16

The `Project` type contains a `RefMap`, which maps reference anchor ids to the `Filename` of the `File` that contains them. This map is then used when rendering references in mdBook.

Unlike in #3814, where the given SRS gets converted to a `Document` type, it now gets [converted to a `Project`](https://github.com/JacquesCarette/Drasil/blob/5f9ba9280a05c8eca11af14c06ef3957e1fdc64f/code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs#L61-L63) using the new [`makeProject`](https://github.com/JacquesCarette/Drasil/blob/5f9ba9280a05c8eca11af14c06ef3957e1fdc64f/code/drasil-printers/lib/Language/Drasil/Printing/Import/Document.hs#L31-L38) function.

## Markdown vs mdBook

I created two new SRS formats. The first being `MDBook` which generates the mdBook project. The other being `Markdown` which generates a single-page SRS in Markdown which can be rendered directly in the GitHub repo. I spoke to @balacij about this, and he seems to agree that having the two separate formats is beneficial. @smiths and @JacquesCarette, what are your thoughts?

The `Markdown` format is not complete. Currently it renders with the exact same syntax as mdBook; however, there are slight differences in syntax between the two. I will not be addressing that issue in this PR/branch as I believe creating a separate issue and branch after this is merged is the best approach. Addressing this `Markdown` format will also help with the issues in the `JSON` format, as the two formats use the same "flavour" of Markdown. 

## Next Steps

- Work with @balacij to get the required assets into the generated `src` directory and host the mdBook SRSs.
- Address the issues in the `Markdown` format.


code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -7,7 +7,8 @@ module Language.Drasil.Generate (
   -- * Generator Functions
   gen, genDot, genCode,
   -- * Types (Printing Options)
-  DocType(..), DocSpec(DocSpec), Format(TeX, HTML, JSON), DocChoices(DC),
+  DocType(..), DocSpec(DocSpec), DocChoices(DC),
+  Format(TeX, HTML, JSON, Markdown, MDBook),
```
samm82, line 11: This may be a bit late, but is there a reason we went with `Markdown` instead of `MD` to be consistent with the use of abbreviations for the other formats? I realize that could lead to confusion with `MD` vs. `MDBook`, which might be enough of a reason.

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -70,7 +85,8 @@ prntDoc' dt dt' fn format body' sm = do
   where getExt TeX  = ".tex"
         getExt HTML = ".html"
         getExt JSON = ".ipynb"
-        getExt _    = error "We can only write in TeX, HTML and Jupyter Notebook (for now)."
+        getExt Markdown = ".md"
+        getExt _    = error "We can only write in TeX, HTML, Jupyter Notebook, Markdown, and mdBook (for now)."
```
samm82, line 89: A comment here saying that the `MDBook` case is handled above is probably a good idea, for those who don't see it in the previous pattern matching and think there's a mistake (couldn't be me…)

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -89,12 +105,27 @@ prntCSS docType fn body = do
   where
     getFD dtype = show dtype ++ "/HTML/"
 
--- | Renders the documents.
+-- | Helper for generating the .toml config file for mdBook.
+printBook :: DocType -> Document -> PrintingInformation -> IO()
+printBook dt doc sm = do
```
samm82, line 110: Why `printBook` instead of `prntBook` (without an "i") for consistency with other functions? Or, better yet, why is `prnt` used in other functions?

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -89,12 +105,27 @@ prntCSS docType fn body = do
...
+writeDoc s _  TeX      _  doc = genTeX doc s
+writeDoc s _  HTML     fn doc = genHTML s fn doc
+writeDoc s dt JSON     _  doc = genJSON s dt doc
+writeDoc s _  Markdown _  doc = genMD s doc
+writeDoc _ _  _        _  _   = error "we can only write TeX/HTML/JSON/Markdown/MDBook (for now)"
+
+-- | Renders multi-page documents.
+writeDoc' :: PrintingInformation -> Format -> Document -> [(String, Doc)]
+writeDoc' s MDBook doc = genMDBook s doc
+writeDoc' _ _      _   = error "we can only write TeX/HTML/JSON/Markdown/MDBook (for now)"
```
samm82, line 128: This error string should be made consistent and reused instead of manually defining it repeatedly

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,111 @@
...
+    fp s = text $ "./" ++ s ++ ".md"
+    rp = fn <> text ("#" ++ ref)
+
+-- | Helper for setting up links to references with additional information.
+reflinkInfo :: RefMap -> String -> Doc -> Doc -> Doc
+reflinkInfo rm rf txt info = reflink rm rf txt <+> info
+
+-- | Helper for setting up links to external URIs
+reflinkURI :: Doc -> Doc -> Doc
+reflinkURI ref txt = if ref==txt then ang ref
```
samm82, line 83:
```suggestion
reflinkURI ref txt = if ref == txt then ang ref
```

I _feel_ like this is better formatting, unless I'm missing something. If I'm not, I'm surprised HLint didn't complain!

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,111 @@
...
+caption = wrapGen hcat Align "p" (text "center") [""]
+
+-- | Helper for setting up headings
+heading ::  Doc -> Doc -> Doc
+heading t l = t <+> br (text "#" <> l)
+
+h :: Int -> Doc
+h n
+  | n < 1     = error "Illegal header (too small)"
+  | n > 4     = error "Illegal header (too large)"
```
samm82, line 101: Why are these too small/large? More context would be helpful to the user

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,111 @@
...
+  | n < 4     = text "#"
```
samm82, line 102: Is this correct? I would have expected instances of `"##"` or and `"###"`, probably created by something like `"#" * n`

samm82 (changes requested):
Didn't have time to look through everything, but here's my progress on my classic nitpicking. 😂 Super minor notes on a really solid PR!

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: In the SRS declaration, the heading weights are defined assuming a single file will be rendered. However, mdBook extracts the subsections and subsubsections into their own separate sections. Thus, if I defined the headings to be `"#" * n`. the subsections and subsubsections (which are now separate sections) will have smaller headings, and the heading sizes will be inconsistent across the various `.md` files.

This will ultimately be addressed when fixing up the rendering of the single-page Markdown SRS, as that will require the headings to be rendered in different sizes, consistent with the SRS declaration.

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/6a86cda80c866f216d7cd663d4876850f4e971fe.

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/f7ba8fda90132e900e5349385606160f9c09ac72.

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/f7ba8fda90132e900e5349385606160f9c09ac72.

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/f7ba8fda90132e900e5349385606160f9c09ac72.

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: There was no particular reason, I just went with the first thing that popped up in my head. I agree, there could be confusion between `MD` and `MDBook`. I am open to suggestions :)

code/drasil-printers/lib/Language/Drasil/Markdown/Book.hs
```diff
@@ -0,0 +1,30 @@
...
+
+import Language.Drasil.Markdown.Print (pSpec)
+import Language.Drasil.Printing.PrintingInformation (PrintingInformation)
+import Language.Drasil.Printing.Import.Sentence (spec)
+
+import Language.Drasil hiding (Expr)
+
+-- | Generates the .toml config file for mdBook.
+makeBook :: Document -> PrintingInformation -> Doc  
+makeBook (Document t _ _ _) sm = vcat [
```
balacij, line 15: Is it worth creating a record data type that carries the [configuration (according to their spec)](https://rust-lang.github.io/mdBook/format/configuration/general.html) before building the TOML?

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,112 @@
...
+li = wrap "li" []
+-- | Unordered list tag wrapper.
+ul = wrap "ul" []
+
+-- | helper for wrapping HTML lists
+wrap :: String -> [String] -> Doc -> Doc
+wrap a = wrapGen hcat Id a empty
+
+-- | Helper for setting up HTML tags
+wrapGen :: ([Doc] -> Doc) -> Variation -> String -> Doc -> [String] -> Doc -> Doc
```
balacij, line 48: Can you please add some more documentation to this?

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,112 @@
...
+import Data.List.Split (splitOn)
+import Data.Map (lookup)
+import Language.Drasil.Printing.Helpers (ast)
+import Language.Drasil.Printing.LayoutObj (RefMap)
+
+data Variation =  Id | Align
+
+-- | Curly braces.
+br :: Doc -> Doc
+br x = text "{" <> x <> text "}"
```
balacij, line 15: Are these functions duplicates of any in `Language.Drasil.Printing.Helpers`?

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,112 @@
...
+h :: Int -> Doc
+h n
+  | n < 1     = error "Illegal header (header weigth must be > 0)."
+  | n > 4     = error "Illegal header (header weight must be < 5)"
+  | n < 4     = text "#"
+  | otherwise = text "####"
+
+-- | Helper for stripping Docs
+stripStr :: Doc -> Doc -> Doc
+stripStr d s = hcat (map text (splitOn (show s) (show d)))
```
balacij, line 108: This seems a bit odd -- it converts from a `Doc` to a String and then re-renders it back into a `Doc`. Is there any way we can avoid this?

code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
```diff
@@ -0,0 +1,344 @@
...
+mkDocDefn :: RefMap -> [(String,[LayoutObj])] -> [[Doc]]
+mkDocDefn rm = map (\(f, d) -> [text f, makeLO rm (f,d)])
+
+-- | Renders the title/header of the definition table
+makeDHeaderText :: RefMap -> [(String, [LayoutObj])] -> Doc -> Doc
+makeDHeaderText rm ps l = defnHTag header <> nl
+  where
+    lo = lookup "Label" ps
+    c = maybe l (\lo' -> makeLO rm ("Label", lo')) lo 
+    header = nl <> text "##" <+> heading c l <> nl
```
balacij, line 256: Instead of using `text "##"`, I think it wold be good to use `h2` or something of the sorts.

code/stable/dblpend/SRS/Markdown/DblPend_SRS.md
```diff
@@ -0,0 +1,1128 @@
...
+|Equation   |\\[\boldsymbol{a}\text{(}t\text{)}=\frac{\\,d\boldsymbol{v}\text{(}t\text{)}}{\\,dt}\\]                                                                                                                                                                         |
+|Description|<ul><li>\\(\boldsymbol{a}\text{(}t\text{)}\\) is the acceleration (\\(\frac{\text{m}}{\text{s}^{2}}\\))</li><li>\\(t\\) is the time (\\({\text{s}}\\))</li><li>\\(\boldsymbol{v}\text{(}t\text{)}\\) is the velocity (\\(\frac{\text{m}}{\text{s}}\\))</li></ul>|
+|Source     |[accelerationWiki](#accelerationWiki)                                                                                                                                                                                                                           |
+|RefBy      |                                                                                                                                                                                                                                                                |
+
+<div align="center">
+
+## Velocity {#TM:velocity}
+
+</div>
```
balacij, line 292: Does this work in most Markdown renderers?

code/stable/dblpend/SRS/Markdown/DblPend_SRS.md
```diff
@@ -0,0 +1,1128 @@
...
+|Equation   |\\[\boldsymbol{a}\text{(}t\text{)}=\frac{\\,d\boldsymbol{v}\text{(}t\text{)}}{\\,dt}\\]                                                                                                                                                                         |
+|Description|<ul><li>\\(\boldsymbol{a}\text{(}t\text{)}\\) is the acceleration (\\(\frac{\text{m}}{\text{s}^{2}}\\))</li><li>\\(t\\) is the time (\\({\text{s}}\\))</li><li>\\(\boldsymbol{v}\text{(}t\text{)}\\) is the velocity (\\(\frac{\text{m}}{\text{s}}\\))</li></ul>|
+|Source     |[accelerationWiki](#accelerationWiki)                                                                                                                                                                                                                           |
+|RefBy      |                                                                                                                                                                                                                                                                |
+
+<div align="center">
+
+## Velocity {#TM:velocity}
```
balacij, line 290: I think that custom IDs are only supported in extended Markdown flavours, so it might be worth noting around the definition of the printer that it generates an extended Markdown with MathJax and custom header links.

code/stable/dblpend/SRS/mdBook/src/SecSysContext.md
```diff
@@ -0,0 +1,21 @@
+# System Context {#Sec:SysContext}
+
+[Fig:sysCtxDiag](./SecSysContext.md#Figure:sysCtxDiag) shows the system context. A circle represents an entity external to the software, the user in this case. A rectangle represents the software system itself (DblPend). Arrows are used to show the data flow between the system and its environment.
+
+<div id="Figure:sysCtxDiag"></div>
+
+![System Context](../../../../datafiles/dblpend/SystemContextFigure.png)
```
balacij, line 7: Did images end up working if they weren't within the `src` folder of the mdBook project?

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: `Jupyter` is actually used as a `DocType` (I presume it is for the lessons). `JSON` is used for the printer format.

https://github.com/JacquesCarette/Drasil/blob/6a86cda80c866f216d7cd663d4876850f4e971fe/code/drasil-printers/lib/Language/Drasil/Format.hs#L4-L10

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: The functions in `Language.Drasil.Printing.Helpers` are `String -> Doc`

https://github.com/JacquesCarette/Drasil/blob/6a86cda80c866f216d7cd663d4876850f4e971fe/code/drasil-printers/lib/Language/Drasil/Printing/Helpers.hs#L48-L53

I can either:
1) Keep it the way it is.
2) Remove the functions in `Language.Drasil.Markdown.Helpers` and use `show` on the parameter being passed in.

Which do you suggest?

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: Added in https://github.com/JacquesCarette/Drasil/pull/3823/commits/3db0792f58ac67a73664cbb1da3e768558f3f447. Additionally, I modified the type signature to remove an unnecessary parameter.

code/stable/dblpend/SRS/Markdown/DblPend_SRS.md
BilalM04: The heading attributes (`{#id}`) are specific to mdBook, other Markdown renderers require an `id` attribute in the tag. Additionally, the definition headings are aligned to the center as tables in mdBook are centered. For other Markdown renderers aligning the heading to the center is not needed, as all elements are left-justified.

code/stable/dblpend/SRS/mdBook/src/SecSysContext.md
BilalM04: Nope, after researching online it seems the only way to render images is by having a copy in the `src` folder. After solving that issue, the generated links will also have to be updated to the new paths.

code/stable/dblpend/SRS/Markdown/DblPend_SRS.md
BilalM04: Note added in https://github.com/JacquesCarette/Drasil/pull/3823/commits/5a025acd47ca6c3df8bf89c42c6dbd1aa7c84a00.

code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
BilalM04: Addressed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/88a109896dbccd86b0a148837adea09cb8b556a7. I created a `h'` function which fixes this issue and also creates a useful function that can be used when fixing the `Markdown` format rendering (also mentioned in https://github.com/JacquesCarette/Drasil/pull/3823#discussion_r1657652245)

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: The function is used to strip `Doc`s of any `\n` and/or `\t`. `Doc`s are stripped of `\n` when rendering the Markdown tables as the `printLO` function prints to separate lines; however, `LayoutObj`s in the tables need to be on a single line in order to be rendered. `Doc`s are stripped of `\t` as when calling `TeX.pExpr`, the TeX expressions return formatted expressions, which look ugly in Markdown.

To answer your question, no. Since the printers all deal with `Doc`s, and there are no existing functions in `Text.PrettyPrint` to strip `Doc`s of certain characters, the conversion is required.

code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
```diff
@@ -0,0 +1,344 @@
...
+pExpr (MO Perc)      = bslash <> printMath (TeX.pExpr (MO Perc))
+pExpr (Fenced l r m) = fence TeX.Open l <> pExpr m <> fence TeX.Close r
+  where 
+    fence _ Abs = text "\\|"
+    fence a b   = printMath $ TeX.fence a b
+pExpr (Font Bold e)  = printMath $ commandD "boldsymbol" (pExpr' e)
+pExpr (Font Emph e)  = pExpr e
+pExpr (Spc Thin)     = text "\\\\,"
+pExpr (Sqrt e)       = printMath $ commandD "sqrt" (pExpr' e)
+pExpr e              = printMath $ TeX.pExpr e
```
BilalM04, line 168: I had a discussion with @balacij about this, I will summarize it here.

At first glance, the code for `pExpr` seems very similar to the one in [`Language.Drasil.TeX.Print`](https://github.com/JacquesCarette/Drasil/blob/88a109896dbccd86b0a148837adea09cb8b556a7/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs#L101-L124) and one may assume that it is duplicated. However, there is a reason for this. The differences in expressions between pure LaTeX and the version of Mathjax that mdBook uses is very minor. Specifically, the differences are extra escape characters, different command names (i.e. boldsymbol), quotes, etc. I took the approach of re-defining all the expressions that are different, then delegating to the TeX function in a catch all statement. But since the `pExpr` functions involve recursive calls I had to re-define those as well. This is because once `TeX.pExpr` is called, there is no way for it to return to the Markdown specific `pExpr`.

The reason I went with this approach rather than another (i.e. adding a new parameter to `TeX.pExpr`) is because it is simple and the least amount of code. Adding an additional parameter to `pExpr` that takes in the recursive function call would result in more functions to be added, such as one to begin the recursion. Additionally since all of the printers leverage `TeX.pExpr`, this change would propagate to all the printers.

I will annotate the code below with my reasoning to why those patterns are re-defined:

```Haskell
-- | Helper for rendering Exprs into mdBook compatiable Mathjax
pExpr :: Expr -> Doc
pExpr (Str s)        = printMath $ toText $ pure $ lq <> text s <> rq -- | quotes are different
  where 
    lq = text "\\\\(\\``\\\\)"
    rq = text "''"
pExpr (Div n d)      = printMath $ command2D "frac" (pExpr' n) (pExpr' d) -- | recursive call
pExpr (Case ps)      = printMath $ mkEnv "cases" (TeX.cases ps lnl pExpr') -- | different parameters to TeX.cases 
pExpr (Mtx a)        = stripStr (printMath $ mkEnv "bmatrix" (TeX.pMatrix a lnl pExpr')) tab -- | different parameters to TeX.pMatrix
pExpr (Row [x])      = br $ pExpr x -- | recursive call
pExpr (Row l)        = foldl1 (<>) (map pExpr l) -- | recursive call
pExpr (Label s)      = printMath $ TeX.pExpr (Label s')  -- | escaping special chars in string
  where s' = replace "*" "\\*" (replace "_" "\\_" s)
pExpr (Sub e)        = bslash <> unders <> br (pExpr e) -- | extra escape char
pExpr (Sup e)        = hat    <> br (pExpr e) -- | recursive call
pExpr (Over Hat s)   = printMath $ commandD "hat" (pExpr' s) -- | recursive call
pExpr (MO Perc)      = bslash <> printMath (TeX.pExpr (MO Perc)) -- | extra escape char
pExpr (Fenced l r m) = fence TeX.Open l <> pExpr m <> fence TeX.Close r -- | different fence and recursive call
  where 
    fence _ Abs = text "\\|"
    fence a b   = printMath $ TeX.fence a b
pExpr (Font Bold e)  = printMath $ commandD "boldsymbol" (pExpr' e) -- | change to 'boldsymbol'
pExpr (Font Emph e)  = pExpr e -- | recursive call
pExpr (Spc Thin)     = text "\\\\," -- | extra escape chars
pExpr (Sqrt e)       = printMath $ commandD "sqrt" (pExpr' e) -- | recursive call
pExpr e              = printMath $ TeX.pExpr e -- | catch all statement
```

The annotation above is not in the actual code, please let me know if it is beneficial to add it to the source code as well, along with key points from my blurb above.

code/drasil-printers/lib/Language/Drasil/Markdown/Book.hs
BilalM04: As of now, all the TOMLs will be identical, apart from the title. I kept the TOML quite simple, only defining the absolute minimum needed for the book to run as desired. I don't anticipate a need to use the other configuration options.

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: Yes, using `Jupyter` for both is a problem. We could use `Jupyter` for `DocType` and `Ipynb` for `Format`.

```
error:
drasil-printers>     Multiple declarations of ‘Jupyter’
drasil-printers>     Declared at: lib/Language/Drasil/Format.hs:7:32
drasil-printers>                  lib/Language/Drasil/Format.hs:10:36
drasil-printers>    |
drasil-printers> 10 | data Format = TeX | Plain | HTML | Jupyter | Markdown | MDBook
drasil-printers>    |                                    ^^^^^^^
```

samm82:
When trying to `checkout` this branch, the wiki bug from #3793 seems to reappear - any ideas why? @BilalM04 @balacij 

```bash
samcr@Apex MINGW64 ~/Desktop/Drasil/code (main)
$ git checkout mdBookPrinterTest 
error: invalid path 'wiki/Attributes-and-Chunks:-Meeting-June-7th,-2023.md'
```

samm82:
> To run the mdBook projects, go to the `build/example_name/SRS/mdBook` directory and enter `mdbook serve --open` in terminal to run a local server and view the SRS.

Could we also generate a Makefile to handle this? Idk how involved that would be 😅

code/drasil-gen/lib/Language/Drasil/Generate.hs
samm82: I personally like that!

BilalM04:
@samm82 the invalid path error is likely due to the fix not being on this branch. I just merged `main` into the `mdBookPrinterTest` branch, which should hopefully resolve the issue.

samm82:
I needed to `git fetch` in order to actually get the new changes locally, as opposed to the `git checkout <branch>; git pull` workflow I'm used to (just thought I'd share my silliness). I'm able to switch over to the branch now! 🙃

BilalM04:
From @samm82's suggestions in https://github.com/JacquesCarette/Drasil/pull/3823#issuecomment-2197367526, I added the mdBook installation instructions to the New Workspace Setup wiki in https://github.com/JacquesCarette/Drasil/pull/3823/commits/b9b2047435320e126c2812f58fa152e1042736af, and added makefile targets for mdBook in https://github.com/JacquesCarette/Drasil/pull/3823/commits/df11d396354854a16b70cb3970e65d65c85f8ebf. @balacij suggested to keep these additions within this PR.

Added makefile targets:

- `make mdBook_build` - Build all generated mdBook projects.
- `make X_mdBook_build` - Build the mdBook SRS of an individual example.
- `make X_mdBook_server` - Launch local server for the mdBook SRS of an individual example.

samm82:
Should these be renamed to `mdbook` (all lowercase)? That's what the command is, which is what I assumed the `make` rule would be called (if there's a reason for using `mdBook`, I'm sure that's fine too!

BilalM04:
I used `mdBook` as that is how it is capitalized in the [mdBook documentation](https://rust-lang.github.io/mdBook/index.html). However, I agree it should be renamed to `mdbook` in the makefile, as the actual command is all lower case. Additionally it will now be consistent with all the other targets (which are also all lower case).

code/drasil-gen/lib/Language/Drasil/Generate.hs
BilalM04: This was discussed in #3815. 

- `DocType`: `Jupyter` will be changed to `Lesson`
- `Format`: `JSON` will be changed to `Jupyter`
- `Markdown` will be extracted into a new data type, representing the different Markdown flavours.

BilalM04: Added in https://github.com/JacquesCarette/Drasil/pull/3823/commits/f987ff95bb87561978b387243c83b98742758ed1

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,127 @@
...
+  | otherwise = h' n
+
+-- | Helper for setting up heading weights in normal Markdown.
+h' :: Int -> Doc
+h' n
+  | n == 1    = text "#"
+  | n == 2    = text "##"
+  | n == 3    = text "###"
+  | n == 4    = text "####"
+  | otherwise = error "Illegal header (header weight must 0 < n < 5)"
```
samm82, line 119:
```suggestion
  | otherwise = error "Illegal header (header weight must be between 0 and 5)"
```
Just to be consistent with the other error messages which don't use `n`, although this could be approached in other ways

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -0,0 +1,127 @@
...
+
+-- | Helper for setting up headings with an id attribute.
+-- id attribute will only work for mdBook.
+heading ::  Doc -> Doc -> Doc
+heading t l = t <+> br (text "#" <> l)
+
+-- | Helper for setting up heading weights in mdBook.
+h :: Int -> Doc
+h n
+  | n < 1     = error "Illegal header (header weigth must be > 0)."
```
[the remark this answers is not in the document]
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/20ebf35520a935854188b3cab2b4600062620bc3.

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/20ebf35520a935854188b3cab2b4600062620bc3.

samm82 (dismissed):
This seems good to me (once @balacij's items have been addressed) 😁

JacquesCarette (approved):

samm82:
@BilalM04 I'd make a separate issue for each of the notes for future work @JacquesCarette made so that we don't lose track of them/can have more detailed (and contained!) discussion on them!

BilalM04:
In regards to the `splitStr` function and `\n` and `\t` appearing in `Doc`s. This was due to the fact `TeX.pExpr` returns case statements and matrices formatted to separate lines. However in mdBook, they need to be in a single line to be rendered (particularly for the tables). Thus, I was naively converting to a `String` and removing all the formatting. In https://github.com/JacquesCarette/Drasil/pull/3823/commits/feec894a051213d89167756eef8a486fa4bf200c, from @balacij's suggestion, I parameterized the concatenation functions in `mkEnv`, `TeX.pMatrix`, and `TeX.cases`, and passed in horizontal concatenation in the Markdown printers.

Additionally, in order to add empty lines in the `Doc`s, I was using `text "\n"`, which I realized to be the wrong approach. In https://github.com/JacquesCarette/Drasil/pull/3823/commits/dc5954be09c49b4fefb7143bf0a9fba109ee472b, I added a new infix operator which concatenates two `Doc`s with an empty line in between them.

code/Makefile
```diff
@@ -371,6 +379,16 @@ $(filter %$(STABILIZE_E_SUFFIX), $(STABILIZE_WEBSITE)): %$(STABILIZE_E_SUFFIX):
...
 
+# Build individual example mdBook SRSs.
+# Targets will have the form exampleName_mdBook_build.
+$(filter %$(MDBOOK_BUILD_E_SUFFIX), $(MDBOOK_BUILD_EXAMPLES)): %$(MDBOOK_BUILD_E_SUFFIX): $(CLEAN_GF_PREFIX)$(LOG_FOLDER_NAME) %$(GEN_E_SUFFIX)
+	- mdbook build "$(BUILD_FOLDER)$(EDIR)/SRS/mdBook"
+
+# Launch local server for individual example mdBook SRSs.
+# Targets will have the form exampleName_mdBook_server.
+$(filter %$(MDBOOK_SERVER_E_SUFFIX), $(MDBOOK_SERVER_EXAMPLES)): %$(MDBOOK_SERVER_E_SUFFIX): $(CLEAN_GF_PREFIX)$(LOG_FOLDER_NAME) %$(GEN_E_SUFFIX)
+	- mdbook serve --open "$(BUILD_FOLDER)$(EDIR)/SRS/mdBook"
```
balacij, line 390: The `-` in front suppresses error codes. I think we want those to be returned (I previously added too many of the `-` in the Makefile, I need to fix those too). So, we should remove the prepended `-`s.

code/drasil-gen/lib/Language/Drasil/Generate.hs
```diff
@@ -52,25 +55,46 @@ prntDoc :: Document -> PrintingInformation -> String -> DocType -> Format -> IO
...
-    TeX     -> do prntDoc' dtype (show dtype ++ "/PDF") fn TeX d pinfo
-                  prntMake $ DocSpec (DC dtype []) fn
-    Jupyter -> do prntDoc' dtype (show dtype ++ "/Jupyter") fn Jupyter d pinfo
-    _ -> mempty
+    HTML              -> do prntDoc' dtype (show dtype ++ "/HTML") fn HTML d pinfo
+                            prntCSS dtype fn d
+    TeX               -> do prntDoc' dtype (show dtype ++ "/PDF") fn TeX d pinfo
+                            prntMake $ DocSpec (DC dtype []) fn
+    Jupyter           -> do prntDoc' dtype (show dtype ++ "/Jupyter") fn Jupyter d pinfo
+    (Markdown GitHub) -> do prntDoc' dtype (show dtype ++ "/Markdown") fn (Markdown GitHub) d pinfo
```
balacij, line 63: `Markdown GitHub` to show that it uses the GHFM spec => Good idea!

code/drasil-printers/package.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
 - lens
 - pretty
 - containers
+- MissingH
```
balacij, line 17: Out of curiosity, where is this used?

code/stable/dblpend/SRS/Markdown/DblPend_SRS.md
balacij: Since we know that this is the GHFM flavour, this works. So, this comment can be dismissed. Thank you @BilalM04

balacij (approved):
Great work! The PR looks like it's in great shape!

code/drasil-printers/package.yaml
BilalM04: It is used [here](https://github.com/JacquesCarette/Drasil/pull/3823/files?diff=unified&w=0#diff-1247766c67d1c8c19d75aacc0f2153498b823b7055a0aac385c064ae5f66cdb5R155). For the `replace` function imported from `Data.List.Utils`.

code/Makefile
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3823/commits/6eaf3dccd8fd7da4b86aa5df17a5bfd496897c68.

code/drasil-printers/package.yaml
balacij: Ah, gotcha!

smiths:
Should we remove the `Makefile` rule `make X_mdbook_server`?  I like that this is implemented and it is cool that we can start the mdbook server for any of the case studies from one central location, but I'm not sure the idea is consistent with the purpose of the main `Makefile`.  The main `Makefile` is for generating the examples, not for running them.  Continuing this new trend, if we have a `make X_mdbook_server`, should we also have a `make X_run` that runs the code for each example?  We have many variations of the code in some cases.  Should we have a `make` run for each variant?

Now that we have the local `Makefile` for each mdbook, I think this is more in keeping with how we've handled doing something with the examples.  The main `Makefile` generates the examples, including the "local" `Makefile`s.  The local `Makefile`s are used to run the examples, compile the pdf from tex and start the mdBook server.

I'm fine to leave `make X_mdbook_server`, but I thought we should talk about it, since if we aren't going to use this rule, we don't want to have to maintain this code going forward.

code/drasil-printers/lib/Language/Drasil/HTML/Helpers.hs
```diff
@@ -94,6 +94,10 @@ wrapGen' sepf Id s ti _ = \x ->
   let tb c = text ("<" ++ c ++ " id=\"") <> ti <> text "\">"
       te c = text $ "</" ++ c ++ ">"
   in sepf [tb s, indent x, te s] 
+wrapGen' sepf Align s ti _ = \x ->
+  let tb c = text ("<" ++ c ++ " align=\"") <> ti <> text "\">"
```
samm82, line 98: Since there is the repeated pattern of `let tb c = text ("<" ++ c ++ " <lowercase Variation>=\"")`, I'm wondering how much of these case could be generated, as well as if we could define a `show` for `Variation` to assist in this 🤔 Don't know how much effort this would be for its benefits though lol

Love this commit regardless!

BilalM04: Improved in https://github.com/JacquesCarette/Drasil/pull/3823/commits/3614102058874ba7755b208978b52b1c005356ff.

BilalM04:
Removed all traces of the `Markdown GitHub` format. It has been completely moved to the [`markdownPrinter`](https://github.com/JacquesCarette/Drasil/tree/markdownPrinter) branch.

JacquesCarette (approved):

JacquesCarette:
Taking the plunge and merging this in!

PR #3826, opened 2024-06-28
Fixed `listSlice` in the other targets

### LINKED ISSUE #3809: Negative Step in `listSlice` (GOOL)

B-rando1:
I updated the function used by Java, C#, and C++ for `listSlice`, so `listSlice` should be consistent now between all 5 targets.

#### Design decisions:
If `beg` or `end` are not given, we find values for them in the same way
as with Swift.

We also have to generate the condition for the `for`-loop:
- If `step` has a literal value, we use that to generate a one-sided
check: either `step < end` or `step > end`
- If `step` does not have a literal value but `beg` and `end` do, then we
use that information to generate a single double-sided bounds check:
either `beg < step < end` or `end < step < beg`.
- If `step` does not have a literal value, and `beg` or `end` doesn't
  either, then we need to check both of the above double-sided checks.

Closes #3809
Builds on #3820


balacij (approved):
That was a rather smooth extension to the other languages! Great work!

JacquesCarette (approved):

Issue #3706, opened 2024-04-01
Make sure "specialized" definitions reference their "primitive" concepts


samm82:
I noticed during #3702 that the definition of "position vector" doesn't reference "position" at all! Regardless of whether this definition is meant to convey something different (i.e., if the fix is renaming it vs. redefining it), an investigation of how we "build up" concepts should probably be done as part of the work from that meeting. Thoughts?

https://github.com/JacquesCarette/Drasil/blob/1ef469d9eefe1dd6b9f8d0ddac665951445d2a8d/code/drasil-data/lib/Data/Drasil/Concepts/Physics.hs#L137-L141

smiths:
Thank you for creating this issue @samm82.  Yes, we need to go through all of our examples and strengthen the concepts.

Xinlu-Y:
It seems that `position` is being used in the key model definitions and equations. Could this be a result of inconsistent naming, or should `positionVec` replace `position` in this context? Additionally, how should I correctly build up `positionVec` to reference `position`?
https://github.com/JacquesCarette/Drasil/blob/8624ac8d9bb18083f1df9265ff9b311b5df737cc/code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs#L97-L112
![Image](https://github.com/user-attachments/assets/e8db1abd-233f-4ee9-9d78-603e2544ebe7)

JacquesCarette:
You are correct that there seems to be a confusion in this example between `position` and `positionVec`. Your instincts are correct that `position` should not be used at all, and `positionVec` should be instead. Is that clear enough?

JacquesCarette:
It does make sense to define the 'concept' of `position`; `positionVec` should, in theory, refer to that. But our current arrangement of Chunks may not really support that at the moment. It's part of the big project of re-arranging chunks more 'semantically'.

PR #3831, opened 2024-07-04
Erroneous use of nounPhraseSP

### LINKED ISSUE #3782: Erroneous use of nounPhraseSP in `drasil-example`
I have noticed that in `glassbr` almost all words defined in `Unitals.hs` are made using `nounPhraseSP` from #3777 and #2299. Is there a reason for this? `nounPhraseSP` blocks words from being plural. For example, the word "explosion" was a `nounPhraseSP` even though it has the plural form "explosions". I think that we should use `cn'` rather than `nounPhraseSP` as it allows plural. This `nounPhraseSP` default is also used a lot in the other examples not just `glassbr`.

| `cn` | Create a self-plural, common-noun NounPhrase | String -> NP |
| `cn'` | Create a common-noun NounPhrase which becomes plural by adding "s" | String -> NP |

image

image

NoahCardoso:
Closes #3782. I have noticed that in `glassbr` almost all words defined in `Unitals.hs` are made using `nounPhraseSP` from https://github.com/JacquesCarette/Drasil/pull/3777 and https://github.com/JacquesCarette/Drasil/issues/2299. `nounPhraseSP` blocks words from being plural. For example, the word "explosion" was a `nounPhraseSP` even though it has the plural form "explosions". I changed some examples of `nounPhraseSP` to the plural form if it made sense based on the definition of the word or if it made sense to describe multiple of the item.


samm82 (changes requested):
Many of these changes would result in an incorrect plural form; below is a list of how (I think) these terms _should_ be pluralized.

- heat flux**es** into the water from the coil**s**
- volume**s** of the cylindrical tank**s**
- change**s** in temperature
- length**s** of tank**s**
- diameter**s** of tank**s**
- temperature**s** of the heating coil**s**
- minimum length**s** of tank**s**
- maximum length**s** of tank**s**
- maximum surface area**s** of coil**s**

Ideally, these terms would be formed in the style of #3706; for example, the term "heat flux into the water from the coil" should be built from the concept of "heat flux" plus the additional suffix "into the water from the coil", so the _phrase_ would have a defined plural of something like `plural heatFlux +:+ into the water from the +:+ plural coil`. Whether or not this is something you want to take a look at is up to you; this PR could also be updated by just "manually" defining these plurals with a custom constructor, which would at least be a step in the right direction.

samm82:
Also, could you merge `main` into this branch (see https://github.com/JacquesCarette/Drasil/pull/3823#issuecomment-2197418910)? I was unable to run the code to verify how these plurals would actually be rendered, so my comment is (informed) speculation!

NoahCardoso:
@samm82 I might have to create a new branch since when I tried to merge it added 93 commits to this pr

NoahCardoso:
After a discussion with @balacij, I am going to remove any complex examples since I think that is a separate issue. This problem was more about fixing the small 1 to 2 word examples.

balacij (approved):
Looks good!

NoahCardoso:
@samm82 It should be up to date with the current version of main now

samm82 (approved):
I'm still unable to checkout this branch 😩 but I'm willing to trust @balacij lol - would you be able to create an issue for these "more involved" future fixes (or make this PR not close the original issue and track these items there)?

NoahCardoso:
@samm82 I rebased, are you able to run it locally now?

balacij:
I think we agreed previously (in the last in-person meeting... ?) that we would just do the simpler ones to close out the ticket. As @smiths mentioned regarding the complex ones (in #3782), there's no point in guessing what the plural forms should be. Anyone trying to use one of the 'bad' ones that has an incorrect plural form will come about it on their own, so a ticket that discusses this won't really get done until someone go through them individually (i.e., it will be very long-standing). An alternative is to switch these bad ones to a `nounPhraseSU(S/P)` (i.e., 'suspicious noun phrase' [lol] or 'noun phrase singular, unknown plural') that makes the plural forms an `error "$uid: $singular_form: is missing a plural form -- please fix!"`, so that all future use will have the issue highlighted. I think I actually like that idea now.

samm82:
Yes, I was able to rebase this! Thanks, and looks good!

JacquesCarette (approved):

PR #3845, opened 2024-07-10
Split `ClassInterface.hs` into separate files.

B-rando1:
I called them `InterfaceCommon.hs` and `InterfaceGOOL.hs`.  I chose to postfix the `Shared` and `GOOL` parts so that the files are together when sorted alphabetically.
- I'm not sure if these are the best names.  `ClassInterface` makes a lot of sense - it contains typeclasses that describe the interface (or syntax) of GOOL.  `ClassInterfaceX`, where `X` is `Proc`, `Common`, or `GOOL`, would be better, but that seems a bit long.  The name isn't used in too many places though, so maybe we should change it.  Let me know what you think.

This change should help to clarify which functions are fully OO and which can be shared with procedural.  It will also help avoid name conflicts when we separate 'parallel' typeclasses in GOOL.

I tried not to make any changes to the interface itself; however I did notice one method I had left in `ValueExpression`, and I moved it to `OOValueExpression` where it belongs.


JacquesCarette (approved):
These names are certainly good enough for now. We can always rename later when things become clearer.

PR #3846, opened 2024-07-10
Add Sets

### LINKED ISSUE #2655: Add back `OneOf` Value constraints

NoahCardoso:
Closes #2655. WIP, I currently have created sets in expr.


code/drasil-lang/lib/Language/Drasil/Constraint.hs
```diff
@@ -21,20 +21,34 @@ data Constraint a where
...
+isPhysRange, isSfwrRange, isPhysElem, isSfwrElem :: Constraint e -> Bool
 
 -- | Helpful for filtering for Physical constraints. True if constraint is 'Physical'.
-isPhysC (Range Physical _) = True
-isPhysC _ = False
+isPhysRange (Range Physical _) = True
+isPhysRange _ = False
+
+isPhysElem (Elem Physical _) = True
+isPhysElem _ = False
```
balacij, line 47: Is the differentiation needed between Range and Elem for this? Why are the `isX` functions needed at all?

code/drasil-lang/lib/Language/Drasil/CodeExpr/Lang.hs
```diff
@@ -70,6 +70,18 @@ data UFuncVV = NegV
 data UFuncVN = Norm | Dim
   deriving Eq
 
+-- | Set + Set -> Set
+data SSet = SUnion
```
balacij, line 74: Following the pattern below, I think this would be `SSSBinOp`? Union is also associative, so you can build it similar to the other associative operations.

code/drasil-lang/lib/Language/Drasil/CodeExpr/Lang.hs
```diff
@@ -70,6 +70,18 @@ data UFuncVV = NegV
...
+  deriving Eq
+
+-- | Element + Set -> Set
+data ESSSet = SAdd | SRemove
+  deriving Eq
+
+-- | Element + Set -> Bool
+data ESBSet = SContains
+  deriving Eq
```
balacij, line 83: Same with above on naming.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -109,7 +120,7 @@ data Expr where
   Case     :: Completeness -> [(Expr, Relation)] -> Expr
   -- | Represents a matrix of expressions.
   Matrix   :: [[Expr]] -> Expr
-
+  Set :: Expr -> Expr
```
balacij, line 123: This constructor should take in a `Set` (or list if we don't have an `Ord` instance for `Expr`) of `Expr`s.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -282,6 +301,12 @@ instance Typed Expr Space where
               (\_ -> all (\ r -> length r == columns && all (== expT) r) sss)
               (const False) expT
         t = fromLeft (error "Infer on Matrix had a strong expectation of Left-valued data.") expT -- This error should never occur.
+        
+  infer cxt (Set e) =
+    case infer cxt e of
+        Left sp -> if S.isBasicNumSpace sp then Left sp else Right (show sp)
```
balacij, line 307: I'm not sure if we need to restrict ourselves to only numeric sets. For example, power sets and string sets are probably things we want. However, we can probably just admit anything in the set, so as long as the elements are of the same type.

balacij (commented):
Shaping up nicely! I think you'll also want to add the new operators to the `Expr` typeclasses that define the smart constructors of the various `Expr`s.

balacij:
Have you also started on the conversion of the Exprs into GOOL?

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
NoahCardoso: Yes, this was left in by accident.

NoahCardoso:
Yes I have started some of the conversions of expr in GOOL. I wanted to ensure I was on the right path before getting too deep into GOOL.

code/drasil-lang/lib/Language/Drasil/CodeExpr/Lang.hs
NoahCardoso: Yes, I agree `___BinOp` would probably be better. For Union I was thinking of doing
data AssocConcatOper = SUnion
        deriving Eq

code/drasil-lang/lib/Language/Drasil/Constraint.hs
NoahCardoso: It is used for formatting constraints. I don't think a distinction between the two would need to be made here
https://github.com/JacquesCarette/Drasil/blob/eacde82083e33e4c566cbac8b68373dfb90d35db/code/drasil-docLang/lib/Drasil/Sections/SpecificSystemDescription.hs#L278-L284
https://github.com/JacquesCarette/Drasil/blob/eacde82083e33e4c566cbac8b68373dfb90d35db/code/drasil-code/lib/Language/Drasil/Chunk/ConstraintMap.hs#L22-L28

balacij:
It would be good to merge in changes from `main` sooner rather than later. I think @B-rando1 is making some large changes, so he can best advise with how things changed.

PR #3850, opened 2024-07-11
Separated more typeclasses into OO and Procedural

B-rando1:
I made parallel versions of `ModuleSym`, `FileSym`, and `ProgramSym`
(along with their respective associated types).  Because `ModuleSym` was
the only thing relying on `ClassSym`, `ClassSym` is now fully-OO; so I
moved it into the OO file.

Note that while Procedural versions of `ModuleSym`, `FileSym`, and
`ProgramSym` exist, nothing uses them, and `GOOL.Drasil` doesn't export
them.  The latter is because they have the same name, and thus need to
be imported with a qualified name if both versions of the typeclasses
are used in the same file.

Builds on https://github.com/JacquesCarette/Drasil/pull/3849


JacquesCarette (approved):

JacquesCarette (approved):

JacquesCarette:
Weird - I've approved this, and it won't let me merge!

B-rando1:
@JacquesCarette could it be blocking the merge because this PR builds on the last one you merged?  I see now it says your approval was dismissed because the merge-base changed.  I don't think I saw that right away, so maybe GitHub needed to do some processing.

If that is what's happening here, this [comment](https://github.com/orgs/community/discussions/57513#discussioncomment-6248715) says that there's a setting to change the behaviour.

JacquesCarette (approved):

JacquesCarette:
I over-rode the settings. I'll figure out these "merge-base changed" issues later!

PR #3851, opened 2024-07-11
Separated a few more GOOL typeclasses

B-rando1:
- Created `OOMethodSym` to hold the OO-only parts of `MethodSym`.
    - I don't like the name.  Ideally we'd split `MethodSym` into `FunctionSym` and `MethodSym`.  Currently `FunctionSym` is taken, but I don't like the name there so maybe at some point we can do a 3-way rename.
- Moved `StateVarSym` and `PermanenceSym` into `InterfaceGOOL`, as all of their dependencies are now in there.

My diagram tells me that the only OO typeclass left in `InterfaceCommon`
is `ScopeSym`, which is still required by `MethodSym` for some reason.
I'll look into how those functions are implemented, and see if we can
get rid of it.  Otherwise we'll need to split `MethodSym` completely
into `OOMethodSym` and `ProcMethodSym`.  We'll see what works.

Oh, and I also got rid of a bunch of whitespace at the end of lines in the C++ renderer.  Probably should have done that in a different PR 😅.

Builds on https://github.com/JacquesCarette/Drasil/pull/3850


B-rando1:
Now that I'm looking at it, removing the whitespace at the end of lines in the C++ renderer makes that diff completely unreadable.  I probably shouldn't have done that...

The good news is, I didn't do anything in the C++ renderer that I didn't do in the others, i.e. I was just moving the imports and typeclasses around to match the updated refactoring.

JacquesCarette (dismissed):

JacquesCarette (approved):
This is an approval, not 'review comments'.... ?

PR #3854, opened 2024-07-16
Julia renderer v1

B-rando1:
The Julia renderer is finally at a point where it's ready for feedback.  This PR will likely require a good number of changes, but it's a good sign of progress.

#### Julia-Specific
- Currently I have the Julia renderer inheriting from `OOProg` rather than `ProcProc`, because `drasil-code` isn't yet able to generate `ProcProg` code.  I left everything that doesn't exist in `ProcProg` unimplemented, though.  The next big step will be to convert the renderer to `ProcProg` and then enable `drasil-code` to generate `ProcProg` code.
- List slicing currently doesn't pass the edge cases.  This shouldn't be too much work, but I didn't want it to hold back the PR.
- I'm not  quite sure how we should do types.  In Julia it is idiomatic to have functions, etc. take the most general possible types for its inputs, to take full advantage of multiple dispatch.  Currently GOOL doesn't really have a sense of sub-types, so in order to be consistent I've used the concrete types everywhere.
- By default Julia uses 64-bit integers, and to create a 32-bit integer you need to explicitly cast it.  For this reason the Julia renderer currently uses 64-bit integers, even though our other targets (except Python) use 32-bit integers.  Should I keep it this way to be more idiomatic, or change it to 32-bit integers to be more consistent with other targets?

#### Other Changes along the Way
- [HelloWorld.hs](https://github.com/JacquesCarette/Drasil/compare/JuliaRenderer?expand=1#diff-275ece6dae080915de747f91efe7524a1dbfb9990c55eb1233640285027f953c): commented out the `objDecDef` line so that the Julia renderer can run it.  This is a temporary fix, and I am hoping to make a better fix at a later date.
- [Projectile/Choices.hs](https://github.com/JacquesCarette/Drasil/compare/JuliaRenderer?expand=1#diff-1802ac70744b53c1c73e476996afaf1c877f6640e5560ad40a860540972b336c): added Julia as a target language for the two `Unbundled` sets of `Choices`.
- Julia and Python have many similarities, including opening files, checking if command line arguments exist, and for-each loops.  I abstracted this functionality into `CommonPseudoOO.hs`.
- Julia and Swift both usually use markdown-style doc comments, so I abstracted that functionality into `CommonPseudoOO.hs`.
- I noticed that the `arrayElem` function also deals with indices, so I added use of the `intToIndex` function to the `languagePolymorphic.hs` implementation.


code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
```diff
@@ -516,7 +526,20 @@ noteDoc = "Note"
...
+varDec v = do
+  v' <- zoom lensMStoVS v
+  modify $ useVarName (variableName v')
+  mkStmtNoEnd empty
+
+varDecDef :: (RenderSym r) => SVariable r -> SValue r -> MSStatement r
+varDecDef v e = do
+  v' <- zoom lensMStoVS v
+  modify $ useVarName (variableName v')
+  IC.assign v e
```
samm82, line 541: This screams `Maybe SValue` to me, but that probably has some design implications that aren't great, so take that with a grain of salt 😅

code/stable/gooltest/julia/HelloWorld/HelloWorld.jl
```diff
@@ -54,7 +54,7 @@ mySlicedList4 = myOtherList[4:2 - 1]
 
 mySlicedList5 = myOtherList[4:-1:2 - 1]
 
-mySlicedList6 = myOtherList[start:-1:2 - 1]
+mySlicedList6 = myOtherList[begin:-1:2 - 1]
```
samm82, line 57: Any reason for this? Just curious!

samm82 (approved):
This looks really nice!

code/stable/gooltest/julia/HelloWorld/HelloWorld.jl
B-rando1: I think I accidentally wrote `start` instead of `begin` a while ago, and just didn't catch it till now 😅

samm82: Is `begin` a keyword in Julia?

B-rando1: yeah, when dealing with lists, `begin` and `end` refer to the first and last indices of the list!

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
B-rando1: I'm sorry, I'm not quite sure what you mean 🤔

Do you mean the type of `e`?  If you want to declare a variable without giving it a value, you can use `varDec` instead.

samm82: I think my real question is "is there a rationale for having `varDec` and `varDecDef` be separate functions?"

I think these could be merged by doing something like:
```haskell
varDec :: (RenderSym r) => SVariable r -> Maybe (SValue r) -> MSStatement r
varDec v e = do
  v' <- zoom lensMStoVS v
  modify $ useVarName (variableName v') $ def e
    where
      def Nothing  = mkStmtNoEnd empty
      def (Just d) = IC.assign v d
```

At the very least, having the above function as a helper (e.g., as `varDecHelper`) that gets called by both `varDec` and `varDecDef` would reduce code duplication if we want to keep the two variants semantically distinct

code/stable/gooltest/python/HelloWorld/HelloWorld.py
```diff
@@ -63,6 +63,8 @@
 
 mySlicedList10 = myOtherList[2::z]
 
+mySlicedList10 = myOtherList[2::z]
```
samm82, line 66: Is this output expected? (I'm confused because of the comment ` -- | Do it again, to make sure a unique variable name for endIdx is being generated` above; does that correspond to this line of generated code? Should these lines be the same?)

B-rando1: Oh yeah, the reason for that is that Python is the one language that does list slicing correctly 'out of the box', so we don't need to create `endIdx` for it.  All the other renderers do need to create `endIdx` for those examples, though.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/CommonPseudoOO.hs
B-rando1: Hmm, I see what you mean, and I like the suggestion for merging the two helper functions.

I'm not so sure about merging them at the typeclass level, though.  Just digging through the codebase, I've found it's not as simple as I expected.  I'll open a discussion post about it, as that change isn't directly related to this PR.  Stay tuned...

code/stable/gooltest/python/HelloWorld/HelloWorld.py
samm82: That's what I get for only checking the Python version cuz it's the easiest 🙃

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -0,0 +1,1064 @@
...
+import Control.Monad.State (modify)
+import Data.List (intercalate, sort)
+import Text.PrettyPrint.HughesPJ (Doc, text, (<>), (<+>), empty, brackets, vcat,
+  quotes, doubleQuotes, parens, equals, colon)
+import qualified Text.PrettyPrint.HughesPJ as D (float)
+
+jlExt :: String
+jlExt = "jl"
+
+newtype JuliaCode a = JLC {unJLC :: a}
```
JacquesCarette, line 98: can probably use `deriving` for `Functor`, `Applicative` and `Monad`.

JacquesCarette (changes requested):
Very nice! Only one minor thing and we can merge this in.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JuliaRenderer.hs
B-rando1: I tried removing the `instance ... JuliaCode` lines for those three and changing line `98` to be

```haskell
newtype JuliaCode a = JLC {unJLC :: a} deriving (Functor, Applicative, Monad)
```

This gave me errors for all 3:
```
    • Can't make a derived instance of ‘Functor JuliaCode’:
        You need DeriveFunctor to derive an instance for this class
    • In the newtype declaration for ‘JuliaCode’
   |
98 | newtype JuliaCode a = JLC {unJLC :: a} deriving (Functor, Applicative, Monad)
   |                                                  ^^^^^^^
```
```
    • Can't make a derived instance of ‘Applicative JuliaCode’:
        ‘Applicative’ is not a stock derivable class (Eq, Show, etc.)
        Try GeneralizedNewtypeDeriving for GHC's newtype-deriving extension
    • In the newtype declaration for ‘JuliaCode’
   |
98 | newtype JuliaCode a = JLC {unJLC :: a} deriving (Functor, Applicative, Monad)
   |                                                           ^^^^^^^^^^^
```
```
    • Can't make a derived instance of ‘Monad JuliaCode’:
        ‘Monad’ is not a stock derivable class (Eq, Show, etc.)
        Try GeneralizedNewtypeDeriving for GHC's newtype-deriving extension
    • In the newtype declaration for ‘JuliaCode’
   |
98 | newtype JuliaCode a = JLC {unJLC :: a} deriving (Functor, Applicative, Monad)
   |                                                                        ^^^^^
```
I added these to the top of the file, as recommended:
```haskell
{-# LANGUAGE DeriveFunctor #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
This fixed the `deriving Functor` error, but then for `Applicative` (and `Monad` was similar) I got:
```
    • Can't make a derived instance of ‘Applicative JuliaCode’
        (even with cunning GeneralizedNewtypeDeriving):
        cannot eta-reduce the representation type enough
    • In the newtype declaration for ‘JuliaCode’
    |
100 | newtype JuliaCode a = JLC {unJLC :: a} deriving (Functor, Applicative, Monad)
    |                                                           ^^^^^^^^^^^
```

So it looks like we can use `deriving Functor`, but we need an instance of `Applicative` and `Monad`.

I tested it with `deriving Functor`, and that seemed to work.  Should we stick with that?

JacquesCarette: Odd - I am quite sure this should work. But it might need more 'hints'. Have you looked at the documentation for GeneralizedNewtypeDeriving ?

B-rando1: After reading the documentation (and a surprisingly well-explained answer from ChatGPT when I was feeling out of my depth), it appears that the `GeneralizedNewtypeDeriving` language extension only works with `newtype`s that take a basic list of type variables.  Since `JuliaCode` has type variables nested within a record type, it doesn't work.

[Here](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/newtype_deriving.html#a-more-precise-specification) is the specification, if you're interested.

JacquesCarette: Ah, thanks for the investigation.

JacquesCarette:
Now needs some merge conflicts resolved before a merge.

PR #3858, opened 2024-07-18
Made `constDecDef` capitalize constants in Python

### LINKED ISSUE #2179: Change constDecDef implementation in Python
Currently `constDecDef`, GOOL's function for generating code for declaring and defining a constant, is implemented in Python by rendering just a plain variable declaration, since the concept of "constants" isn't actually supported by Python.

Since `constDecDef` is not actually doing what it is seems like it should in Python, we may want to change the implementation. We could:
1. Have the Python instance of `constDecDef` throw an error.
2. Follow the Python convention of naming variables that are supposed to be constant with all capital letters. In this case, the Python renderer would still generate regular variable definitions for `constDecDef`, but would rename the variable to be all caps.

I think I prefer option 2, because with option 1 we would not be able to generate Python code for any example where we have chosen `Const` as the `ConstantRepresentation`. Thoughts, @JacquesCarette and @smiths?

B-rando1:
It capitalizes them, and if the capitalized name shadows another name in the scope it throws an error.

You can see the changes to `HelloWorld.py`.  I also did a test where I declared a second constant `MYCONSTANT` (capitalized in the GOOL code), and the renderer threw an error, as intended.

Closes #2179


B-rando1:
I got everything working roughly how it should.  I'll just restate my couple of comments from the issue:

> - While the Python renderer does check to see if making a constant's name uppercase would conflict with any existing variable names, it does not check against variable names declared _after_ the constant. So it's still not perfectly safe. With that said, I think we do have all the tools we need to put in more rigorous checking for name conflicts, which would enable greater safety.
> - I'm not sure how I feel about having `staticVar` hold the original `variableName` while making the Doc that it uses to render itself all uppercase. It's probably mostly harmless, but there's always a chance of things going wrong down the road when we start having parts of the program lie to other parts. Again, it might be harmless, but it can be hard to tell.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs
```diff
@@ -535,7 +537,14 @@ instance DeclStatement PythonCode where
...
   arrayDecDef = listDecDef
-  constDecDef = varDecDef
+  constDecDef v e = do
+    v' <- zoom lensMStoVS v
+    let n = map toUpper (variableName v')
+        newConst = constant n (pure (variableType v'))
+    available <- varNameAvailable n
+    if available
+      then varDecDef newConst e
+      else error "Cannot safely capitalize constant."
```
samm82, line 547: Do we want this to be an error? We could just warn the user and potentially just use the lowercase version of it (assuming that _it_ is available)…

code/stable/pdcontroller/src/python/Constants.py
```diff
@@ -5,5 +5,5 @@
 
 ## \brief Structure for holding the constant values
 class Constants:
-    AbsTol = 1.0e-10
-    RelTol = 1.0e-10
+    ABSTOL = 1.0e-10
+    RELTOL = 1.0e-10
```
samm82, line 9: What I'd _really_ expect here is `ABS_TOL` and `REL_TOL`, to preserve the separation between the two parts of the variable name. Is there a reasonably straightforward way of doing this? (If not, this is likely getting "into the weeds"; NoPCM explicitly uses underscores for its variable names, so we may just want to do that for Projectile as well, potentially enforcing variable names to be all lowercase?)

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs
B-rando1: That's a good point.  It would be better to do it that way.  The only thing I'm not sure about is, how do we keep track of whether or not the renderer used the modified name?  When the `constant` function shows up later, it needs to know whether or not to modify the variable name.
- We _could_ have it infer whether the modified or unmodified name was used by checking `varNameAvailable (variableName v')` - if the original name was used, we can probably assume that it was used for the constant in question, else we must have modified it.  This would work, but it feels a little sketchy to me.
- A better way might be to have a mapping from GOOL variable name to rendered variable name floating around in the State.  Then in the `constant` function, we could look up the name to use, and not have to worry about any misplaced assumptions.  This would be a bit of extra work to implement, though.

What are your thoughts?  I'd rather not do option 2 until @JacquesCarette can take a look.  Would it be better to do option 1 for now, as it should be a small change?

code/stable/pdcontroller/src/python/Constants.py
B-rando1: That's a good point!  It should be possible to do better.  Off the top of my head, I think if we just insert a `_` wherever there's a lowercase letter on the left and an uppercase letter on the right, it would match what we expect in just about every case.  I'll see what functions are available, and get something working.

B-rando1: I made new function to do the renaming, and it handled the few examples we have well.

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs
B-rando1: As a side note, what's the best way to print a warning in Haskell?  I'm not actually sure if there's anywhere else in GOOL that currently prints warnings.  It seems like a shame to wrap all of GOOL in `IO String` just for this, is there a better way?

samm82: Honestly, I'm thinking that "error vs. warning" may be a variability we'd want to introduce at some point. Since this is likely going to be a bigger change, I would create a new issue for tracking our design decisions and working on it in a separate PR (if at all!)

No idea about your warning question 😅

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs
```diff
@@ -1032,3 +1032,10 @@ pyDocComment :: [String] -> Doc -> Doc -> Doc
 pyDocComment [] _ _ = empty
 pyDocComment (l:lns) start mid = vcat $ start <+> text l : map ((<+>) mid . 
   text) lns
+
+toConstName :: String -> String
+toConstName (s:s':ss) = if isLower s && isUpper s'
+                          then toUpper s : '_' : toUpper s' : toConstName ss
```
samm82, line 1038:
```suggestion
                          then toUpper s : '_' : s' : toConstName ss
```

Super minor refactoring that should work (😅), since we know that `isUpper s' == True`

Not sure if this was intentional, but as far as I can tell, this function would format both `myId` _and_ `myID` as `MY_ID`, which I think is ideal!

A nice next step would be handling strings of all caps in the input, but then the user input of either `inputHTMLcode` or `inputHTMLCode` would be ambiguous, so I think what you have is sufficient for now!

samm82 (approved):

code/drasil-gool/lib/GOOL/Drasil/LanguageRenderer/PythonRenderer.hs
B-rando1: Good catch!

Yeah true, I guess there'll always be a point where smart-renaming fails.  We _could_ do something like look for two capitals followed by a lowercase, and add an `_` between the two capitals.  That would work with `inputHTMLCode` (which seems more conventional to me than `inputHTMLcode`), but that might be a bit too 'clever' and lead to unexpected variable names under some conditions.

JacquesCarette (approved):

JacquesCarette:
Some conflicts need resolved before I can merge.

B-rando1:
Just to summarize the developments, there are a couple of things that aren't ideal with the current state of this implementation:
- We can't currently be certain when renaming variables that the new name doesn't create a conflict.  This is being discussed in #3867.
- If we detect a conflict, currently the best we can do is throw an error.  There are two reasons for this:
  - As far as I know, GOOL doesn't have the ability to do IO, so we can't issue warnings.  I _might_ be wrong here, but that's my current understanding.
  - Even if we were able to issue a warning (and pick a different name), we'd have no way of telling GOOL that we picked a different name.  For this we'd want to maintain a mapping from inputted variable name to rendered variable name in the `MethodState`.

Based on this, what do we want to do with this PR?  Is it good enough to merge in its current form, or would it be better to wait for now, and get these safety checks in place first?

JacquesCarette (approved):

JacquesCarette:
More conflicts to resolve.

I am happy to pull this in as-is, and defer the resolution of name conflicts to a different pass (i.e. that conflicts throw a hard error for now).

JacquesCarette (approved):

PR #3859, opened 2024-07-18
Removed outer parenthesis from `ifCond` in Python

### LINKED ISSUE #3855: Remove outer parenthesis in Python if-statements
I noticed that in the Python renderer, the conditional of an if-statement is wrapped in parenthesis.  E.g. in `HelloWorld.py`:
https://github.com/JacquesCarette/Drasil/blob/e8797118de04997a450458cbfd6e98f881559275/code/stable/gooltest/python/HelloWorld/HelloWorld.py#L166-L171

I was wondering if we want to remove them.  There are a few reasons why I think this would be a good change:
- For other structures (`for`, `while`, etc.), the Python renderer doesn't wrap the conditional in outer parenthesis. 
- Our Swift renderer doesn't wrap the conditional of `if` statements in outer parenthesis.
- It's been a while since I've used PyCharm, but I believe its linter tells you not to use outer parenthesis.
- I checked [PEP 8](https://peps.python.org/pep-0008/), but it doesn't seem to have any stance on this.  This StackOverflow [article](https://stackoverflow.com/questions/4740419/parentheses-in-python-conditionals#answer-59170904) seems to say the same thing, that PEP 8 doesn't give any rule, but it is more common to exclude the parenthesis.

B-rando1:
It was a _very_ simple change to make.  I tested most of the generated code, and it has the same behaviour.

Closes #3855


samm82 (approved):

JacquesCarette (approved):
Have to wonder if there are cases where this will go wrong (precedence and all that). But if the code works, this is indeed nicer.

Issue #2836, opened 2021-09-13
`scripts/pr_label_update.py` left unused (intended to note CI step failures on PRs via labels)


balacij:
Related to #2287 

As a result of our transition from Travis to GitHub Actions, `code/scripts/pr_label_update.py` file is unused. We should figure out a way to bring it back into usage.

balacij:
On second thought, when we're developing and we see CI errors, we typically jump to the logs or to the discussion. So, is this really needed now? Should we just simplify and close the ticket as 'won't do'?

smiths:
I'm fine with closing this ticket, but we should wait to see what @JacquesCarette says.

JacquesCarette:
What did `pr_label_update.py` do? (I'm not against deleting it, I just want to be more informed first).

balacij:
The script added labels to PRs for failed steps in the run. Since GitHub Actions is so integrated here, I don't think it's necessary. A quick peek into the logs will show the failed build step.

JacquesCarette:
I think the idea was to automate that "quick peek into the logs" - which still seems like a good idea! It would just need to be ported to be compatible with Github Actions, if that's possible.

balacij:
With #5271, our CI will be split into: Build & Test, Haddock, Compile GOOL, LaTeX, and mdBook (jobs). So, I'm not sure if resurrecting `pr_label_update.py` is necessary anymore. Instead of labels being added/removed from PRs (which can be noisy), we would just have an Actions history quickly viewable on the PRs themselves.

JacquesCarette:
I agree - please just delete (and then close this issue).

Issue #3756, opened 2024-05-28
Change Non-Functional Requirement (NFR) labels in the SRS to be consistent with the template


BilalM04:
This issue is a result of issue #3753 

Currently, the NFR labels are inconsistent between examples. For some examples, the NFR labels end in `"-able"`. Ex. maintainable, understandable, usable, etc. While for others, it ends in `"-ability"`. Ex. maintainability, understandability, usability, etc. 

Change the all the NFR labels to be consistent with the [Smith et al SRS template](https://github.com/smiths/capTemplate/blob/main/docs/SRS/SRS.pdf). This means we should be using the labels ending in `"-ability"`.

Change Should be made in the following examples:
- [ ] [dblpend](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/srs/DblPend_SRS.html#Sec:NFRs)
- [ ] [GlassBR](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/srs/GlassBR_SRS.html#Sec:NFRs)
- [ ] [SWHSNoPCM](https://jacquescarette.github.io/Drasil/examples/swhsnopcm/SRS/srs/SWHSNoPCM_SRS.html#Sec:NFRs)
- [ ] [PDController](https://jacquescarette.github.io/Drasil/examples/pdcontroller/SRS/srs/PDController_SRS.html#Sec:NFRs)
- [ ] [Projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:NFRs)
- [ ] [SglPend](https://jacquescarette.github.io/Drasil/examples/sglpend/SRS/srs/SglPend_SRS.html#Sec:NFRs)
- [ ] [SSP](https://jacquescarette.github.io/Drasil/examples/ssp/SRS/srs/SSP_SRS.html#Sec:NFRs)
- [ ] [SWHS](https://jacquescarette.github.io/Drasil/examples/swhs/SRS/srs/SWHS_SRS.html#Sec:NFRs)

smiths:
@BilalM04 I do not know how the NFRs are implemented, but if it is feasible, it makes sense for me for maintainability, usability etc to be concepts.   It seems odd to me that the concepts have different string outputs in the different examples.  They really should all be the same (even if we later change our mind on what they should be) across all examples.

BilalM04:
@smiths the NFRs are implemented as follows:

[...]

PR #3862, opened 2024-07-19
Issue #3757: Clarify portability requirements for Windows, Mac OSX, and Linux.

### LINKED ISSUE #3757: Fix the portability non-functional requirements descriptions
This issue is a result of issue #3753 

Fix the portability requirements. The current requirements are ambiguous. We shouldn't say the code can be ported to multiple operating systems, we should say which ones. In the future we might want to talk about the version of the big three OSs, but for now, let's say the code should be portable to work with Windows, Mac OSX and Linux.

Sample Description: "The code shall be portable to multiple Operating Systems, particularly Windows, Mac OSX, and Linux."

The code should be implemented similar to the [maintainability](https://github.com/JacquesCarette/Drasil/blob/5b5faadb1a130b24225106f4f71fcc9b65afe369/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs#L102-L113) NFR. Additionally the OSes should be parametrized, so that different examples can pick a subset of the three OSes.

TRISHI-L:
Contribute to #3756
Close #3757

**Changes:**
- Updated the portability requirements to explicitly state that the code should be portable to Windows, Mac OSX, and Linux.
- Modified the examples to ensure they reflect the updated requirements and parameterized the operating systems so that different examples can select a subset of the three specified operating systems.
- Change all of the labels to be consistent with the original Smith et al template for requirements. (Using ilities like Maintainability, Usability, etc.)

**Affected Examples:**

- DblPend
- GlassBR
- PDController
- Projectile
- SglPend


code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
```diff
@@ -111,6 +112,16 @@ mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
   S "assuming the same development resources are available"
   ]) lbl nonFuncReqDom
 
+mkPortableNFR :: String -> [String] -> String -> ConceptInstance
+mkPortableNFR refAddress osList lbl
+  | null osList = cic refAddress (S "No operating systems specified.") lbl nonFuncReqDom
```
BilalM04, line 117: We should probably error out on an empty list as it doesn't make sense to have a requirement with no operating system defined. That being said, @smiths would be a better source of reference for this.

BilalM04 (changes requested):
Looks good! You should also update stable by running `make stabilize` whenever there are changes to generated artifacts.

BilalM04:
**Minor note**: this PR should close #3757 and contribute to #3756.

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
samm82: I personally agree with this!

code/drasil-example/dblpend/lib/Drasil/DblPend/Requirements.hs
```diff
@@ -49,9 +50,7 @@ correct :: ConceptInstance
...
  atStartNP' (output_ `the_ofThePS` code), S "have the", 
  namedRef (propCorSol [] []) (plural propOfCorSol)]
- ) "Correct" nonFuncReqDom
+ ) "Correctness" nonFuncReqDom
 
 portable :: ConceptInstance
-portable = cic "portable" (foldlSent [
-  atStartNP (the code), S "is able to be run in different", plural environment])
-  "Portable" nonFuncReqDom
+portable = mkPortableNFR "portable" ["Windows", "Mac OSX", "Linux"] "Portability"
```
samm82, line 56: At some point, having "established" OSes that can be selected is probably a better idea, but that could potentially wait for a future PR to be done after more discussion.

samm82 (commented):
Other than @BilalM04's comments, this PR looks really good! Great job on your first one (assuming I haven't missed one 😅)

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
smiths: I agree as well.

smiths:
I will wait for stable to updated before reviewing.

TRISHI-L:
All suggested changes have been implemented and updated. Thank you!

TRISHI-L:
> I will wait for stable to updated before reviewing.

I have already made the updates. Thank you for your patience!

code/SRS/PDF/Makefile
samm82, line 1: Where did these files (i.e., in `code/SRS/`) come from? My guess is that this accidentally got added to a commit, since the changes to `code/stable` were also added to this PR.

samm82 (changes requested):

code/SRS/PDF/Makefile
JacquesCarette: I agree - these 5 extra files shouldn't be here.

code/drasil-example/dblpend/lib/Drasil/DblPend/Requirements.hs
```diff
@@ -49,9 +50,7 @@ correct :: ConceptInstance
 correct = cic "correct" (foldlSent [
  atStartNP' (output_ `the_ofThePS` code), S "have the", 
  namedRef (propCorSol [] []) (plural propOfCorSol)]
- ) "Correct" nonFuncReqDom
+ ) "Correctness" nonFuncReqDom
 
 portable :: ConceptInstance
```
JacquesCarette, line 55: This change makes it clear that several requirements (like `portable`) are repeated. As a subsequent PR, these should be pulled out of the example into `drasil-data`.

JacquesCarette (changes requested):
Looking good - one small thing to change, and some future work.

smiths:
The changes to stable look good to me.  (There are some other concerns mentioned above that still need to be addressed.)

@TRISHI-L can you please add a link to our [benefit of Drasil wiki page](https://github.com/JacquesCarette/Drasil/wiki/Benefits-of-Drasil) that points to this example and briefly discusses it?  (Could you also please fix the spelling mistake for "Cosmetic" in the other example at this page?)  When you are done this should be a nice example of how making a small change, like the name Portable to Portability is fixed everywhere in the documentation.  You don't edit the wiki directly, but rather edit the wiki files and create a pull request.

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
```diff
@@ -111,6 +112,16 @@ mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
...
+  | length osList == 1 = cic refAddress (S $ "The code shall be portable to " ++ head osList) lbl nonFuncReqDom
+  | otherwise = cic refAddress (foldlSent [
```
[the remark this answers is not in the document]
BilalM04: The case statement is better; but, you can avoid it completely by pattern matching on the function parameters directly. For example:

```Haskell
mkPortableNFR _          []     _   = ...
mkPortableNFR refAddress [os]   lbl = ...
mkPortableNFR refAddress osLIst lbl = ...
```

code/drasil-docLang/lib/Drasil/Sections/Requirements.hs
BilalM04: When you use `cic`, you are still creating a `ConceptInstance`. We want to use `error "text"` instead. Where "text" is your error message, in this case something like "No operating system(s) specified."

JacquesCarette (approved):

JacquesCarette:
Could @BilalM04 and @samm82 re-review please?

BilalM04 (approved):

JacquesCarette:
What I don't understand is that the build associated to this PR seems to have failed, even though it worked locally? @balacij can you look into that please?

BilalM04:
@JacquesCarette I noticed you mentioned @balacij, but I decided to investigate the build failure myself. The issue arises because mdBook was merged into the `main` branch, but the branch in this PR does not include those changes. As a result, the mdBook SRSs in stable are missing the updates from this PR. I’ve created #3874 to stabilize these changes, which should resolve the build failure.

balacij:
Yep, it looks like we had 2 branches that were merged into `main` which edited `stable/`. Specifically, mdBook's stable used and old version of the NFRs, while this branch used the new one.

When we merge commits into `main` that edit `stable/`, we need to make sure that all open PRs are synced with `main`. For files that are already tracked and conflicting, GitHub will report the _git_ conflicts on the PR. However, for new files, GitHub  doesn't find any conflicts with tracked files. So, we end up in an awkward situation where the CI hits an edge case, and fails.

I think there are a few things we can do:
1. Add a `git pull origin main` step early on in the CI to make sure that when PRs are opened/contributed to while `stable/` was potentially edited after opening/before a new contribution, we make sure to force errors to come up.
2. When a PR is merged (that affects stable), we need to re-run all CI-passing PRs that don't have a merge conflict with `main` to make sure that with `main` pulled in again, the PRs still pass the CI.

(2) might make (1) slightly redundant, but that's fine. 

Thinking out loud: with (2), to save CI minutes, we could split up our main Build & Test workflow into:
1. Build code, test `stable/` diff, test Haddock docs, test `stable-website/` diff.
2. If `stable/` has TeX changes, build TeX using `stable/` (not using our current `build/`-based builds!). 
3. If `stable/` has mdBook changes, build mdBooks with the same `stable/`-based scheme as (2).
4. If `stable/` has changes to generated code, build the generated code with the same `stable/`-based scheme as (2).

However, that leaves our deployment in an awkward position where it would either have to rebuild everything from 1-4 or use cached results from 2-4 (which is a bit more difficult). 

I like the idea of splitting up the CI into a few files (that would conveniently resolve #2836 for us too, I think), but I'm not quite happy with how website deployments would work later on. There would either be some redundancy or there would be hacky 'retrieve cached results or do the redundant rebuild again'-style work needed. I suppose the redundancy is better, however. 

Anycase, I've rambled enough here. Time for a new issue.

JacquesCarette:
Thanks for diving right in @BilalM04 !

@balacij I think your solution feels like overkill. This is rare enough that we can handle this via process rather than automation. i.e. when new PRs are merge that add new files to `stable/`, PRs after that must do a manual merge of `main` before being pulled in.

PR #3868, opened 2024-07-22
Switched the Julia renderer to `ProcProg`

B-rando1:
After all the changes in the past few weeks, GOOL is finally separated enough that I was able to move the Julia renderer from `OOProg` to `ProcProg`.

Notable changes:
- Created a new file, `AbstractProc.hs`, to hold any procedural-only functionality.  Currently it holds a few functions that have `OORenderSym` counterparts in `LanguagePolymorphic.hs` or `CommonPseudoOO.hs`, implemented under `ProcRenderSym`.
- Commented out all reference to Julia outside `drasil-gool` and `drasil-code/test`, as `drasil-code` doesn't know how to render `ProcProc` programs yet.
- Made `ProcProg` versions of all applicable test files in `drasil-code/test`.
- Made a `ProcProg` version of `CodeInfo.hs`.

Next steps:
- Continue resolving issues with the Julia renderer's implementation as they come up.  The larger ones have their own issues, but there are a couple still relevant smaller ones I mentioned in #3854:
  - Julia defaults to using 64-bit integers, so I went with that so that the code wouldn't have casts all over the place.  This does cause a potential discrepancy between Julia and other targets (except Python), though.
  - Julia makes a big deal about multiple-dispatch, and using generic types, sub-types, etc. seems to be a big part of that.  Since GOOL doesn't have  any of that, all type declarations need to be concrete, which isn't ideal.
- Teach `drasil-code` to generate `ProcProg` programs so that we can generate the projectile examples again.
  - There'll definitely be some code duplication due to the different types involved, but it's hard to say how much.  Because `OOProg` and `ProcProg` both inherit from `SharedProg`, we'll hopefully be able to reuse anything that is purely imperative.

Builds on #3876


JacquesCarette:
Can you rebase and deal with the merge conflicts first? Then I'll review.

B-rando1:
Sounds good.  Is it alright if I do the requested changes from #3854 here, since #3863 is merged already?

JacquesCarette:
Yes, sure.

JacquesCarette (commented):
This is looking excellent, but I don't really want to merge it in, as it is commenting out all Julia results. Was this on purpose?

B-rando1:
I'm glad you like it!

I commented out Julia because `drasil-code` doesn't know what to do with `ProcProg` yet, so it throws an error now if we try to generate Julia code.  we need to add support for `ProcProg` before before we can generate Julia code again.

code/stable/gooltest/swift/HelloWorld/main.swift
```diff
@@ -120,6 +120,7 @@ else if b == 5 {
     c -= 1
     b -= 1
     var myList: [Int] = []
+    var myObj: Character = "o"
```
samm82, line 123: Is this intentional or a side effect of something else? Not saying it's wrong, just curious

JacquesCarette:
> I commented out Julia because drasil-code doesn't know what to do with ProcProg yet, so it throws an error now if we try to generate Julia code. we need to add support for ProcProg before before we can generate Julia code again.

So maybe it makes sense to wait for that before pulling this in?

code/stable/gooltest/swift/HelloWorld/main.swift
B-rando1: That's there because before Julia, `HelloWorld.hs` was producing that line, using `objDecDef` or something like that.  When I added Julia, which does not implement `objDecDef`, I commented out the line since it's the only line in the program that isn't procedural-friendly.  Now that I've moved Julia over, there need to be two versions of `HelloWorld.hs` - one for OO languages and one for procedural ones.  Because of that, I removed the commented-out line in the procedural version , and uncommented it in the OO version.

B-rando1:
> So maybe it makes sense to wait for that before pulling this in?

I guess you're probably right.  That's the next big thing on my to-do list, so it hopefully won't be too long.

B-rando1:
I made the necessary changes to `drasil-code` to have us generating Julia code again.  This process wasn't as straightforward as I had hoped, and required making `SharedProg` versions of a few dozen `OOProg` functions in `drasil-code`.  The reason for this is that a lot of underlying types in Drasil have OO options, so the functions need to differ for those cases.  Still, I think we should be able to merge some of them by either making some assumptions or changing those types.  I'll keep you posted in #3871.

JacquesCarette (approved):

PR #3870, opened 2024-07-22
Fixed PDF deployment

### LINKED ISSUE #3704: SRS PDFs have disappeared from the website
I suppose the culprit is #3696. The solution should be not deleting the previous PDFs before what code previously re-built them. That means our old code was always building a fresh set of PDFs for each commit? It might be worth squashing some of the commit history for our `gh-pages` branch to save space.

BilalM04:
Closes #3704 

- Changed the conditions for when the TeX examples are compiled.
  1. On a pull request if there is a change in a TeX file within Stable.
  2. On a push to `main` (deployment environment).
- Separated the different SRS formats into their respective directories. Previously, all SRS artifacts were copied to `SRS/srs` directory within the `gh-pages` branch. Instead I separated it into `SRS/HTML` and `SRS/PDF`.
  - Updated the website with these link changes.
- Added `-j` flag when calling the `make` target to compile the TeX files.
  - The `-j` flag allows the parallel execution of jobs. This allows several TeX files to compile concurrently leading to faster times for the overall workfow.
  - [**No `-j` Flag**](https://github.com/BilalM04/Drasil/actions/runs/10048556021/job/27772779560): TeX takes ~13 min for a total workflow time of ~24 min
  - [**With `-j` Flag**](https://github.com/BilalM04/Drasil/actions/runs/10048133543/job/27771431028): TeX takes ~6min for a total workflow time of ~18 min

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -86,8 +86,8 @@ jobs:
         run: make codegenTest
 
       - name: "Compile generated TeX artifacts"
-        run: make tex SUMMARIZE_TEX=yes
-        if: steps.changes.tex.src == 'true'
+        run: make -j tex SUMMARIZE_TEX=yes
+        if: ${{ steps.changes.outputs.tex == 'true' || fromJSON(env.is_deployment) }}
 
       - name: "Compile generated software artifacts"
         run: make gool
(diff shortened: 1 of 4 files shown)


balacij (approved):
Great work, @BilalM04 ! :smile:

PR #3878, opened 2024-07-24
Adding Sets

### LINKED ISSUE #356: Implementing Discrete Sets
In commit 5876e62, `DiscreteI`, `DiscreteD`, and `DiscreteS` were created. Upon implementation, however, the following error was showing up: 
image. In commit 90ed590, a `Discrete` space was created similar to the `Vect` space, however, the same error showed up with this implementation as well...

The reason new Spaces were being created is because 
> type Set = Space

from (Expr.hs), and the sets for **nominal thickness** and **glass type** (see screenshot below) are not currently represented or stored as discrete sets (but rather are stored as lists of Strings/Double and displayed as Sentences).

image

@JacquesCarette @szymczdm 
If possible, can I get some clarification on what is the best way to implement such discrete sets in Drasil? Or are they already considered discrete sets from their current definition in the constraints of each `ConstrainedChunk`?
image

### LINKED ISSUE #2655: Add back `OneOf` Value constraints

NoahCardoso:
This pr adds sets as a constraint, space and set operations. It also adds sets in GOOL and the ability to check if an element is in a set. Closes #2655. Closes #356. 

Added
- Sets as a constraint for chunks
- Added Set operations in `drasil.lang` (Add, Remove, Union and Contains)
- Added Sets as a Space
- Updated physical and software constraints
- Added Sets in expr and set constructor 
- Changed `nomThick` back to contained input in `glassbr`
- Proper rendering of Sets 
- Added the `contains` and `in` operation to GOOL
- Added `litSet`, `printSet` and other Set support in GOOL
- Added Sets in all generated langs other than Julia 

Also
- Removed `DiscreteX` space types
- Added `OneOf` constraint to `glassType`


NoahCardoso:
@balacij in this branch I have `set`s working in Drasil and the ability to check for membership of a `set`

I would still like to:
-finish adding more `set` operations in GOOL like adding and union of `set`s
-finish assigning a `set` to a variable before use

Do you think I should focus on getting this pr merged first?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -471,10 +469,19 @@ vecVecNumBfunc Dot = error "convExpr DotProduct"
...
+elementSetSetBfunc SRemove = error "convExpr Removing an Element to a Set"
+
+-- Maps a 'ESSBinOp' to it's corresponding GOOL binary function.
+elementSetBoolBfunc :: (SharedProg r) => ESBBinOp -> (SValue r -> SValue r -> SValue r)
+elementSetBoolBfunc SContains = contains
+
 -- medium hacks --
 
 -- | Converts a 'Mod' to GOOL.
-genModDef :: (OOProg r) => Mod -> GenState (OO.SFile r)
```
JacquesCarette, line 477: I would want @B-rando1 to double-check this change.

JacquesCarette:
Can you please augment (i.e. edit) the description with *all* the items that you ended up adding? There is quite a lot of good stuff here, but we need to have a more external write-up of what you did (way more than just adding `Set` !).

code/drasil-gool/lib/GOOL/Drasil/ClassInterface.hs
```diff
@@ -0,0 +1,518 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module GOOL.Drasil.ClassInterface (
```
JacquesCarette, line 3: I want @B-rando1 to also check this. I wonder if this is somehow an artifact of merge order, i.e. this file has moved and shouldn't be re-added anew.

code/drasil-lang/lib/Language/Drasil/Expr/Class.hs
```diff
@@ -174,13 +177,15 @@ class ExprC r where
   
   -- | Smart constructor for indexing.
   idx :: r -> r -> r
-  
+
+  idxOf :: r -> r -> r
```
JacquesCarette, line 181: This needs a line of comment. I don't really know what it is / does / supposed to be.

code/drasil-lang/lib/Language/Drasil/Expr/Class.hs
```diff
@@ -114,7 +116,8 @@ class ExprC r where
   ($=>), ($<=>) :: r -> r -> r
   
   ($&&), ($||) :: r -> r -> r
-  
+
+  in' :: r -> r -> r
```
JacquesCarette, line 120: This also needs a comment. I think I know what it means (it is the set-theoretic membership relation), and I think I understand why it has a funny name (else it's a conflict with Haskell), but still, this should all be documented.

code/drasil-lang/lib/Language/Drasil/Expr/Class.hs
```diff
@@ -205,6 +210,7 @@ class ExprC r where
   -- | Create a matrix.
   matrix :: [[r]] -> r
 
+  sSet :: [r] -> r
```
JacquesCarette, line 213: Here too: what is an `sSet`?

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
```diff
@@ -388,9 +388,7 @@ <h2>Table of Symbols</h2>
                 </tr>
                 <tr>
                   <td><em>t</em></td>
-                  <td>
-                    Nominal thickness <em>t&thinsp;&isin;&thinsp;{2.5,2.7,3.0,4.0,5.0,6.0,8.0,10.0,12.0,16.0,19.0,22.0}</em>
-                  </td>
+                  <td>Nominal thickness</td>
```
JacquesCarette, line 391: Hmm, why did we lose some information from your changes? Did it just move?

code/stable/glassbr/src/cpp/InputParameters.cpp
```diff
@@ -236,6 +238,20 @@ void InputParameters::input_constraints() {
         std::cout << "." << std::endl;
         throw("InputError");
     }
+    if (!(set<double>{2.5, 2.7, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 12.0, 16.0, 19.0, 22.0}.find(this->t) != set<double>{2.5, 2.7, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 12.0, 16.0, 19.0, 22.0}.end())) {
```
JacquesCarette, line 241: This set shouldn't be inlined 3 times. The "error" here is in the specification, it should name this set. This shouldn't be done automatically by the framework which is likely doing exactly what it is being asked to do.

In fact, we should probably make it an error to output a concrete value of a compound type in this way.

code/stable/glassbr/src/cpp/input.txt
```diff
@@ -10,7 +10,7 @@
 1.0
 # glass type
 HS
-# nominal thickness t is in {2.5, 2.7, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 12.0, 16.0, 19.0, 22.0} (mm)
+# nominal thickness (mm)
```
JacquesCarette, line 13: There's loss of information here too.

JacquesCarette (changes requested):
This is generally fantastic work! It's quite a large piece of work that touches many different places - with quite clean code.

There are still a few things to settle, but this is a very impressive piece of work.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
B-rando1: It looks like the imports were changed up here:

I'm not sure, it seems like in general we avoid importing everything like that.  It might be better to only import what we need.  Especially when dealing with `Drasil.GOOL` and `Drasil.GProc`, which export a lot of things with the same name.

Either way, I would say that `SFile` should be imported qualified, so that all of its occurrences are either `OO.SFile` or `Proc.SFile`.

code/drasil-gool/lib/GOOL/Drasil/ClassInterface.hs
B-rando1: That would be my guess.  This file, as well as the folders it's in (`GOOL/Drasil` was switched to `Drasil/GOOL`), are from the older GOOL structure and should be removed.  It looks like it isn't referenced anywhere, so hopefully that's an easy step.

code/stable/glassbr/src/cpp/InputParameters.cpp
B-rando1: I know we discussed in #3869 what would be involved to create a variable above the `if`-statement.  The types and control-flow involved make it harder than it probably should be, but it's definitely doable.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
```diff
@@ -1246,14 +1261,7 @@ instance MathConstant CppSrcCode where
...
 instance VariableValue CppSrcCode where
-  valueOf vr' = do
-    vr <- vr'
-    its <- zoom lensVStoMS getIter
-    let namevr = variableName vr
-    if namevr `elem` its then
-      mkStateVal (iterator $ toState $ variableType vr)
-        (parens $ cppDeref <> RC.variable vr)
-      else G.valueOf vr'
+  valueOf = G.valueOf
```
B-rando1, line 1264: I'm just curious, what was the reason for this change?  To be honest I don't exactly know what the original code does. 
 It looks like it's checking for some edge case before calling `G.valueOf`.  Is it safe to get rid of the check?

code/stable/glassbr/src/cpp/InputParameters.cpp
JacquesCarette: I agree that this very hard to do post-facto, i.e. in GOOL itself. It should be easier to do in `drasil-code` and potentially should be done even further upstream than that.

Basically: sets should never be anonymous values. But we shouldn't fix that "just too late" but ask that whoever uses sets, does it 'right'.

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
balacij: I don't think it's something we're 'losing' necessarily, but moving to where it should be (or at least closer to where it should be, but at least in the form it should be in) -- a constraint. The reason why it was in the description of nominal thickness to begin with was because we previously removed 'enumerated reals/integers' constraints.

balacij:
Great work, @NoahCardoso ! :smile: I will wait until you've worked on the earlier feedback before I have a deeper look at it as well.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -102,13 +100,12 @@ tNT = uvc "tNT" (nounPhraseSP "TNT equivalent factor")
...
-    sfwrc $ Bounded (Inc, sy sdMin) (Inc, sy sdMax)] (exactDbl 45)) defaultUncrt
+    sfwrRange $ Bounded (Inc, sy sdMin) (Inc, sy sdMax)] (exactDbl 45)) defaultUncrt
 
 nomThick = cuc "nomThick" 
-  (nounPhraseSent $ S "nominal thickness" +:+ displayDblConstrntsAsSet 
-    nomThick nominalThicknesses)
+  (nounPhraseSent $ S "nominal thickness t is in" +:+ eS (Spc $ DiscreteD nominalThicknesses))
   lT millimetre {-Discrete nominalThicknesses, but not implemented-} Rational 
-  [{- TODO: add back constraint: enumc nominalThicknesses -}] $ exactDbl 8
+  [sfwrElem $ mkSet (map dbl nominalThicknesses)] $ exactDbl 8 -- for testing
```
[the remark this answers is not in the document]
JacquesCarette: Right, that's good -- but I don't see, other than in the code, where this information re-appears in the SRS itself. I only see it disappear!

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
```diff
@@ -388,7 +388,9 @@ <h2>Table of Symbols</h2>
                 </tr>
                 <tr>
                   <td><em>t</em></td>
-                  <td>Nominal thickness</td>
+                  <td>
+                    Nominal thickness t is in <em>{2.5,2.7,3.0,4.0,5.0,6.0,8.0,10.0,12.0,16.0,19.0,22.0}</em>
```
JacquesCarette, line 392: Hurray!

JacquesCarette (commented):
This is not final approval, but it is quite close.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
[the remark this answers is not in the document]
balacij: It probably would be good to show it somewhere in the SRS too. I think we have a constraints table that it is supposed to be shown in. @NoahCardoso can you confirm please? CCing @smiths because he probably has suggestions too.

code/stable/glassbr/src/cpp/input.txt
NoahCardoso: I managed to add it back to the definition of nominal thickness. However, as Jason mentioned above I am not sure if we still want it to be a part of the definition.

code/drasil-lang/lib/Language/Drasil/Expr/Class.hs
NoahCardoso: It was just the name that I chose initially to avoid any name conflicts, I changed it to `set'`, it creates a set.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
NoahCardoso: When I tried to use this it assumed the variable I wanted the valueOf was a pointer and this was causing issues so I changed it.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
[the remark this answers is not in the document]
NoahCardoso: I added back the constraint to the definition of nomThick

NoahCardoso:
I am having an issue in `Modules.hs` when I try to create the variables for the sets to be used in the conditional. This function is causing errors but I am unsure of how to proceed. @JacquesCarette @balacij @B-rando1 do you have any suggestions for what the issue might be?
https://github.com/JacquesCarette/Drasil/blob/9a2d15a0dda085e11919ee18ee4656f54ed56eb3/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs#L377-L382
https://github.com/JacquesCarette/Drasil/blob/9a2d15a0dda085e11919ee18ee4656f54ed56eb3/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs#L400-L406
image

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
B-rando1: It looks like this has been resolved.

B-rando1:
@NoahCardoso I just noticed that `stable/gooltest` needs to be updated (our CI doesn't check this, hopefully we get that working soon).  Could you run `make codegenTest_gen`, then copy `build/gooltest` into `stable/gooltest`?

code/drasil-gool/lib/GOOL/Drasil/ClassInterface.hs
NoahCardoso: I have removed the file and folder

code/stable/glassbr/src/cpp/InputParameters.cpp
NoahCardoso: Sets are now defined in a variable before being used

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -378,9 +395,9 @@ constraintViolatedMsg q s c = do
     print v,
     printStr $ ", but is " ++ s ++ " to be "] ++ pc
 
--- | Generates statements to print descriptions of constraints, using words and
```
JacquesCarette, line 381: Why did these comments get deleted?

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CommonPseudoOO.hs
```diff
@@ -571,7 +598,7 @@ listSize :: (CommonRenderSym r) => SValue r -> SValue r
 listSize l = do
   f <- S.listSizeFunc l
   mkVal (RC.functionType f) (RC.function f)
-
+
```
JacquesCarette, line 601: extra space should be deleted (but not the blank line)

code/stable/gooltest/java/HelloWorld/HelloWorld/HelloWorld.java
```diff
@@ -6,9 +6,9 @@
     \note Generated by Drasil v0.1-alpha
 */
 
+import java.util.*;
```
JacquesCarette, line 9: Is this really the right thing to import? Doesn't this make the imports below this redundant?

JacquesCarette (changes requested):
The code is excellent. Just a few minor things to take care of.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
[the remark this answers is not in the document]
smiths: I've looked at the current location of the `nomThick` information and it looks fine to me.  It would be nice in the future if our SRS let us define types and then nominal thickness could be a type.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
NoahCardoso: I am not sure, it may have happened during a merge conflict. I will add back these comments

code/stable/gooltest/java/HelloWorld/HelloWorld/HelloWorld.java
NoahCardoso: This is from old changes I made and is redundant now, I will remove it.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
B-rando1: From what I can tell, I believe you're right that we can safely remove that line.  It looks like you changed the C++ implementation of `forEach` (for the better imo) so that the check isn't needed.

I noticed though, with that change the `getIter` function isn't used at all across GOOL.  This leads me to believe that we can remove completely the `_iterators` field in [State.hs](https://github.com/JacquesCarette/Drasil/blob/8ca26be6df6d3f4c24627e76d051637b0ff85b16/code/drasil-gool/lib/Drasil/GOOL/State.hs#L124).  It's only used by C++, which doesn't seem to need it anymore.

That change might be somewhat large, so it might be better to look into it after this PR is merged.

B-rando1 (commented):
It all looks really good to me.  Just a couple of things I noticed.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/LanguagePolymorphic.hs
```diff
@@ -348,6 +352,9 @@ loopStmt = S.stmt . setEmpty
 emptyStmt :: (CommonRenderSym r) => MSStatement r
 emptyStmt = mkStmtNoEnd empty
 
+emptyValStmt :: (CommonRenderSym r) => MSStatement r
```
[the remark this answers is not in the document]
NoahCardoso: Its not used in GOOL, it is only used in drasil-code. Since emptyStmt didn't exist in interfaceCommon I added this so I could use it in drasil-code

code/drasil-gool/lib/Drasil/GOOL/InterfaceCommon.hs
```diff
@@ -349,6 +357,7 @@ type MSStatement a = MS (a (Statement a))
 class (ValueSym r) => StatementSym r where
   type Statement r
   valStmt :: SValue r -> MSStatement r -- converts value to statement
+  emptyValStmt :: MSStatement r
```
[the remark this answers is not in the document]
B-rando1: Ah, after looking more into it I see my assumptions weren't quite right.  `G.emptyStmt` isn't exported by `Drasil.GOOL`, and anyway it requires an instance of `CommonRenderSym`, which is internal to GOOL.

I also wasn't thinking about the `CodeInfo` instances.  Since they don't instantiate `CommonRenderSym`, their implementations _are_ different.

Honestly, I was just surprised we didn't already have something like this.  I think if we need access to an empty statement in `drasil-code`, this is the best way to do it.

I did have one other thought.  I'm not exactly sure if it would work, but could we just have [exc](https://github.com/JacquesCarette/Drasil/blob/fc1e8e48e80c40a4990f47930a2499cbe6342aea/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs#L377-L384) return `[]` instead of `[emptyValStmt]`?  I'm just thinking that if we can get away with not extending GOOL when there's another way, it might be nice.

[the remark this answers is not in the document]
NoahCardoso: I agree

NoahCardoso:
This is how the `contains` constraint looks in the generated code
image
image

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
JacquesCarette: That seems like a good way forward - could either @NoahCardoso or @B-rando1 write an issue for this, so it doesn't get lost?

B-rando1: Done!

code/drasil-gool/lib/Drasil/GOOL/InterfaceCommon.hs
```diff
@@ -294,6 +297,11 @@ class (ValueSym r) => List r where
   --   Arguments are: List, Value
   indexOf :: SValue r -> SValue r -> SValue r
 
+class (ValueSym r) => Set r where
+  --fromList :: SValue r -> SValue r -> SValue r
```
JacquesCarette, line 301: please delete dead/commented out code

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CLike.hs
```diff
@@ -88,6 +94,8 @@ andOp = andPrec "&&"
 orOp :: (Monad r) => VSOp r
 orOp = orPrec "||"
 
+inOp :: (Monad r) => VSOp r
+inOp = inPrec ""
```
JacquesCarette, line 98: A comment for why the empty string here is the right thing would be good.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -86,9 +85,10 @@ getActorName :: Space -> String
 getActorName (Actor n) = n
 getActorName _         = error "getActorName called on non-actor space"
 
--- | Gets the inner 'Space' of a vector.
+-- | Gets the inner 'Space' of a vector or set.
 getInnerSpace :: Space -> Space
 getInnerSpace (Vect s) = s
+--getInnerSpace (Set s) = s
```
JacquesCarette, line 91: Why is this commented out?

code/stable/glassbr/src/swift/InputParameters.swift
```diff
@@ -736,6 +736,26 @@ class InputParameters {
             print(".")
             throw "InputError"
         }
+        let set: Set<Double> = ["AN", "FT", "HS"]
```
JacquesCarette: Oops... `Set<Double>` definitely won't work here!

JacquesCarette (changes requested):

code/stable/glassbr/src/swift/InputParameters.swift
NoahCardoso: This was just changed in my last commit

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -284,6 +283,13 @@ genInOutFunc f docf n desc ins' outs' b = do
     bComms bothVs) bod else f inVs outVs bothVs bod
 
 -- | Converts an 'Expr' to a GOOL Value.
+convExprSet :: (OOProg r) => CodeExpr -> GenState (SValue r)
```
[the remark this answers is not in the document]
NoahCardoso: This function is no longer used, it has been deleted

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -366,6 +381,15 @@ constrExc c = do
   msgs <- mapM (constraintViolatedMsg q "expected") cs
   return $ map (bodyStatements . (++ [throw "InputError"])) msgs
 
+exc :: (OOProg r) => (CodeVarChunk, ConstraintCE) ->
+  GenState [MSStatement r]
+exc (v, Elem _ e) = do
```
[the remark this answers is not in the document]
NoahCardoso: I can pattern match on `e` to check that it is a `Set` then otherwise return an error

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -332,18 +334,30 @@ physCBody cs = do
...
+interleave xs [] = xs
+interleave (x:xs) (y:ys) = x : y : interleave xs ys
+
 -- | Generates conditional statements for checking constraints, where the
 -- bodies depend on user's choice of constraint violation behaviour.
 chooseConstr :: (OOProg r) => ConstraintBehaviour ->
   [(CodeVarChunk, [ConstraintCE])] -> GenState [MSStatement r]
 chooseConstr cb cs = do
+  let ch = transform cs
+  varDecs <- mapM exc ch
```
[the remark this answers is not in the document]
NoahCardoso: This is the only time `exc` is used. It was originally intended to work like printConstraint.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
JacquesCarette: I feel that this shouldn't error-out at all, because whoever calls `exc` should already know it's fine, and thus only pass in the necessary information (so that `exc` doesn't have to pattern match again)

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CLike.hs
NoahCardoso: This was originated used in python but this has been deleted

code/drasil-lang/lib/Language/Drasil/Space.hs
NoahCardoso: I was testing in Space and must have forgot about this, it has been added back

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
NoahCardoso: I see. Are you saying that `exc` should be directly given the `set` or that it should be able to work if the chunk is constrained to be an `Elem` of something other than a `set`?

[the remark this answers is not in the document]
JacquesCarette: `exc` should be given the `set`. Whether you want to make it more general is for you to decide if that would be useful.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -354,6 +363,10 @@ convExpr (Matrix [l]) = do
                                     -- hd will never fail here
   return $ litArray (fmap valueType (head ar)) ar
 convExpr Matrix{} = error "convExpr: Matrix"
+convExpr (S.Set l) = do
```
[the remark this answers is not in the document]
NoahCardoso: It could be added to `Set`s I'm just not sure if it is necessary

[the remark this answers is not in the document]
JacquesCarette: I'm not sure either... but doing (late) type reconstruction is usually a bad idea.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
JacquesCarette: You would need the single caller of `exc` to do that job itself (rather than a separate function). This change should "propagate upwards" through the code.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -381,26 +406,31 @@ constraintViolatedMsg q s c = do
 -- | Generates statements to print descriptions of constraints, using words and
 -- the constrained values. Constrained values are followed by printing the
 -- expression they originated from, using printExpr.
-printConstraint :: (OOProg r) => ConstraintCE ->
+printConstraint :: (OOProg r) => CodeVarChunk -> ConstraintCE ->
```
JacquesCarette, line 409: This feels like a bad change. For one, you're forcing a dummy `CodeVarChunk` be created when you're printing a `Range`. Second, you don't want a `CodeVarChunk`, all you want is a name -- why don't you instead ask for a name?

This is also a case where you might want to look upstream at who calls `printConstraint` and see if the design of all of this is correct.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CLike.hs
```diff
@@ -72,6 +72,12 @@ listType lst t' = do
   typeFromData (List (getType t)) (lst 
     `containing` getTypeString t) $ text lst <> angles (RC.type' t) 
 
+setType :: (OORenderSym r) => String -> VSType r -> VSType r
```
JacquesCarette, line 75: What is this function for?

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
NoahCardoso: I can change it to use a String not a CodeVarChunk since I still need some way to get the set's name.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
NoahCardoso: Something like this?
image

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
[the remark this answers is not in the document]
NoahCardoso: Is there any other way to do the type reconstruction other then giving a space to set?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
JacquesCarette: Something like this. But it might make sense to filter `ch` for `Elem` instead of returning all those `emptyStmt`? In fact, it might make sense to filter *cs* (which is the raw-er input to `chooseConstr`) instead.

Don't think of it as "getting the job done" but rather "putting the code in the most logical place". You're putting the code a little too late in the logic.

Also, you should go back to the top of this particular sub-conversation, which was about digging out the inner list!

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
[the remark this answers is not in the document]
JacquesCarette: Sorry, I was too obtuse: you shouldn't be doing 'type reconstruction' at all. You should know, right when the set is created, what kind of thing it is going to hold. You shouldn't be asking the "values" in the set what type they are, because you should already know it.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
JacquesCarette: `String` would be a good start. Using `UID` for digging out a `String` is an anti-pattern, BTW. That's what things like `shortname` are for.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CLike.hs
JacquesCarette: Makes sense. But also see above comments about figuring out the inner type.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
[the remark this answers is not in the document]
NoahCardoso: If I change `set` to 'Set :: Space -> [Expr] ->Expr' I can use the `Space` to store the type of the set

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
NoahCardoso: If I don't add `emptyStmt` I would have to define all of the set variables before the block of if statements. Otherwise, the order in which the set variables are declared will not line up with the order that the constraints appear.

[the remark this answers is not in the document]
NoahCardoso: With regards to your top comment, why would I need to pattern match on `e`?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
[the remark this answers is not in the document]
JacquesCarette: Yes, that is what I was suggesting.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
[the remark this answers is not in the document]
JacquesCarette: Re: `emptyStmt` - that makes sense, but the generation of "dummy code" still worries me. Does anything match on `emptyStmt` later on to choose not to emit anything? That would be bad. The information about whether to emit a declaration or not should be directly visible instead of being indirectly encoded.

The pattern match on `e` seems to happen later in the code, when it tries to figure out if things are an explicit list or something else. Again that knowledge should be "more visible" in the data itself.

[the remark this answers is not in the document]
NoahCardoso: The `emptyStmt` is only used once and I change `exc` to `constrVarDec` which experts a `set` and it doesn't use pattern matching. Would it be better to declare all of the variables beforehand so that having to emit a declaration is unnecessary?

JacquesCarette (changes requested):
A few changes still to go (and a merge too.) Getting very close.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -399,6 +410,9 @@ renderRealInt s (UpTo    (Exc, a))          = sy s $<  a
 renderRealInt s (UpFrom  (Inc, a))          = sy s $>= a
 renderRealInt s (UpFrom  (Exc, a))          = sy s $>  a
 
+renderSet :: (HasUID c, HasSymbol c) => c -> CodeExpr -> CodeExpr
+renderSet e s = in' (Variable ("set_" ++ showUID e) s) (sy e)
```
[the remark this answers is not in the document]
NoahCardoso: I changed it so that it gets the `Symbol` of c for the name rather than the `UID`

code/stable/glassbr/src/cpp/InputParameters.cpp
```diff
@@ -238,29 +238,29 @@ void InputParameters::input_constraints() {
         std::cout << "." << std::endl;
         throw("InputError");
     }
-    set<string> set_glassTy = {"AN", "FT", "HS"};
-    if (!(set_glassTy.find(this->g) != set_glassTy.end())) {
+    set<string> set_g = {"AN", "FT", "HS"};
```
JacquesCarette, line 241: wait, this diff seems to go in the opposite direction than I expected? `glassTy` is nicer than `g`? 

Now, it may well be that `glassTy` was the UID and `g` is the symbol name - is that so?

JacquesCarette:
Also, I think that `force-push` is a bad idea, it makes the diffs all weird.

code/stable/glassbr/src/cpp/InputParameters.cpp
NoahCardoso: Yes that is the case

JacquesCarette: This feels like a regression to me. *However* you are doing what is being asked of you in this PR, so it's not "your fault", as it were.

So: I'll let this PR in as is, but could you please open an issue where we discuss these names? We have different ways of associating names to chunks, and we might want to use a different one here (UID was definitely wrong, `symbol` is better for sure, but maybe `shortname` would be better still, but that is not in-scope for this PR).

JacquesCarette:
There's a new conflict - can you fix it? And then, as I said above, once I see the issue has been created, I'll merge this.

code/stable/glassbr/src/cpp/InputParameters.cpp
NoahCardoso: I opened an issue. I think it would be best to use `shortname`. I also don't think it would take very long to do.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -471,6 +489,15 @@ vecVecNumBfunc Dot = error "convExpr DotProduct"
 numVecVecBfunc :: NVVBinOp -> (SValue r -> SValue r -> SValue r)
 numVecVecBfunc Scale = error "convExpr Scaling of Vectors"
 
+-- Maps a 'ESSBinOp' to it's corresponding GOOL binary function.
+elementSetSetBfunc :: ESSBinOp -> (SValue r -> SValue r -> SValue r)
+elementSetSetBfunc SAdd = error "convExpr Adding an Element to a Set"
```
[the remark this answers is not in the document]
NoahCardoso: Yes since adding sets has not been implemented yet

JacquesCarette (approved):

JacquesCarette:
Almost there... there's a conflict to resolve!

JacquesCarette (approved):

PR #3881, opened 2024-07-25
Updated `Mul` operator rendering

### LINKED ISSUE #468: Rendering multicharacter symbols in LaTeX
For symbols that consist of multiple characters, like min, max, MINFRACT, etc, when they are rendered in LaTeX equations they would look better in an \mbox.  Rather than LaTeX interpreting min as m times i times n, a \mbox{min} would achieve a more natural spacing.

BilalM04:
Closes #468 

Modified the `Mul` operator to be rendered as a thin space (`\,`). See https://github.com/JacquesCarette/Drasil/issues/468#issuecomment-2248881260 for explanation.


BilalM04:
@smiths would this be a good example to put in [Benefit of Drasil](https://github.com/JacquesCarette/Drasil/wiki/Benefits-of-Drasil). This simple change resulted in ~1360 modification in 67 files.

JacquesCarette (approved):

smiths:
Yes, @BilalM04 this is a great example to add to the Benefits of Drasil wiki!

PR #3885, opened 2024-07-25
Fixed placement of figures in TeX PDF

### LINKED ISSUE #3882: Figures in the PDF SRS are Scattered
The figures in the PDF SRS are scattered. In LaTeX, figures can sometimes float to unexpected locations because LaTeX tries to optimize the layout. I've noticed that often times figures get placed in a location that disrupts the flow of the document. For example, the trace graphs in [GlassBR](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/PDF/GlassBR_SRS.pdf) get placed in the middle of a list, same for the system context figure in [Projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/PDF/Projectile_SRS.pdf).

One possible option is to use the `float` package. We would add the following to the preamble:

```LaTeX
\usepackage{float}
```

Then modify the code for figures by adding a placement specifier. In this case, `H`, for "here".

```LaTeX
\begin{figure}[H]
...
\end{figure}
```

I manually tested the above code for GlassBR and it seems to work quite nicely.

BilalM04:
Closes #3882 

Fixed the placement of figures in the TeX/PDF by using the `float` package.


JacquesCarette (approved):
Nice.

Issue #3821, opened 2024-06-26
Local scope in Julia


B-rando1:
I've been working some more on the Julia render, and I discovered that it's stricter with local scope than other languages.  A minimal example is the following:

```julia
a = 0

while a < 5
    println("a = $a")
    a += 1;
end
```

This throws an error, because the `while` loop introduces a 'soft local' scope (see [here](https://docs.julialang.org/en/v1/manual/variables-and-scoping/#local-scope) for details), and local scopes cannot refer to global variables without explicitly stating that they are.  There are two ways to get the intended behaviour:

###### 1. Using `global` declaration
```julia
a = 0

while a < 5
    global a
    println("a = $a")
    a += 1;
end
```

###### 2. Using a function to define functions in a local scope
If variables are defined in a local scope, then nested local scopes are free to access them.
```julia
function main()
    a = 0

    while a < 5
        println("a = $a")
        a += 1;
    end
end

main()
```

#### Analysis
Both options have difficulties.
- Option 1 is more straightforward, and probably slightly more idiomatic as well.  We might be able to use `static` vs `dynamic` to keep track of which variables are global vs local.  I think that this wouldn't work for any logic occurring in the global scope, though.
- Option 2 would make it easy to get rid of global variables in program files, but modules would still have global variables to deal with.
- I think if we combine both options, there might be something there.  This will require a bit more investigation into what information GOOL stores about variables and how it uses that information.  It will also require understanding how `State` can be used to add `global` declarations to the top of a scope, similar to how we do imports currently.

balacij:
It sounds like option 1 is the 'right' choice in the sense that if we go with option 2, we wouldn't be able to generate purely executable Julia scripts. Am I understanding this correctly?

B-rando1:
> It sounds like option 1 is the 'right' choice in the sense that if we go with option 2, we wouldn't be able to generate purely executable Julia scripts. Am I understanding this correctly?

I'm not quite sure what you mean.  As long as we add `main()` to the bottom of the file, it would still run with the same behaviour, i.e. an outside user wouldn't be able to tell that the program is inside a function.

If you mean that scripts require a main function and won't run in the global scope, then you are correct.  This might conflict with #3512, but at the same time it's no different than the C++ target.  Let me know if you think this is something we should avoid.

[...]

PR #3888, opened 2024-07-26
Finished tracing out `scope` to its sources in `drasil-code`

B-rando1:
For the most part this went pretty smoothly.  There was definitely ample opportunity for mistakes though, so hopefully we can find a way to spot them.

In order to test them, I'm thinking of modifying one of the renderers to prepend `global.`, `local.`, or `mainFn.` to each variable.  This will let us test everything and give us some empirical assurance.

Contributes to #3821

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
@@ -32,30 +32,30 @@ import Control.Monad.State (get)
 -- | Generates calls to all of the input-related functions. First is the call to
 -- the function for reading inputs, then the function for calculating derived
 -- inputs, then the function for checking input constraints.
-genAllInputCalls :: (OOProg r) => GenState [MSStatement r]
-genAllInputCalls = do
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):
So I'm approving, in that the code is clean and works, but I'm not going to merge as I want a further investigation first: since this is all done in `GenState`, maybe this could / should be done that way? i.e. add a notion of scope to `DrasilState` and then use get/set to work with them. That would end up requiring many fewer code changes.

B-rando1:
That's a good point.  I had it in the back of my mind that there's probably a better way to do this, but I guess I didn't want to lose momentum.  I agree that integrating it into `GenState` would lead to a cleaner result.

B-rando1:
I implemented these changes using `GenState`, as you recommended.  The PR is here: #3899.

balacij:
Should we delete the branch? It has unmerged commits.

B-rando1:
That should be fine.  It was a step in the wrong direction, so we don't want the changes.  It was helpful in that it helped me know where the scope needs to be changed in #3899, but since that's done we should be good to delete it.

PR #3890, opened 2024-07-26
Created `ScopeData` for Julia's Scope

B-rando1:
I created `ScopeData`, which is intended to hold information about a variable's scope.  I also switched Julia's `ScopeSym` representation to use it.

Currently nothing is done with it, but I believe we need to integrate it into the AST for Julia, so that we can get the information back out of the variable as needed.

Contributes to #3821

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
@@ -32,30 +32,30 @@ import Control.Monad.State (get)
 -- | Generates calls to all of the input-related functions. First is the call to
 -- the function for reading inputs, then the function for calculating derived
 -- inputs, then the function for checking input constraints.
-genAllInputCalls :: (OOProg r) => GenState [MSStatement r]
-genAllInputCalls = do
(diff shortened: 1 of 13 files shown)


JacquesCarette (commented):
While this seems reasonable, as it is built on top of a PR that I'm not sure is the right way to go, I won't approve until that is investigated.

This change is quite likely independent of that one, and should likely have been done that way.

B-rando1:
You're right, this should have been on its own branch from the start.  I've rebased it to `main`.

B-rando1:
I added `ScopeData` to the other renderers as well, abstracting the functions into their appropriate places.  I was initially hoping that we could just use `ScopeData` in the Julia renderer where it's used, but it looks like because we need it in the state, all of the renderers need to use the same type.
- We need variables' scope to be in the state so that we can extract a variable's state when creating a new variable in a few places.  E.g.
https://github.com/JacquesCarette/Drasil/blob/150ea021ceba51074ed98621e13985b3023053a3/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JuliaRenderer.hs#L706

JacquesCarette (approved):

PR #3897, opened 2024-07-30
Add `weeder` to make file

NoahCardoso:
Contributes to #3710. Adds `weeder` to make file. Also updated `ghc 9.2.8` to `ghc 9.6.6` as weeder does not support `ghc 9.2.8`. When you run `weeder` it gives you a list of potential unused code. It also gives you error 228, meaning that it found multiple places with "weeds" which it thinks could be unused code.

### DIFF
--- a/code/stack.yaml
+++ b/code/stack.yaml
@@ -74,7 +74,7 @@ extra-deps:
 - unicode-properties-3.2.0.0@sha256:239766a6ac4322329353f6b9cd546024fd8c5f0235c8e32b3cba2d6bd245a699,1000
 - transformers-0.6.1.1@sha256:80d128763c6965dfcddca2155b3dec3bb1970d476f0c8cceb4bb28608ab75e7c,3243
 
-# The following packages are dependencies needed for 'stan', specifically tested against GHC 9.4.8.
+# The following packages are dependencies needed for 'stan' and 'weeder', 'stan' specifically tested against GHC 9.4.8.
 - clay-0.14.0@sha256:a50ba73137a39c55e89f24a7792107ec40ba07320b2c5ff7932049845c50ffc9,2204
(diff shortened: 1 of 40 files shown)


balacij:
There is a script in the scripts folder  that should be used to update our GHC version or else things will break eventually in the CI. Can you please use that?

NoahCardoso:
I am not sure which scripts need to be changed in order to update the ghc version

code/stack.yaml
```diff
@@ -15,7 +15,7 @@
 # resolver:
 #  name: custom-snapshot
 #  location: "./custom-snapshot.yaml"
-resolver: lts-20.26
+resolver: lts-22.31
```
JacquesCarette, line 18: the build weirdly broke, and I suspect that this may have something to do with it?

code/stack.yaml
```diff
@@ -92,6 +100,8 @@ allow-newer: true
 # In this case, it's allowing newer versions of the 'stan' package.
 allow-newer-deps:
 - stan
+- wedder
```
JacquesCarette, line 103: Also: this is a typo! Should be 'weeder'

JacquesCarette (changes requested):

code/stack.yaml
NoahCardoso: I changed this since weeder is not supported on `lts-20.26`

JacquesCarette: I'm not saying that you shouldn't have changed this (clearly, you have no choice), just that this is something to investigate as a potential source of breaking things -- the error message was quite obscure. It might have been because of the typo below too!

NoahCardoso: I also updated the version using the scripts so it should be good now

JacquesCarette (approved):

Issue #3252, opened 2023-01-31
Rework `Sentence`s to build them "properly"


samm82:
In #3199 and #3237, the `Sentence`s used for the new constructors were just done by wrapping the string with `S`. These now need to be reworked to use our concepts appropriately

balacij:
Can you please clarify how we should use the new concepts appropriately?

samm82:
Sure! For example, the following sentence (for its definition; `S "the ratio ..."`) is simply a string wrapped in a `Sentence` constructor, but it should be built similarly to how other `Sentence`s are built (for example, it should use the `glass` concept, shown below, using the `phrase` function). This ensures that we are using our concepts consistently and makes the document more maintainable.

https://github.com/JacquesCarette/Drasil/blob/24046be82e94e62acba7ccc73169c76a21de73b2/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L28-L30

https://github.com/JacquesCarette/Drasil/blob/24046be82e94e62acba7ccc73169c76a21de73b2/code/drasil-example/glassbr/lib/Drasil/GlassBR/Concepts.hs#L62

As for the specifics on how to build the sentences, there are numerous examples in the other examples of how to do this. The main ones I use are `+:+` (for joining two `Sentence`s with a space), `:+:` (for joining two `Sentence`s without a space), `+:+.` (for joining two `Sentence`s with a space and adding a period to the end), and `foldSent` and its variants, which converts a `[Sentence]` to a `Sentence`; this one uses `+:+` between each `Sentence` and adds a trailing period.

daijingz:
Ready

daijingz:
@samm82 I see your meaning. So we need to use the given classes' functions to define strings, not just directly use `S "..."`. However, what is the range of my changes? I see the String `S...` occurs a lot of places, including websites and different examples. 
![JPNBXM}@3R))TA}UY)0GQX1](https://github.com/JacquesCarette/Drasil/assets/61481741/0c5746dd-e975-4d89-85f3-8f428ed4cb79)
Do I only need to change the `glassbr` example, or all examples? Do I need to change String examples in the website code?

@JacquesCarette Before I have already created a branch to include `glassbr` changes (not all changes), since its volume is heavy. Like other issues, I think I have been very closed to the solution. (Have not been submitted)

samm82:
[...]

PR #3902, opened 2024-08-01
Update stack.yaml

daijingz:
#3252 
Update stack.yaml to fix one compatible error.
Errorenous Command: `make dblpend_diff`

Error message: 
```
Error: [S-4804]
       Stack failed to construct a build plan.

       While constructing the build plan, Stack encountered the following
       errors:

       In the dependencies for MissingH-1.6.0.1:
           time must match >=1.4 && <1.15, but this GHC boot package has been
                pruned (issue #4510); you need to add the package explicitly to
                extra-deps (latest matching version is 1.14)
       needed due to dblpend-0.1.24.0 -> MissingH-1.6.0.1

       In the dependencies for aeson-2.1.2.1:
           time must match >=1.6.0.1 && <1.13, but this GHC boot package has
                been pruned (issue #4510); you need to add the package
                explicitly to extra-deps (latest matching version is 1.12.2)
       needed due to dblpend-0.1.24.0 -> aeson-2.1.2.1

       In the dependencies for directory-1.3.8.5:
           Win32 must match >=2.13.3 && <2.15, but this GHC boot package has
                 been pruned (issue #4510); you need to add the package
                 explicitly to extra-deps (latest matching version is 2.14.0.0)
           time must match >=1.8.0 && <1.15, but this GHC boot package has been
                pruned (issue #4510); you need to add the package explicitly to
                extra-deps (latest matching version is 1.14)
       needed due to dblpend-0.1.24.0 -> directory-1.3.8.5

       In the dependencies for drasil-gen-0.1.3.0:
           time needed, but this GHC boot package has been pruned (issue #4510);
                you need to add the package explicitly to extra-deps (latest
                matching version is 1.14)
       needed due to dblpend-0.1.24.0 -> drasil-gen-0.1.3.0

       In the dependencies for hslogger-1.3.1.1:
           time must match >=1.2 && <1.15, but this GHC boot package has been
                pruned (issue #4510); you need to add the package explicitly to
                extra-deps (latest matching version is 1.14)
       needed due to dblpend-0.1.24.0 -> hslogger-1.3.1.1

       In the dependencies for process-1.6.20.0:
           Win32 must match >=2.4 && <2.15, but this GHC boot package has been
                 pruned (issue #4510); you need to add the package explicitly to
                 extra-deps (latest matching version is 2.14.0.0)
       needed due to dblpend-0.1.24.0 -> process-1.6.20.0

       In the dependencies for time-compat-1.9.6.1:
           time must
                match >=1.5 && <1.7 || >=1.8 && <1.9 || >=1.9.2 && <1.9.4 || >=1.10 && <1.10.1 || >=1.11 && <1.11.2 || >=1.12 && <1.13,
                but this GHC boot package has been pruned (issue #4510); you
                need to add the package explicitly to extra-deps (latest
                matching version is 1.12.2)
       needed due to dblpend-0.1.24.0 -> time-compat-1.9.6.1
make: *** [Makefile:305: dblpend_gen] Error 1
```

This error leads to unsuccessful `make` compiling.

Previous Fixing Attempts:
1. Use the `stack clean` command, then re-make.

My final solution is adding two dependency lines inside the "stack.yaml" file.

@JacquesCarette This problem occurs after updating the stack version, and it is serious.


JacquesCarette (commented):
I don't know if this is a good change on or. @balacij (and @BilalM04 ) may know better.

daijingz:
@JacquesCarette @balacij @BilalM04 This improvement is not unnecessary, because I just confronted this compilation problem after updating the stack version. Without this change, my compilation cannot get success. 
I feel strange about why nobody else has the same problem.

BilalM04:
The make targets seem to work for me just fine. I'm not sure what could be causing this issue.

code/stack.yaml
```diff
@@ -94,6 +94,8 @@ extra-deps:
 - unix-2.8.5.1@sha256:3f702a252a313a7bcb56e3908a14e7f9f1b40e41b7bdc8ae8a9605a1a8686f06,9808
 - Cabal-3.12.1.0@sha256:08be296ddd941b3f9be4bd125fbb3e6857f5c707b8d10464f6f1d30c91ca3e5f,13551
 - Cabal-syntax-3.12.1.0@sha256:6dbb06fb08ff77520947fb4c1ef281c9cea5b8dc7fd9a41ad62273fa916cf4b2,7407
+- time-1.12.2@sha256:88e8493d9130038d3b9968a2530a0900141cd3d938483c83dde56e12b875ebc8,6510
+- Win32-2.14.0.0
```
balacij, line 98: @daijingz is on a 32bit machine. I wouldn't be surprised if that is the discrepancy between your two machines.

Considering [Win32](https://hackage.haskell.org/package/Win32) is an official library of Haskell.org, I think it's okay to add it without too much concern. That being said, we should also pin the checksum in this line.

samm82: I can confirm that I had this same issue, even though I'm on a 64-bit machine (I can drop more specs if that would help).

The specific line that was recommended to me was the following (my time-1.12.2 line is the same as @daijingz's):
```yaml
- Win32-2.14.0.0@sha256:e34af84fec733b5c0c8f052ec39499785e719e2fbbe308983adf26c82ea3704d,5942
```

JacquesCarette: Thanks for making this more precise. So I'll accept this one it is pinned.

daijingz: Updated.

JacquesCarette:
I don't really have the expertize to review this - I'll wait for @balacij and @samm82 .

samm82 (approved):
This works for me!

JacquesCarette (approved):

PR #3903, opened 2024-08-01
Used the `Scope` in generated Julia code

B-rando1:
This shows what we are currently able to do with the `Scope` given.  We still need to work out some decisions in #3899, but this indicates that the general direction seems to be right.

#### Changes:
- I made it so that `constDecDef` uses the `const` keyword when in the `global` scope.
- I also made it so that whenever assigning, incrementing, or decrementing a variable, if that variable is a `global` variable, the `global` keyword is used with it.
  - This is a bit overkill, and the generated code doesn't look very idiomatic because of it.  It's a good start though, and it's enough to prevent `HelloWorld.jl` from crashing when run.

#### Next steps:
- As I mentioned, we are adding way more `global` declarations than we need to.  We only need a `global` declaration if we are assigning to a `global` variable from a `local` context.  Also, I think it would look better if we add the `global` declaration once, at the top of the outermost `local` scope, rather than inline with the assignment.
- The problem is that GOOL doesn't seem to differentiate very well what kind of scope it's in.  Functions and classes are fine, but loops are just `MSStatement`s, which are too general.

Builds on #3891
Contributes to #3821

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
@@ -32,30 +32,30 @@ import Control.Monad.State (get)
 -- | Generates calls to all of the input-related functions. First is the call to
 -- the function for reading inputs, then the function for calculating derived
 -- inputs, then the function for checking input constraints.
-genAllInputCalls :: (OOProg r) => r (Scope r) -> GenState [MSStatement r]
-genAllInputCalls scp = do
-  gi <- genInputCall scp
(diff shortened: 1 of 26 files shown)


B-rando1:
From our discussion elsewhere (see my [comment](https://github.com/JacquesCarette/Drasil/issues/3821#issuecomment-2263077522)) there's a chance that we need to rethink some of the details of this.  It's probably smart not to merge any changes until we know what direction we're taking.

B-rando1:
From our recent discussions, these changes don't seem to be relevant anymore.  The equivalent of these changes were reproduced in #3899.

Issue #3524, opened 2023-07-10
Does `uid` have to be a lens?


hrzhuang:
`uid` being a lens means that the UID of all instances of `HasUID` can be both read and written to. I `grep`ed for all occurrences of `uid` in the code base and the only place where it is actually used to set the UID is `addRelToCC`, which is easy to remove.

Making `uid` read-only would allow some UIDs to be computed on the fly. For example, currently the UID of a data definition is the same as the UID of the underlying concept, and in the meeting we decided that we should change the UID of the data definitions (see #3518 for a bit of context). However, the type `DataDefinition` doesn't have a field for its own UID, its `uid` simply points to the `uid` of the underlying concept contained inside. Allowing computed UIDs would enable us to prepend something like `dd:` or `theory:` to the UID of the underlying concept in the `uid` of a data definition.

balacij:
It's definitely an oddity. Since we manually set `UID`s, making `HasUID`s `uid` read-only won't stop anyone from doing the same overriding (if that even exists). Ideally, when someone changes something about another chunk (without a theory refinement), the `UID` would change to indicate that they're both different. I don't quite understand your point about making `uid` read-only would allow some `UID`s to be computed on the fly, can you elaborate? How is it different than adding an extra `UID` field for `DataDefinition`s?

hrzhuang:
Adding a UID field to `DataDefinition` should also work. Computing the UID on the fly is just an alternative solution where we are not storing the same string in two places. I don't think this alone justifies making `uid` read-only, but I don't think `uid` should be a lens in the first place (why should `HasUID` allow the UID to be modified?), and this is an example of the kind of things we could do if `uid` was read-only.

balacij:
`HasUID` exposing a writable `uid` is useful for when we want to build chunks that are slightly remixes of others. For example, different English definitions, etc. These areas should change both the `UID` and the data to make sure it's different. Since `UID` is meant to be 1 level 'above' the Haskell code (e.g., implicitly done well, but there is an explicit way to do it too if we didn't have to manually write the Haskell code) but isn't, there are other ways to create the same 'problematic chunks' that having the writable `UID` allows -- for example, record syntax overrides allows the same overwrites, and manually creating new chunks with the 'right' chunk constructors both allow the same functionality. For that reason, I think it would be better to keep it but restrict usage heavily at the code-review level.

JacquesCarette:
[...]

PR #3907, opened 2024-08-02
Make the `HasUID` lens read-only

Xinlu-Y:
As discussed in https://github.com/JacquesCarette/Drasil/issues/3524 , the `uid` field was previously implemented as a `Lens`, which allowed for modification. However, after switching `uid` to a `Getter`, the `uid` field became read-only, which affected functions that relied on setting the `uid` value, such as `addRelToCC`.

### Key Changes:
- Removed `addRelToCC` Function: The `addRelToCC` function, which previously allowed modification of the `uid` field using the `set` function, has been removed due to its incompatibility with the read-only `Getter` implementation.

- Direct RelationConcept Creation: Modified the `mobShr` definition to create RelationConcept directly with appropriate properties using functions like `makeRC` instead of relying on the removed `addRelToCC` function.

### DIFF
--- a/code/drasil-example/ssp/lib/Drasil/SSP/GenDefs.hs
+++ b/code/drasil-example/ssp/lib/Drasil/SSP/GenDefs.hs
@@ -150,7 +150,8 @@ resShrDeriv = mkDerivNoHeader [foldlSent [S "Derived by substituting",
 
 --
 mobShr :: RelationConcept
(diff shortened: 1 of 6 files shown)


JacquesCarette (approved):

balacij:
For the future, writing `Closes #X` at the top of the PR description will create a link to said issue (`#X`) that closes the issue if/when the PR is merged.

Issue #718, opened 2018-06-27
Formatting Long Expressions in Latex


niazim3:
(Similar in purpose to #346)

`Breqns` was removed from the import list for the latex files (reasoning for which is discussed in #701).

Implementation for formatting of long equations is necessary to make the output look better.

Specific places where this is necessary: 

#### Terminology and definitions
![image](https://user-images.githubusercontent.com/28247301/41984158-fda875a0-79fd-11e8-9c3c-d7518c2833e9.png)

#### Data definitions
![image](https://user-images.githubusercontent.com/28247301/41984231-2e2d19c4-79fe-11e8-8545-8e6e6809bc29.png)
![image](https://user-images.githubusercontent.com/28247301/41984264-43197512-79fe-11e8-9d3c-18a67f9301be.png)

#### Instance models
![image](https://user-images.githubusercontent.com/28247301/41984400-96b0de40-79fe-11e8-9f6f-603cad1971ce.png)
![image](https://user-images.githubusercontent.com/28247301/41984461-beb3b732-79fe-11e8-964c-4803d7dc3d74.png)

balacij:
From meeting: For the text, we might be able to use "parbox"s. For the equation, we will need to look into it a bit more to understand what is going on.

balacij:
I don't remember what from the meeting today reminded me of this ticket, but something did. Another possible way to wrap the "block long expressions" (long expressions that aren't inlined) is to use a table to format them, splitting at each operator-precedence level and tabbing as needed. For example, if we had the equation $a = b \cdot{} c \cdot{} d + (e + f - g) / h + log(i) \cdot{} j - k + \dots{}$ where $a, b, c, \dots{}$ are long expressions as well, we might write it as:

$$
\begin{align}
a = & \ b \cdot{} c \cdot{} d \\
  & + (e + f - g) / h \\
  & + log(i) \cdot{} j \\
  & - k + \dots{}
\end{align}
$$

(alignment is bad and poorly formatted because I couldn't use the `tabular` environment here on GitHub)

Then, if terms that are escaped to the next line are also too long, we can escape their terms to the next line as well, with the same pattern of escaping by the operators according to precedence. However, each one should be given an extra tab or something for readability.

For inlined expressions on the other hand, the "parbox" solution should work fine.

Of course, with both of these ideas, we might still need to delve into gathering font and font metric information so that we can really ensure that overflows never happen.

smiths:
[...]

PR #3910, opened 2024-08-06
Resize long expressions in LaTeX to prevent overflow

BilalM04:
Contributes to #718 

- Detailed description in https://github.com/JacquesCarette/Drasil/issues/718#issuecomment-2263312386.
- Adjusts block equations within definition tables to 80% of the page width if they overflow.
- Adjusts standard block equations to the full page width if they overflow.

To save you the trouble of compiling, here is a precompiled example of [SSP_SRS.pdf](https://github.com/user-attachments/files/16512476/SSP_SRS.pdf) for your reference.

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
@@ -7,7 +7,7 @@ import Text.PrettyPrint (text)
 import Language.Drasil.Printing.LayoutObj (LayoutObj(..))
 import Language.Drasil.TeX.Monad (D, vcat, (%%))
 import Language.Drasil.TeX.Helpers (docclass, command, command1o, command2, command3,
-  usepackage, command0)
+  usepackage, command0, lbrace, rbrace)
 import Language.Drasil.Printing.Helpers (sq, brace)
 
 import Language.Drasil.Config (hyperSettings, fontSize, bibFname)
@@ -94,14 +94,14 @@ addDef SymbDescriptionP1 = command3 "newlist" "symbDescription" "description" "1
 addDef SymbDescriptionP2 = command1o "setlist" (Just "symbDescription") "noitemsep, topsep=0pt, parsep=0pt, partopsep=0pt"
 addDef SaveBox       = command "newsavebox" "\\mybox"
 addDef ResizeExpr    = vcat [
-    command "newcommand" "\\resizeExpression" <> pure (sq "2") <> pure (text "{"),
+    command "newcommand" "\\resizeExpression" <> pure (sq "2") <> lbrace,
     command2 "savebox" "\\mybox" "$#1$",
     command0 "ifdim" <> command0 "wd" <> command0 "mybox" <> pure (text ">#2") <> command0 "linewidth",
     command3 "resizebox" "#2\\textwidth" "!" ("\\usebox" ++ brace "\\mybox"),
     command0 "else",
     command "usebox" "\\mybox",
     command0 "fi",
-    pure (text "}")
+    rbrace
   ]
 
 -- | Generates LaTeX document preamble.
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
@@ -7,7 +7,7 @@ import Text.PrettyPrint (text)
 import Language.Drasil.Printing.LayoutObj (LayoutObj(..))
 import Language.Drasil.TeX.Monad (D, vcat, (%%), nest)
 import Language.Drasil.TeX.Helpers (docclass, command, command1o, command2, command3,
-  usepackage, command0, lbrace, rbrace)
+  usepackage, command0, lbrace, rbrace, ExprScale(InDef, OutDef))
 import Language.Drasil.Printing.Helpers (sq, brace)
 
 import Language.Drasil.Config (hyperSettings, fontSize, bibFname)
@@ -81,6 +81,8 @@ data Def = Bibliography
          | SymbDescriptionP1
          | SymbDescriptionP2
          | SaveBox
+         | InDefScale
+         | OutDefScale
          | ResizeExpr
          deriving Eq
 
@@ -93,6 +95,8 @@ addDef SetMathFont   = command "setmathfont" "Latin Modern Math"
 addDef SymbDescriptionP1 = command3 "newlist" "symbDescription" "description" "1"
 addDef SymbDescriptionP2 = command1o "setlist" (Just "symbDescription") "noitemsep, topsep=0pt, parsep=0pt, partopsep=0pt"
 addDef SaveBox       = command "newsavebox" "\\mybox"
+addDef InDefScale    = command0 "def" <> command (show InDef)  "0.8"
+addDef OutDefScale   = command0 "def" <> command (show OutDef) "1.0"
 addDef ResizeExpr = vcat [
     command "newcommand" "\\resizeExpression" <> pure (sq "2") <> lbrace,
     nest 2 $ vcat [
@@ -138,6 +142,6 @@ parseDoc los' =
     parseDoc' Header{}     = ([], [])
     parseDoc' Paragraph{}  = ([], [])
     parseDoc' List{}       = ([EnumItem], [])
-    parseDoc' EqnBlock{}   = ([Graphicx, Calc], [SaveBox, ResizeExpr])
+    parseDoc' EqnBlock{}   = ([Graphicx, Calc], [InDefScale, OutDefScale, SaveBox, ResizeExpr])
     parseDoc' Cell{}       = ([], [])
     parseDoc' CodeBlock{}  = ([], [])
--- a/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
+++ b/code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
@@ -3,10 +3,12 @@ module Language.Drasil.TeX.Preamble (genPreamble) where
 
 import Data.List (nub)
 
+import Text.PrettyPrint (text)
 import Language.Drasil.Printing.LayoutObj (LayoutObj(..))
 import Language.Drasil.TeX.Monad (D, vcat, (%%))
 import Language.Drasil.TeX.Helpers (docclass, command, command1o, command2, command3,
-  usepackage)
+  usepackage, command0)
+import Language.Drasil.Printing.Helpers (sq, brace)
 
 import Language.Drasil.Config (hyperSettings, fontSize, bibFname)
 
@@ -35,6 +37,8 @@ data Package = AMSMath      -- ^ Improves information structure for mathematical
              | EnumItem     -- ^ Contol basic list environments.
              | SVG          -- ^ For rendering svg diagrams.
              | Float        -- ^ For enhanced control over placement of figures and tables.
+             | Graphicx     -- ^ Resizebox for large expressions.
+             | Calc         -- ^ Perform arithmetic operation in LaTeX commands.
              deriving Eq
 
 -- | Adds a 'Package' to the LaTeX document.
@@ -66,6 +70,8 @@ addPackage Unicode   = usepackage "unicode-math"
 addPackage EnumItem  = usepackage "enumitem"
 addPackage SVG       = usepackage "svg"
 addPackage Float     = usepackage "float"
+addPackage Graphicx  = usepackage "graphicx"
+addPackage Calc      = usepackage "calc"
 
 -- | Common LaTeX commands.
 data Def = Bibliography
@@ -74,6 +80,8 @@ data Def = Bibliography
          | SetMathFont
          | SymbDescriptionP1
          | SymbDescriptionP2
+         | SaveBox
+         | ResizeExpr
          deriving Eq
 
 -- | Define common LaTeX commands.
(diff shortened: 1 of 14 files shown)


smiths:
The new pdfs are an improvement over the previous ones.  I like this change to the generated stable files.  @JacquesCarette or @balacij should review the Drasil code changes.

code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
```diff
@@ -84,6 +92,17 @@ addDef LessThan      = command2 "newcommand" "\\lt" "\\ensuremath <"
...
+addDef SaveBox       = command "newsavebox" "\\mybox"
+addDef ResizeExpr    = vcat [
+    command "newcommand" "\\resizeExpression" <> pure (sq "2") <> pure (text "{"),
+    command2 "savebox" "\\mybox" "$#1$",
+    command0 "ifdim" <> command0 "wd" <> command0 "mybox" <> pure (text ">#2") <> command0 "linewidth",
+    command3 "resizebox" "#2\\textwidth" "!" ("\\usebox" ++ brace "\\mybox"),
+    command0 "else",
+    command "usebox" "\\mybox",
+    command0 "fi",
+    pure (text "}")
```
samm82, line 104: Do we have any helper functions for wrapping content in curly braces?

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -76,10 +76,22 @@ lo (Graph ps w h c l)    _  = toText $ makeGraph
 lo (Cell _) _               = empty
 lo (CodeBlock _) _          = empty
 
+-- | Helper for converting layout objects into a more printable form.
+-- This function is specific to definitions.
+lo' :: LayoutObj -> PrintingInformation -> D
+lo' (HDiv _ con _)      sm = print' sm con
+lo' (EqnBlock contents) _  = makeEquation contents 0.8
+lo' obj sm                 = lo obj sm
```
samm82, line 84: Minor nitpick for alignment 😅

```suggestion
lo' obj                 sm = lo obj sm
```

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -76,10 +76,22 @@ lo (Graph ps w h c l)    _  = toText $ makeGraph
...
+
 -- | Converts layout objects into a document form.
 print :: PrintingInformation -> [LayoutObj] -> D
 print sm = foldr (($+$) . (`lo` sm)) empty
 
+-- | Converts layout objects into a document form.
+-- Specific to Definitions.
+print' :: PrintingInformation -> [LayoutObj] -> D
+print' sm = foldr (($+$) . (`lo'` sm)) empty
```
samm82, line 93: If this is `Specific to Definitions`, could it have a more meaningful name, such as `printDef`?

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -345,17 +357,17 @@ makeDefTable sm ps l = mkEnvArgBr "tabular" (col rr colAwidth ++ col (rr ++ "\\a
 
 -- | Helper that makes the rows of a definition table.
 makeDRows :: PrintingInformation -> [(String,[LayoutObj])] -> D
-makeDRows _  []         = error "No fields to create Defn table"
-makeDRows sm ls    = foldl1 (%%) $ map (\(f, d) -> dBoilerplate %%  pure (text (f ++ " & ")) <> print sm d) ls
+makeDRows _  []      = error "No fields to create Defn table"
+makeDRows sm ls      = foldl1 (%%) $ map (\(f, d) -> dBoilerplate %%  pure (text (f ++ " & ")) <> print' sm d) ls
   where dBoilerplate = pure $ dbs <+> text "\\midrule"
```
samm82, line 362: Is `dBoilerplate` reused anywhere? If not, it might make more sense to just inline it, but others may disagree

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -76,10 +76,22 @@ lo (Graph ps w h c l)    _  = toText $ makeGraph
 lo (Cell _) _               = empty
 lo (CodeBlock _) _          = empty
 
+-- | Helper for converting layout objects into a more printable form.
+-- This function is specific to definitions.
+lo' :: LayoutObj -> PrintingInformation -> D
+lo' (HDiv _ con _)      sm = print' sm con
+lo' (EqnBlock contents) _  = makeEquation contents 0.8
```
samm82, line 83: Where does `0.8` come from? Is this just hardcoded? Would it ever change?

samm82 (commented):
Minor, likely non-blocking notes on another solid PR!

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
BilalM04: The value `0.8` specifies the fraction of the page width to which the expression size is limited. Expressions within definition tables have a smaller area compared to those outside, requiring a reduced scale. I have added information about this parameter in the code documentation for [`makeEquation`](https://github.com/JacquesCarette/Drasil/pull/3910/files#diff-59ffd95c9dbdc3081ed3fc775fd70ca76f460454be7cc730a0f73b280caeb5f4R369). Below, I provide a visual representation showing that expressions in definition tables are restricted to `0.8 * page-width`.

**Would it ever change?** I don't anticipate this to change as all definition tables have the same left column for all the examples, and to my knowledge, there is no existing functionality to modify what those fields are. 

image

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3910/commits/f78f58a2d6d93b74e7b82afe12e536ffedbacf8c.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3910/commits/f78f58a2d6d93b74e7b82afe12e536ffedbacf8c.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3910/commits/f78f58a2d6d93b74e7b82afe12e536ffedbacf8c.

code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
BilalM04: There are functions to wrap items in curly braces; however, they all concatenate the items horizontally. Whereas in this particular instance, they need to be concatenated vertically. Here are some of the helper functions:

https://github.com/JacquesCarette/Drasil/blob/8ca26be6df6d3f4c24627e76d051637b0ff85b16/code/drasil-printers/lib/Language/Drasil/Printing/Helpers.hs#L62

https://github.com/JacquesCarette/Drasil/blob/8ca26be6df6d3f4c24627e76d051637b0ff85b16/code/drasil-printers/lib/Language/Drasil/Printing/Helpers.hs#L80

https://github.com/JacquesCarette/Drasil/blob/8ca26be6df6d3f4c24627e76d051637b0ff85b16/code/drasil-printers/lib/Language/Drasil/TeX/Helpers.hs#L28-L31

JacquesCarette: The 'right' thing to do would be extract
```haskell
lbrack = text "{"
rbrack = text "}"
```
and use that in making the above code, i.e. both the new code and the places where it is currently 'hard coded'.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -76,10 +76,22 @@ lo (Graph ps w h c l)    _  = toText $ makeGraph
 lo (Cell _) _               = empty
 lo (CodeBlock _) _          = empty
 
+-- | Helper for converting layout objects into a more printable form.
+-- This function is specific to definitions.
+lo' :: LayoutObj -> PrintingInformation -> D
+lo' (HDiv _ con _)      sm = printDef sm con
```
JacquesCarette, line 82: why does this dispatch to `printDef` while below it goes to `lo` ?

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
```diff
@@ -23,6 +25,15 @@
 \newcommand{\lt}{\ensuremath <}
 \newlist{symbDescription}{description}{1}
 \setlist[symbDescription]{noitemsep, topsep=0pt, parsep=0pt, partopsep=0pt}
+\newsavebox{\mybox}
+\newcommand{\resizeExpression}[2]{
+\savebox{\mybox}{$#1$}
+\ifdim\wd\mybox>#2\linewidth
```
JacquesCarette, line 31: can you make this format nicely, i.e. indent the stuff in the 'then' and 'else' branches?

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
```diff
@@ -311,7 +322,7 @@ \subsubsection{Theoretical Models}
         
 \\ \midrule
 Equation & \begin{displaymath}
-           \symbf{a}\text{(}t\text{)}=\frac{\,d\symbf{v}\text{(}t\text{)}}{\,dt}
+           \resizeExpression{\symbf{a}\text{(}t\text{)}=\frac{\,d\symbf{v}\text{(}t\text{)}}{\,dt}}{0.8}
```
JacquesCarette, line 325: We can predict that we'll likely want more processing of expressions in the future, so may as well abstract from that now. In other words, introduce a `\formatExpr` macro that wraps expressions, which would then just be `resizeExpression{#1}{0.8}`.

JacquesCarette (changes requested):
Nice, but since you've opened this can of worms, may as well do it right!

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
BilalM04: First, I'll provide what the difference is between `lo` and `lo'`.

`lo` is for printing `LayoutObj`s normally (i.e. anything that is not inside of a definition table).
`lo'` is for printing `LayoutObj`s that are inside a definition table.

This distinction had to be made because `EqnBlock`s that are not in a definition table can span the entire width of the page (i.e. their `scale` parameter is `1`). Whereas `EqnBlock`s inside definition tables are restricted to 0.8 * page width (i.e. their `scale` parameter is `0.8`)

```Latex
\resizeExpression{...}{1.0} % Outside definition table
\resizeExpression{...}{0.8} % Inside definition table
```

---

The reason the `HDiv` pattern match dispatches to `printDef` (i.e. uses `lo'`) is because `con` is of type `[LayoutObj]`, thus those could also contain `EqnBlock`s. 

The last pattern match is a "catch all" statement, where if the `LayoutObj` is not an `EqnBlock`, it will print the same as `lo`.

code/drasil-printers/lib/Language/Drasil/TeX/Preamble.hs
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3910/commits/581da6258114fbbfaabcb8e68c906064b4e0e850. I could not find any other instances where this is present as they all make use of horizontal concat when wrapping items in `{}`.

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
BilalM04: Fixed in https://github.com/JacquesCarette/Drasil/pull/3910/commits/3f29a2c9a1fed602ce065c77e3afc4b78af7ed87.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
JacquesCarette: Very thorough explanation, thank you.

Having `lo`/`lo'` will do. The right solution would be to make our printers use an environment that could carry context (i.e. inside of definition table or not) instead of doing that via different functions.

For the scale parameter, I think it would be good to generate cleaner LaTeX, i.e. generate two variable definitions (at the 'top' of the LaTeX file) for each of the scales, and then use those variables instead of hard-coded numbers. LaTeX is a programming language, after all, and we should use it with the same care as we do other things.

BilalM04: Added in https://github.com/JacquesCarette/Drasil/pull/3910/commits/820619edf8387cfd0ecf5a9477cb19232f4aa403 and https://github.com/JacquesCarette/Drasil/pull/3910/commits/845174b40bbf98016564d9fe30f2a4bb0c866ad6.

The following will be defined in the preamble:

```LaTeX
\def\inDefScale{0.8}
\def\outDefScale{1.0}
```

`\resizeExpression` will be used as follows:

```Latex
\resizeExpression{...}{\outDefScale} % Outside definition table
\resizeExpression{...}{\inDefScale} % Inside definition table
```

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
BilalM04: ### To clarify, are you suggesting the following?

In addition to the existing variables/commands I created in this PR, which are as follows:

```LaTeX
\def\inDefScale{0.8}
\def\outDefScale{1.0}

\newsavebox{\mybox}

\newcommand{\resizeExpression}[2]{
  \savebox{\mybox}{$#1$}
  \ifdim\wd\mybox>#2\linewidth
    \resizebox{#2\textwidth}{!}{\usebox{\mybox}}
  \else
    \usebox{\mybox}
  \fi
}
```

we want the following macro:

```LaTeX
\newcommand{\InDef}{true}
\newcommand{\OutDef}{false}

\newcommand{\formatExpr}[2]{
  \if#2
    \resizeExpression{#1}{\inDefScale}
  \else
    \resizeExpression{#1}{\outDefScale}
  \fi
}
```

Thus, expressions will be printed as follows:


```Latex
\formatExpr{...}{\outDef} % Outside definition table
\formatExpr{...}{\inDef} % Inside definition table
```

That’s how I understand it. Please let me know if I’ve misunderstood anything.

JacquesCarette (approved):

Issue #3387, opened 2023-05-08
Have some independent tests for GOOL hooked in to Drasil


JacquesCarette:
As per #3372 , @bmaclach comments that there exists GOOL tests - but these are not part of the Drasil test suite right now.

This should be fixed. And likely the GOOL test suite should be expanded too.

B-rando1:
As of the two PRs linked above, the GOOL tests are much better hooked into Drasil's testing.

What is still somewhat lacking is a consistent, comprehensive test suite of everything in GOOL.  @bmaclach had a solid start with [HelloWorld.hs](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/test/HelloWorld.hs) and the other files in [drasil-code/test](drasil-code/test), but I think there's more we can do.  These test files provide a good summary of the most used features in GOOL, but they aren't exhaustive of all of GOOL's features.  They also aren't very systematic when dealing with edge-cases.

Another issue is that these files rarely test the actual values returned by GOOL code.  Many values aren't even printed to the screen, and many more are only printed to the screen.  @Xinlu-Y's work in #3911 is improving that, however.

I think there are two ways we can move forward here:
- Design a new suite of tests that covers GOOL in a more systematic way.  @samm82 do you have any advice for this?
- Increase the helpfulness of the tests by continuing to add `assert` statements wherever it makes sense.  As I mentioned in #3911, `listSlice` is one obvious choice to add `assert` to; and I'm guessing there are others.

samm82:
A really good technique for designing test cases is "equivalence partitioning": essentially dividing the (likely) infinite set of possible test values/configurations into "partitions" that behave similarly, then only testing a handful of values from these partitions (usually one from the middle and a few from the edges). For example, some good list values to use for testing list indexing are an empty list, a list with one item, and a list with many items. A list with 7 items and a list with 8 items are likely going to behave similarly, while a list with 0 items and with 1 item will behave quite differently, even though they both have a difference of 1 item. My guess is that many (if not all) of the existing test cases will be useful in this and will only need to be augmented with more test cases for some of the edge-cases (as you mentioned). I think that most, if not all, values that get printed to the screen (or that probably should be) could be replaced with an `assert` statement to further automate this testing.

PR #3912, opened 2024-08-06
Added better support for GOOL tests in Makefile

B-rando1:
- I moved the `diff` check added in #3909 from `codegenTest_gen` to `codegenTest_diff`.  This is more consistent with other `make` rules, and is needed for `codegenTest_stabilize`.
- I added `codegenTest_diff` to `make all`.  This means it will be run whenever `make` and `make pr_ready` are run, and ensures that the GOOL tests are checked when we run the main Drasil tests.
- I created `codegenTest_stabilize`.  This allows us to easily update `stable/gooltest` when we need to.

Contributes to #3387

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -184,7 +184,7 @@ BATCHTERM=dumb
 
 # Default `make` will just run examples and test them against stable. See `test`
 # target for more details.
-all: test test_website codegenTest_gen ##@Examples Run examples and test against stable.
+all: test test_website codegenTest_diff ##@Examples Run examples and test against stable.
 
 install: ##@Examples Install all example project binaries into your local binary path (see $(stack path) for local-bin-path).
 	stack install $(stackArgs)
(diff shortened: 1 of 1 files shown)


JacquesCarette:
So I seem to remember that we had not turned on some of the tests because they didn't work on all platforms (i.e. because some languages didn't work on them at all.)  But maybe I have this wrong, and this isn't trying to compile / run the generated code, but doing something else?

B-rando1:
@JacquesCarette I'm not exactly sure what you're referring to, but I can give a summary of what the different `codegenTest` commands do (including the changes in this PR):
- `codegenTest_gen` simply generates the target code from the tests in `drasil-code/test`.
- `codegenTest_diff` generates the target code, and compares it with `stable`.  With this PR, it runs when `make` and `make pr_ready` are run.
- `codegenTest_stabilize` generates the target code and uses it to update `stable`.
- `codegenTest` generates the target code, compares it with `stable`, and also runs `make` to compile each generated program.  This command runs as part of our CI.  Running `make` has been going on for a while I believe, and comparison with `stable` is new as of #3909.

I'm not quite sure what you mean that not all of the examples work with all the targets.  In the CI for this PR ([here](https://github.com/JacquesCarette/Drasil/actions/runs/10273438020/job/28427889902?pr=3912) under `Compile GOOL examples`) for example, you can see it compiles the Java, C++, C#, and Swift without issues (other than some warnings that the test programs are weird programs 😄).  It doesn't do anything with the Python or Julia code, since they aren't compiled.  It is true that the current version of `HelloWorld.jl` would crash if run, but as we're only compiling them that shouldn't be an issue.

JacquesCarette:
The main points were, if I remember well, that
1. Drasil developpers should have to have java, c++, ... installed to work on Drasil
2. That Swift, at some point, wasn't Windows/Mac/Linux.

B-rando1:
If I understand you right, your main concern is that if any of our target languages are not supported by one of our platforms, then some of our `make` commands would be useless to that platform.  I have a couple of notes on that:
- You mentioned that Swift didn't use to be available on all platforms.  I wonder if that's changed since then.  I got it working on my WSL shell (not the easiest install if I remember correctly, but still doable), and from Googling it, it sounds like it's available on Windows too.
- Even if we get languages at some point that cause problems on certain platforms, it is only the `codegenTest` command that would break.  The `_gen`, `_diff`, and `_stabilize` variants don't compile generated code, so they wouldn't be affected.  I think `codegenTest` is mainly intended to be used for CI, so as long as all our targets work on the server we should be alright.

JacquesCarette (approved):

PR #3916, opened 2024-08-07
Added "Figure: " prefix to captions in both HTML and Markdown outputs

### LINKED ISSUE #702: HTML: Figures should have `Figure X: ` in their `figcaption` and references to figures should be prettier.
EDIT: Take [Projectile's SRS](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html) for example. Specifically, the [System Context Figure](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Figure:sysCtxDiag). We have ~2-3 issues:
1. Figures are missing the `Figure: ` text in their captions.
2. References to figures use the internal HTML IDs! For example, in the [System Context section](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:GenSysDesc), we write `Fig:sysCxtDiag`.
3. Instead of writing the references, we should either use a short-caption (i.e., a short-title where available) OR use a numbering system. 

@smiths Do you have any thoughts on whether this we should use a numbering system or a short-caption?

Original OP:
* LaTeX gets figure prefixes (i.e. "Figure xx: ") for free in captions, whereas HTML does not. This results in either the HTML not having "Figure xx:" prefixed to captions or LaTeX having "Figure xx: Figure xx: " prefixed to its captions.
* ~~HTML is not centering figures. LaTeX encloses figures in a `center` environment. ~~

Xinlu-Y:
Closes #702 
- HTML Output: Updated the image function to include "Figure: " before the caption text in the <figcaption> element.
- Markdown Output: Modified the image function to prepend "Figure: " to the caption in the generated Markdown content.


code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -62,7 +62,7 @@ reflinkURI ref txt = if ref == txt then ang ref
 
 -- | Helper for setting up figures
 image :: Doc -> Doc -> Doc
-image f c =  text "!" <> reflinkURI fp c $^$ bold c
+image f c =  text "!" <> reflinkURI fp (text "Figure: " <> c) $^$ bold (text "Figure: " <> c)
```
BilalM04, line 65: This might just be me being picky, but the parameter for `reflinkURI` is the alt text, which I don't think should have "Figure:" appended to it. If we do, then we should add it to the HTML alt text as well.

code/stable-website/index.html
```diff
@@ -32,7 +32,7 @@ <h1></h1>
         <div id="Figure:Drasil">
           <figure>
             <img src="./images/Icon.png" alt="" width="50.0%" >
-            <figcaption></figcaption>
+            <figcaption>Figure: </figcaption>
```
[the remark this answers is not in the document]
Xinlu-Y: Changing the `image` function will also affect `drasil-website`. Maybe we should add a caption to drasil's icon to ensure that every image has a caption?

code/stable-website/index.html
```diff
@@ -697,7 +697,7 @@ <h2>Package Dependency Graphs</h2>
             <div id="Figure:Figuredrasil-website">
               <figure>
                 <img src="graphs/drasil-website.png" alt="Package: drasil-website" >
-                <figcaption>Package: drasil-website</figcaption>
+                <figcaption>Figure: Package: drasil-website</figcaption>
```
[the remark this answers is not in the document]
Xinlu-Y: After adding the "Figure:" prefix, the original description doesn't look very nice. We can improve the brief captions in these situations.

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
Xinlu-Y: Sure, I will wait for the above questions to be answered and then modify it together.

code/stable-website/index.html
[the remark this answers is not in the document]
smiths: It does look odd to have two full colons.  Is it possible to change the capstone to Package - drasil-website? (That is, replace the : with a -)

code/stable-website/index.html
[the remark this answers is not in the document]
BilalM04: That particular image is for the header of the website, which, in my opinion, wouldn't make sense to have a caption for.

image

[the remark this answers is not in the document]
BilalM04: We can either include a caption for every image, as you suggested, or modify the image function to append `"Figure:"` only if the caption is non-empty. If we choose the first option, we enforce that *all* images *must* have a caption. In this case, we should handle empty captions by generating an error, as the issue of `"Figure:"` appearing with an empty caption would persist since there is nothing stopping a user from defining an empty caption.

That being said, others may have differing opinions, so we should wait for their input.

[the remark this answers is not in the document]
smiths: Great observation and solution @BilalM04.  I agree that there are cases where we don't want a caption and in those cases the term `Figure:` should not appear.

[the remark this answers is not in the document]
JacquesCarette: The solution there would be to use a `Maybe` so that `Nothing` would indicate an empty caption -- that way it is clear that this is *intentional*. An empty string caption might well be an accident.

code/drasil-printers/lib/Language/Drasil/HTML/Helpers.hs
```diff
@@ -140,14 +140,19 @@ reflinkURI rf txt = text ("<a href=\"" ++ rf ++ "\">") <> txt <> text "</a>"
...
+image f c 100 =
+  figure $ vcat $ img [("src", f), ("alt", c)] : figCap c
 image f c wp =
-  figure $ vcat [
-  img [("src", f), ("alt", c), ("width", text $ show wp ++ "%")],
-  figcaption c]
+  figure $ vcat $ img [("src", f), ("alt", c), ("width", text $ show wp ++ "%")] : figCap c
+
+-- | Helper function for figcaption.
+figCap :: Doc -> [Doc]
```
JacquesCarette, line 149: Seems like this choice is done 'too late' (i.e. it is already a `Doc`). I guess this could become a `Maybe Doc`. But basically we don't want to "dig into" a `Doc` to figure out if it is empty, but pass that information down explicitly.

code/drasil-printers/lib/Language/Drasil/Markdown/Helpers.hs
```diff
@@ -62,9 +62,16 @@ reflinkURI ref txt = if ref == txt then ang ref
 
 -- | Helper for setting up figures
 image :: Doc -> Doc -> Doc
-image f c =  text "!" <> reflinkURI fp c $^$ bold c
+image f c =  text "!" <> reflinkURI fp c $^$ figCaption c
   where
     fp = text $ "./assets/" ++ takeFileName (show f)
+    figCaption cap
```
JacquesCarette, line 68: The reasoning for `figCaption` should be the same here as for HTML, no? Perhaps even share the same code for the part that emits "Figure: " ?

JacquesCarette (changes requested):

code/drasil-printers/lib/Language/Drasil/HTML/Print.hs
```diff
@@ -337,7 +337,7 @@ pItem (Nested s l) = vcat [pSpec s, makeList l]
 -----------------------------------------------------------------
 -- | Renders figures in HTML.
 makeFigure :: Doc -> Doc -> Doc -> L.MaxWidthPercent -> Doc
-makeFigure r c f wp = refwrap r (image f c wp)
+makeFigure r c f wp = refwrap r (image f (if c == mempty then Nothing else Just c) wp)
```
JacquesCarette, line 340: As I said in a previous commit, `c` should not be of type `Doc` here -- the information of whether it is empty or not should come from the callee instead.

code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
```diff
@@ -307,7 +307,7 @@ item rm (Nested s l) = vcat [pSpec rm s, makeList rm l 0]
 
 -- | Renders figures in Markdown
 makeFigure :: Doc -> Doc -> Doc -> Doc
-makeFigure r c f = centeredDivId r (image f c)
+makeFigure r c f = centeredDivId r (image f (if c == mempty then Nothing else Just c))
```
JacquesCarette, line 310: same here

JacquesCarette (changes requested):

Xinlu-Y:
I've updated the relevant code to ensure that the emptiness of `c` is checked before calling `makeFigure`. However, I'm unsure about the potential impact of changing the `Caption` type to `Maybe Caption` within the `Figure` constructor. It appears that two `Nothing` checks are necessary in `printLo`: one to handle the validity of `Maybe Spec`, and another to validate the content within `Spec` itself.

JacquesCarette:
The change should go even further up: `Figure` as defined in `drasil-lang` (i.e. in `Language.Drasil.Document.Core`) should already have the information in it whether there is a caption or not. Might as well introduce a new data-type (`HasCaption`) with cases `NoCaption` and `Caption` for that. Then propagate this information through.

code/drasil-lang/lib/Language/Drasil/Document.hs
```diff
@@ -133,11 +133,11 @@ getSecCons (Sub sec) = getSec sec
 getSecCons (Con _)   = []
 
 -- | 'Figure' smart constructor with a 'Lbl' and a 'Filepath'. Assumes 100% of page width as max width.
-fig :: Lbl -> Filepath -> RawContent
+fig :: Lbl -> Filepath -> HasCaption -> RawContent
```
JacquesCarette, line 136: Sorry to cause yet more rework, but I think the 'right' thing to do would be to have `fig`, the smart constructor, default to `WithCaption` and then have a `figNoCap` constructor for the other case.

code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
```diff
@@ -94,7 +94,7 @@ printLO rm (EqnBlock contents)   = text "\\\\[" <> rndr contents <> text "\\\\]"
 printLO rm (Table _ rows r b t)  = makeTable rm rows (pSpec rm r) b (pSpec rm t)
 printLO rm (Definition _ ssPs l) = makeDefn rm ssPs (pSpec rm l)
 printLO rm (List t)              = makeList rm t 0
-printLO rm (Figure r c f _)      = makeFigure (pSpec rm r) (pSpec rm c) (text f)
+printLO rm (Figure r c f _) = makeFigure (pSpec rm r) (fmap (pSpec rm) c) (text f)
```
JacquesCarette, line 97: you should line up the `=` like they were before

JacquesCarette (changes requested):

JacquesCarette (approved):

PR #3926, opened 2024-08-13
Brandon's Poster

### LINKED ISSUE #3921: Sam's review of Brandon's poster
Here are my (rough) notes on your (so far, very good!) poster - feel free to pull any of these out to separate issues if they'll require more discussion

- [x] I would explicitly define what you mean by "shared" earlier (i.e., in the Goal section)
- [x] What's the significance of "OO typeclasses can inherit from Shared typeclasses."? (this may be addressed by the above point)
- [x] What do outline colours mean in your graphs?
- [x] Captions should be sentence case and end with a period

Minor things:
- [x] "5" should be spelled out in the Introduction
- [x] "E.g." shouldn't start a sentence; use "For example" instead
- [x] I'd add a space before the line colour note and re- add the "Legend" heading, and I don't think the parentheses are needed
- [x] Explicitly explaining that OO = object- oriented might help? idk if this is assumed knowledge
- [x] Is there a reason "Scientific Computing" is capitalized?
- [x] Could you use the colours from the Julia logo in your graphs? That might look cool

B-rando1:
Feel free to add any feedback!  Note that the poster is due for printing by today (August 13) at 5, so making changes after that will be difficult.

Current poster location is [here](https://github.com/JacquesCarette/Drasil/blob/Poster/People/Brandon-Bosman/Julia%20Poster/Poster.pdf).

Closes #3921


samm82:
My only notes:
- [ ] I don't think you need to have "Julia is not OO" twice in the Problem section (once as a heading and once in the bullet point)
- [ ] I'm not sure what you mean by "GOOL is too flexible" - if this is the case, I would think that it would able to handle anything, including procedural languages! Did you mean "GOOL is not flexible enough?"
- [ ] Could you move Figure 3 to the right and have the bullets of Designing GProc wrap to the left? This would reduce "white" space

Other than that, these look great!

smiths:
I agree with the feedback from @samm82.  Otherwise, it looks great!

JacquesCarette (approved):

Issue #3918, opened 2024-08-08
Add ODE Solvers to Julia


B-rando1:
Most of the examples in Drasil require an ODE solver.  In Drasil we don't have information set up about an ODE library for Julia.  Thus, enabling `Bundled` inputs in Julia (via structs) isn't actually enough to get these examples up and working - we also need the ODE libraries working.

There doesn't seem to be much documentation that I could find on how ODEs are represented in Drasil.  I found a wiki [page](https://github.com/JacquesCarette/Drasil/wiki/ODE-Definitions) giving a theory recap and a few issues discussing them (e.g. #394), but I still don't really know how they work.

I have two main questions:
- What should I look for in an ODE solver for Julia?  I don't have much experience with ODEs, so I'm not quite sure where to start.
- What does Drasil need to know about the ODE solvers we find?

smiths:
@B-rando1, you should have a look at [Dong's report](https://github.com/JacquesCarette/Drasil/blob/main/People/Dong/Thesis_Main.pdf) on solving ODEs in Drasil.  I only did a quick google, but [DifferentialEquations.jl](https://docs.sciml.ai/DiffEqDocs/stable/) looks pretty good.  They have [translations of standard libraries](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/#Translations-from-MATLAB/Python/R), which probably look like the code we already have for Python.

Don't worry about anything clever for the algorithms.  Our current ODEs examples are well-behaved and not likely to require any numerical tricks for stiff systems.

B-rando1:
I created a case study for Julia's ODE solver with `swhsnopcm`, viewable [here](https://github.com/smiths/caseStudies/pull/251).  It's not exactly the same as Python's, but it's pretty close.

B-rando1:
I'm closing #3931, and putting that knowledge here.

The branch with the ODE changes is [JuliaODE](https://github.com/JacquesCarette/Drasil/tree/JuliaODE), with the original draft PR #3931. I also have a [PR](https://github.com/smiths/caseStudies/pull/251) in the caseStudies repo explaining how Julia's ODE library itself works, and how to install it.

I don't remember the details so this is mainly a summary of the diff, but it might still be worth writing down here. The main change is in `drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs`. We have to add an `ODEPackg` and an `ODESymbols`. Here's what that looked like in the branch:

```hs
[...]

PR #3931, opened 2024-08-21
WIP making ODEs work in Julia

B-rando1:
Unfortunately we can't quite finish getting ODEs in Julia, but this should guide the way when we're ready.

The reason we can't finish is because the ODE library uses structs, and the Julia renderer doesn't support structs yet.

Other changes:
- I changed how Julia imports libraries, as that was previously incorrect.  I will create a separate PR for this.
- I changed how Julia represents empty lists.  Previously any empty list was simply `[]`; now it is `Type[]` - e.g. an empty list of doubles is `Float64[]`.  The reason for this is because Julia assigns `[]` the type `Array{Any}`, which causes problems when you need it to be type `Array{Float}`, for example.  I will also create a separate PR for this.
- I created the `populateSolList'` function, which acts like the `populateSolList` but without accessing a field.  I think that once we have structs in Julia, we won't need this function.
- I removed the `[]` case from `applyWithNamedArgs`, as it was causing problems.

I commented most of the things that need to be changed once we have structs.  You can also refer to my test case [here](https://github.com/smiths/caseStudies/pull/251) to compare.

@JacquesCarette @smiths @balacij feel free to give any feedback or ask any questions you have.

Contributes to #3918


JacquesCarette:
Why did you change Bundled to Unbundled in one example? Also, if you could rebase, so that it's easier to see what's new in just this PR?

B-rando1:
I changed the example to `Unbundled` since Julia can't support `Bundled` until we add `Struct`s.  Basically I just needed a way to see/show what code it was outputting for one of the ODE examples.  Since adding `Struct`s to Julia should unblock both ODEs and `Bundled`, we can probably put it back to `Bundled` by the time this PR is finished.

B-rando1:
Closed in favour of #3918

balacij:
Should we delete the branch?

B-rando1:
Sure, I captured all the important details in #3918, so I can delete it.

PR #3932, opened 2024-08-21
Made fixes to the Julia renderer

B-rando1:
There were two problems I discovered while implementing ODEs in Julia:
- Importing external libraries didn't work correctly.
- The types of empty lists was `Any` instead of its specific type.
  - Julia automatically assigns `[]` the type `Array{Any}`, which is problematic if you need its type to be `Array{Float64}`, for example.  I fixed it to add the type if the list is empty.  For example, an empty list of doubles is now represented as `Float64[]`.  This is in line with how the Julia REPL prints lists, and removes the potential of type errors.

This PR fixes those two issues.

Note on the `listDec` from `CommonPseudoOO.hs`: I'm not exactly sure why this only affects Julia, but previously it was giving lists the wrong level of type nesting (e.g. `List<List<Integer>>` instead of `List<Integer>`.  This PR fixes that, and also simplifies the implementation overall.


JacquesCarette (approved):
Nice. Love these tiny but important steps forward.

PR #3935, opened 2024-08-21
Switched LaTeX package from `resizebox` to `adjustbox` for resizing expressions

BilalM04:
Contributes to #718
Builds on #3910

In #3910, I used the `resizebox` package to resize the block equations; however, I have switched to the `adjustbox` package as I found it to be better overall.

### Pros of using `adjustbox` instead:

- No need for a scale parameter.
  - `adjustbox` is "smarter" and can automatically determine when scaling is necessary. 
    - This resulted in the removal of the `if` conditions in the custom LaTeX command.
    - We also don't need `lo'` and the `InDef` and `OutDef` params as the package automatically handles the scaling factor.
- `adjustbox` can resize an expression directly; no need for a `savebox`. 

Rendered PDF remains the same as #3910. Switching to the `adjustbox` package cleaned up both the generated TeX code and the written Drasil code.


JacquesCarette (approved):
Very very nice! I really like that you kept going until you were satisfied that this was good code.

PR #3938, opened 2024-08-23
removed `iterators` from `MethodState`

### LINKED ISSUE #3919: Remove `iterators` from `MethodState`
As we found in a [discussion](https://github.com/JacquesCarette/Drasil/pull/3878#discussion_r1705735756) in #3878, the `iterators` field in `MethodState` will be unused once that PR is merged.  It is currently only used in one place in the C++ renderer, and @NoahCardoso changed the implementation so that it is no longer needed there.

Once #3878 is passed, we should do a quick double-check that `iterators` isn't used anywhere else; then we should be safe to remove it.

NoahCardoso:
I removed `iterators` from `MethodState` since as said in #3919 the changes that I made to the Cpp render `forEach` loop makes the iterators field in `MethodState` unnecessary.

Closes #3919


JacquesCarette (approved):
In some sense, this is the opposite than what we'd want: we would want more semantic-oriented iteration concepts in the upper levels of Drasil, and then get loops (or iterators) out as appropriate.

But, be that as it may, this is the right code cleanup for how things are right now.

PR #3939, opened 2024-08-23
Adding more Set Opertions in GOOL

NoahCardoso:
This pr adds `setAdd`, `setRemove` and `setUnion` to GOOL but I did not have time to finish set union in C++ but I don't think it would be too hard to do.


B-rando1:
It might be helpful to rebase on #3942 so that you can add the new functions to Julia with that implementation in mind.

JacquesCarette (approved):
Very nice! Good team work, and nice effectiveness (i.e. you implemented new features very quickly).

JacquesCarette (approved):

JacquesCarette:
Serves me right for approving before the build finished... And I've just done it again.

B-rando1:
Note also that `setUnion` is not implemented in C++ yet.

JacquesCarette (approved):

PR #3940, opened 2024-08-23
Added basic set tests to `HelloWorld.hs`

B-rando1:
This ensures that sets are being tested.

It also shows that sets are not implemented properly in Julia, as `HelloWorld.jl` crashes when  run.  I will create a separate issue for this.


B-rando1:
P.S. once #3939 is merged, we should add lines checking `setAdd` and `setRemove`.

JacquesCarette (approved):

JacquesCarette:
So, I shouldn't merge this until that's done?

B-rando1:
I'd say it's up to you.  These changes are fine as-is.  I'd just want to add a line for each of the new functions added in #3939, with `assert` statements checking that they work correctly.

Since #3939 is waiting on `union` in C++, it might be good to merge this and make the further changes after it gets merged.  Or you could add the changes directly to #3939.  That might be better, since currently that PR has no code actually testing the functionality being added.

Issue #1867, opened 2019-09-03
GOOL code optimizations


bmaclach:
As discussed at the Aug. 29, 2019 meeting, GOOL should perform some optimization prior to generating code. For example:
- remove import statements unless they are needed
- remove identity operations (multiplication by 1, etc.)

To achieve this, GOOL will need to "know" more about the code it is generating. For example, at the point where GOOL is generating import statements, it needs to know whether those libraries were used in that file to determine whether to import them. Up to now, we've added tags to GOOL's underlying data types in an ad-hoc fashion whenever we needed to retain a piece of information for GOOL to use later (ex. tagging a value with its type). We'd like to have a more systematic way to do this. We should attempt to find patterns in the sorts of tags we've created so far which we may be able to take advantage of. @JacquesCarette mentioned the possibility of having an AST for use internally in GOOL which may be nearly identical to our AST in Drasil (the AST with `Mod` as the root node, I assume?) but will store the information GOOL needs. Likely we will want GOOL to perform multiple "passes", where earlier passes collect information to be used to actually generate code in later passes (having multiple passes should also help with generating code to interface with ODE solvers or other external libraries for #394). 

We should also be able to turn these optimizations on or off.

JacquesCarette:
Generally I agree. Though for the first example (removing import statements that are not needed), I would much prefer to do it the other way around: only include what is needed. In other words, the optimizer should have no work to do here [and possibly shouldn't even try]. For 'arithmetic' kinds of optimizations, that's a different matter.  While it might be possible to try harder to not generate silly code, this is insufficient and eventually an actual optimizer will be needed.

bmaclach:
So I've looked at all of our current tags in an attempt to find a pattern that will guide us on what information we need to store in an AST at each level. My hypothesis was that the pattern would be that each underlying GOOL type would store everything passed to the most generic constructor for that type plus a Doc. 

It ended up being not so simple for a number of reasons.
[...]

PR #3945, opened 2024-08-24
Introduced `LitValue` for expanded literal value handling

Xinlu-Y:
As discussed in #1867, `smartAdd` and `smartSub` are currently using literal integer values as `SValue` to perform simple optimizations. By creating a new datatype `LitValue`, we could extend support to more simple types of literal values (such as floats, doubles, and strings, which I believe would be useful) for addition and subtraction operations.

However, after making these changes, I noticed that the generated test code for `listSlice` has been altered. This change seems to occur specifically when the `step` is a variable and the issue arises from how I'm handling the return value of `valueInt` in the instances of `ValueElim`. 
image
The problem might stem from how these values are being unpacked and interpreted in the context of `listSlice` which remains somewhat puzzling to me.


code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/Constructors.hs
```diff
@@ -36,11 +36,11 @@ mkStmtNoEnd = flip stmtFromData Empty
 
 -- | Constructs a value in a stateful context
 mkStateVal :: (CommonRenderSym r) => VSType r -> Doc -> SValue r
-mkStateVal = valFromData Nothing Nothing
+mkStateVal = valFromData Nothing (LitInt 0)
```
JacquesCarette, line 39: This is suspicious to me: why is the static value defaulting to 0? (same in other places in this PR)

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -381,7 +381,21 @@ instance RenderValue JuliaCode where
 
 instance ValueElim JuliaCode where
   valuePrec = valPrec . unJLC
-  valueInt = valInt . unJLC
+  valueChar sc = case litVal (unJLC sc) of
```
JacquesCarette, line 384: I think you should be able to abstract out the
```
case litVal x of
   LitChar c -> Just c
   _  -> Nothing
```
part in a separate function so that it can be re-used across the different renderers

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/LanguagePolymorphic.hs
```diff
@@ -134,16 +134,26 @@ smartAdd v1 v2 = do
   v1' <- v1
   v2' <- v2
   case (RC.valueInt v1', RC.valueInt v2') of
-    (Just i1, Just i2) -> litInt (i1 + i2)
-    _                  -> v1 #+ v2
+    (Just i1,Just i2) -> litInt (i1 + i2)
```
JacquesCarette, line 137: I wonder if `smartAdd` should be "more typed", so that this cascade of checking doesn't happen.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/LanguagePolymorphic.hs
```diff
@@ -278,7 +288,7 @@ lambda f ps' ex' = do
   ps <- sequence ps'
   ex <- ex'
   let ft = IC.funcType (map (return . variableType) ps) (return $ valueType ex)
-  valFromData (Just 0) Nothing ft (f ps ex)
+  valFromData (Just 0) (LitInt 0) ft (f ps ex)
```
JacquesCarette, line 291: Here again, why default to 0?

code/drasil-gool/lib/Drasil/GOOL/RendererClassesCommon.hs
```diff
@@ -148,11 +148,15 @@ class RenderValue r where
   -- calls.
   call :: Maybe Library -> Maybe Doc -> MixedCall r
 
-  valFromData :: Maybe Int -> Maybe Integer -> VSType r -> Doc -> SValue r
+  valFromData :: Maybe Int -> LitValue -> VSType r -> Doc -> SValue r
```
JacquesCarette, line 151: The previous type was more precise! Perhaps `LitValue` is not the best way to go, but instead a synonym for `Maybe a` would be, where the type could be more precise.

code/stable/gooltest/cpp/HelloWorld/HelloWorld.cpp
```diff
@@ -126,33 +126,31 @@ int main(int argc, const char *argv[]) {
     mySlicedList7 = temp5;
     
     vector<double> temp6(0);
-    for (int i4 = 3; 0 < i4 && i4 <= 3 && z < 0; i4 += z) {
+    for (int i4 = 3; i4 < 0; i4 += z) {
```
JacquesCarette, line 129: I would not be surprised if this change was due to your defaulting some things to 0!

It is also strange that this changed, as you don't have any code for simplifying constraints / boolean expressions. So I really suspect it's the defaults.

B-rando1: These should be the relevant lines of code:
https://github.com/JacquesCarette/Drasil/blob/e72dd927fbcc94ffdf0eacda6252826d9b74bcff/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/Macros.hs#L98-L109

I believe @JacquesCarette was right: setting values to `litInt 0` makes the line `100` case trigger instead of the line `101` case.

Xinlu-Y: In this case, where `step` is a variable with a negative value, `valueInt stepV` will return `Nothing`, making `mbStepV = Nothing`, which I believe would trigger the logic starting at line 101. However, in this code, how does the default value `litInt 0` affect `mbStepV` and result in `mbStepV` getting a value of `0`, which subsequently generates the code `for (int i4 = 3; i4 < 0; i4 += z)`?

B-rando1: This is the line creating the `listSlice` in the example:
https://github.com/JacquesCarette/Drasil/blob/e72dd927fbcc94ffdf0eacda6252826d9b74bcff/code/drasil-code/test/HelloWorld.hs#L170-L173

I believe that the specific reason that line `100` is triggering instead of line `101` is that the value of `step` uses `valueOf`, which gives `litInt 0` as its literal value.

Variables aren't literal values; they hold values.  Since GOOL doesn't know what the value they're holding is, the correct `litVal` for them should be `Nothing`, as they aren't literal values.

JacquesCarette:
Looking at this code again, I don't think this is quite the right approach -- it loses too much type information. I think the 'static' information should be tracked in a different way. Most likely abandoning this PR and re-starting will be the easiest way forward.

Issue #3948, opened 2024-08-28
Adding `Scope` and `Motivation` Fields and Defining the Role of the `What` Section in README Files


Xinlu-Y:
This issue relates to both issue #3159 and #3248. Based on the comments above, generating a comprehensive software review isn't feasible for us at the moment. The main goal is to ensure that the `scope` and `motivation` are available in both the `SystemInformation` and in the `README` files.
- We can update the following sections in the generated `README` files for each example's src directory:
  -  **Example name**
  - **Authors**
  - **Motivation**
  - **Purpose**
  - **Scope**
  - **Making Examples**
  - ...
- For instance, taking the `Projectile` example, the `README` could look like this, where `Purpose`, `Motivation`, and `Scope` are optional subsections under `introInfo`:
  - **Projectile**
  - **Authors:** Samuel J. Crawford, Brooks MacLachlan, W. Spencer Smith
  - **Motivation:** Projectile motion is a common problem in physics.
  - **Purpose:** Predict whether a launched projectile hits its target.
  - **Scope:** The scope of the requirements includes the analysis of a two-dimensional (2D) projectile problem with constant acceleration.
  - **Making Examples**
How to Run the Program: In your terminal command line, enter the same directory as this README file. 
  - ...
- Additionally, a `What` section has already been created in the current code, but currently, it holds no information in the generated `README` file. https://github.com/JacquesCarette/Drasil/blob/e72dd927fbcc94ffdf0eacda6252826d9b74bcff/code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs#L33-L35
- I'm considering whether we could use this section to include relevant background information. If so, the `What` section might look like this:
  - ...
  - **Scope:** The scope of the requirements includes the analysis of a two-dimensional (2D) projectile problem with constant acceleration.
  - **What:** Common examples of projectile motion include ballistics problems (missiles and bullets) and the flight of the balls in various sports (baseball, golf, football, etc.).
  - **Making Examples**
  - ...

We need to decide whether it's necessary to keep the `What` section and, if so, what information should be included in it. One possibility is to move some of the content from `introInfo` to `whatInfo`. For example, we could have `introInfo` focus on answering the "why" questions (`motivation`, `purpose`), while `whatInfo` could address the "what" questions (`background`, `scope`).

JacquesCarette:
This is an excellent proposal! Yes, please start designing exactly this.

PR #3950, opened 2024-09-03
Implement Proposal #3948: Enhance README Generation

Xinlu-Y:
This PR implements the design outlined in proposal #3948.  The main objective is to ensure that the scope and motivation are included in both the SystemInformation and the generated README files.

Key Updates:
- Optional Subsections: `Motivation` an `Purpose` are now optional subsections under `introInfo`.
- 'What' Section: `background` and `scope` are now optional subsections under `whatInfo`.

Closes: #3948, #3159, #3248, #3971


code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -52,6 +52,10 @@ type CaseName = String
 type ExamplePurpose = String
 -- | Description of example
 type ExampleDescr = String
+-- | Motivation of example
```
JacquesCarette, line 55: While it is good that you are following the established pattern in this code, the problem is that the pattern is wrong... in that GOOL shouldn't know anything at all about 'examples'.

The real problem isn't these types, but `ReadMeInfo` which shouldn't be defined inside `GOOL` at all!

So I think the first step of the implementation, as a separate PR, would be to first fix this and move this information out of `GOOL`.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -150,6 +150,10 @@ genPackage unRepr = do
         (foldlSent $ purpose $ codeSpec g)
       bckgrnd = show $ sentenceDoc db Implementation OneLine
         (foldlSent $ background $ codeSpec g)
+      mtvtn = show $ sentenceDoc db Implementation OneLine
```
JacquesCarette, line 153: In that same refactor, I think all the stuff that concerns `README` generation should be moved out of this file. It could go into `Language/Drasil/Code/Imperative/Generator/README`.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -47,6 +47,10 @@ data CodeSpec where
   purpose :: Purpose,
   -- | Example Background.
   background :: Background,
+  -- | Example Scope.
```
JacquesCarette, line 50: I don't think this information should be in `CodeSpec` either. It is fine to have it in `SystemInformation` though.

JacquesCarette (changes requested):
Your actual changes are actually fine - but you are doing them against a bad design... so the right thing to do is to first fix the design, then apply these changes again.

Now, if you want, I would be fine with accepting the code as-is and have you fix the design after. I suspect that that is actually more work!

Xinlu-Y:
I got stuck after representing the fields `purpose, background, motivation, and scope` as `ConceptInstances`. By doing this, these pieces of information are stored in the `sysinfodb`, which is then copied into `CodeSpec`. However, in the `genPackage` function, I need assistance with retrieving a `ConceptInstance` by its `UID`. Any guidance from those familiar with this process would be appreciated.

JacquesCarette:
If @balacij can't help you tonight, then I will be at McMaster tomorrow and should be able to help (likely time: early afternoon).

code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
```diff
@@ -40,7 +41,7 @@ unmodularDesc = do
       implTypeStr Program = "program"
       implTypeStr Library = "library"
   return $ show $ sentenceDoc (sysinfodb spec) Implementation OneLine $ capSent $
-    foldlSent ([S "a", S (implTypeStr (implType g)), S "to"] ++ purpose spec)
+    foldlSent ([S "a", S (implTypeStr (implType g)), S "to"] ++ codeSpec g ^. sysInfo .purpose)
```
JacquesCarette, line 44: You should not need to "double lens" (i.e. `.sysInfo .purpose`). If you make the right instance, `.purpose` should be enough.

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -41,36 +40,3 @@ class AuxiliarySym r where
 
   auxHelperDoc :: r (AuxHelper r) -> Doc
   auxFromData :: FilePath -> Doc -> r (Auxiliary r)
-
```
JacquesCarette, line 44: Oh, this is a nice clean-up!

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/PythonRenderer.hs
```diff
@@ -5,8 +5,9 @@ module Language.Drasil.Code.Imperative.GOOL.LanguageRenderer.PythonRenderer (
   PythonProject(..)
 ) where
 
-import Language.Drasil.Code.Imperative.GOOL.ClassInterface (ReadMeInfo(..),PackageSym(..),
-  AuxiliarySym(..))
+import Language.Drasil.Code.Imperative.GOOL.ClassInterface (PackageSym(..), AuxiliarySym(..))
+import Language.Drasil.Code.Imperative.ReadMe.Import (ReadMeInfo(..))
```
JacquesCarette, line 9: Can you explain to me why all the language renderers need access to `ReadMeInfo`?

code/drasil-code/lib/Language/Drasil/Code/Imperative/WriteReadMe.hs
```diff
@@ -5,7 +5,7 @@ module Language.Drasil.Code.Imperative.WriteReadMe (
 import Language.Drasil.Choices (ImplementationType(..))
 import Language.Drasil.Printers (makeMd, introInfo, verInfo, unsupOS,
     extLibSec, instDoc, endNote, whatInfo)
-import Language.Drasil.Code.Imperative.GOOL.ClassInterface (ReadMeInfo(..))
+import Language.Drasil.Code.Imperative.ReadMe.Import (ReadMeInfo(..))
```
JacquesCarette, line 8: So much better!

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -70,7 +67,9 @@ data CodeSpec where
   -- automatically define.
   mods :: [Mod],  -- medium hack
   -- | The database of all chunks used in the problem.
-  sysinfodb :: ChunkDB
+  sysinfodb :: ChunkDB,
```
JacquesCarette, line 70: I thought the aim was to rename this `CodeSpec` to `OldCodeSpec` and then make `CodeSpec` has just 2 fields?

JacquesCarette (changes requested):

JacquesCarette:
@Xinlu-Y can you please fix the conflict that was recently created?

Xinlu-Y:
@JacquesCarette Dr. Carette, Could you please review the latest changes when you have time? I have addressed the earlier feedback and resolved the conflicts. Thank you!

JacquesCarette (approved):

PR #3952, opened 2024-09-08
#3252 Sentence Reworking (for the Glassbr Example) 

daijingz:
Now these changes are much more worthy.

@JacquesCarette @samm82 I split all updates by each Drasil example, and now you can see these updates become much more worthy-to-change.
For the hghc example, I only found a few (very few) constructed sentences in the folder `drasil-example`. (I did not find any potential places to change) I just skip this example, is that OK?


JacquesCarette (approved):
I don't know about "much more worthy", but they are legitimate small improvements.

PR #3953, opened 2024-09-08
#3252 Sentence Reworking (for the GamePhysics Example, Part 1)

daijingz:
Update some sentence re-works in the Gamephysics example.
@JacquesCarette @balacij @samm82 Because the GamePhysics examples still have lots of sentences, I am going to submit another PR by the end of today.


JacquesCarette (approved):
Some more tiny changes that are fine.

PR #3967, opened 2024-09-20
#3252 Sentence Reworking (for the ssp Example, Part 2)

daijingz:
Continue with #3964


JacquesCarette:
Can you do your future work on clean branches (and not main)? This is making the history quite dirty.

PR #3993, opened 2024-12-30
take care of all the comments from Jason

JacquesCarette:
As per the comments on the first PR on this, aka #3978 .

### DIFF
--- a/wiki/What-If.md
+++ b/wiki/What-If.md
@@ -34,17 +34,34 @@ often are not. A non-exhaustive list:
 
 **N**: And what do you mean by *generating*? Like templates?
 
-**S** Not like templates. Like a program that uses a more semantic source for
-all the artifacts, selects the appropriate information for each piece, and
-writes each out in its *native format*. The basic idea is that each
-artifact contains a quite particular *view* of the total information behind
-a single software product, and so the generation process is one of
-filtering and translation.
+**S** Sort of. Many templates are simplistic fill-in-the-hole, and
+that's not what is meant here. What we mean is turning a semantic
+description of all of the artifacts, and then *rendering it* it in
+one (or more) programming languages. We see each artifact that is generated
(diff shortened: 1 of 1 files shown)


wiki/What-If.md
```diff
@@ -34,17 +34,34 @@ often are not. A non-exhaustive list:
...
-filtering and translation.
+**S** Sort of. Many templates are simplistic fill-in-the-hole, and
+that's not what is meant here. What we mean is turning a semantic
+description of all of the artifacts, and then *rendering it* it in
+one (or more) programming languages. We see each artifact that is generated
+as a particular *view* onto the semantic information behind a single
+software product, so the generation process is more one of filtering and
+translation.
+
+In other words, we want to *compute* the results.
```
balacij, line 45: So, I'm quite happy with the way this was written. It definitely makes sense _to me_. I think that a figure would go a long way to make it understandable to others.

I'm thinking a figure approximately with a two-column layout:
1. The template side, that shows what a typical template looks like. Pulling a slightly complicated example from sqlc: [SQL query Golang code generation](https://github.com/sqlc-dev/sqlc/blob/main/internal/codegen/golang/templates/pgx/queryCode.tmpl) with a configuration file (they have a `sqlc.yaml` file that projects need to use) and an input SQL file (e.g., their [booktest](https://github.com/sqlc-dev/sqlc/blob/main/examples/booktest/postgresql/query.sql) example). Maybe it can be a three-column layout with a really dead simple example of template usage and another column using this example.
2. Our side, which would be similar to the sqlc example, except for the depth/number of "templates" used and how mapping/rendering/transforming (what term should we use here?) higher-level concepts into fragments related to the templates.

JacquesCarette: yes, if we can illustrate things that way, that would be great!

Issue #3992, opened 2024-12-28
Issues with `forRange`


B-rando1:
As I noted in #3991, Julia's ranges include the `end` value, whereas other languages do not. At first I thought we could get away with a quick fix of subtracting `1` from the given `end` value, but then I remembered that negative `step` values exist, and if `step` is negative we need to add `1` instead.

Furthermore, I looked at other generated code with negative step values (test cases for this do not currently exist; we will need to add some to `main`) in other langauges and realized that three other languages does not handle negative step values correctly either. Basically it's a similar issue to what we had in #3809, where our current implementations do not account for negative `step`.

What needs to be done:
- Julia: if `step` is positive, subtract `1` from `end`; if `step` is negative, add `1` to `end`.
- Java, C#, C++: change the inequality checking `i` so it's `<` when `step` is positive and `>` when `step` is negative.
- Python is fine, and Swift appears to be fine as well. I don't have Swift installed on my current machine, but I can install it to double-check at some point.

As with `listSlice`, we can make it so that if the value of `step` is known at generation-time, we generate more concise code. Otherwise we need to generate code to do that logic at runtime.

JacquesCarette:
It feels to me that we need to look more deeply into what `forRange` means (and ought to mean) to make the rendering in each language more transparent.

I definitely agree with the last paragraph: we need to be clearer about what information is known when (this is something that @Xinlu-Y was working on, but it got very complicated rather quickly). We need to 'teach' our representations more about staging.

B-rando1:
I made a PR to fix Julia's `forRange` for positive `step` values. Beyond that, there seems to be something of a rabbit-hole, so it might be best to leave it for now.

PR #3994, opened 2024-12-31
Fixed `forRange` in Julia for positive `step` values

B-rando1:
Julia's ranges include the end value, and other languages don't. I updated the Julia renderer to subtract `1` from the end value to match the other languages.

Note that while this fixes it for positive `step` values, it does not for negative `step` values. More work needs to be done here.

Contributes to #3992


B-rando1:
I'm not sure what's going on with HLint. The error in the log is `E: Unable to locate package libtinfo5`.

code/stable/gooltest/julia/VectorTest/VectorTest.jl
```diff
@@ -2,11 +2,11 @@ module VectorTest
 
 global v1 = [1.0, 1.5]
 global v2 = [0.0, -1.0]
-for i in 0:1:length(v1) - 1
+for i in 0:1:length(v1) - 1 - 1
```
JacquesCarette, line 5: This seems odd -- length - 2 ?  This looks like it is over-compensating!

JacquesCarette (changes requested):

code/stable/gooltest/julia/VectorTest/VectorTest.jl
B-rando1: Good catch. I should have noticed that. It looks like I accounted for the range difference in the `thunkAssign` function, which uses the `forRange` function. Because this change is upstream of it (most likely the 'correct' place), it was affected. My new commit should remedy that.

JacquesCarette (approved):
I'm approving now, but won't merge in quite yet, as the linter did not run -l ooks like an upstream problem though, i.e. not your fault.

JacquesCarette:
#3998 should fix that, but I'm waiting for a review.

balacij (approved):
Pending the CI, this PR LGTM!

JacquesCarette (approved):

PR #3998, opened 2025-01-01
We don't need the hack anymore

JacquesCarette:
and it is broken -- `libtinfo5` is not found anymore on Ubuntu LTS 24.04.

We can probably simplify this further (see https://github.com/haskell-actions/hlint-setup ), but that can be a subsequent PR.


balacij:
Nice!

Regarding the follow-up, we already use hlint-setup, this hack was only necessary for earlier versions of HLint.

balacij (approved):

PR #4002, opened 2025-02-04
Remove unused code: `Language.Drasil.URI.AST`

balacij:
Contributes to #2885


JacquesCarette:
My guess is that, some time in Drasil's future, URIs will come back. But I agree, that might be far enough in the future that deletion for now is the better call.

JacquesCarette (approved):

PR #4009, opened 2025-02-19
Geometric Algebra Types

CSchank:
Hi Dr. @JacquesCarette,

I was hoping you could give me some feedback on what I have so far. It's definitely far from complete, this is not really meant to be merged as-is but is a very early WIP. I have not actually removed anything yet, just started to build up what I hope is the right infrastructure to internally represent clifs. I'm hoping you can tell me if I'm on the right track and answer some questions:

- [ ] Should Clifs be `Expr` constructors directly instead of their own type?
  - [ ] Should Clifs take in `Expr` values instead of `Clif` values?
- [ ] Should I only have `NGradeC` and not have the others (like `Vector`, `Bivector`, etc)? Dr. Smith thought we maybe should have the "usual" cases, but those could be smart constructors.
- [ ] How should the basis be incorporated? I have something here based on what we talked about but I'm not sure if I'm on the right track.
- [ ] Where do smart constructors go? I believe in `module Language.Drasil.Expr.Class`.
- [ ] For `Space`s, should Clifford spaces be able to take in other `Space` values, like you've done for these:`Set Space | Matrix Int Int Space | Array Space`?
- [ ] We talked about implementing matrices as functions from `Vector -> Vector`, how do we represent these in the current `Expr`, or do I need more constructors?
  - [ ] I'm thinking maybe I should replace anything with a "V" (e.g. `UnaryOpVV`) in it with clifs instead of vectors.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,20 @@
+{
```
JacquesCarette, line 1: any reason to add this file?

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -160,6 +160,38 @@ data Expr where
   Operator :: AssocArithOper -> DiscreteDomainDesc Expr Expr -> Expr -> Expr
   -- | A different kind of 'IsIn'. A 'UID' is an element of an interval.
   RealI    :: UID -> RealInterval Expr Expr -> Expr
+  -- | A clif is an expression
+  Clif     :: Clif -> Expr
+
+-- | The basis in which to project clifs
+data Basis where
+  -- | ℝⁿ
+  Rn :: Int -> Basis
```
JacquesCarette, line 169: I think we have the opportunity here to use `Natural` or `Nat` from `Data.Nat`. Let's build more invariants into our representations!

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -160,6 +160,38 @@ data Expr where
...
+
+-- | A clif is a clif, or the addition of clifs, or the scalar multiplication or geometic product of clifs
```
JacquesCarette, line 171: I think it might make sense to separate base `Clif` from `Clif`-forming operations.

Also, in the AST, I don't think it makes a lot of sense to have all those special cases - go straight for `NGradeC`. We can have the external API provide all these special cases (both for input and output).

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -160,6 +160,38 @@ data Expr where
...
+  -- | A scalar clif
+  ScalarC     :: Clif
+  -- | A vector clif with a given dimension
+  VectorC     :: S.Dimension -> Clif
+  -- | A bivector clif with a given dimension
+  BivectorC   :: S.Dimension -> Clif
+  -- | A trivector clif with a given dimension
+  TrivectorC  :: S.Dimension -> Clif
+  -- | An n-grade clif with a given dimension
+  NGradeC     :: Int -> S.Dimension -> Clif
```
JacquesCarette, line 182: `Nat` or the like here too.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -43,6 +44,30 @@ data Space =
...
   | Function (NE.NonEmpty Primitive) Primitive
   | Void
+  -- | Clifford algebra objects (Clifs)
+  | Clif ClifShape
+  deriving (Eq, Show)
+
+-- | The dimension of a clif
+data Dimension where
+  -- | Fixed dimension
+  Fixed :: Int -> Dimension
```
JacquesCarette, line 54: Again, `Nat`.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -43,6 +44,30 @@ data Space =
...
+  -- | Variable dimension
+  VDim  :: String -> Dimension
+  deriving (Eq, Show)
+
+-- The shape of a clif object
+data ClifShape where
```
JacquesCarette, line 60: Let's have a smaller `Shape` type without as much redundancy. There will be ample time to have more shapes of things (especially when it comes to representations), not point making this complicated.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -43,6 +44,30 @@ data Space =
...
+  -- Scalars
+  ScalarCS    :: ClifShape
+  -- Vectors with a given dimension
+  VectorCS    :: Dimension -> ClifShape
+  -- Bivectors (geometric extent 2) with a given dimension
+  BivectorCS  :: Dimension -> ClifShape
+  -- Trivectors (geometric extent 3) with a given dimension
+  TrivectorCS :: Dimension -> ClifShape
+  -- Clifs (geometric extent n) with a given dimension
+  NGradeCS    :: Int -> Dimension -> ClifShape
```
JacquesCarette, line 70: `Nat`

JacquesCarette (changes requested):

JacquesCarette:
As to your questions:
- A single constructor to embed a `Clif` into an `Expr` is better
- Let's wait on that design. It will be easiest to figure out what to do with bases when we try to actual enter a `Clif`.
- yes, `Language.Drasil.Expr.Class` is where the other smart constructors seem to be, currently
- well `Space` is our name for `Type`. So we need to understand what the `Type` of a `Clif` ought to be. So yeah, we do need to know the 'contents' of a `Clif`. 
- no, matrices *represent* linear functions. the last thing we want to do is represent them that way! We want them to be "arrays of numbers" in the most common case. There's 2-levels of "represent" here: matrices *represent* linear functions, and we use arrays to *represent* matrices. (And yes, anything with a V should get ported to clifs)

.devcontainer/devcontainer.json
CSchank: Oh sorry, I should have explained. This was so I could develop the project on GitHub Codespaces (instead of using more disk space on my laptop). It's up to you if you want to merge it in or not, I can remove it from any future mergeable PRs. It's actually not foolproof because it doesn't install Stack by default yet. If you're interested I'll get it fully working, if not, I'll leave it as-is/remove it before we merge.

CSchank:
@JacquesCarette Thanks, I'll get to work on those things! 

So, representing matrices as arrays makes sense, and it along the lines of what you're already doing. I guess what I'm still not sure about is how we unify the idea of clifs and arrays (since matrices can be represented as arrays)? How we represent a vector is pretty obvious: it's a vector (grade-1) clif. But how do we represent multi-dimensional arrays (and therefore, matrices) as clifs? Or do we keep around the idea of arrays? But that feels weird because I was under the impression that we're trying to unify the whole understanding of vectors/matrices/etc under Clifford/geometric algebra.

.devcontainer/devcontainer.json
JacquesCarette: On that topic, I'll want the opinion of @balacij .

JacquesCarette:
Arrays already exist -- in the backend. They are "implementation structures". We don't need to talk about them at the theory / specification stage.

If you start trying to implement vectors (and just vectors, but as clifs) and try to push them through the whole codebase, I think it will all become much clearer. Let Haskell lead you to where things break.

Pithy motto: vectors (clifs) in the math, arrays in the CS. Drasil's pipeline actively transforms math to CS.

JacquesCarette:
The latest commit is a nice way to track down everything that needs to be touched.

CSchank:
> The latest commit is a nice way to track down everything that needs to be touched.

Thanks, my plan is to try to get it all ported over to the new exprs/"shapes" and have it compile. I just committed what I had when I stopped last night to "save" it even though it doesn't currently compile.

CSchank:
Finished doing a bunch of refactoring. In addition to looking at it, I had some questions (some of which correspond to TODOs in the code):

- Inner data types: changed all V’s to C’s
- Smart constructors: changed c’s to v’s, but added aliases for v* as well
- I am a bit confused why I am getting some type errors (even prior to my changes), e.g.:  
```
=====[ Start type checking ]=====
theory:imPDRC` exposes ill-typed expressions! - ERROR: Inferred type `Vect Real` does not match expected type `Real`
- ERROR: Inferred type `Vect Real` does not match expected type `Real` ```
- Restricted many clif operations to vectors only for now.
```
- In Physics.hs : currently acceleration is a vector of real, but without a specific dimension. But new constructor requires a dimension. We may need to rethink.
- ReadInput.hs: 
    - `getDimension`: Does this make sense? I think maybe we’re overloading the term “dimension” now.
    - `strListAsExpr`: Not sure how to implement this
    - `strList2DAsExpr`: Not sure how to implement this
- ODELibraries.hs:
    - What dimensions do we use? In the old type, you don’t specify a dimension. But now we have to.
- GlassBR:
    - I assume it’s 3D?
    - ModuleDefs.hs
        - extractColumnCT: correct implementation?

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -443,15 +443,15 @@ odeint = externalLib [
       customObjArg [] "Class representing an ODE system" ode odeCtor
         (customClass [constructorInfo odeCtor [] [],
           methodInfoNoReturn odeOp "function representation of ODE system"
-            [unnamedParam (Vect Real), unnamedParam (Vect Real), lockedParam t]
+            [unnamedParam undefined, unnamedParam undefined, lockedParam t] -- TODO: what do we put here? We need specific dimensions now.
```
JacquesCarette, line 446: Design decision needs to be made: either we support "arbitrary" dimensional vector spaces (i.e. where the dimension is not mentioned" or we support "specific but unknown" (i.e. symbolic) dimension. Or both.

code/drasil-code/lib/Language/Drasil/Code/Code.hs
```diff
@@ -23,7 +23,7 @@ spaceToCodeType S.Rational       = [Double, Float]
 spaceToCodeType S.Boolean        = [Boolean]
 spaceToCodeType S.Char           = [Char]
 spaceToCodeType S.String         = [String]
-spaceToCodeType (S.Vect s)       = map List (spaceToCodeType s)
+spaceToCodeType (S.ClifS _ _ s)  = map List (spaceToCodeType s)
 spaceToCodeType (S.Matrix _ _ s) = map (List . List) (spaceToCodeType s)
```
JacquesCarette, line 27: I guess you're choosing not to replace `Matrix` with `ClifS` this go around?

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
```diff
@@ -7,7 +7,7 @@ import Language.Drasil hiding (Data, Matrix, CodeVarChunk)
 import Language.Drasil.Code.DataDesc (DataDesc'(..), Data'(..), DataItem'(..),
   Delimiter, dataDesc, junk, list, singleton')
 import Language.Drasil.Chunk.Code (CodeVarChunk)
-import Language.Drasil.Expr.Development (Expr(Matrix))
+import Language.Drasil.Expr.Development (Expr(Matrix, Clif)) -- TODO: remove Matrix entirely
```
JacquesCarette, line 10: Ah, if I had not read the code sequentially, I would have seen this!

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
```diff
@@ -83,7 +83,7 @@ strAsExpr _        _ = error "strAsExpr should only be numeric space or string"
 
 -- | Gets the dimension of a 'Space'.
 getDimension :: Space -> Int
-getDimension (Vect t) = 1 + getDimension t
+getDimension (ClifS _ _ s) = 1 + getDimension s -- TODO: Does this make sense? Maybe we're overloading the term "dimension" now.
```
JacquesCarette, line 86: I agree, this is weird. Can you figure out who calls `getDimension` and what they need from it?  It looks like this is always going to return '1' for a `Vector` and 0 in all other cases!

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
```diff
@@ -98,12 +98,12 @@ splitAtFirst = splitAtFirst' []
         dropDelim [] s = s
         dropDelim _ [] = error "impossible"
 
--- | Converts a list of 'String's to a Matrix 'Expr' of a given 'Space'.
+-- | Converts a list of 'String's to a Clif 'Expr' of a given 'Space'.
 strListAsExpr :: Space -> [String] -> Expr
-strListAsExpr (Vect t) ss = Matrix [map (strAsExpr t) ss]
```
JacquesCarette, line 103: you shouldn't do `Matrix` partially; either try for all cases or none, but not just some!

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
```diff
@@ -98,12 +98,12 @@ splitAtFirst = splitAtFirst' []
...
+strListAsExpr (ClifS gr d s) ss = undefined -- TODO: fill this in
 strListAsExpr _ _ = error "strListsAsExpr called on non-vector space"
 
--- | Converts a 2D list of 'String's to a Matrix 'Expr' of a given 'Space'.
+-- | Converts a 2D list of 'String's to a Clif 'Expr' of a given 'Space'.
 strList2DAsExpr :: Space -> [[String]] -> Expr
-strList2DAsExpr (Vect (Vect t)) sss = Matrix $ map (map (strAsExpr t)) sss
+strList2DAsExpr (ClifS gr0 d0 (ClifS gr1 d1 s)) sss = undefined -- TODO: fill this in
```
JacquesCarette, line 108: Trying to fill these particular TODOs will be very instructive.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
```diff
@@ -152,7 +152,7 @@ findCT = funcDef "find"
 
 extractColumnCT :: Func
 extractColumnCT = funcDef "extractColumn" "Extracts a column from a 2D matrix" 
-  [mat, j] (Vect Real) (Just "column of the given matrix at the given index")
+  [mat, j] (vect3D Real) (Just "column of the given matrix at the given index") -- TODO: is this correct?
```
JacquesCarette, line 155: while the above `3D`s are likely right, I agree, this one is less clear. You'll have to see where `extractColumnCT` is used to see what the context says.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
```diff
@@ -415,7 +415,7 @@ outputs = [tempW, tempPCM, watE, pcmE]
 tempW = cuc' "tempW"
   (nounPhraseSP "temperature of the water")
   "the average kinetic energy of the particles within the water" 
-  (sub (eqSymb temp) lWater) centigrade (Vect Real)
+  (sub (eqSymb temp) lWater) centigrade (vect undefined Real) -- TODO: dimension of vector?
```
JacquesCarette, line 418: I believe this is 3D

JacquesCarette (changes requested):

JacquesCarette:
The type checking errors is because... things are currently broken! Our use of vectors is quite bad, i.e. type incorrect. And there was no point in fixing it in the current state.

We will indeed encounter even worse bugs. Like you have in `Physics` that will require us to think and make changes. This is expected.

For the other things you ask, I left comments in the PR itself so that they are 'in context'. The way you asked with insufficient context made it impossible for me to understand what the questions really were.

code/drasil-code/lib/Language/Drasil/Code/Code.hs
CSchank: Oops - I should have included that as another question. I didn't because I wasn't quite sure what kind of clif it should be represented as. Do you have thoughts? Would an m-by-n matrix be represented as something like an n-grade clif in an m-dimensional space? Dr. Smith and I did some research and there's not much about using clifs to represent matrices. The geometric algebra people have this attitude of "we don't need matrices" so they seem not to even think about them.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
CSchank: Yes, as above, I would like some thoughts on what to replace it with.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
CSchank: Well, we had things like `Vect (Vect Natural)` which would return 2.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
CSchank: This is another question I had: does geometric algebra have such a thing as "clifs of clifs"?

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
CSchank: Yes. Currently, I support specific dimensions and symbolic dimensions (currently, very simple ones: strings). We could add more such as "dimension is not mentioned" (not sure the ramifications of that for the operations), and things like addition of dimensions.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
CSchank: Yeah, so this is to extract a given column in the matrix. It's used as part of this:

```haskell
interpMod :: Mod
interpMod = packmod "Interpolation" 
  "Provides functions for linear interpolation on three-dimensional data" []
  [linInterpCT, findCT, extractColumnCT, interpY, interpZ]
```

So yes, 3D.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
smiths: I'm having trouble thinking of what the dimension of tempW should be.  The temperature of the water is a scalar value.  The problem is 1D.  There is only one temperature for the water.  It varies over time, but not over space.  Even if temperature varies over space, it is still a scalar.  The only reason I can think for a vector is that in solving the ODE we use a vector, but Tw is one of the components of the vector, not a vector itself.

.devcontainer/devcontainer.json
balacij: I'm not against this, but I'm hesitant to say we should merge until (at least):
1. We've cleaned up our own project configuration a little bit.
2. We can install all dependencies (for which we have many) through the file.
3. We write instructions on how to use it.
4. We confirm that it works for all of usual tasks and that we have a usecase. The [pricing of Codespaces](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) doesn't look too bad, which is certainly appealing if we ever have students with low-spec computers!

Does HLS work with this? I would imagine that GitHub Codespaces has fairly limited resources?

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
```diff
@@ -44,7 +44,8 @@ acceleration, angularAccel, angularDisplacement, angularVelocity, chgInVelocity,
   yVel, momentum, moment, moment2D, fOfGravity, positionVec, tension, angularFrequency, 
   period, frequency, chgMomentum :: UnitalChunk
 
-acceleration           = uc CP.acceleration           (Concat [vec lA, label "(", lT, label ")"]) (Vect Real) accelU
+-- TODO: what should the type here be?
+acceleration           = uc CP.acceleration           (Concat [vec lA, label "(", lT, label ")"]) undefined   accelU
```
balacij, line 48: Taking a look at usage in GamePhysics, a `Real -> Real` function seems more likely than a `Vect Real`.

balacij: Note that functions are still mostly unsupported as well, hence why the time variable is manually added to the symbol of acceleration.

smiths: Another option for the type is `Real -> Vect Real`.  In 3D the acceleration is a function of time.  In general, at any time, we can have 3 components of acceleration.  We also have a 1D version of acceleration in Game Physics which would have the type `Real -> Real`.

balacij:
Out of curiosity, which resources are we using to pull this from? I'm unfamiliar with Clifford algebra.

Also, assuming you're still designing the Clif data type, it _might_ be worth implementing a prototype externally first (because Drasil can be difficult to work with) with:
* A stripped down copy of Drasil's `Expr`.
* A simple type-checker on said `Expr` lang.

Afterwards, you can:
1. Extend `Expr` with your Clifford algebra work (along with the type-checker). 
2. Figure out how simple expressions in `Expr` involving Clif deal should be translated to, say, Java (or any one of Drasil's currently supported languages) -- without a code prototype.

Afterwhich, we can figure out how Drasil needs to be pushed around to implement the Clifford algebra and have the Clif code generation work.

If you chose to do this, nothing in this PR would be lost.

(I mention this in part because I intend to soon edit nearby pieces of code, so I want to avoid merge conflicts as much as possible.)

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
JacquesCarette: This is a symptom of a deeper problem in Drasil: acceleration should exist in many different ways (as a concept, as an instance of properties for several objects, but also as something that may be represented in different ways!)

So `acceleration` can be `Real`, `Vect Real` and `Real -> Real` depending on the case at hand. Thus we need to be able to do that.

.devcontainer/devcontainer.json
JacquesCarette: So we should definitely not merge this in via this particular PR.

smiths:
@balacij you can learn more about Clifford Algebra from Chris's [SRS slides](https://github.com/CSchank/TensorCodeGen/blob/main/docs/SRS/SRS.pdf).  The slides cite a good [introduction to Clifford Algebra](https://www.av8n.com/physics/clifford-intro.htm).

An external prototype is an idea that is definitely worth thinking about.

JacquesCarette:
While I agree that working directly in Drasil is difficult, shaking out the design will only be possible within the full Drasil. Until it proves too difficult, I think it best to stay the current course.

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
JacquesCarette: Let's try to handle this particular case with symbolic dimensions and see what that means, i.e. what else that forces us to think about.

code/drasil-code/lib/Language/Drasil/Code/Code.hs
JacquesCarette: https://math.stackexchange.com/questions/850869/linear-algebra-without-matrices is one place to start reading. It points to the nice book "Linear Algebra Done Right" which is really quite a gem. [In a sense, chapter 1 of https://arkadiusz-jadczyk.eu/docs/clifford.pdf says that concrete Clifford Algebras are equivalent to some spaces of matrices...]

However, we will still want to be able to enter 'concrete' matrices (in a basis). 

The book "Guide to Geometric Algebra in Practice " is available for download from McMaster's library - I think that one would help most.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
JacquesCarette: Ah, yes. Horrible hack that. See if you can get rid of it?

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
JacquesCarette: No, GA does not nest.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
JacquesCarette: Right, so this is 'different' in the sense that this isn't a semantic operation but rather this is a programming-level operation where an array is used for packed storage and then its columns are extracted.

So in a separate language (`ModelExpr` ?), we need ways to talk about storage and concrete representations. It's a typical mistake to conflate these.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
JacquesCarette: So this is a bug in our 'formalization' of `tempW`?

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
JacquesCarette: So this is where @CSchank is likely going to need to rely on @balacij sorting out some stuff in our chunks.

For this one: I don't think the *concept* of `acceleration` should be a `UnitalChunk` because its 'units' will actually depend on the setting in which it is used, i.e. acceleration gets a unit once the object of concern is known.

We could treat it as a vector of unknown dimension if we insisted on giving it units 'early', thus indicating that it is a *dependent* concept.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
smiths: I don't know if it is a bug, or a misunderstanding on my part.  The generated documentation for `tempW` has been reviewed many times, so I believe it is correct.  The definition of `tempW` looks good (the average kinetic energy part) and I believe that part appears in the generated documentation.  I don't know if the type of `tempW` is ever displayed or used.  If the "formalization" of tempW is not used to do anything, it could be wrong and we'd never know because there are no consequences of a mistake.  If someone could explain the consequences of the formalization in the generated artifacts, I would help judge whether they are correct or not.  If they really aren't used, then I'm fairly sure they are currently incorrect.  :smile:

.devcontainer/devcontainer.json
CSchank: @balacij HLS works on Codespaces just fine, the smallest one has 2 vCores, 8gb of RAM and 32gb of storage. But I agree we shouldn't merge this yet.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
CSchank: Didn't think so. So, if Clifs take `Expr` as an argument, we will have to have type checking to make sure they don't take another clif as an argument.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
smiths: @JacquesCarette should this separate language be within the scope of the project by @CSchank?  Ideally I would like us to include at least parts of this other language as needed for the current examples, but I don't have a feel for how difficult a task this is.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
```diff
@@ -151,8 +151,8 @@ pOps Dim      = command "mathsf" "dim"
 pOps Exp      = pure $ text "e"
 pOps Neg      = pure hyph
 pOps Cross    = texSym "times"
-pOps VAdd     = pure pls
-pOps VSub     = pure hyph -- unfortunately, hyphen and - are the same
+pOps CAdd     = pure pls
```
ptrandev, line 154: I see that this PR uses Clifs to supersede Vectors in some fashion. Is the goal to allow reverse-compatible representations of Vectors and Vector math while also allowing for representations of Scalars, Bivectors, etc. via Clifs?

code/drasil-lang/lib/Language/Drasil.hs
```diff
@@ -269,6 +270,7 @@ module Language.Drasil (
 
   -- Language.Drasil.Space
   , Space(..)
+  , vect2D, vect3D, vect
```
ptrandev, line 273: I also see that you are referencing vect, vect2D, vect3D here.

Do these correspond to vectors, bivectors, and trivectors as defined in the SRS and would be represented as clifs? I.e. under the hood, they are represented as Clif, but I am also able to rely on the vect, vect2D, vect3D type in Drasil?

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
ptrandev, line 10: When you are talking about removing Matrix entirely, will this type no longer exist in Drasil? Or will it be represented as a Clif in the backend, but any existing code relying upon the Matrix type will still work as usual?

ptrandev (commented):
Coming from https://github.com/CSchank/TensorCodeGen/issues/13

I'm a relative novice with Haskell, but adding feedback to strengthen the conceptual understanding of what this PR is achieving.

.devcontainer/devcontainer.json
CSchank: Also, there is a way to get "unlimited" (for practical purposes) Codespaces for free for education settings. It is nice for helping people get started for sure.

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
CSchank: @balacij any chance we can work on this?

CSchank:
The other day I:

- Investigated how to represent general-dimension vector of temperatures in SWHS example
- Converted all examples to use vectors of length “n”

This raised some questions we have to answer:
- Does it make sense to be using strings to represent variable-length dimensions for vectors (and other clifs)? I followed closely to how variables were defined in other parts of Drasil.
- How will we support operations on vector length? (addition, subtraction, etc e.g. vector insertion makes the dimension into `n + 1`, and vector "concatenation" of length `m` and `n` gives `m+n`)

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -129,10 +130,10 @@ data Expr where
   UnaryOp       :: UFunc -> Expr -> Expr
   -- | Unary operation for @Bool -> Bool@ operations.
   UnaryOpB      :: UFuncB -> Expr -> Expr
-  -- | Unary operation for @Vector -> Vector@ operations.
-  UnaryOpVV     :: UFuncVV -> Expr -> Expr
-  -- | Unary operation for @Vector -> Number@ operations.
-  UnaryOpVN     :: UFuncVN -> Expr -> Expr
+  -- | Unary operation for @Clif -> Clif@ operations.
+  UnaryOpCC     :: UFuncCC -> Expr -> Expr
+  -- | Unary operation for @Clif -> Clif@ operations.
```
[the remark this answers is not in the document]
CSchank: Thank you!

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
CSchank: Good question :)

code/drasil-lang/lib/Language/Drasil.hs
CSchank: They represent vectors of different dimensions (1D, 2D, 3D, etc). The _grade_ of a clif determines if it's a scalar, vector, bivector, trivector, etc. Clifs of different grades can be embedded in various dimensions. For example, we can have a _vector_ in 3 dimensions or 2 dimensions, or a _bivector_ in 2 dimensions, 3 dimensions, 4 dimensions, etc. The only stipulation appears to be that a grade-n clif must have dimension $d \leq n$. A scalar of dimension 3 (or 2, or 1, for that matter), obviously, makes no sense. Similarly, a vector of dimension 0 makes no sense either.

code/drasil-printers/lib/Language/Drasil/TeX/Print.hs
CSchank: Yes, in essence. My idea here was to replace any references to "vector" in the lower-level `Expr` language with "clif" (so V's become C's). Vector addition is ultimately just Clif addition on grade-2 clifs.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
```diff
@@ -415,7 +415,7 @@ outputs = [tempW, tempPCM, watE, pcmE]
 tempW = cuc' "tempW"
   (nounPhraseSP "temperature of the water")
   "the average kinetic energy of the particles within the water" 
-  (sub (eqSymb temp) lWater) centigrade (Vect Real)
+  (sub (eqSymb temp) lWater) centigrade (vectND "n" Real) -- TODO: investigate the implications of this
```
CSchank, line 418: I spoke to Dr. @smiths today about the SWHS example. We think that the type of the [tempW constraint](https://github.com/JacquesCarette/Drasil/blob/905a96bb60f83e627241c0a5fbc419c7fe87b146/code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs#L415) is wrong. It should probably be `Real` if the variable is seen as a single Celsius temperature (which it how it's defined earlier on). But it gets a bit confusing because some parts of the document refers to it as T_w(t), meaning the temperature over time, which implies that it is a function. In code this function may be implemented as a discretized vector of temperatures at different time intervals, which may be where the confusion came from.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
JacquesCarette: Correct.

code/drasil-example/glassbr/lib/Drasil/GlassBR/ModuleDefs.hs
JacquesCarette: It should be in-scope. It's a much simpler language.

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
JacquesCarette: @balacij might now be free, given that his Comp II was earlier today.

code/drasil-code/lib/Language/Drasil/Code/Imperative/ReadInput.hs
JacquesCarette: We're still figure out what "a representation of a linear operator on a Cliff" is.

There will still be matrices in the generated code.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
JacquesCarette: Wonderful! I was sure that work like this was going to find more bugs.

code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
balacij: Hi, sorry @CSchank! I will message you on Discord!

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
CSchank: Yes, I came to the same conclusion. This appears to result in the output constraints table (below), but it was likely never caught because the type is not printed, as you say.

image

smiths: I'm fine with using `Real` for Tw just to move on, but the constraint isn't on a single value of Tw; it is on every value of the function for Tw between the start of the simulation and the conclusion of the simulation.  The constraint should be something like:

$$ \forall (t: \mathbb{R} | 0 \leq t \leq t_\text{final} : T_\text{init} \leq T_w(t) \leq T_c)$$

Since we don't handle types consistently right now, I don't think we should make any changes.  It also seems out of scope of the work by @CSchank to explore this.  :smile:

CSchank: @smiths well, I now have a partial answer why this was listed as a vector type. I tried running the example again changed to `Real` and I got an error produced by [this line of code](https://github.com/JacquesCarette/Drasil/blob/f8e88e58c964af867c62f70c445e9ec3e5de5b11/code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs#L650), producing the following error:

```
=====[ Start type checking ]=====
`theory:eBalanceOnWtrRC` exposes ill-typed expressions!
  - ERROR: Associative arithmetic operation expects all operands to be of the same expected type.
    Received:
    - ERROR: List accessor expects a list/vector, but received `Real`.
`heatEInWtrIM` exposes ill-typed expressions!
  - ERROR: Associative arithmetic operation expects all operands to be of the same expected type.
    Received:
    - Real
    - ERROR: Function application on non-function `tempW` (Real).
`wMass` OK!
`wVol` OK!
`tankVol` OK!
`tauW` OK!
=====[ Finished type checking ]=====
swhsnopcm: Type mismatch between variable and value in declare-define FuncStmt
  error, called at lib/Language/Drasil/Code/Imperative/Import.hs:650:28 in drasil-code-0.1.9.0-BZ5f3ZI9HUq1TjClKZN633:Language.Drasil.Code.Imperative.Import
```

From what I can understand from that linked line of code, this section of the code checks an expression that is a function definition to ensure that if the output type is a matrix then the input type must be either a list or an array. At some point vectors (now clifs) get converted into arrays for their concrete code representation. `Real`s, of course, do not get converted to arrays or lists. So this caused this error. So obviously this type _is_ factoring in, albeit in a pretty roundabout way. So I'll have to switch it back to `vectND` (a clif) for now to appease avoid this error for now as, like you said, this issue is beyond my scope.

CSchank:
> Strings here represent _variable names_. That makes sense. If 'dimensions' will be more than _variable names_ then no, that will not be adequate.

Currently we support variable-length (`String`) and fixed-length (`Natural`) dimensions. 

> If we need to support insertion and concatenation, then vector length will have to be an `Expr`.

Is this possible? That would require Drasil to support dependent types -- does it? I avoided using `Expr`s as the grade of a clif for this reason (used Natural instead). But it would be more flexible that way. You could, for example, "programmatically" choose a grade in a grade selection (which I just added) that way.

JacquesCarette:
The current type system does not yet support dependent types, because it had no reason to. I guess it will very soon. Let @balacij handle that part.

@balacij : our design will be to residualize constraints, we don't want to start proving anything.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
smiths: Thank you for the explanation @CSchank.  I agree with your solution.  In some sense, it is a good error.  The generated code will not be a single scalar for the output.  The types of the output in the code and the output in the SRS should be the same, or at least have a translation between them, like between a continuous function and a discrete sequence of output.  Since we don't have types fully and consistently used right now, we should move on from this instance and let it be sorted out in the future when types are handled more consistently and completely.

JacquesCarette:
> so I guess I'll just leave this as-is for now then (using the "meta" idea of Naturals instead of an embedded one)?

Yes, for now.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
CSchank: Sounds good, thank you!

CSchank:
Hi @JacquesCarette, @balacij, and @ptrandev! I have completed my latest milestone, which completes redefining vectors as Clifs. Code generation is not working yet, but the code compiles and type-checking within Drasil works. In fact, I was able to fix some type errors that were showing up! I'll try to summarize my changes and design decisions. I am very open to feedback on the code and the general design.

- **`Space.hs`**: The types (`Space`) for Clif is:
```haskell
ClifS    :: Dimension -> Space -> Space
```
As you can see, a Clifford space has a dimension (currently, variable or fixed to a `Natural`). An n-dimensional Clifford space has basis vectors `e0..en`, and to fully-define such a space in a given basis we need `2 ^ n` components (see https://www.av8n.com/physics/clifford-intro.htm#sec-dimensions). `vectNDS` allows you to specify the type of a vector that has `n` components. I redefined any physics concepts, etc that previously didn't have the correct types. We also have `vect2Ds` and `vect3DS` which are helper functions giving you the relevant `ClifS` definitions.
- **`Expr` data type**: As I learned more about Clifford algebra, I realized my previous definitions didn't make much sense. I was, broadly-speaking, defining homogeneous clifs when I realized I should have just been defining a general Clifford space. So the new constructor is:
```haskell
Clif     :: S.Dimension -> BasisBlades Expr -> Expr
```
Obviously, clifs need to be in a dimension, but then I realized we need to be able to store many expressions inside the clif; as I mentioned, up to `2 ^ n` of them. But, this doesn't take much advantage of sparsity, as was pointed out in the wonderful new paper @JacquesCarette sent me (Roelfs 2025, https://arxiv.org/pdf/2503.10451). For example, if I am just adding together 3D vectors or doing other operations on them, I only need 3 components, not 8! And you can imagine this would get very important for n-dimensional vectors. If I am modelling a slope using 250 slices, we certainly don't want to have to store 2^256 components! So, I used their approach and decided we should be storing only the components we need in a dictionary:
```haskell
type BasisBlades e =
  Map BasisKey e
```
and following Roefls, `BasisKeys` are basically fancy binary numbers:
```haskell
data BasisKey =
    Y BasisKey
  | N BasisKey
  | E
  deriving (Eq, Ord, Show)
```
which define the different components. There are some examples in the code.
- **Helper functions**: There is a helper function currently to make vectors of a given dimension (called `vect`, in code/drasil-lang/lib/Language/Drasil/Expr/Class.hs), and then I redefined the `vec2D` helper on top of that.
- Unfortunately, I didn't get to code generation for as I'd hoped. My new goal for the project is to get the projectile example working, and then hopefully examples with variable numbers of components. But I realized that my original goal of getting the SSP example working was too ambitious. I want to try to follow the Roelfs approach and try to generate only the minimum amount of code needed for each operation. This will be tricky, but an interesting problem. Technically, for Python, we could just use their library :)
- One general problem I'd like input on is the fact that since we represent for instance vectors as clifs, how do we "retain" that knowledge so even though we're representing a vector as a clif, it really "is" a vector and is restricted to those components only.
- I still don't know how we're going to handle linear transformations (matrices).
- Another thing I learned in Roelfs 2025 was about positive, negative, and null basis vectors, which was actually in Denker under the names spacelike, timelike, and null, but I previously didn't understand it. My implementation currently makes no judgments on whether the basis vectors are spacelike (they square to positive, e.g. e0⋅e0 = 1), timelike (square to negative, e.g. e0⋅e0 = -1) or null (square to 0) (see https://www.av8n.com/physics/clifford-intro.htm#sec-spacelike-timelike). Once I do generate code I will have to make a decision, but ideally we would support all combinations, like Roefls does. Timelike vectors are important to support things like special relativity. My first pass will likely assume all are spacelike as this is the most common space we're working in. In fact, my comment in the code stating that I assume spacelike is a bit wrong, because, as I said, until I implement things like geometric product, we currently make no such assumption. It will matter once those operations are defined though! I'll have to generalize the `Expr` type, and the `Space` type to do so, which shouldn't be too hard, but is not done yet.

Happy to hear any and all thoughts on this! I will add more notes here if I think of them; I've probably forgotten something, but I'm sure you'll make notes on it in the code!

JacquesCarette:
This is looking good!

> since we represent for instance vectors as clifs, how do we "retain" that knowledge so even though we're representing a vector as a clif, it really "is" a vector and is restricted to those components only.

I was hoping that the additional information in the clif would be enough to know that that is the case? Not to Haskell statically (as you're using a `Map`) but it should nevertheless be "true". Teaching Haskell more about clifs would lead into "light dependent types" which, while great, is beyond the scope of this.

> My first pass will likely assume all are spacelike

good, that's what I was going to recommend.

I still have to review the details of the code though. Hopefully soon.

code/drasil-lang/lib/Language/Drasil/ModelExpr/Lang.hs
```diff
@@ -171,6 +179,15 @@ data ModelExpr where
...
   ForAll   :: UID -> Space -> ModelExpr -> ModelExpr
 
+  -- | A clif of arbitrary dimension. The Maybe [Expr] determines the
+  --   components of the clif projected in a basis. If this is `Nothing`,
+  --   then the clif has not been projected into a particular basis. 
+  --   If this `isJust`, the number of components must be 2 ^ d where
+  --   d is the dimension of the clifford space.
+  -- All Clifs are currently assumed to be embedded in a space defined by spacelike 
+  -- basis vectors (e.g. Euclidean space) for now.
+  Clif     :: S.Dimension -> BasisBlades ModelExpr -> ModelExpr
```
ptrandev, line 189: I see that you have defined `Clif` in `ModelExpr` and `ClifS` in `Space`.

Is there a reason you have decided to call this a `ClifS` in `Space` instead of `Clif`? Perhaps avoiding a potential name collision is more important than having a concise naming scheme.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -31,25 +31,44 @@ import           Numeric.Natural           (Natural)
...
+vect2DS :: Space -> Space
+vect2DS s | checkClifSpace s = ClifS (Fixed 2) s
+
+vect3DS :: Space -> Space
+vect3DS s | checkClifSpace s = ClifS (Fixed 3) s
+
+vectS :: Natural -> Space -> Space
+vectS n s | checkClifSpace s = ClifS (Fixed n) s
+
+vectNDS :: String -> Space -> Space
```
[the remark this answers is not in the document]
CSchank: Yes, this defines an n-dimensional clif that stores some space (which is supposed to be a real number). The `n` variable represents the variable name. This could be `"n"` or `"m"` or whatever you want it to be. This is used for clifs where we don't know their dimension ahead of time. For instance, this could be a vector where we don't know how big it is when we specify it, but we want to still ensure we restrict the operations on that vector. For example, in the Slope Stability Problem we have a vector of size `n` representing the different slices. There are a bunch of vector operations (mostly additions). In our specification, we'd like to be able to check that only vectors of the same size (e.g. `n`) can be added together. But we don't know `n` until the runtime of the program.

code/drasil-lang/lib/Language/Drasil/ModelExpr/Lang.hs
CSchank: This is a good question, and one I posted above. I named it `ClifS` to avoid collisions, but I think it could be renamed as `Clif` just like the `Expr`, since we usually import `Space` qualified (`as S`) anyway, so right now you need to say `S.ClifS` in most place, which is of course redundant. But I wanted some input on this from @JacquesCarette and others, since I don't believe there is another example where we make such a "pun". It's an easy change if needed.

JacquesCarette: I think this particular pun is harmless, especially as types and constructors live in different name spaces.

code/drasil-code/lib/Language/Drasil/Code/Code.hs
JacquesCarette: We should have a page on our Wiki for that too - maybe @balacij can show @CSchank how to do a PR for that?

JacquesCarette (commented):
This looks like continued good progress.

Are there blockers? Is "what is a matrix" still a problem?

smiths:
@CSchank what is the status of this PR?  It would be great to merge at least a subset of your work into the Drasil repo.  We have a summer research assistant that might be able to continue this work, but we need to better define what has been done and what remains to be done.  I know you are busy with many other tasks, but we don't want to wait too long because branches have a tendency to go stale.

smiths:
@CSchank great to see this pull request.  It looks like the generated version does not match stable and that the linter failed.  Can you please look into these CI problems?  If you need help, we can possibly enlist the summer research assistants.

CSchank:
Hi @JacquesCarette, @balacij, @smiths, et al. I believe this PR is ready for review. The goal of this PR is to replace vectors with the geometric algebra (clif) constructor/space. It looks like there are some differences in some of the examples, but I don't believe I caused them? Please look them over and let me know.

This currently has a few limitations:
- Currently, clifs are rendered the same way as vectors were in the $\LaTeX$. But since clifs are implemented as maps, currently I'm printing out all the components that are non-zero using `Map.toList`. This is kinda iffy, but works for when you're only using vectors. We need a more generalized solution; i.e. one where you can select whether to print out the full clif with its zero components.
- Some things like the `strList2DAsExpr` I'm not sure how to handle yet. Clifs don't nest in this way. I think we need to figure out matrices before this can make sense.
- Currently, generalized dimensions are handled as simple string variable names. We probably eventually want to change this to handle addition, subtraction, etc.
- This PR doesn't have code generation. That's [here](https://github.com/CSchank/Drasil/pull/1) instead.

smiths:
@CSchank and @balacij, did you get a chance to meet on Thursday, like you discussed?  Has a plan been formulated for getting this PR mergeable?

balacij:
Unfortunately we hadn't met yesterday. We're going to meet this afternoon to discuss it.

smiths:
Sounds good @balacij.

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -443,15 +443,15 @@ odeint = externalLib [
       customObjArg [] "Class representing an ODE system" ode odeCtor
         (customClass [constructorInfo odeCtor [] [],
           methodInfoNoReturn odeOp "function representation of ODE system"
-            [unnamedParam (Vect Real), unnamedParam (Vect Real), lockedParam t]
+            [unnamedParam (vectNDS "n" Real), unnamedParam (vectNDS "n" Real), lockedParam t] -- TODO: what do we put here? We need specific dimensions now.
```
balacij, line 446: @CSchank Do you think you could give an example please? Of the change you'd like to see, that is.

code/drasil-lang/lib/Language/Drasil/Expr/Class.hs
```diff
@@ -388,6 +408,23 @@ instance ExprC Expr where
...
+    in
+      Clif (S.Fixed d) $ Map.fromList $ mapWithIndex vectComp es
+
+mapWithIndex :: (Natural -> a -> b) -> [a] -> [b]
+mapWithIndex f xs = 
+  mapWithIndex' 0 xs
+  where
+    mapWithIndex' _ [] = []
+    mapWithIndex' n (x:xs) =
+      f n x : mapWithIndex' (n+1) xs
```
balacij, line 427: Needs to be moved to `drasil-utils`.

balacij (changes requested):
GHC reports non-exhaustive patterns on the rendering functions. We should add errors to those. The CI passes however, which is odd.

I'll work on this PR. I'll have a look at the changes to stable/ and the typing changes soon.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -14,37 +14,70 @@ module Language.Drasil.Space (
...
+  -- | Vect Space -- TODO: Length for vectors?
+    Set      :: Space -> Space
+    Matrix   :: Int -> Int -> Space -> Space
+    Array    :: Space -> Space
+    Actor    :: String -> Space
+    Function :: (NE.NonEmpty Primitive) -> Primitive -> Space
+    Void     :: Space
+  -- | Clifford algebra objects (Clifs) with a dimension
+  -- TODO: Can this be just called `Clif`, depsite shadowing the binding in `*Expr`, since we `import qualified Space as S` usually?
+    ClifS    :: Dimension -> Space -> Space
```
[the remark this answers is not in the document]
JacquesCarette: 1. I don't have a strong opinion on the name. It might be premature to change it until we fully settle on what we're representing.
2. If my understanding is correct, we need a "dimension" and a way to represent what is there at each dimension. Yes, it is the number of components of the formal sum.
3. No, I don't think so. And, even if we did, `BigWedge` would have to contain more data than that. `BigWedge` is a type-former over vector spaces. We don't want to manipulate these 'raw', instead this should be baked into the type for `Clifs` (or whatever it will be called)
4. `Clifs` at a particular dimension (one), with a particular signature (no component of grade 0) *are* vectors. So we shouldn't need vectors are they are a special case. We do need the signature part of `Clifs` to be known data to be able to recognize vectors 'statically'.
5. Yes, type-checking will need to learn about all of this. That's why the signature part will have to be static, else type-checking will be impossible.

code/drasil-printers/lib/Language/Drasil/Printing/Import/ModelExpr.hs
```diff
@@ -192,7 +192,8 @@ modelExpr (ForAll c s de)            sm = P.Row [
   ]
 -- TODO: Fix this to be more specific to Clifs
 -- TODO: How do we control whether to print all the components or just a subset (e.g. only the vector components)?
-modelExpr (Clif _ es)                sm = P.Mtx [map (`modelExpr` sm) $ Map.elems es]
+modelExpr (Clif _ es)                sm = P.Mtx $ map ((:[]) . (`modelExpr` sm)) $ Map.elems es
```
balacij: It looks like the issue was a misunderstanding in how `Mtx` is supposed to be used. Rows vs columns. A single inner list is a single row in the matrix. However, we seem to want column vectors appearing, so we need multiple rows, only 1 column (i.e., many one-element inner lists).

I think this approach only works for existing case studies, however (in general, i.e., using a matrix here). If one of the basis blades were $\mathbb{1}$ or $e_1e_2e_3$, this would still render as a matrix with 1 column, when it should be a sum of scalar with a bivector.

JacquesCarette: This fix is an improvement, thanks for tracking down this bug.

code/drasil-utils/lib/Data/Map/Ordered.hs
```diff
@@ -0,0 +1,39 @@
...
+import Prelude hiding (lookup, null)
+
+import qualified Data.Map as M
+import Data.List (sortOn)
+import Data.Foldable (Foldable(foldl'))
+
+-- | Simple ordered map. Based on 'container's 'Data.Map'. Note: Deletions are
+-- unsupported.
+newtype OrderedMap k v = OM { im :: M.Map k (Int, v) }
+  deriving (Show, Eq)
```
balacij, line 14: I just wrote this instead of using an external library. I think we have something similar in our `ChunkDB` -- though I don't remember why we need to recall order of insertion of things into the `ChunkDB`, but if this is merged, we can use this one as well.

This was necessary for the rendering of basis blades to be done in order of insertion.

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
sarrasoussia: Shouldn't we specify the dimension in each example to match its requirements? So here it must remain abstract?

JacquesCarette: Yes!!!

JacquesCarette:
We will implement a subset of Geometric Algebra, i.e. Exterior Algebra. And we'll try to do it more incrementally. Though, in the end, it will look like this a fair bit.

PR #4013, opened 2025-02-27
Remove unused SystemInformation field.

balacij:
In our latest meeting, @szymczdm, @JacquesCarette and @smiths mentioned you found one unused field in the `SystemInformation`. I think this is the one.


JacquesCarette (approved):
I'm guessing that this was partly-done work (because of `Block`) that was going to be useful... but I can't at all remember it. So better to remove it, we can reconstruct later when it comes up again.

PR #4016, opened 2025-02-28
`drasil-lang`: clean up `UID`-related stuff and move it away from the other modules

balacij:
Contributes to #2885 

Once #2873 is merged, `UID` will move there, so I'm at least moving the module to a different namespace now, locally in `drasil-lang`.


code/drasil-lang/lib/Drasil/Database/UID.hs
```diff
@@ -23,51 +23,42 @@ class HasUID c where
...
 instance ToJSON UID where
-  toJSON = toJSON . fullName
+  toJSON = toJSON . show
 
 instance ToJSONKey UID where
   toJSONKey = toJSONKeyText (pack . show)
 
 instance Show UID where
-  show = intercalate ":" . fullName
+  show u = intercalate ":" $ u ^. baseName : u ^. namespace
```
balacij, line 41: Note that I inlined `fullName` here, but I also moved the "basename" to the front.

code/drasil-lang/lib/Drasil/Database/UID.hs
```diff
@@ -23,51 +23,42 @@ class HasUID c where
...
 mkUid :: String -> UID
 mkUid s = UID { _namespace = [], _baseName = s }
-  -- '►' `elem` s = error $ "► not allowed in UID " ++ show s -- FIXME: Need to implement other constructors before we can use this.
-  -- null s       = error "UID must be non-zero length" -- FIXME: See Drasil.DocumentLanguage.TraceabilityGraph (uses an empty UID)
-  -- otherwise    = UID s
 
--- | Nest UID under a namespace
+-- | Nest a 'UID' under a namespace.
 nsUid :: String -> UID -> UID
 nsUid ns = over namespace (ns:)
```
balacij, line 49: This nesting structure is the reason why I moved the basename to the front. Consider the case where we nest something $n$ times. If it starts in `A`, and then we nest it in `B` within `A`, the list is reversed, so appending the final UID to the end is confusing. e.g., `[D,C,B,A]` means we first nested in `A` (the "largest" thing), and then nested the concept within `B`, etc. So, if a chunk has title $T$, we want to see $T:D:C:B:A$.

This might not have been strictly necessary, in part because I think we're going to end up redoing `UID`s entirely, but I _think_ this implementation makes more sense.

JacquesCarette:
I don't quite understand: #2873 is not a PR but an issue? Is there a PR that this is waiting on?

code/drasil-lang/lib/Drasil/Database/UID.hs
JacquesCarette: Agree with this choice.

code/stable/dblpend/TraceyGraph/allvsall.dot
```diff
@@ -1,17 +1,17 @@
 digraph allvsall {
-	theory:v_x1 -> velocity;
-	theory:v_x1 -> p_x1;
-	theory:v_y1 -> velocity;
-	theory:v_y1 -> p_y1;
-	theory:v_x2 -> velocity;
-	theory:v_x2 -> p_x2;
-	theory:v_y2 -> velocity;
-	theory:v_y2 -> p_y2;
+	v_x1:theory -> velocity;
```
JacquesCarette, line 2: in this case, `theory:v_x1` and `v_x1:theory` both make little sense... curious where it comes from. The reasoning above is sound, so it's super weird that the `theory` now comes at the end!

JacquesCarette (approved):

balacij:
Sorry, I mean "once the prototype code written for #2873 is merged, `UID` will live in `drasil-database` much closer to the upgraded `ChunkDB`".

code/stable/dblpend/TraceyGraph/allvsall.dot
balacij: That UID is for:
https://github.com/JacquesCarette/Drasil/blob/1330bc431605eb31bb8d7d9bca90da02cb5b9696/code/drasil-example/dblpend/lib/Drasil/DblPend/GenDefs.hs#L33-L34

Relevant 'backend' code:
https://github.com/JacquesCarette/Drasil/blob/1330bc431605eb31bb8d7d9bca90da02cb5b9696/code/drasil-theory/lib/Theory/Drasil/ModelKinds.hs#L89-L91

https://github.com/JacquesCarette/Drasil/blob/1330bc431605eb31bb8d7d9bca90da02cb5b9696/code/drasil-theory/lib/Theory/Drasil/ModelKinds.hs#L58-L59

balacij: i.e., `theory` is a namespace

JacquesCarette: So you want namespaces to appear on the right?

balacij: If we have `nest C $ nest B $ nest A X`, the UID before this PR ends up as: `C:B:A:X`. I read the relationships between `C`, `B`, and `A` as $C \subset B \subset A$. The code would then be read as "nest $X$ in $A$, then create a new namespace for $X$ within $A$, etc.".  In which case, the update would make more sense because it shows $X$ near the smallest namespace, `C` (e.g., `X:C:B:A`).

Thinking about it again... maybe I was confused? The documentation is unclear about the relationship between the nested namespaces. I'm not sure, and I'm impartial because I'm fairly convinced we're going to have to completely rework the `UID` system once we get to the real work on editing the chunk graph, theory refinements, etc.

I'll undo it for now.

balacij: Or did you mean that you agreed with my view and just wanted the larger namespaces to appear leftmost rather than rightmost? e.g., `A:B:C:X`.

JacquesCarette: I want our name-spacing conventions to be the same as Haskell's and Agda's, so `Data.List.Membership` for example. I don't see this as having anything to do with subsets, but rather is all about hierarchical classification.

balacij: Yeah, that does make more sense. I just updated that now in 361b507d0d72b7749ad99bd22b11d461e530ddfd and e2f687cda3ae5c9e1ebe01b26308a25ed7d36117 .

JacquesCarette (approved):

PR #4022, opened 2025-03-15
Demote references table to 'not a chunk table' in `ChunkDB`

balacij:
Contributes to #2873 

This is a bit of a silly PR, and it might not even be the correct way to go about things. The title sounds big and mighty, but it's really not doing much.

This PR:
* Reorders the tables in the `ChunkDB`, sorting by `Chunk` vs `Not a Chunk`.
* Removes automatic extraction of `Reference`s from `Sentence`s rendered in the SRS.
* Removes `Reference`s from being considered as part of "knowledge maps" we dump in `Dump.hs`, hence excluding `Reference`s when calculating conflicting `UID`s.

The goal of this PR is to remove as many UID conflicts of `Reference`s with other chunks as possible. This PR does that by silently ignoring them, kicking the rock of actually fixing this down the road.

The demotion of `Reference` to "not a chunk" is more of a semantic change. As I mentioned in https://github.com/JacquesCarette/Drasil/issues/2873#issuecomment-2693129320, `Reference`s are somewhat confused data types. Up to naming, they confuse internally generated intermediate references to displayed knowledge with purely external (actual) references. By ...., I mean ....:
* "internally generated intermediate references to displayed knowledge" (IGIRDP... 🫤 [because there's no :sigh: emoji, a double :sigh:]), I mean reference handles to chunks we render at some point in artifacts, which we can direct a readers' attention to somehow [e.g., clickable link or just plain text], such as HTML IDs, LaTeX labels, ~~URIs,~~ (URIs are NOT an example of an IGIRDP) and [not that we do this one, but...] import statements in code.
* "purely external (actual) references" (PEAR), I mean actual references. Things that we have no real internal representation of. It seems like the existing examples of (actual) references include: local file handles (e.g., images we embed) and URLs (e.g., clickable links and "sources"). These are things we have very limited control over. I think a better way of thinking of these is: "Citations that we try to pull something from."

The terminology surrounding "Reference" in Drasil is quite overloaded, and I might have gotten confused along the way, but this is the design that I see as the "right one going forward:"
* Move the IGIRDPs off into the printers (continuing to think of them as "not a chunk"). This would involve moving it to `drasil-printers` most likely, an attempt that I've already tried but failed because our printers are not ready to handle this situation in a reasonable fashion (e656c6f3b6d20bce13e97763f60402ec570efd0e -- see https://github.com/JacquesCarette/Drasil/commit/e656c6f3b6d20bce13e97763f60402ec570efd0e#r153777701 for why it fails). We need a stateful renderer that can keep track of all the chunks we render. One issue with our existing approach to creating the IGIRDPs is that we just assume that all chunks are going to be rendered (?!) so create a bunch of invalid IGIRDPs.
* Figure out an actual design for the PEARs. This might involve dealing with creating `Citation`s instead of `Reference`s and creating some sort of `CiteExternalContent` chunk that references both a citation and provides some content we can use in the renderer (that comes from the citation directly).

Because I believe this is the 'right thing to do going forward', I removed the automatic extraction of `Reference`s from `Sentence`s rendered in the SRS (via theories primarily...) so that we know which things to convert into actual chunks later on.

### DIFF
--- a/code/drasil-database/lib/Database/Drasil/Dump.hs
+++ b/code/drasil-database/lib/Database/Drasil/Dump.hs
@@ -1,7 +1,7 @@
 module Database.Drasil.Dump where
 
 import Language.Drasil (UID, HasUID(..))
-import Database.Drasil.ChunkDB (refTable, labelledcontentTable, 
+import Database.Drasil.ChunkDB (labelledcontentTable, 
   conceptinsTable, theoryModelTable, gendefTable, insmodelTable, dataDefnTable,
   unitTable, UMap, ChunkDB(termTable, symbolTable))
 
@@ -28,5 +28,4 @@ dumpChunkDB cdb =
     $ insert "theoryModels" (umapDump $ cdb ^. theoryModelTable)
     $ insert "conceptInstances" (umapDump $ cdb ^. conceptinsTable)
     $ insert "labelledContent" (umapDump $ cdb ^. labelledcontentTable)
-    $ insert "references" (umapDump $ cdb ^. refTable)
       mempty
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -160,8 +160,6 @@ fillReferences dd si@SI{_sys = sys} = si2
     -- get refs from SRSDecl. Should include all section labels and labelled content.
     refsFromSRS = concatMap findAllRefs allSections
     -- get refs from the stuff already inside the chunk database
-    inRefs  = concatMap dRefToRef ddefs ++ concatMap dRefToRef gdefs ++
-                concatMap dRefToRef imods ++ concatMap dRefToRef tmods
     ddefs   = map (fst.snd) $ Map.assocs $ chkdb ^. dataDefnTable
     gdefs   = map (fst.snd) $ Map.assocs $ chkdb ^. gendefTable
     imods   = map (fst.snd) $ Map.assocs $ chkdb ^. insmodelTable
@@ -173,7 +171,7 @@ fillReferences dd si@SI{_sys = sys} = si2
     -- search the old reference table just in case the user wants to manually add in some references
     refs    = map (fst.snd) $ Map.assocs $ chkdb ^. refTable
     -- set new reference table in the chunk database
-    chkdb2 = set refTable (idMap $ nub $ refsFromSRS ++ inRefs
+    chkdb2 = set refTable (idMap $ nub $ refsFromSRS
       ++ map (ref.makeTabRef'.getTraceConfigUID) (traceMatStandard si) ++ secRefs -- secRefs can be removed once #946 is complete
       ++ traceyGraphGetRefs (programName sys) ++ map ref cites
       ++ map ref conins ++ map ref ddefs ++ map ref gdefs ++ map ref imods
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
@@ -38,7 +38,7 @@ import Drasil.GamePhysics.Concepts (gamePhysics, acronyms, threeD, twoD)
 import Drasil.GamePhysics.DataDefs (dataDefs)
 import Drasil.GamePhysics.Goals (goals)
 import Drasil.GamePhysics.IMods (iMods, instModIntro)
-import Drasil.GamePhysics.References (citations)
+import Drasil.GamePhysics.References (citations, uriReferences)
 import Drasil.GamePhysics.Requirements (funcReqs, nonfuncReqs, pymunk)
 import Drasil.GamePhysics.TMods (tMods)
 import Drasil.GamePhysics.Unitals (symbolsAll, outputConstraints,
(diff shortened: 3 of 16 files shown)


code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -155,7 +155,7 @@ symbMap = cdb (map (^. output) iMods ++ map qw symbolsAll) (nw gamePhysics :
 
   -- | Holds all references and links used in the document.
 allRefs :: [Reference]
-allRefs = [externalLinkRef, pymunk] ++ offShelfSolRefs
+allRefs = [externalLinkRef, pymunk] ++ uriReferences ++ offShelfSolRefs
```
balacij, line 158: These are "actual" references.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -160,8 +160,6 @@ fillReferences dd si@SI{_sys = sys} = si2
     -- get refs from SRSDecl. Should include all section labels and labelled content.
     refsFromSRS = concatMap findAllRefs allSections
     -- get refs from the stuff already inside the chunk database
-    inRefs  = concatMap dRefToRef ddefs ++ concatMap dRefToRef gdefs ++
-                concatMap dRefToRef imods ++ concatMap dRefToRef tmods
```
balacij, line 164: This is where I remove the automatic gathering of the actual references.

code/drasil-database/lib/Database/Drasil/Dump.hs
```diff
@@ -28,5 +28,4 @@ dumpChunkDB cdb =
     $ insert "theoryModels" (umapDump $ cdb ^. theoryModelTable)
     $ insert "conceptInstances" (umapDump $ cdb ^. conceptinsTable)
     $ insert "labelledContent" (umapDump $ cdb ^. labelledcontentTable)
-    $ insert "references" (umapDump $ cdb ^. refTable)
```
balacij, line 31: This is the effective "demotion" of references to "not a chunk" that removes it from consideration in the problematic chunks.json file generation.

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/References.hs
```diff
@@ -43,3 +43,15 @@ sciComp2013 = cArticle
...
+uriReferences = [accelGravitySrc, impulseSrc]
+
+accelGravitySrc :: Reference
+accelGravitySrc = makeURI "accelGravitySrc" "https://en.wikipedia.org/wiki/Gravitational_acceleration" $
+  shortname' $ S "Definition" `S.of_` S "Gravitational Acceleration"
+
+
+impulseSrc :: Reference
+impulseSrc = makeURI "impulseSrc" "http://www.chrishecker.com/images/e/e7/Gdmphys3.pdf" $
+  shortname' $ S "Impulse for Collision Ref"
```
balacij, line 57: I moved the PEARs to the "References" file because I think that's most appropriate for where they belong. The PEARs should also contain authorship information, which is why they're next to the citations.

code/drasil-example/swhs/lib/Drasil/SWHS/References.hs
```diff
@@ -1,6 +1,4 @@
-module Drasil.SWHS.References (citations, bueche1986, incroperaEtAl2007,
-  koothoor2013, lightstone2012, parnasClements1986, smithEtAl2007,
-  smithLai2005, smithKoothoor2016) where
+module Drasil.SWHS.References where
```
JacquesCarette, line 1: why remove the explicit list of exports?

code/drasil-example/swhs/lib/Drasil/SWHS/TMods.hs
```diff
@@ -27,7 +27,7 @@ import Data.Drasil.Quantities.Thermodynamics (boilPt, heatCapSpec,
 import Drasil.SWHS.Assumptions (assumpHTCC, assumpTEO)
 import Drasil.SWHS.Concepts (transient)
 import Drasil.SWHS.DataDefs (ddMeltFrac)
-import Drasil.SWHS.References (incroperaEtAl2007)
+import Drasil.SWHS.References
```
JacquesCarette, line 30: This is moving in the wrong direction. Explicit imports are much better than imports that may, randomly in the future, cause conflicts in the namespace.

JacquesCarette (changes requested):
I think the general idea is good (there are different kinds of knowledge we encode, and not all should be stored in the same database(s) in the same way.)

But: regression on good Haskell style.

code/drasil-example/swhs/lib/Drasil/SWHS/References.hs
balacij: In all honesty, that was just laziness w.r.t. sorting them out. Some of these exports were re-exports of things from `drasil-data`. I can add them back.

Regarding codestyle, unless there's a reason to manually write the export lists, isn't "export all by default" okay for modules like this one, where everything is exported anyways (and where we expect all future content to be too)?

JacquesCarette: If you leave a comment saying "exports everything, on purpose", then that's fine.

JacquesCarette (approved):

PR #4024, opened 2025-03-17
Ensure Theory UIDs are always in the `theory:` namespace

balacij:
Contributes to #2873 

I don't necessarily agree with this change, nor the whole "namespace" system and how its used right now, but this at least ensures that theory UIDs are consistently placed with a "Theory" prefix.


JacquesCarette (approved):

JacquesCarette:
It's easy enough to revisit this later. At least now things are consistent.

PR #4025, opened 2025-03-17
Add new database dump file: things that are _just_ terms.

balacij:
Contributes to #2873 

This PR merely lets us see what chunks added to the ChunkDB are "just" terms.

My suspicion is that a "TermTable" shouldn't exist and that we should instead think of it as a "term cache" until the upgraded chunkdb prototype is merged and we start working on replacing `TypeRep` with something else.

_Why?_ A TermTable contains `IdeaDict`s:
https://github.com/JacquesCarette/Drasil/blob/cd996e49b222deaf25bac195b16d34fd611138b2/code/drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs#L46-L56

Realistically, this chunk is just shy of empty. It contains a UID, a noun phrase, and maybe a short name. In some sense, this is great because, long-term, I think we want something like this (not necessarily a chunk) before we create actual meaningful chunks that prescribe meaning to "noun phrases" for specific domains. Think `ConceptChunk` and `ConceptInstance` here, except that `IdeaDict` would sit atop `ConceptChunk` too. So, a `ConceptChunk` would be an instance of a noun phrase for a specific domain with an abstract definition, and a `ConceptInstance` would be a specific instance of a noun phrase for a specific with a concrete definition.

So what makes the term map a "term cache"? Well, that's because we use it like one. The vast majority of our usage of it is as follows:
https://github.com/JacquesCarette/Drasil/blob/cd996e49b222deaf25bac195b16d34fd611138b2/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/Body.hs#L205-L213
We use `nw` to extract `IdeaDict`s from the actually meaning-dense chunks. All use of this map is restricted to "getting a nounphrase or maybe a shortened abbreviation." So, in some sense this is just a convenient map for caching terms so that we didn't need to know which specific map to look into previously when rendering `Sentences` and needing a term's noun phrase.

I would like to see all of this `map nw` and manual `nw` usage disappear long-term. It should either be automated or not happen at all. In a follow-up message, I will write some more thoughts on how we can proceed with "de-duplicating UIDs in the chunkdb."

### DIFF
--- a/code/drasil-gen/lib/Language/Drasil/Dump.hs
+++ b/code/drasil-gen/lib/Language/Drasil/Dump.hs
@@ -12,7 +12,7 @@ import qualified Data.ByteString.Lazy.Char8 as LB
 import qualified Data.Map.Strict as SM
 
 import Utils.Drasil (invert, atLeast2)
-import Database.Drasil (traceTable, refbyTable)
+import Database.Drasil (traceTable, refbyTable, ChunkDB (termTable))
 import Control.Lens ((^.))
 import System.Environment (lookupEnv)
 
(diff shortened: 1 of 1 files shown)


JacquesCarette:
See discussion #4027 for much of my thoughts on the underlying commentary here.

JacquesCarette (approved):

JacquesCarette:
Even though I think the commentary in this PR is *wrong*, the PR itself is a valuable addition to our toolkit.

PR #4023, opened 2025-03-17
Demote 'LabelledContent' to 'not a chunk' until they're dealt with properly

balacij:
Contributes to #2873

This is similar to #4022. I will annotate the PR.

Here is an excerpt from `DblPend`'s output debug information, and unfortunately, this information is largely unused as a 'chunk'.

```
LabelledContent:
----------------------------------------------------------------------------------------------------
UID                           Short Name                    Content Type
----------------------------------------------------------------------------------------------------
aR                            aspectRatio                   Definition or Model
demandq                       calofDemand                   Definition or Model
eqTNTChar                     eqTNTW                        Definition or Model
glassTypeFac                  gTF                           Definition or Model
loadDurFactor                 loadDurFactor                 Definition or Model
minThick                      minThick                      Definition or Model
stdOffDist                    standOffDist                  Definition or Model
doc:InDataConstraints         InDataConstraints             Table
doc:OutDataConstraints        OutDataConstraints            Table
doc:ReqAssignments            ReqAssignments                Table
doc:ReqInputs                 ReqInputs                     Table
doc:ReqOutputs                ReqOutputs                    Table
doc:TAbbAcc                   TAbbAcc                       Table
doc:TAuxConsts                TAuxConsts                    Table
doc:ToS                       ToS                           Table
doc:ToU                       ToU                           Table
doc:TraceGraphAllvsAll        TraceGraphAllvsAll            Figure
doc:TraceGraphAllvsR          TraceGraphAllvsR              Figure
doc:TraceGraphAvsA            TraceGraphAvsA                Figure
doc:TraceGraphAvsAll          TraceGraphAvsAll              Figure
doc:TraceGraphRefvsRef        TraceGraphRefvsRef            Figure
doc:TraceMatAllvsR            TraceMatAllvsR                Table
doc:TraceMatAvsA              TraceMatAvsA                  Table
doc:TraceMatAvsAll            TraceMatAvsAll                Table
doc:TraceMatRefvsRef          TraceMatRefvsRef              Table
doc:demandVSsod               demandVSsod                   Figure
doc:dimlessloadVSaspect       dimlessloadVSaspect           Figure
doc:physSystImage             physSystImage                 Figure
doc:sysCtxDiag                sysCtxDiag                    Figure
theory:dimlessLoad            dimlessLoad                   Definition or Model
theory:isSafeLR               isSafeLR                      Definition or Model
theory:isSafeLoad             isSafeLoad                    Definition or Model
[...]

PR #4026, opened 2025-03-17
Change the UID of DataDefinitions and add their "terms" to the term cache

balacij:
Contributes to #2873 

Builds on #4023, #4024, #4025 

Hopefully we'll get to properly explaining this in our meeting in 13min, otherwise I will update this properly later.


code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
```diff
@@ -108,23 +109,26 @@ instance RequiresChecking DataDefinition Expr Space where
 
 -- * Constructors
 
+ddUid :: HasUID c => c -> UID
+ddUid = nsUid "dataDefn" . (^. uid)
```
balacij, line 113: 1. Here we are moving DataDefinitions to the "dataDefn" namespace.

I don't think this is the right naming scheme ultimately. I want to see `UID`s rethought later, but for now, this at least lets us remove the UID duplication between `dataDefn`s and everything else but the term cache.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -213,6 +213,8 @@ cdb s t c u d ins gd tm ci lc r =
     _refbyTable = Map.empty,
     _refTable = idMap r
   }
+  where
+    termsHACK = map nw d
```
balacij, line 217: 2. When we change the `UID`s of the existing data definitions, we see runtime errors asking for missing `UID`s. That's caused by missing `IdeaDict`s. Here, we are creating them.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -191,15 +191,15 @@ makeLenses ''ChunkDB
...
+cdb :: (Quantity q, MayHaveUnit q, Concept c, IsUnit u) =>
+    [q] -> [IdeaDict] -> [c] -> [u] -> [DataDefinition] -> [InstanceModel] ->
     [GenDefn] -> [TheoryModel] -> [ConceptInstance] ->
     [LabelledContent] -> [Reference] -> ChunkDB
 cdb s t c u d ins gd tm ci lc r = 
   CDB {
     -- CHUNKS
     symbolTable = symbolMap s,
-    termTable = termMap t,
+    termTable = termMap $ t ++ termsHACK,
```
balacij, line 202: 3. Here we are adding the terms of the data definitions to the term cache.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -191,15 +191,15 @@ makeLenses ''ChunkDB
 --     * 'TheoryModel's (for 'TheoryModelMap'),
 --     * 'ConceptInstance's (for 'ConceptInstanceMap'),
 --     * 'LabelledContent's (for 'LabelledContentMap').
-cdb :: (Quantity q, MayHaveUnit q, Idea t, Concept c, IsUnit u) =>
-    [q] -> [t] -> [c] -> [u] -> [DataDefinition] -> [InstanceModel] ->
+cdb :: (Quantity q, MayHaveUnit q, Concept c, IsUnit u) =>
+    [q] -> [IdeaDict] -> [c] -> [u] -> [DataDefinition] -> [InstanceModel] ->
```
balacij, line 195: 4. Constraining to the `Idea t => t` was kind of unnecessary because it was exclusively filled out with `IdeaDict`s, so here we are just removing it so that we know which monomorphic type to cache the `DataDefinition`'s terms as.

code/stable/dblpend/TraceyGraph/allvsr.dot
```diff
@@ -18,17 +18,17 @@ digraph allvsr {
...
 	}
 
-	velocity	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionGDD"];
-	p_x1	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionXDD1"];
-	p_y1	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionYDD1"];
-	p_x2	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionXDD2"];
-	p_y2	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionYDD2"];
-	acceleration	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:accelerationGDD"];
-	force	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:forceGDD"];
+	dataDefn:velocity	[shape=box, color=black, style=filled, fillcolor=paleturquoise1, label="DD:positionGDD"];
```
balacij, line 21: This and everything below is the result of the change.

code/drasil-theory/lib/Theory/Drasil/DataDefinition.hs
JacquesCarette: agree that this is likely temporary, but useful nevertheless.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
JacquesCarette: If I was sure that we were going to keep `cdb` as is, I would predict that you'd come to regret this change. I was actually going to suggest that we move some other routines in the exact opposite direction!

Philosophically what you're doing is moving to a specific representation (`IdeaDict`) where before this worked based on an interface (`Idea t`). From an SE perspective, working to interfaces is better (except for efficiency, but not even here) than to specific implementations.

JacquesCarette:
@balacij can you merge master in here and resolve the conflicts? That should make the actual changes done here simpler for me to review.

balacij:
I'm kind of surprised github reported any merge errors... I didn't see any locally. I intentionally used a merge locally on those branches to specifically avoid a conflict. I fixed it right now by:
1. `git fetch`
2. `git checkout main` && `git pull`
3. `git checkout whatsNext` && `git merge main`
4. No conflicts... so... `git push`

JacquesCarette (approved):

code/stable/gamephysics/SRS/PDF/GamePhysics_SRS.tex
```diff
@@ -913,7 +913,7 @@ \subsubsection{Data Definitions}
            \end{displaymath}
 \\ \midrule
 Description & \begin{symbDescription}
-              \item{${\symbf{v}\text{(}t\text{)}_{\text{B}}}$ is the velocity at point B ($\frac{\text{m}}{\text{s}}$)}
+              \item{${\symbf{v}\text{(}t\text{)}_{\text{B}}}$ is the Chasles' theorem ($\frac{\text{m}}{\text{s}}$)}
```
JacquesCarette, line 916: as noted elsewhere, this is wrong.

JacquesCarette (changes requested):

code/stable/gamephysics/SRS/PDF/GamePhysics_SRS.tex
balacij: It looks like we've run into #2788... I'm surprised this issue only appears here because `mkQuantDef'` is used in nearly all of our examples. So why does this issue only appear for `mkQuantDef'`?! It looks like an issue with nesting and inheritance of fields w.r.t. QuantityDicts, QDefinitions, and DataDefinitions.

balacij: Ok, I believe I fixed the issue. That being said, the type classes around `QDefinition`s+related chunks are all pretty odd. We definitely need to revisit them. I guess our conversation earlier today was right on time!

code/drasil-docLang/lib/Drasil/DocumentLanguage/Definitions.hs
```diff
@@ -203,7 +203,7 @@ mkIMField _ _ l _ = error $ "Label " ++ show l ++ " not supported " ++
 firstPair' :: InclUnits -> DataDefinition -> ListTuple
 firstPair' IgnoreUnits d  = (P $ eqSymb $ d ^. defLhs, Flat $ phrase d, Nothing)
 firstPair' IncludeUnits d =
-  (P $ eqSymb $ d ^. defLhs, Flat $ phrase d +:+ sParen (toSentenceUnitless $ d ^. defLhs), Nothing)
+  (P $ eqSymb $ d ^. defLhs, Flat $ phrase (d ^. defLhs) +:+ sParen (toSentenceUnitless $ d ^. defLhs), Nothing)
```
JacquesCarette, line 206: This fixes the *symptom* but not the bug! It's nice that the fix is fairly generic though. So I'll accept this as a proper fix for the purposes of this PR, but it is not really a fix in that the 'phrase' in the given Chunk should never have been "Chasles' theorem", which points to bad re-use.

JacquesCarette (approved):

code/stable/gamephysics/SRS/PDF/GamePhysics_SRS.tex
JacquesCarette: Can you point to the actual use of `mkQuantDef'` that is at fault here?

PR #4033, opened 2025-03-31
Remove unused _concepts field in SystemInformation

balacij:
It's not used, and it's not clear how it should be used.


JacquesCarette (approved):
When we figure out a use for this information, we can reconstruct it. In the meantime, I agree, best remove it.

PR #4034, opened 2025-04-01
Fix symbol sub- and sup-script rendering

### LINKED ISSUE #3188: Fix subscript and superscript display for variables that have both
When we have a variable with a both subscript and a superscript it displays in awkward way.  The [symbols in Projectile](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#Sec:ToS) are an example.  For instance, we have `a_y^c` (y-component of constant acceleration).  In both LaTeX and html the display is a_y with the superscript c.  Instead of a with the subscript y and the superscript c.  In pdf generated by LaTeX the difference looks like this:

image

The Drasil-generated LaTeX code is:

`${{a_{\text{y}}}^{\text{c}}}$`

what we want is:

`${a_{\text{y}}^{\text{c}}}$`

Hopefully, this is an easy fix, but I wonder if it was done the "wrong" way because the "right" way isn't intuitive?

We can function with the examples as they are, so I'll set this as an enhancement.

balacij:
Closes #3188 

Contributes to #1606 -- Maybe closes?


balacij:
I think this works as-is, but the diff is too huge. I'll need to audit it later.

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -460,12 +460,12 @@ <h2>Table of Symbols</h2>
                   <td><em>\(\frac{\text{m}}{\text{s}}\)</em></td>
                 </tr>
                 <tr>
-                  <td><em><b>v</b>(t)<sub>f</sub><sup>AB</sup></em></td>
+                  <td><em><b>v</b>(t)<sup>AB</sup><sub>f</sub></em></td>
```
JacquesCarette, line 463: one of the reasons for the huge change is this: sub and sup have changed order. That seems weird - why would they? I would track that down to stabilize this.

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -904,7 +904,7 @@ <h3>Theoretical Models</h3>
                     </tr>
                     <tr>
                       <th>Equation</th>
-                      <td>\[{\symbf{F}_{1}}=-{\symbf{F}_{2}}\]</td>
+                      <td>\[\symbf{F}_{1}=-\symbf{F}_{2}\]</td>
```
JacquesCarette, line 907: Here the change seems like an improvement, but I also don't know why it happened! Why were there extra {} before? Sometimes they were needed to make some things behave better. But I have no idea why your changes would have cause this too.

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -952,7 +952,7 @@ <h3>Theoretical Models</h3>
                     <tr>
                       <th>Equation</th>
                       <td>
-                        \[\symbf{F}=G\,\frac{{m_{1}}\,{m_{2}}}{\|\symbf{d}\|^{2}}\,\symbf{\hat{d}}=G\,\frac{{m_{1}}\,{m_{2}}}{\|\symbf{d}\|^{2}}\,\frac{\symbf{d}}{\|\symbf{d}\|}\]
+                        \[\symbf{F}=G\,\frac{m_{1}\,m_{2}}{\|\symbf{d}\|^{2}}\,\symbf{\hat{d}}=G\,\frac{m_{1}\,m_{2}}{\|\symbf{d}\|^{2}}\,\frac{\symbf{d}}{\|\symbf{d}\|}\]
```
JacquesCarette, line 955: This is one of the cases where the old one might have been better? `\,{m_{2}}` might be better latex than `\,m_{1}` ?

code/drasil-printers/lib/Language/Drasil/Printing/Import/Symbol.hs
```diff
@@ -17,13 +17,18 @@ symbol (Integ    n) = P.Int (toInteger n)
...
+  let
+    mergeSymbols :: [Symbol] -> P.Expr
+    mergeSymbols [s'] = symbol s'
+    mergeSymbols ss   = P.Row $ map symbol ss 
+
+    renderSubSup :: [Symbol] -> [Symbol] -> P.Expr -> [P.Expr]
+    renderSubSup [] [] _ = []
+    renderSubSup us [] x = [x, P.Sup $ mergeSymbols us]
+    renderSubSup [] ds x = [x, P.Sub $ mergeSymbols ds]
+    renderSubSup us ds x = [x, P.Sup $ mergeSymbols us, P.Sub $ mergeSymbols ds]
```
JacquesCarette, line 30: I'm guessing this is the line than switches the order of Sup and Sub "by mistake".

code/drasil-printers/lib/Language/Drasil/Printing/Import/Symbol.hs
```diff
@@ -17,13 +17,18 @@ symbol (Integ    n) = P.Int (toInteger n)
...
-symbol (Corners [] [] [x] [] s) = P.Row [P.Row [symbol s, P.Sup $ symbol x]]
-symbol (Corners [] [] [] [x] s) = P.Row [P.Row [symbol s, P.Sub $ symbol x]]
-symbol (Corners [_] [] [] [] _) = error "rendering of ul prescript"
-symbol (Corners [] [_] [] [] _) = error "rendering of ll prescript"
-symbol Corners{}                = error "rendering of Corners (general)"
+symbol (Corners ul dl ur dr s)  =
+  let
+    mergeSymbols :: [Symbol] -> P.Expr
+    mergeSymbols [s'] = symbol s'
+    mergeSymbols ss   = P.Row $ map symbol ss
```
JacquesCarette, line 24: I'm guessing this flattening (instead of a nested `P.Row`) is what causes a level of `{}` to go away. I'm pretty sure the nested `P.Row` was inserted explicitly for the latex to have more `{}` in it at some point!!!!!

JacquesCarette (changes requested):

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
balacij: Oh, I did that intentionally. The order doesn't necessarily matter to LaTeX, so I just made a decision based on my usual way.

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
balacij: The extraneous `P.Row`s were the culprit for that. My changes try to use them as frequently as possible.

code/drasil-printers/lib/Language/Drasil/Printing/Import/Symbol.hs
balacij: Yep! Do you prefer the other way?

code/drasil-lang/lib/Language/Drasil/Symbol/Helpers.hs
```diff
@@ -53,7 +55,8 @@ subStr sym substr = sub sym $ Label substr
 -- | Helper for creating a symbol with a superscript to the right.
 -- Arguments: Base symbol, then superscripted symbol.
 sup :: Symbol -> Symbol -> Symbol
-sup b ur = Corners [] [] [ur] [] b
+sup (Corners ul dl ur dr b) p = Corners ul dl (ur ++ [p]) dr b
+sup b p                       = Corners [] [] [p] [] b
```
balacij, line 59: These changes to sub and sup are what cause the extraneous interior curly braces to be removed.

For example, if we previously had `sub lX f`, the result would be a `Corners [] [] [] [lX] f`. Then, if we had a `sup lI $ sub lX f`, the result would be a `Corners [] [] [lI] [] (Corners [] [] [] [lX] f)`. So, `Corners` wasn't really being used the way I _believe_ was intended (I assume this because of the errors previously thrown). With this change, we're merging the `Corners`, so that the `sup ... $ sub ...` produces `Corners [] [] [lI] [lX] f`. This way the generator can recognize that the symbol is intended to be one single unit rather than a compound symbol. That being said, there is a usecase for a compound symbol: exponents. For example, if we have `n_x^2` where `^2` is meant to represent "squared," I think it would be better presented as `(n_x)^2`.

code/drasil-printers/lib/Language/Drasil/Printing/Import/Symbol.hs
balacij: I think this was just incomplete. Searching through the issue tracker leads to #1606.

balacij: Lo and behold, that exact issue happens: Screenshot 2025-04-01 at 3 33 04 PM

balacij:
Pre-PR Projectile's HTML SRS "Detailed derivation of GD:posVec":

Screenshot 2025-04-01 at 3 39 51 PM

Post-PR:

Screenshot 2025-04-01 at 3 39 55 PM

The diff is a lot more conservative since you've last seen it, @JacquesCarette, but now there's an issue with symbol rendering in GOOL-generated code.

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
JacquesCarette: I'm fairly sure that we introduced those extra curly braces to fix a rendering bug. The fix was probably overly brutal, and I agree that fewer of them is a real improvement. Maybe try to use `git blame` to figure out the history?

code/drasil-printers/lib/Language/Drasil/Printing/Import/Symbol.hs
JacquesCarette: Ah ah! So we do need to take some care.

code/drasil-lang/lib/Language/Drasil/Symbol/Helpers.hs
JacquesCarette: This is most definitely an improvement, I agree.

JacquesCarette:
Post-PR is definitely an improvement, thank you.

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
```diff
@@ -866,15 +868,15 @@ <h4>
                     <p class="paragraph">
                       We also know the horizontal position that is defined in <a href=#DD:positionXDD1>DD:positionXDD1</a>
                     </p>
-                    \[p_{\text{x}1}=L_{1}\,\sin\left(θ_{1}\right)\]
+                    \[{p_{\text{x}1}}={L_{1}}\,\sin\left({θ_{1}}\right)\]
```
JacquesCarette, line 871: Looks like the change added too many curlies, especially at the 'top level', as can be seen here.

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
```diff
@@ -67,13 +67,13 @@ \subsection{Table of Symbols}
 {colspec={l l l}, rowhead=1, hline{1,Z}=\heavyrulewidth, hline{2}=\lightrulewidth}
 \textbf{Symbol} & \textbf{Description} & \textbf{Units}
 \\
-$a_{\text{x}1}$ & Horizontal acceleration of the first object & $\frac{\text{m}}{\text{s}^{2}}$
+${a_{\text{x}1}}$ & Horizontal acceleration of the first object & $\frac{\text{m}}{\text{s}^{2}}$
```
JacquesCarette, line 70: Same here. So I think the logic of when to add curlies still needs tweaking.

JacquesCarette (changes requested):

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
balacij: I _think_ that this comment (and the one above) were made on a commit that reverts my earlier removal of those curly braces. The current diff doesn't show these changes happening, so it's what's currently on `main`.

I agree that this is bad however. I want the extra curly braces removed. This is where an actual design is needed (#1606).

balacij (commented):
Unfortunately, I think we should close the PR and delete the branch in favour of tackling #1606 properly along with properly documenting how `P.Expr` is supposed to work and be used by the SRS and code generator. The small change I made to the `symbol -> P.Expr` function can be commented into #1606.

Why do I think this? Primarily because I wrote `P.Label ""` in the change, and this is what results in a change to the code. (It kind of screams "hack" alone, so it makes sense that it results in issues.)

To 'properly fix' that, the `symbol` printer needs to be aware of the printing context and switch its rendering strategy. It needs to neither actually have native prefix sub/super-script support and that adding such is hacked in with conventionally agreed knowledge. For example, for LaTeX, the output format should be `{}_{<DOWN_LEFT>}^{<UP_LEFT>}<BASE_SYMBOL>_{<DOWN_RIGHT>}^{<UP_RIGHT>}`. For code, ~`<DOWN_LEFT>_<UP_LEFT>_<BASE_SYMBOL>_<DOWN_RIGHT>_<UP_LEFT>`. Of course, if `<DOWN_LEFT>` is empty, then it would be removed along with the symbol it added to place it in the "down and left" corner position. However, we can't encode these formats at this level. It's in a weird spot in the rendering pipeline to make this decision because it isn't aware of where the rendered symbols will end up. That being said, I'm not so sure that the same symbol language should be used for both the SRS and the code generators. Maybe there should be an earlier step that generates the code symbols from the SRS symbols rather than generated when rendering the final artifacts.

Another issue with this PR is that it doesn't fix the issue for the symbols rendered in the `Table of Symbols` sections of the HTML-rendered SRS. For some reason, that table renders symbols using HTML-native nodes rather than MathJaX, so on top of it looking different, we can't even produce the sub/supscript symbols placed atop each other. `<sub> ... </sub><sup> ... </sup>` is not automatically placed the way we want them to be (see [Mozilla Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sub#see_also), "See Also" pt. 1).

code/stable/dblpend/SRS/PDF/DblPend_SRS.tex
balacij: I _think_ that this comment (and the one above) were made on a commit that reverts my earlier removal of those curly braces. The current diff doesn't show these changes happening, so it's what's currently on `main`.

I agree that this is bad however. I want the extra curly braces removed. This is where an actual design is needed (#1606).

code/stable/pdcontroller/src/csharp/Calculations.cs
```diff
@@ -20,8 +20,8 @@ public class Calculations {
     */
     public static List<double> func_y_t(double K_d, double K_p, double r_t, double t_sim, double t_step) {
         List<double> y_t;
-        Func<double, Vector, Vector> f = (t, y_t_vec) => {
-            return new Vector(y_t_vec[1], -(1.0 + K_d) * y_t_vec[1] + -(20.0 + K_p) * y_t_vec[0] + r_t * K_p);
+        Func<double, Vector, Vector> f = (t, y_tvec) => {
+            return new Vector(y_tvec[1], -(1.0 + K_d) * y_tvec[1] + -(20.0 + K_p) * y_tvec[0] + r_t * K_p);
```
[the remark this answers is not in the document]
balacij: Note: the _direct_ cause of this is that we now fuse nested `Corners`. Previously, this worked fine because it would be rendered separately. Now, not.

balacij:
If we want to force this PR to get merged, I think the course of action is to push `Corners` into `P.Expr` as well. This way, the final renderer of `P.Expr -> SRS|Code` can pick the right format as needed. I think this way would work, but I can't confidently say I designed a solution because the design of `P.Expr` doesn't immediately speak to me (and the docs are incomplete).

JacquesCarette:
I agree that closing this PR is likely the best solution. I'm fine with thinking of it as a "get your hands dirty exploration" of the issues with proper rendering.

That pretty-printers need more context is well-known. So of our printers have been so modified, but not all. So for sure that's in the cards.

That we need to redesign our layout primitives also doesn't surprise me too much. What the input for rendering into SRS and/or Code should be, is an interesting question. I think the current P.Expr, modeled on Presentation MathML, really should be fine for SRS. That it might not be fine for Code, that could well be!

balacij:
Moved the content to #1606. Closing the PR and deleting the branch.

PR #4035, opened 2025-04-07
#3529 Full Name Update

### LINKED ISSUE #3529: Add Abbreviation Introduction for Software Names
We would like to introduce the abbreviation for Software Names in the Introduction where it is first mentioned. For example, 

> Solar Water Heating System (SWHS)

daijingz:
Update full name on Drasil website, resolves #3529 
@JacquesCarette @balacij There are a lot of places I feel I should not change, so I just keep it. Please have some discussions with me.

@JacquesCarette For SSP, I did not change the `NP` capital letters, as it has some side effects on multiple MARKDOWN files. There are many similar places I feel hesitate modifying. I suggest you, once you find some places worthy to change, give me some information.


JacquesCarette:
I think these changes might make sense, but I think there should have been a discussion first #3529 doesn't cover all the changes, i.e. there should have been an issue to discuss the name changes. Also, there should have been 1 PR per name change, not rolling them up together like this.

Issue #3864, opened 2024-07-21
`State`ful artifact rendering: Verify it makes sense to implement (and implications), design, and implement


balacij:
(TODO: @balacij will return to this ticket on Mon. July 22nd, 2024 and write an _actual_ description).

balacij:
Finally returning to this, I'll start with a code snippet:

https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L193-L211

It seems harmless, right? The answer should be "yes" because that's one of the key "transformation/render" points where we generate the SRS. The issue is that we abuse it:

```console
λ ~/Programming/Drasil/code/ noInterMapDupe rg "mkSections" -ths
drasil-docLang/lib/Drasil/DocumentLanguage.hs
71:  mkSections fullSI l where
96:    allSections = concatMap findAllSec $ mkSections si $ mkDocDesc si dd -- FIXME: `mkSections` on something particularly large that is immediately discarded is a sign that we're doing something wrong. That's in addition to `mkDocDesc`...
128:    allSections = mkSections si $ mkDocDesc si dd
```

We have 3 references to it!

## 3 References to `mkSection`

### (1) Generating the Abstract SRS Documents

https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L66-L73

### (2) Finding all "Sections and LabelledContent" for inputting into the `ChunkDB`

https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L89-L118

### (3) Finding all "References" for inputting into the `ChunkDB`

https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L120-L149

balacij:
Now let's look at: #1235 and #1661. This issue is also related to why we have `usedinfodb` (also discussed in #3260 and more in https://github.com/JacquesCarette/Drasil/issues/1661#issuecomment-1021450950) -- a hack solution that requires _manual collection_ of the terms with acronyms presented in the generated SRS documents.

[...]

Issue #1661, opened 2019-07-03
Automate Table of Abbreviations and Acronyms


Mornix:
Related #698.

Currently `usedinfodb` is used to extract acronyms to print.
https://github.com/JacquesCarette/Drasil/blob/f43b3c189ac90af4681e5f1250077817a0f2e86a/code/drasil-docLang/Drasil/DocumentLanguage.hs#L275-L276

And the items pulled from the `usedinfodb` are manually populated in the examples.

What are the steps to automate this?
1. Use the `short` constructor to construct abbreviations (in the examples) as opposed to `getA`. The `short` smart constructor maintains chunk information of what is being abbreviated.
2. Write a plate extension for `sentencePlate`
https://github.com/JacquesCarette/Drasil/blob/f43b3c189ac90af4681e5f1250077817a0f2e86a/code/drasil-docLang/Drasil/ExtractDocDesc.hs#L180
to extract the abbreviations from `Sentence` constructors.
3. Expand `renderRef` from using raw strings to `Sentence` (makes reference pretty names more flexible too). Ensure `renderRef` embeds the chunk of the abbreviation (and not just the string). `ConceptInstance` already does this for `renderRef` with `Deferred`.
https://github.com/JacquesCarette/Drasil/blob/f43b3c189ac90af4681e5f1250077817a0f2e86a/code/drasil-theory/Theory/Drasil/Theory.hs#L65
4. Extend the mapping from `sentencePlate` done in step 2, to extract abbreviations from references.
5. Lower Table of Abbreviation and Acronym (and really all subsections in RefSec) to a later pass. (As things like Traceability Matrices are generated from the `DocDesc`, there will likely be references to likely changes section, which wouldn't be present prior to generation of the Traceability Matrix section.
6. Finally, enable expansion to only lookup terms which have been gathered.
https://github.com/JacquesCarette/Drasil/blob/f43b3c189ac90af4681e5f1250077817a0f2e86a/code/drasil-docLang/Drasil/DocumentLanguage.hs#L275-L276

balacij:
This was discussed in the [Dec 20th meeting](https://mcmasteru365.sharepoint.com/:v:/r/sites/SummerDrasil/Shared%20Documents/General/Recordings/Drasil%20Research%20Group%20Meeting-20211220_105536-Meeting%20Recording.mp4?csf=1&web=1&e=KgEnHL) (starting at approximately 6:30).

[...]

Issue #1235, opened 2019-04-26
Table of Acronyms includes unused acronyms


bmaclach:
Every `CI` in Drasil has a short-form, but that short-form is not necessarily used anywhere in the SRS. For example in SSP, the phrase "three-dimensional" is used, and the `threeD` `CI` is used for this. However, the associated acronym "3D" is never used in SSP. Despite this, "3D" still shows up in the Table of Acronyms. The table shows any term in the `TermTable` that has an acronym, regardless of whether the acronym was ever used.

PR #4037, opened 2025-04-09
WIP: No Duplicate Chunks in ChunkDB

### LINKED ISSUE #4036: `ChunkDB`: error out on duplicate UID insertion.
Everything added to the `ChunkDB` should be added only once. Currently, we seem to be overwriting data by inserting over `UID`s.

balacij:
Closes #4036 

Unfortunately, this is already appearing to be a challenge and is also digging up some issues.

I'm currently working towards fixing Projectile, but I've come across the following error
```console
projectile: Same symbols for different quantities found: [pressure,scalarPos] all have symbol p
[projSpeed,linearVelocity] all have symbol v(t)
[launchangle,angularDisplacement] all have symbol θ
```

### DIFF
--- a/code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
+++ b/code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
@@ -167,7 +167,7 @@ tMods :: [TheoryModel]
 tMods = [accelerationTM, velocityTM]
 
 symbMap :: ChunkDB
-symbMap = cdb (qw pi_ : map qw physicscon ++ unitalQuants ++ symbols)
+symbMap = cdb (qw pi_ : symbols)
   (nw projectileTitle : nw mass : nw inValue : [nw errMsg, nw program] ++
     map nw doccon ++ map nw doccon' ++ map nw physicCon ++ map nw physicCon' ++
     map nw physicscon ++ map nw mathcon ++ [nw algorithm] ++ concepts ++ 
@@ -279,10 +279,10 @@ physSystParts = map (!.)
 -- Various gathered data that should be automated --
 ----------------------------------------------------
 symbols :: [QuantityDict]
-symbols = qw gravitationalAccelConst : unitalQuants ++ map qw constants ++
-  map qw [acceleration, constAccel, iPos, iSpeed, iVel, ixPos,
-  iyPos, ixVel, iyVel, position, scalarPos, projPos, projSpeed, time, velocity, xAccel,
-  xConstAccel, xPos, xVel, yAccel, yConstAccel, yPos, yVel]
+symbols = unitalQuants 
+       ++ [tolQD]
+       ++ map qw [projPos, projSpeed]
+       ++ map qw physicscon
 
 constants :: [ConstQDef]
 constants = [gravitationalAccelConst, piConst, tol]
--- a/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
+++ b/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
@@ -35,6 +35,7 @@ import Data.List (sortOn)
 import Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Map as Map
 import Utils.Drasil (invert)
+import Debug.Trace (traceWith)
 
 -- | The misnomers below (for the following Map types) are not actually a bad thing. We want to ensure data can't
 -- be added to a map if it's not coming from a chunk, and there's no point confusing
@@ -84,7 +85,7 @@ cdbMap mn fn = Map.fromListWithKey (\k _ _ -> error $ "'" ++ show k ++ "' is ins
 
 -- | Smart constructor for a 'SymbolMap'.
 symbolMap :: (Quantity c, MayHaveUnit c) => [c] -> SymbolMap
-symbolMap = cdbMap "SymbolMap" qw
+symbolMap cs = cdbMap "SymbolMap" qw $ traceWith (show . map (^. uid)) cs
 
 -- | Smart constructor for a 'TermMap'.
 termMap :: (Idea c) => [c] -> TermMap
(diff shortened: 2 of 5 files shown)


JacquesCarette:
Quantities can have the same symbols, as long as they are not used within the same example. Because, for example, `pressure` does not occur in Projectile, then the real bug is that it is being inserted in Projectile's database.

The overlap of `projSpeed` and `linearVelocity` seems to be deeper that needs more investigation.

Duplicate UIDs are a much bigger problem. I

JacquesCarette:
Also: I'm surprised that these changes of yours have 'surfaced' these duplicate symbols when this wasn't an error before? That is most definitely quite odd.

balacij:
> Quantities can have the same symbols, as long as they are not used within the same example.

Agreed, with the caveat that "same example" means "same `SI`, excluding the ChunkDB." I think we should be able to throw anything/everything into the `ChunkDB` without that affecting anything currently generated by Projectile. Unfortunately, that's not the case right now.

> Because, for example, pressure does not occur in Projectile, then the real bug is that it is being inserted in Projectile's database.

Ah, I guess we're on different pages. Why we shouldn't be able to throw `pressure` into Projectile's `ChunkDB` if Projectile's `SI` never directly/indirectly references `pressure`? If memory is a concern, then that's okay, but otherwise?

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -279,10 +279,10 @@ physSystParts = map (!.)
...
 ----------------------------------------------------
 symbols :: [QuantityDict]
-symbols = qw gravitationalAccelConst : unitalQuants ++ map qw constants ++
-  map qw [acceleration, constAccel, iPos, iSpeed, iVel, ixPos,
-  iyPos, ixVel, iyVel, position, scalarPos, projPos, projSpeed, time, velocity, xAccel,
-  xConstAccel, xPos, xVel, yAccel, yConstAccel, yPos, yVel]
+symbols = unitalQuants 
+       ++ [tolQD]
+       ++ map qw [projPos, projSpeed]
+       ++ map qw physicscon
```
balacij, line 285: > The overlap of projSpeed and linearVelocity seems to be deeper that needs more investigation.
> Also: I'm surprised that these changes of yours have 'surfaced' these duplicate symbols when this wasn't an error before? That is most definitely quite odd.

Info. dump:
1. `linearVelocity :: UnitalChunk` is imported through `physicscon :: [UnitalChunk]` (an import from `drasil-data`).
2. `symbols :: [QuantityDict]` looks no different than any of other "aggregate all chunks of this particular type" lists (following the comment right above it too!).
3. `symbMap :: ChunkDB` carries our list of "all symbols" and currently carries both `QuantityDict`s extracted from both `linearVelocity` and `projSpeed :: UnitalChunk`.
4. `symbMap` adds things on top of `symbols` before registering them -- probably just something someone never fixed, right?
5. ^ No! `symbols` also fills in the `_quants` field of the `SI`. Adding `linearVelocity` to `symbols` means it's also added to the `_quants` field.
6. `_quants` is used in code generation. Specifically, it is the list of symbols that are shown in the "Table of Symbols" shown in the SRS (relevant code is `drasil-docLang/lib/Drasil/Sections/TableOfSymbols.hs`'s `table` function).
7. `table` ensures that a table of symbols only contains unique symbols. Hence, the error.

balacij: This is a very similar issue to https://github.com/JacquesCarette/Drasil/issues/1661, which works around a similar issue through `usedinfodb` (a second `ChunkDB` we hold in our `SI` so that our "Table of Acronyms" only carry the ones shown in the document [manually created list of acronyms]).

JacquesCarette: I think this means that we should deprecate hand-assembled things in favour of grabbing what's actually used.

balacij: I certainly agree, and I realize I linked the wrong ticket (actually, they are duplicates of the same root issue). I intended to link #1235 which I link to from #3864 (the solution I expect will help us).

JacquesCarette (commented):
Needs a merge. How close is this to be ready?

balacij:
When I said it out loud in the meeting that we need compile -> analyze error -> fix -> compile -> analyze error -> etc., I felt a bit silly. Early, I didn't expect there to be many inter-map `UID` conflicts, which is why I had made the `cdb` function error out on those instances. With the latest commit, @jackwyand, I made it dump all the `UID` conflicts. Now you should be able to do the analysis first without much headache.

jackwyand:
Thank you @balacij. After the meeting I realized I was a little unclear as to how we were going to utilize this branch. Should I revert all the changes I made to the duplicates in order to use this branch as a way to identify the duplicate UIDs and then use separate branches to analyze and fix duplicates in their own PRs?

balacij:
That's fine, we didn't really know how things would go, and I jumped the gun a bit with this PR because I was not expecting this many duplicates.

We can slowly pull the 'obvious' things into other PRs. For example, the `1x` and `2x` UID suffixes are easy to pull out into other PRs. The other things can be slowly analyzed and then pushed into new tickets or new PRs.

This PR can be kept to keep track of how many issues we have left, and it can eventually be turned into a PR that enables the hard error again.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -31,8 +31,9 @@ modElas = uc' "modElas" (nounPhraseSP "modulus of elasticity of glass")
 {--}
 
 constrained :: [ConstrainedChunk]
-constrained = map cnstrw inputDataConstraints ++
-  [cnstrw probBr, cnstrw probFail, cnstrw stressDistFac, cnstrw nomThick, cnstrw glassTypeCon]
+constrained = map cnstrw dataConstraints ++ [nomThick, cnstrw glassTypeCon]
```
balacij, line 34: `map cnstrw dataConstraints` == input constraints ++ derived input constraints

`[nomThick, cnstrw glassTypeCon]` == quasitype constraints

jackwyand:
> @jackwyand we don't usually force-push. Any reason you are using that method?

@JacquesCarette I accidentally included extra commits while rebasing to squash some of my old ones, so I rewrote the commit history to clean it back up. Jason and I talked about it in person and made sure everything was okay.

jackwyand:
Just saw your comment now, sounds good @balacij.

jackwyand:
Yep, sounds good. Closing it now.

Issue #2911, opened 2021-12-02
What do we want from our `UID`s?


balacij:
This kind of hack strongly tells me that we need to revamp our `UID` implementation to something more structured. I know MMT uses URIs for doing something similar. I don't like that either, but it is (slightly) better than name concatenation.

I'm okay with keeping the hack for now, as we can easily find it by looking for `+++`.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/2904#discussion_r760502476_

JacquesCarette:
UID = "Unique IDentifier". It is a way to refer to a piece of 'data', and gives an 'identity' to data.

Using strings for this was a horrible hack that worked while we were figuring things out. At least now, we kind of have an abstract type for it, so that the outside world shouldn't know our internal representation. Which shouldn't be a string.

In the web world, [URIs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) are a 'solution' to provide [addresses](https://www.w3.org/Addressing/) for resources. Related are [URNs](https://en.wikipedia.org/wiki/Uniform_Resource_Name).

Although URIs and URNs are **encoded** as strings, they possess a grammar so that their parts can be recovered (by an easy parse). The problem with that is such a scheme means that the names of 'inner' items either need to be restricted (to avoid 'code' characters) or these need to be escaped. This is useful for external transmission - but since UIDs are completely internal, we are free to design out own Abstract Data Type.

The point is that our 'resources' (knowledge, information, in large part) do happen to live in contexts. If these contexts have names, then we can indeed use *name spaces* as a means to have short local referents (our current UIDs) that are still unique because they sit in their own name space.

Part of the problem is that we have **derived information**, for which we don't have a good scheme that generates names for them. For example, we could have many things have the same 'name', but if they had different 'types', and those types were part of our own UID/URI/URN scheme, then they would not be the same anymore. This is what the appending hack tries to fake.

> I'm a bit confused by our existing ConceptDomains; I'm having a hard time understanding which "level" we exactly want to call a conceptual domain for something. Do you think a new UID scheme could/should address this?

[...]

PR #4039, opened 2025-04-29
Demote IdeaDicts/'terms' table to 'not a chunk'

balacij:
Contributes to #2873

The `IdeaDict`s map is essentially a cache of known 'terms'. This PR removes it from consideration of being a 'chunk' because they contain only information about a single word. In the future, we can also remove the `IdeaDict` type entirely in favour of just referring to the 'term's themselves (via `NP`), but I'm not yet confident we want that.

With this PR, the majority of the problematic conflicting `UID`s appearing in the `problematic_seeds.json` file is from conflicts between `Concept`s and `ConceptInstance`s.

### DIFF
--- a/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
+++ b/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
@@ -165,7 +165,6 @@ asOrderedList = map fst . sortOn snd . map snd . Map.toList
 data ChunkDB = CDB {
   -- CHUNKS
     symbolTable           :: SymbolMap
-  , termTable             :: TermMap 
   , defTable              :: ConceptMap
   , _unitTable            :: UnitMap
   , _dataDefnTable        :: DatadefnMap
@@ -175,6 +174,7 @@ data ChunkDB = CDB {
   , _conceptinsTable      :: ConceptInstanceMap
   , _citationTable        :: CitationMap
   -- NOT CHUNKS
+  , termTable             :: TermMap 
   , _labelledcontentTable :: LabelledContentMap -- TODO: LabelledContent needs to be rebuilt. See JacquesCarette/Drasil#4023.
   , _refTable             :: ReferenceMap -- TODO: References need to be rebuilt. See JacquesCarette/Drasil#4022.
   , _traceTable           :: TraceMap
@@ -202,7 +202,6 @@ cdb s t c u d ins gd tm ci lc r cits =
   CDB {
     -- CHUNKS
     symbolTable = symbolMap s,
-    termTable = termMap $ t ++ termsHACK,
     defTable = conceptMap c,
     _unitTable = unitMap u,
     _dataDefnTable = idMap d,
@@ -212,6 +211,7 @@ cdb s t c u d ins gd tm ci lc r cits =
     _conceptinsTable = idMap ci,
     _citationTable = idMap cits,
     -- NOT CHUNKS
+    termTable = termMap $ t ++ termsHACK,
     _labelledcontentTable = idMap lc,
     _traceTable = Map.empty,
     _refbyTable = Map.empty,
--- a/code/drasil-database/lib/Database/Drasil/Dump.hs
+++ b/code/drasil-database/lib/Database/Drasil/Dump.hs
@@ -3,7 +3,7 @@ module Database.Drasil.Dump where
 import Language.Drasil (UID, HasUID(..))
 import Database.Drasil.ChunkDB (conceptinsTable, theoryModelTable, gendefTable,
   insmodelTable, dataDefnTable, unitTable, citationTable, UMap,
-  ChunkDB(termTable, symbolTable))
+  ChunkDB(symbolTable))
 
 import Data.Map.Strict (Map, insert)
 import qualified Data.Map.Strict as SM
@@ -19,7 +19,6 @@ umapDump = map ((^. uid) . fst) . SM.elems
 dumpChunkDB :: ChunkDB -> DumpedChunkDB
 dumpChunkDB cdb = 
       insert "symbols" (umapDump $ symbolTable cdb)
-    $ insert "terms" (umapDump $ termTable cdb)
     $ insert "concepts" (umapDump $ cdb ^. conceptinsTable)
     $ insert "units" (umapDump $ cdb ^. unitTable)
     $ insert "dataDefinitions" (umapDump $ cdb ^. dataDefnTable)


JacquesCarette:
I think that terms are very important! They may not contain much, but they are our basic vocabulary. I don't see the advantage of demoting `IdeaDict` to not be a chunk.

We want these to be unique too. So a global database of them helps with that.

balacij:
I agree, terms are very important. And I think they should be a built-in feature of `ChunkDB` because of how prominent they are. However, designing that is too large of a project for right now. Further considering we don't have an urgent need for it, it's a rock I want to kick down the road. 

## Why do I think they should be a built-in feature?

Bear with me for a long-winded explanation that took me far too long to type up. Here's our chunk graph related to Projectile:

Screenshot 2025-05-01 at 11 13 14 AM

Everything but the following chunk types have a `term` they define:
* `Reference`
* `LabelledContent`
* `Section`
* `DecRef`
* `Citation`

Now, two questions.

## Two Questions

1. Do `IdeaDict`s within a `ChunkDB` declare unique `term`s?
2. Do all chunks define `term`s?

### Hypothetical Answers

Let's "case" on the answer.

#### (Yes, Yes)

Then both `term`s and `UID`s are unique, but they have a 1-1 relationship, as well as a 1-1 relationship to a unique chunk. So, a `term` is also a kind of `UID` (with just one extra level of indirection to the chunk)! In which case, we can flatten and remove `IdeaDict`s as well as `UID`s in favour of `term`s.

#### (Yes, No)

There should still be a better design. If all `IdeaDict`s define unique `term`s (up to the `ChunkDB`) but not all chunks define `IdeaDict`s/`term`s, then we can still remove `IdeaDict` in favour of a `term` cache (in the `ChunkDB`) of known terms and use said terms as alternative `UID`s to chunks. Everything else would only be referable by `UID`.

#### (No, Yes)

If `IdeaDict`s don't define unique `term`s (up to the `ChunkDB`) and all chunks define terms, then we can expect `term` conflicts. i.e., ambiguity. Do we really want to permit that? It depends on how we feel about term overloading. My kneejerk reaction is that it's a good feature, but then I think that `term`s in `IdeaDict`s should still be unique up to some criteria (e.g., domain?). For example, "chunk" means something different in "Drasil" and "psychology." The `ConceptChunk`s that define "chunk" (in English) for these domains respectively, should also note which domain it is defining the term for. This is what `ConceptChunk` already does via `ConceptDomain`! `ConceptChunk` further let us define `term`s for multiple domains using the same definition.

Okay, this is good. Now, a fairly natural query we might have for Drasil is: "what do we know about 'chunk' (w/ all disambiguated definitions)?" How would our implementation work?

Our implementation would need to look into _all_ `IdeaDict`s, whose respective `term`s are "chunk," and gather their `UID`s. Querying _all_ hurts me a bit. How can we avoid this situation?

We can replace `UID`s with `term`s $\times$ `domain`s. This way, `UID`s remain unique, but we can give a bit more structure to our `ChunkDB`: from `UID -> Chunk` to `term -> domain -> Chunk`.

One issue here is that we're going to have nested maps (excess memory use for something fairly common, but that I wouldn't expect common enough to warrant the excess memory use).

Regardless, it remains that we can redesign `IdeaDict` by essentially making it a built-in feature of `ChunkDB`.

#### (No, No)

This is the ugly duckling (easiest to answer): it's a mixture of the previous two answers. The `UID`s for the things that don't define `term`s would be purely random.

### All Hypothetical Answers Converge

Now, in any of these situation, I think we can (and should) redesign `IdeaDict` along with `ChunkDB`. So, what's the 'right' answer for the two questions?

### But what is the real answer?

I believe the answer is actually none of the above, but instead: (Yes [up to domains at least], Yes).

As I mentioned at the top of this comment, we already have some chunk types that don't define `term`s. So, the answer to the second question is No. However, I would argue that it should be Yes!

#### Why should the answer to Q2 be "Yes"?

First, let's look at the chunk types that don't define `term`s:
* `Reference`, `LabelledContent`, and `DecRef`: Commenting on these 3 is a bit difficult because I think they're all confused in various ways. I think that they should all be redesigned first to truly answer definitively.  However, I'm quite convinced that these don't violate the rule. Thoughts:
  * `DecRef`: This one annotates a `Reference` with information about what content the reference is specifically directing attention towards, e.g., "equation $x" or "page $x." There are two other variants: `None` and `RefNote String`. `None` is the default, meaning directing attention towards the whole. `RefNote` is exclusively used for added a note "(with PCM removed)" to a source in SWHSNoPCM's [IM:eBalanceOnWtr](https://jacquescarette.github.io/Drasil/examples/swhsnopcm/SRS/HTML/SWHSNoPCM_SRS.html#IM:eBalanceOnWtr). That's a little bit hand-wavy to me. I wasn't able to immediately figure out what that directs attention to looking at [Koothoor's MASc work](https://macsphere.mcmaster.ca/bitstream/11375/13266/1/fulltext.pdf). I think the note needs to be "fixed" somehow, and I'm a bit skeptical of the necessity of "RefNote." It really seems like a hack. Also, it's a heavily underused chunk type because we hack in the same "decorations" (references) with `Sentence`: https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-lang/lib/Language/Drasil/Reference.hs#L47-L68 Regardless, this is no different than $p_x$ ("thing @ equation $x"), $p_y$ ("thing @ page $x"), or even $n+1$ (`RefNote`) -- we know that the creation of them as symbols is a wart, they should merely be atoms that we can extract terms for if we so choose, but they're not really "chunks" -- they're atoms.
  * `Reference`: It is used to refer to other labelled things, and duplicates their/contains projections of their short names (e.g., IM:$T, etc.). We use it to refer to specific bits of the 'reference' via the aforementioned `DecRef` and the `Sentence` hacks. I don't think this should be a chunk at all. I see it as no different than `Expr`'s symbol references (`Sy` reference to `QuantityDict`s; an atom). So, I don't think should be a chunk at all.
  * `LabelledContent`: This one defines the content that goes into a reference. e.g., a table, figure, equation block, table, definition, figure, graph, code block, paragraph, derivation. All of these define a term that's relative to a file/document, its sections, and/or naming scheme. For example, "Table [optional: $n]: $title", "Figure [optional: $n]: $title", ..., "Paragraph $n [optionally added: of Section $m]", etc. Note: all of these names should be generated and made relative to various files/documents. Arguably, these titles indicate that these are really references. So, I think they all define a unique term. We just don't record it.
* `Section`: We don't record their terms, but they certainly have one. In the context of a `ChunkDB` building up "Projectile", each section should refer to `$file's section $x.y.z, "$title"` in the "Projectile" domain.
* `Citation`: I think this should have one as well. All bibtex entries should have at least a title, author list, and year recorded. So, for example, `Koothoor's "Document driven approach to certifying scientific computing software"` (perhaps with `paper (2013)` appended) sounds like a valid `term` that it could/should have. Uniqueness of this term might be an issue if authors published $n$-works in the same year with the same title and format, but I think that should force disambiguation in the title. For example, if one was a poster vs a paper, they should be added to the title. The `term` should be a rigid designator for the exact work we're trying to build a bibtex-like citation for.

Thus, I believe that all "All chunks should define a term" is a fair requirement given that all of our existing chunks either define a term already or should have one. 

A weaker, alternative rationale for "yes" is that we use terms to uniquely direct each other to snippets of code implicitly, only through the terms. Thus, I don't believe there should exist any chunks that we can't speak about. Even the things that are pure meaningless garbage (e.g., `a;lkdhasiufhjakfhgiauu08rhhjkjbfkjabfkjabh`) should still have a unique term (e.g., `the text spam in $document's paragraph $n`).

One hint this gives as well: terms have structure (which we have been trying to extract via the various sentence and name combinators we have!!! +1!!!).

Now returning to Q1: The answer is "no" right now, and that's just because that's how the code operates. However, I believe the answer to Q1 should be "Yes, up to domains at least."

#### Why should the answer to Q1 be "Yes, up to domains at least?"

The "chunk" example already explains the former "domains" requirement, but I'll reiterate quickly. We're going to run into situations where we want to prescribe a term multiple definitions. There's no way getting around this. We can cheat that in with `IdeaDict`s as they are, but then we run into the wasteful resource use issue I mentioned earlier for the case of querying based on term. Is that a big deal? I believe it is. I think that Drasil really wants to become something we interactively work with and program rather than have deeply embedded in our existing Haskell code. Should that happen, this feature will be important.

However, an issue arises: we might want to have duplicate definitions within the same domain of the same term. The alternative definition will be based on the axiomatization. For example, acceleration can be Cartesian-based or Polar-coordinate-based. Here, the axiomatization differs between them: the Cartesian-based system uses a vector of acceleration in each axis while the Polar coordinate only has the two components.

Another issue: identity. We might want to refer to the identity of $+$, $\times$, or my identity (or the type of identities; `Identity`). In other words, axiomatizations ~~~~ "chunk _types_" might also be a part of this "up to" criteria. Regardless, I think overloading is a good thing, or else we need to remember a million names, but also that `term`s should be able to uniquely refer to chunks.

## Returning to the Original Question

Why do I think they should be a built-in feature? That's because I believe the answer to the above two questions is: (Yes [with an unclear caveat], Yes) whilst allowing term overloading, and the work involved with all of that involves more closely integrating `IdeaDict` into `ChunkDB`, likely removing `IdeaDict` altogether. However, designing how to build `ChunkDB` for our unknown requirements is not an easy task, which is likely easier to deal with after having dealt with #2873.

balacij:
Thank you, @smiths !

Yes, I believe the `term` is the "name" (noun phrase / the `NP` type) that a chunk can "defines." When we define `IdeaDict`s, all we're doing is declaring that a `term`/`NP`/name exists.

That wiki page looks out-of-date. `NamedChunk`s do not exist now. What that wiki page shows is a `NamedChunk` is now an `IdeaDict` (funny enough, our constructors for `IdeaDict` still use the `nc` acronym -- and this is the first time I realized why!). What the wiki pages shows is an `IdeaDict` appears to be replaced with [`CI`](https://github.com/JacquesCarette/Drasil/blob/390cd3276819387a10b9313853c96a3172c82281/code/drasil-lang/lib/Language/Drasil/Chunk/CommonIdea.hs#L22-L28).

I agree with you. I prefer `term` $<$ `noun phrase` $\le$ `idea` with slight preference to `idea` with an understanding that an idea is just an alias for noun phrase. Then we can say that all of our chunks are really "conceptualizations," or, "concepts" that give meaning to "ideas." I like these discussions of idea vs concept: https://philosophy.stackexchange.com/a/76403 and https://philosophy.stackexchange.com/a/92173 (but there are definitely better ones out there). I forget where I saw this discussion from a real/academic source that we can reasonably take at face value -- so take this with a grain of salt. Also, I think this discussion is particularly important because we're redesigning the chunks and the current names are very confusing. 

Thank you! I think the acceleration example certainly throws a wrench in the system. I'm not entirely sure how this works yet, but I'm expecting we'll need to consider subdomains in addition to domains. The SWHS[NoPCM] examples are also similar. I'll need to think about this, and have a proper read through of SWHS[NoPCM]. I need to expedite working on re-writing Projectile on pen and paper...

JacquesCarette:
There is *way* too much implicit stuff in the above, i.e. presuppositions that we might not agree on, which would undermine the whole analysis. So here we go.

- What is a 'chunk'? It is a small amount of re-usable information. It is meant to be unique, and the UID serves as our means to do that.
  - note that many chunks have 3 'names' associated to them:
    1. a Haskell variable name, in Drasil code
    2. a UID, meant to be used for internal consistency and uniqueness; it is not meant for humans
    3. a 'term', meant to be meaningful to human beings (and only to humans)
- What is a 'UID'? (See also above): an internal, machine-readable key associated to all information we store.
- What is a 'term'? (See also above): a human-readable 'phrase' which is the name of a concept. Its *type* is NounPhrase because that is the grammatical category that it falls in.
- What is an 'IdeaDict'? It is our currently means of storage for a re-usable term.

So if we now turn to the questions:
1. Do `IdeaDict`s within a `ChunkDB` declare unique `term`s? There are two ways to interpret that: First, unique UIDs, second, unique NounPhrase. By definition the UID should be unique, otherwise it's a bug. But the embedded `NounPhase` does not need to be unique!!! Human language overloads things a lot, so there should be no assumption of uniqueness at this level.
2. Do all chunks define `term`s. Here the answer should be a clear 'no'. There are many pieces of re-usable information that do not have an associated single piece of English.

So I'd say the only interesting case is (no, no). And the analysis is somehow flawed because it basically does not consider this case as valid?

Why do we diverge so much? I think because @balacij may have a really different idea of "What is a Chunk" than I do -- but leaves that implicit.

Now, I think what @balacij might be arguing for is that all re-usable information should be somehow 'renderable' in English (and he uses `term` as a proxy for that). I don't believe that's necessary.

Lastly, as to the design question (should `term` be built-in to `ChunkDB`), until I know the *purpose* of `ChunkDB`, I can't even form an opinion on that. To me `ChunkDB` is a database of re-usable information with multiple tables but all of them having a unique key (`UID`). Why would it also have another colum (`term`) exist in *all* tables? That seems to overly restrict the kinds of information that we can reuse! So I continue to be thoroughly unconvinced.

smiths:
@JacquesCarette I like your discussion above.  I especially like that you defined chunk, UID, term etc.  You are correct that we can't communicate properly without the same definitions.  We saw a similar problem when we were  using the word theory when we weren't all thinking the same thing.

@balacij has pointed out that our existing documentation in the Drasil wiki is out of data.  I find I go to our documentation often, and I know our summer students use it.  We should update the documentation in the wiki to match our current understanding and definitions.  @balacij could you take on the task of updating our documentation?  I don't mean everything.  I just mean not losing the current conversation.  Remove the pieces of documentation that are out of date, like NamedIdea, and copy the definition of chunk, UID etc that @JacquesCarette has provided.  We can then point to those definitions whenever we need them. 

@JacquesCarette, on the question "do all chunks define `terms`s"? you say there are pieces of useful information that do not have an associated piece of English.  I feel like if we wanted to, we could add a piece of English to any chunk.  To talk about that knowledge, we give it a name, so we could have a `term` for it.  Maybe we don't want to, but I feel like we could.  If we insisted that every chunk have a `term`, I think we could do that.  Am I wrong about that?  Is there an example where we couldn't?  If I am right and we can add a `term`, what is the advantage of not doing so?  By having a term we could take an English view of any chunk, in case we want to talk about it in some text that we generate.  I guess I'm wondering whether adding a `term` is small enough a task that we should require it for all chunks, just in case we need it?

balacij:
Thank you for clarifying the working definitions. I believe I'm operating on the same definitions.

> So I'd say the only interesting case is (no, no). And the analysis is somehow flawed because it basically does not consider this case as valid?

I agree with you, actually. I acknowledged that this is our existing situation, albeit in separate sentences:
> As I mentioned at the top of this comment, we already have some chunk types that don't define terms. So, the answer to the second question is No. However, I would argue that it should be Yes!
&
> Now returning to Q1: The answer is "no" right now, and that's just because that's how the code operates. However, I believe the answer to Q1 should be "Yes, up to domains at least."

I see now that I wrote the following sentence poorly, however:
> I believe the answer is actually none of the above, but instead: (Yes [up to domains at least], Yes).

I meant to say "I think the answer _should be_ none of the above." I apologize for that. I went through 2 phases of rewriting my original write-up to fix holes, and the last copy was a bit rushed. Regardless, I acknowledge that the current answer is (No, No). The reason why I idealize to (Yes, Yes) is because:
1. Q1: I believe that every chunk (that is, recorded in the ChunkDB) should be uniquely referrable/queryable through a (structured) English statement (which might involve other internal DSLs) without using a `UID`. That is why I talk about how the existing data structure would not necessarily be up to standards, and we would need to rethink how the `ChunkDB` is made and how `IdeaDict` is worked with. I also acknowledge that overloading terms is a good, desirable thing, and that's why I added the discussion about "domain, type, at least."
2. Q2: This goes to your comment:
  
  > Now, I think what @balacij might be arguing for is that all re-usable information should be somehow 'renderable' in English (and he uses term as a proxy for that). I don't believe that's necessary.
  
  Yes, I believe that, and it sounds like this is a particularly major difference. If something weren't in any way/shape/form renderable in English (or whatever human-readable natural language we choose, which may contain other DSLs), then what are we saying about Drasil's scope of well-understood knowledge [I'm not saying this is a contra-point to your argument]? Can you give me an example of something that you couldn't assign a unique term to? I've racked my brain trying yesterday when writing this all up. One note here: I'm not saying that all terms will be useful/used in the end for our purposes.

balacij:
> @balacij has pointed out that our existing documentation in the Drasil wiki is out of data. I find I go to our documentation often, and I know our summer students use it. We should update the documentation in the wiki to match our current understanding and definitions. @balacij could you take on the task of updating our documentation? I don't mean everything. I just mean not losing the current conversation. Remove the pieces of documentation that are out of date, like NamedIdea, and copy the definition of chunk, UID etc that @JacquesCarette has provided. We can then point to those definitions whenever we need them.

Yep, I think I can work on that slowly.

> "is a term" and "has a term" are quite different things. Same difference as "is a description" and "has a description". Or "is a value" and "has a value".
> We need to be more precise. "Has an English description" is a laudable goal -- and is very different than "term".

I acknowledge and agree with both of these.

> Lastly, as to the design question (should term be built-in to ChunkDB), until I know the purpose of ChunkDB, I can't even form an opinion on that. To me ChunkDB is a database of re-usable information with multiple tables but all of them having a unique key (UID).

Since we're pinning down our definitions, I'll write how I see `ChunkDB`: Ignoring the extra niceties that a `ChunkDB` currently has -- aka traceability matrix, etc., we may think of `ChunkDB` as one big map: $k \times ty \to ty$, where $k$ is a `UID`/key and $ty$ is the type of the chunk. This holds under the current design ($k$ is a `UID` where the "$\times ty$" part involves picking the right 'lookup' function that spits out the right $ty$ type) and the new ($k$ remains a `UID` but there is only one lookup function that looks for the unique chunk with the `UID` and tries to cast it to $ty$). 

So, I think we see it the same.

> Why would it also have another colum (term) exist in all tables? That seems to overly restrict the kinds of information that we can reuse!

The way I see it is that there is nothing ineffable in the LLWUS scope, so I don't quite see the issue. I admit that for generated chunks, we would perhaps be generating a lot of 'junk', but any of the things we refine should have meaningful ones. I think of our recent "the _ of the _" discussion here, as well as more complex long-in-the-future statements, such as "the output of rendering Projectile as a software via the Java ICO software generator against user-designated choices."

What's interesting about querying this last statement is that we could be telling the `ChunkDB` implicitly to make an inference (i.e., run the generator itself). But now this is 10 steps ahead of myself, top-down, and assuming that Projectile is an SI, SI is a _chunk_ (which I know we've been saying is not), "Java ICO software generator" is a chunk, and "user-designated choices" may or may not be a chunk.

The way I see (Yes [up to domains, types, at least], Yes) going involves potentially giving some more structure to the `UID`s using the `term`s, and the domain information, type information, etc., similar to how we were discussing the structure of terms of various chunks. So, it might not end up that `term`s are added as a 'column' to all of our 'tables'.

Maybe I'm confusing myself now, I know I'm doing too much top-down thinking, but this is also why I want to kick the 'terms' rock down the road a bit. If we want, we can also hold off on committing to kicking it down the road once we've guaranteed that there is at most on chunk conflict with `IdeaDict`s table.

Linking the existing ticket on `UID`s: #2911 (we apparently haven't spoken on that ticket since 2022!).

JacquesCarette:
> I believe that every chunk (that is, recorded in the ChunkDB) should be uniquely referrable/queryable through a (structured) English statement (which might involve other internal DSLs) without using a UID.

Why? I really don't see the need for that as a requirement. That we could *output* a structured English statement from all chunks is something I can see (though I'm still not convinced of that; I can see instances of having some 'black boxes' still exist uniquely and be referable and re-usable, i.e. fit what we use chunks for). So if you really believe these two items (English for input, English rendering) you need to (strongly) justify each.

[See also a different post about the need for defining things; I think we're miscommunicating in part because we don't seem to have a shared definition of 'term'.]

I'll add a rephrasing of the (extra) usefulness of `ChunkDB`: at the level of constants, it insures that we can have a "global theory" of the current system where everything is unique, at least internally. So it's a 'flattening' in a way.

PR #4051, opened 2025-05-08
Enable StrictData extension in drasil-gool via package.yaml

sarrasoussia:
After turning on StrictData in drasil-gool, the program crashed during codegenTest because it tried to use values that aren’t numbers in a math expression. I updated the numType function to show the actual types that caused the problem. This makes it easier to understand what went wrong and where to fix it, since the error message now tells us which types were used incorrectly.


JacquesCarette:
The change is good, but obviously I can't merge it, as it would break the build! Can you document on here what the results of the debugging information is?

PR #4074, opened 2025-05-16
New Workspace Setup: Add information about permanently switching to UTF-8 encoding on windows

### LINKED ISSUE #4069: Windows unicode support permanent fix for new workspace setup
The current [new workspace setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup) states

>each time you open up Cygwin/Git Bash, you will need to run chcp.com 65001 or chcp 65001 so as to change the encoding of the terminal to UTF-8.

To make this permanent you can add `chcp.com 65001 > /dev/null` to the `.bashrc` file for Cygwin/Git Bash, running the command and hiding it from the output using `/dev/null`. The `/dev/null` part is technically optional but I think it would be cleaner to not have the extra output every time on startup. I think that the new workspace setup should mention this as I often found myself forgetting to run the command, waiting for a full `make` to finish, realizing, and having to do it again.

jackwyand:
Closes #4069


JacquesCarette (approved):
I'll pull this in because it is an improvement, but in general I seriously dislike 'here' links. In this instance, I would have preferred "this stackoverflow post" as the text for the link.

Issue #3238, opened 2023-01-15
Should we be explicitly building `ConceptChunk`s when making `UnitalChunk`s or letting the constructors do it implicitly?


samm82:
_This issue was migrated from #3199 and will be updated upon its completion. It makes #3074 obsolete._

---

We have constructors that build a `UnitalChunk` from a `Concept c` and constructors that build it from "a given 'UID', term, and definition". It definitely makes sense to have constructors of the first type (since some concepts are reused independently of their `UnitalChunk` versions, but I'm not sure if we should keep the constructors that implicitly build a `Concept` from its parts. It might be good to force the user to create the `Concept` explicitly to encourage reuse, support future change, and make code less cluttered, but it also means that information is more spread out. These implications are similar to those discussed in the [wiki page for the `DefinedQuantityDict` chunk investigation](https://github.com/JacquesCarette/Drasil/wiki/DefinedQuantityDict#conceptchunks-made-only-to-be-used-in-a-dqd). Since `ucs` (which builds from the parts of a `Concept`) is used in `cuc'`, we would potentially need to tweak it as well.

https://github.com/JacquesCarette/Drasil/blob/09365d7d1f5dc9b54e84a8d64cb601184456ea0c/code/drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs#L125-L130

Alternatively, we could also move one of these sets of constructors to `drasil-utils`.

JacquesCarette:
As discussed at the meeting, this is part of a bigger design question. A `Unital` is really an instance of its underlying `Concept`; so the `Concept` should be "by reference" instead of being *contained*.

Can you please add (or point me to if it already exists) a list of all the calls to `cuc'` to see what they "really are"?

balacij:
Returning to this: Brilliant find, @samm82 !

I need to return to this soon. For now, some notes:

We recently discussed stateful chunk smart constructors that only create chunks relative to a chunk database. If we have this, we might be able to erase all manually written `UID`s in favour of ones randomly assigned by the chunk database on registration.

Pseudocode:
```haskell
type Chunk a = State ChunkDB a

unassignedUID :: UID
unassignedUID = error "A chunk was used before insertion into the ChunkDB!"

insert :: HasUID a => a -> Chunk a
insert chunkWithInvalidUID = do
  nextUID <- generateUID
  let brandedChunk = chunkWithInvalidUID & uid .~ nextUID
[...]

PR #4079, opened 2025-05-23
Remove stale wiki page documenting `UnitalChunk` usage

balacij:
This wiki page appears to have been created specifically for redesign ticket (#3199) which was closed out.


JacquesCarette:
Some times wiki pages were created exactly to document things that would otherwise be lost on closed tickets / PRs! Can you double-check that this information is more than 'stale' but is actively misleading? "Stale" is too low a bar for deletion, those pages should be updated. "Wrong" and "useless" are different matters.

balacij:
That wiki page appears to have served its purpose -- it also brought about a really interesting ticket -- #3238 . For any future redesign of UnitalChunk (which I think should happen), I don't think there's any utility in referring to the wiki page, because it would all need to be double-checked and any "real" analysis done by anyone else would probably need them to interact with the code actively -- in which case, they should be able to retrieve the same information using `rg` or `grep`. With `rg`:

```console
λ  rmUCWikiPage* rg "cuc'[^']" -ths --stats
drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
186:pendDisAngle = cuc' "pendDisAngle"

drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
415:tempW = cuc' "tempW"
422:tempPCM = cuc' "tempPCM"
429:watE = cuc' "watE" (nounPhraseSP "change in heat energy in the water")
435:pcmE = cuc' "pcmE" (nounPhraseSP "change in heat energy in the PCM")

drasil-lang/lib/Language/Drasil.hs
135:  , cuc, cvc, constrained', cuc', cuc'', constrainedNRV'

drasil-lang/lib/Language/Drasil/Chunk/UncertainQuantity.hs
11:import Language.Drasil.Chunk.Constrained (ConstrConcept(..), ConstrainedChunk, cuc', cnstrw, cvc)
120:uqc nam trm desc sym un space cs val = uq (cuc' nam trm desc sym un space cs val)
125:uqcND nam trm sym un space cs val = uq (cuc' nam trm "" sym un space cs val)

drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs
9:  cnstrw', constrained', constrainedNRV', cuc', cuc'') where
69:-- | Creates a constrained unitary chunk from a 'UID', term ('NP'), 'Symbol', 'Space', 'Constraint's, and a 'Maybe' 'Expr' (Similar to 'cuc' but no units).
127:cuc' :: (IsUnit u) => String -> NP -> String -> Symbol -> u
129:cuc' nam trm desc sym un space cs rv =


13 matches
13 matched lines
5 files contained matches
455 files searched
1146 bytes printed
2793463 bytes searched
0.003628 seconds spent searching
0.044862 seconds
```

JacquesCarette (approved):

PR #4082, opened 2025-05-27
Remove outdated Typed-Exprs.md wiki page, and add reference documents to comments

balacij:
Rationale for the design is somewhat unnecessary for users -- irrelevant to the wiki of Drasil. Maybe we should instead move it to another wiki page? For example, the DrasilOrg/Notes repo might be a good candidate to move this information if we wanted to keep it.


JacquesCarette:
The wiki could be split into user-oriented information and requirements+design information. Or the latter could be moved to `DrasilOrg/Notes`. I agree, not deleted.

PR #4088, opened 2025-05-30
(dbl|sgl)pend: Avoid duplicate `UID`s in `A:startOrigin`

balacij:
Two concept instances contained the same `UID`. ~~They are carefully only used in SglPend and DblPend, but it should be changed regardless.~~

Contributes to #4036


code/drasil-example/dblpend/lib/Drasil/DblPend/Assumptions.hs
```diff
@@ -1,34 +1,30 @@
...
-twoDMotion, cartSys, cartSysR, yAxisDir, startOriginSingle, startOriginDouble,
+twoDMotion, cartSys, cartSysR, yAxisDir, startOriginDouble,
   firstPend, secondPend:: ConceptInstance 
 
 twoDMotion        = cic "twoDMotion"    twoDMotionDesc          "twoDMotion"    assumpDom
 cartSys           = cic "cartSys"       cartSysDesc             "cartSys"       assumpDom
 cartSysR          = cic "cartSysR"      cartSysRDesc            "cartSysR"      assumpDom
 yAxisDir          = cic "yAxisDir"      yAxisDirDesc            "yAxisDir"      assumpDom
-startOriginSingle = cic "startOrigin"   startOriginDescSingle   "startOrigin"   assumpDom
-startOriginDouble = cic "startOrigin"   startOriginDescDouble   "startOrigin"   assumpDom
```
JacquesCarette, line 31: wow, what a clear instance of bad duplicate UID!

code/drasil-example/dblpend/lib/Drasil/DblPend/Assumptions.hs
```diff
@@ -1,34 +1,30 @@
...
+startOriginDouble = cic "startOrigin2x" startOriginDescDouble   "startOrigin"   assumpDom
```
JacquesCarette, line 27: This change ("startOrigin" to "startOrigin2x") creates a lot of noise in the output. Which is sad, as this means that UIDs are used for way way too much. But let's go with that for now, as at least it's clear what happened.

JacquesCarette (approved):

Issue #2468, opened 2021-05-14
We shouldn't sort references by UID


JacquesCarette:
The internal UID isn't necessarily stable or meaningful. We should probably sort by visible label instead.

PR #4111, opened 2025-06-03
Give all ConceptInstances a unique UID.

balacij:
Contributes to #2873

Clears out a few duplicate `UID`s for GamePhysics.

The massive number of files changed is because:

1. We check in `.dot` files based on `UID`s.
2. We abuse `UID`s for sorting (#2468).


smiths:
I went through the changes to stable and everything looks fine from my point of view.  Some of the items in the traceability lists swapped order, but nothing was removed or added.

code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
```diff
@@ -50,4 +50,5 @@ cw c = ConDict (nw c) (c ^. defn) (cdom c)
 -- Reference Address ('String'), a definition ('Sentence'), 
 -- a short name ('String'), and a domain (for explicit tagging).
 cic :: Concept c => String -> Sentence -> String -> c -> ConceptInstance
-cic u d sn dom = ConInst (ccs (nc u $ pn sn) d [dom]) u $ shortname' (S sn)
+cic u d sn dom = ConInst (nsUid "instance" $ icc ^. uid) icc u $ shortname' (S sn)
```
JacquesCarette, line 53: This is an (understandable) hack that I'd like to have disappear over time. It does get around the deeper bug that `cic` is abusing the UID notion. So the hack is an improvement in that sense.

code/drasil-lang/lib/Language/Drasil/Chunk/Concept/Core.hs
```diff
@@ -54,13 +54,16 @@ instance ConceptDomain ConceptChunk where cdom = cdom'
 -- Ex. Something like the assumption that gravity is 9.81 m/s. When we write our equations,
 -- we can then link this assumption so that we do not have to explicitly define
 -- that assumption when needed to verify our work.
-data ConceptInstance = ConInst { _cc :: ConceptChunk , ra :: String, shnm :: ShortName}
+data ConceptInstance = ConInst { _ciuid :: UID
```
JacquesCarette, line 57: Ah, and this is a valuable step forward. If an "instance" really is that, then indeed it should *refer* to a `Concept` and not contain it.

PR #4116, opened 2025-06-04
PDController: Give `processVariable` system goal a unique UID.

balacij:
Contributes to #2873

Previously, it collided with `Concepts.hs` L62-L64:

```haskell
processVariable
  = dcc "processVariable" (nounPhraseSP "Process Variable")
      "The output value from the power plant"
```

PR #4117, opened 2025-06-04
PDController: Give `setPoint` system assumption a unique UID.

balacij:
Builds on #4116

Previously, it collided with `Concepts.hs` L70-73:

```haskell
setPoint
  = dcc "setPoint" (nounPhraseSP "Set-Point")
      ("The desired value that the control system must reach. This also knows "
         ++ "as the reference variable")
```


smiths:

JacquesCarette (changes requested):
This needs a merge - and probably should not have been built on top of the previous PR but as separate.

balacij:
It's dependant because they both affect `stable/`. Do you want me to de-couple them? Or for next time (seeing as you already accepted the dependency PR)?

JacquesCarette:
For next time, do indeed decouple.

JacquesCarette (approved):

JacquesCarette:
For some reason, my re-review isn't good enough for github -- @smiths can you do the honours please?

smiths (approved):

smiths (dismissed):
The system says my previous approval was stale, so I'm trying again.

JacquesCarette:
Why did you self-request a review @smiths *after* you reviewed it? Now it's blocked on that!

smiths:
GitHub isn't accepting my approval either.  It says my review and that of @JacquesCarette are stale because they were submitted before the merge base changed.  I don't understand, since I literally just approved it?

balacij:
Screenshot 2025-06-04 at 5 08 12 PM

That's the first time I've ever seen this!

JacquesCarette:
> The merge-base changed after approval.

So we shouldn't do that...

PR #4121, opened 2025-06-04
Differentiate between second (unit) and second (order) UIDs

### LINKED ISSUE #4109: Cannot find chunk after changing UID
I was attempting to change a UID in order to fix an issue directly related to #4036, however I ran into a strange error. When changing an `IdeaDict` with the UID `second` defined in `doccon`, I get the following output:
```console
=====[ Finished type checking ]=====
'requirement' is inserted twice in 'TermMap'!
dblpend.EXE: Term: second not found in TermMap
  error, called at lib\Database\Drasil\ChunkDB.hs:125:24 in drasil-database-0.1.1.0-4jdMY7TWows9S8CmX9loPI:Database.Drasil.ChunkDB
make: *** [Makefile:311: dblpend_gen] Error 1
```
Stating that `second` can no longer be found in the `TermMap`. Here is the relevant code:

`second` definition https://github.com/JacquesCarette/Drasil/blob/74b4803c9e91668eb41c696fae15a8fb4ceb6410/code/drasil-data/lib/Data/Drasil/Concepts/Documentation.hs#L207

Where the error is called: https://github.com/JacquesCarette/Drasil/blob/74b4803c9e91668eb41c696fae15a8fb4ceb6410/code/drasil-database/lib/Database/Drasil/ChunkDB.hs#L122-L125

For reference, I changed the `second` definition to the following:
```haskell
second_         = nc "ordSecond"      (cn'    "second"             ) 
```

I imagine that this is caused by the UID being searched directly somewhere, rather than just using it as unique label. The fix will involve finding where this may be happening and referencing the chunk's UID, not just searching for the UID itself.

jackwyand:
Closes #4109, contributes to #4036

Unblocks an issue related to trying to fix a duplicate UID, ensuring that the correct one is being found in the chunk database (either the unit or the ordering)

### DIFF
--- a/code/drasil-data/lib/Data/Drasil/SI_Units.hs
+++ b/code/drasil-data/lib/Data/Drasil/SI_Units.hs
@@ -21,28 +21,28 @@ siUnits = map unitWrapper fundamentals ++ map unitWrapper derived
 -- * Fundamental SI Units
 
 metre, kilogram, second, kelvin, mole, ampere, candela :: UnitDefn
-metre    = fund "metre"    "length"               "m"
-kilogram = fund "kilogram" "mass"                 "kg"
-second   = fund "second"   "time"                 "s"
-kelvin   = fund "kelvin"   "temperature"          "K"
-mole     = fund "mole"     "amount of substance"  "mol"
-ampere   = fund "ampere"   "electric current"     "A"
-candela  = fund "candela"  "luminous intensity"   "cd"
+metre    = fund "unit:metre"    "length"               "m"
+kilogram = fund "unit:kilogram" "mass"                 "kg"
+second   = fund "unit:second"   "time"                 "s"
+kelvin   = fund "unit:kelvin"   "temperature"          "K"
+mole     = fund "unit:mole"     "amount of substance"  "mol"
+ampere   = fund "unit:ampere"   "electric current"     "A"
+candela  = fund "unit:candela"  "luminous intensity"   "cd"
 
 -- * Commonly Defined Units
 
 degree :: UnitDefn --FIXME: define degree in terms of radians and pi
 -- degree = UD (dcc "degree" (cn' "degree") "angle") (BaseSI (US [(Special Circle,1)])) ["degree"]
-degree = fund' "degree" "angle" (Special Circle)
+degree = fund' "unit:degree" "angle" (Special Circle)
 
 -- Some of these units are easiest to define via others less common names, 
 -- which we define first.
 s_2 :: UnitDefn
-s_2 = newUnit "seconds squared" $ second ^: 2
+s_2 = newUnit "unt:seconds squared" $ second ^: 2
 
 m_2, m_3 :: UnitDefn
-m_2 = newUnit "square metres"   $ metre ^: 2
-m_3 = newUnit "cubic metres"    $ metre ^: 3
+m_2 = newUnit "unit:square metres"   $ metre ^: 2
+m_3 = newUnit "unit:cubic metres"    $ metre ^: 3
 
 -- And now for the ones with 'common' names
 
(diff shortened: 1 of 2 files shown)


code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
```diff
@@ -143,13 +143,16 @@ symbolsAll = symbols ++ scipyODESymbols ++ osloSymbols ++ apacheODESymbols ++ od
...
-   map nw physicscon ++ concepts ++ map nw physicalcon ++ map nw acronyms ++ map nw symbols ++ map nw [metre, hertz] ++
-   [nw algorithm] ++ map nw compcon ++ map nw educon ++ map nw prodtcon)
-  srsDomains (map unitWrapper [metre, second, newton, kilogram, degree, radian, hertz])
-  dataDefs iMods genDefns tMods concIns [] allRefs citations
+  (nw newtonSLR : nw progName : nw mass : nw len : nw kilogram : nw inValue
+   : nw newton : nw degree : nw radian : nw unitVect : nw unitVectj
+   : [nw errMsg, nw program] ++ map nw symbols ++ map nw doccon
+   ++ map nw doccon' ++ map nw physicCon ++ map nw mathcon ++ map nw mathcon'
+   ++ map nw physicCon' ++ map nw physicscon ++ concepts ++ map nw physicalcon
+   ++ map nw acronyms ++ map nw symbols ++ map nw fundamentals
```
jackwyand, line 151: This diff seems large because I reformatted this code but all that was changed is just adding `fundamentals`.

code/drasil-data/lib/Data/Drasil/SI_Units.hs
```diff
@@ -21,120 +21,120 @@ siUnits = map unitWrapper fundamentals ++ map unitWrapper derived
...
 degree :: UnitDefn --FIXME: define degree in terms of radians and pi
 -- degree = UD (dcc "degree" (cn' "degree") "angle") (BaseSI (US [(Special Circle,1)])) ["degree"]
-degree = fund' "degree" "angle" (Special Circle)
+degree = fund' "unit:degree" "angle" (Special Circle)
 
 -- Some of these units are easiest to define via others less common names, 
 -- which we define first.
 s_2 :: UnitDefn
-s_2 = newUnit "seconds squared" $ second ^: 2
+s_2 = newUnit "unt:seconds squared" $ second ^: 2
```
JacquesCarette: typo `unt:`

code/drasil-data/lib/Data/Drasil/SI_Units.hs
```diff
@@ -21,120 +21,120 @@ siUnits = map unitWrapper fundamentals ++ map unitWrapper derived
...
 
 metre, kilogram, second, kelvin, mole, ampere, candela :: UnitDefn
-metre    = fund "metre"    "length"               "m"
-kilogram = fund "kilogram" "mass"                 "kg"
-second   = fund "second"   "time"                 "s"
-kelvin   = fund "kelvin"   "temperature"          "K"
-mole     = fund "mole"     "amount of substance"  "mol"
-ampere   = fund "ampere"   "electric current"     "A"
-candela  = fund "candela"  "luminous intensity"   "cd"
+metre    = fund "unit:metre"    "length"               "m"
```
balacij: Wouldn't we be able to automate this by altering `fund`?

JacquesCarette: Yes! Well spotted.

jackwyand:
I implemented some changes which I think align with what @balacij suggested, feedback would be appreciated :)

JacquesCarette (approved):
Yes, that's good.

PR #4124, opened 2025-06-04
Cleanup GlassBR duplicate UIDs

jackwyand:
Contributes to #4036 

~~I _think_ with this all of the GlassBR duplicates have been resolved.~~
Edit: Not quite, but getting close!

### DIFF
--- a/code/drasil-example/dblpend/lib/Drasil/DblPend/Concepts.hs
+++ b/code/drasil-example/dblpend/lib/Drasil/DblPend/Concepts.hs
@@ -1,7 +1,6 @@
 module Drasil.DblPend.Concepts where
 
 import Language.Drasil
-import Data.Drasil.Domains (physics)
 import Data.Drasil.Concepts.Documentation (first, second_, object)
 import Data.Drasil.Theories.Physics (newtonSLRQD)
 import Language.Drasil.Chunk.Concept.NamedCombinators (compoundNC)
--- a/code/drasil-example/dblpend/package.yaml
+++ b/code/drasil-example/dblpend/package.yaml
@@ -11,6 +11,10 @@ language: Haskell2010
 default-extensions:
 - StrictData
 
+ghc-options:
+- -Wall
+- -Wredundant-constraints
+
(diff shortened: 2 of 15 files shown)


code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -235,7 +235,7 @@ sflawParamM = unitary "sflawParamM" (nounPhraseSP "surface flaw parameter") --pa
 {-Quantities-}
 
 unitless :: [QuantityDict]
-unitless = [riskFun, isSafePb, isSafeProb, isSafeLR, isSafeLoad,
+unitless = [riskFun, isSafeProb, isSafeLoad,
```
JacquesCarette: But `isSafePb` and `isSafeLR` *are* unitless. So if this list is used for anything other than populating `ChunkDB`, things will go wrong.

Did we fix one bug just to have it show up somewhere else later? [Maybe not, it depends on what the point of this list ist.]

JacquesCarette (commented):
There should a 'question that need answered first' as a review - comment will do.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
jackwyand: At the moment it is just added to the ChunkDB and not used elsewhere. The original purpose of the list I am not sure of, but I do see how removing things could cause issues. It does complicate fixing the bug however, since the other place they exist is in `outputs` which I imagine would have the same issue?

JacquesCarette: So if we don't use `unitless` for anything, then I would prefer to delete that list altogether, rather than having it as a ticking time-bomb. We still need to insert the pieces into the ChunkDB, just not via a named list.

JacquesCarette:
I will wait until the build comes back clean before merging.

PR #4131, opened 2025-06-06
Ensure `-Wall` is is still enabled when stackArgs is passed in to Makefile.

### LINKED ISSUE #4130: CI no longer fails when GHC warnings appear in the build
As per the title.

balacij:
Closes #4130

The CI _should_ fail if I did this right.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -162,8 +162,8 @@ PROFEXEC := --profile --rts-options '-xc -P'
 # GHC build options
 GHCTHREADS ?= 2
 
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
let's see if it does! And it did, actually. So I should not merge this?

balacij:
We should only merge this after #4132 and #4133 are merged into `main` and this has `main` pulled back in to verify that a build on `main` won't break. If we merge this as-is, then we won't be able to update the website, and our CI would just fail for all branches.

JacquesCarette:
Those two are merged now.

PR #4139, opened 2025-06-06
`termResolve`: Generate synthetic `IdeaDict`s to fill in the gaps of unregistered `IdeaDict`s

balacij:
Contributes to #4126

### DIFF
--- a/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
+++ b/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
@@ -130,7 +130,20 @@ symbResolve m x = uMapLookup "Symbol" "SymbolMap" x $ symbolTable m
 
 -- | Looks up a 'UID' in the term table from the 'ChunkDB'. If nothing is found, an error is thrown.
 termResolve :: ChunkDB -> UID -> IdeaDict
-termResolve m x = uMapLookup "Term" "TermMap" x $ termTable m
+termResolve m x -- = uMapLookup "Term" "TermMap" x $ termTable m
+  | (Just (t, _)) <- Map.lookup x (termTable m) = t
+  -- SUPER HACK: The extras below allow any 'DataDefinition' or 'QuantityDict'
+  -- or ... to have their terms auto-extracted and used. Generates a fake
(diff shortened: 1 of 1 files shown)


balacij:
With #4080, we now have a use of `termResolve` that also needs the abbreviation, but I agree with you -- it returning a faked `IdeaDict` is icky. Should we make it return a tuple instead? It's better than returning an `IdeaDict`, but seems only marginally better.

Another idea could be that it also takes a function as input that would process the 'term' and 'maybe abbreviation' that the `IdeaDict` would contain to produce whatever thing necessary. e.g.,

```haskell
termResolve :: ChunkDB -> UID -> IdeaDict

termResolve' :: (NP -> Maybe String -> c) -> ChunkDB -> UID -> c
```

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -128,22 +129,27 @@ uMapLookup tys ms u t = getFM $ Map.lookup u t
...
-  | (Just (t, _)) <- Map.lookup x (symbolTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (conceptChunkTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_unitTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_dataDefnTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_insmodelTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_gendefTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_theoryModelTable m) = nw t
-  | (Just (t, _)) <- Map.lookup x (_conceptinsTable m) = nw t
-  | otherwise = error $ "Term: " ++ show x ++ " not found in TermMap"
+data TermAbbr = TermAbbr { longForm :: NP, shortForm :: Maybe String }
```
balacij, line 132: I left this in `ChunkDB.hs` for now because this is where the 'resolver's live.

JacquesCarette (approved):

Issue #4097, opened 2025-06-02
`Quantity` Typeclass


balacij:
<img width="1056" alt="Image" src="https://github.com/user-attachments/assets/6b3b10f0-55e8-4649-ae81-385a73ded123" />

Immediate question: The note in the above mentions that adding `MayHaveUnit` would cause an import cycle and/or orphaned instances, but thinking about it, I don't quite see how this could happen. This could be attempted again to clarify what's happening.

`Quantity` is a typeclass that indicates some chunk is 'quantity-like', including:

1. `QuantityDict`
2. `ConstrainedChunk`
3. `ConstrConcept`
4. `QDefinition`

Now, immediately two questions (to look into):

1. Why is `QDefinition` an instance of it? A `QDefinition` only provides a formula for one such chunk that would satisfy `Quantity`. It is not a quantity itself.
2. Why isn't `DefinedQuantityDict` an instance of it? `ConstrConcept` is an instance of it, and its instance of all the requirements are based on the `DefinedQuantityDict`.

Designwise, I know we've discussed that an empty typeclass is a "smell" of sorts. Here, I wonder if it questions if `HasSpace` and `HasSymbol` are worth defining on their own at all:

```console
λ rg "instance\s+HasSymbol" -ths --sort-files -l
drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
drasil-code/lib/Language/Drasil/Chunk/NamedArgument.hs
drasil-code/lib/Language/Drasil/Chunk/Parameter.hs
drasil-lang/lib/Language/Drasil/Chunk/CodeVar.hs
drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs
drasil-lang/lib/Language/Drasil/Chunk/DefinedQuantity.hs
drasil-lang/lib/Language/Drasil/Chunk/Eq.hs
drasil-lang/lib/Language/Drasil/Chunk/Quantity.hs
drasil-lang/lib/Language/Drasil/Chunk/UncertainQuantity.hs
drasil-lang/lib/Language/Drasil/Chunk/Unital.hs
drasil-lang/lib/Language/Drasil/Chunk/Unitary.hs
drasil-theory/lib/Theory/Drasil/MultiDefn.hs
```

```console
λ rg "instance\s+HasSpace" -ths -l --sort-files
drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
[...]

PR #4140, opened 2025-06-06
Remove Quantity typeclass instance of QDefinition.

balacij:
Contributes to #4097

A `QDefinition` is not a `Quantity`. A `QDefinition` only pairs a 'quantity' with a 'defining formula'.


JacquesCarette:
So what do you think a `Quantity` *is*? I agree that a `QDefinition` pairs a 'quantity' with a 'defining formula'. The real question: is the 'defining formula' a mere annotation, thus making a `QDefinition` a `Quantity`, or something entirely different, thus changing the very nature of `QDefinition`?

balacij:
I finished the work, and I've also marked the PR as "ready for review" so that the CI can run. However, the PR is not quite "ready to merge" because your comment has me second-guessing.

I see your point, and I subscribe to [your view](https://github.com/JacquesCarette/Drasil/issues/4097#issuecomment-2931055470) on what a 'Quantity' is.

Brain dump: I think of 'QDefinition's as specific solutions to calculating a quantity. i.e., its nature is about calculation, not about defining a symbol. If a 'QDefinition' were a 'Quantity', then could we say that a 'QDefinition' is equal to a 'UnitalChunk' about them being 'Quantities'? Would that relate to them having the same symbol, UID, space, formula (the 'UnitalChunk' doesn't have a formula)?

If a 'QDefinition' provides an approximate solution to a QuantityDict, does it change the definition of the QuantityDict? Of course, I'm not talking about mutation here, but about a cloned QuantityDict that explicitly mentions it is approximated, potentially with a slightly different symbol.

I think we can also look at the "names" (I'll refrain from saying "terms" here) of these things. Wouldn't we want the names of the QuantityDict to be different from the QDefinition? We currently have this in Drasil with our IMs showing "Calculation of $x$" -- that is information embedded in the QDefinition, arguably hacked in through its internal DefinedQuantityDict and sometimes overridden by the carrier EquationalModel. For example, Projectile has:

```haskell
timeIM :: InstanceModel
timeIM = imNoRefs (equationalModelN (nounPhraseSP "calculation of landing time") timeQD)
  [qwC launSpeed $ UpFrom (Exc, exactDbl 0)
  ,qwC launAngle $ Bounded (Exc, exactDbl 0) (Exc, half $ sy pi_)]
  (qw flightDur) [UpFrom (Exc, exactDbl 0)]
  (Just timeDeriv) "calOfLandingTime" [angleConstraintNote, gravitationalAccelConstNote, timeConsNote]
```

Overall, I still tend towards a 'QDefinition' not being a 'Quantity', but I'm not entirely convinced anymore. What do you think?

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
```diff
@@ -374,7 +374,7 @@ coilHTC = uqc "coilHTC" (nounPhraseSP
   (sub (eqSymb htTransCoeff) lCoil)
   UT.heatTransferCoef Real
   [gtZeroConstr,
-  sfwrRange $ Bounded (Inc, sy coilHTCMin) (Inc, sy coilHTCMax)] (exactDbl 1000) defaultUncrt
+  sfwrRange $ Bounded (Inc, sy $ coilHTCMin ^. defLhs) (Inc, sy $ coilHTCMax ^. defLhs)] (exactDbl 1000) defaultUncrt
```
balacij, line 378: I like that that `^. defLhs` is appearing here. But I'm confused now: Should this quantity ~~definition~~ _declaration_ be dependant on `QDefinition`s?

code/drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
```diff
@@ -57,9 +58,12 @@ instance DefiningCodeExpr CodeDefinition where codeExpr = def
 qtoc :: (Quantity (q Expr), MayHaveUnit (q Expr), DefiningExpr q) => q Expr -> CodeDefinition
 qtoc q = CD (codeChunk $ quantfunc q) (expr $ q ^. defnExpr) [] Definition
 
+qtoc' :: QDefinition Expr -> CodeDefinition
```
JacquesCarette: I don't like this signature. The previous one was nicely polymorphic, this one is monomorphic. That's a regression.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -193,13 +193,13 @@ oldcodeSpec SI.SI{ SI._sys = sys
       const' = map qtov (filter ((`Map.notMember` conceptMatch (maps chs)) . (^. uid))
         cnsts)
       derived = map qtov $ getDerivedInputs ddefs inputs' const' db
-      rels = (map qtoc (getEqModQdsFromIm ims ++ mapMaybe qdEFromDD ddefs) \\ derived)
+      rels = (map qtoc' (getEqModQdsFromIm ims ++ mapMaybe qdEFromDD ddefs) \\ derived)
```
JacquesCarette: This is kind of good - not the `qtoc'` call per se, but the fact that the information is actually coming from IMs and DDs. What's wrong is how the code is extracting the information. This seems to indicate we have the 'wrong' lenses for that purpose.

JacquesCarette (changes requested):
I think this PR has some good things, and some dubious things.

I think part of the issue is that we don't know what the `Quantity` constraint *means*. Does it mean "is a quantity"? Or does it mean "represents a quantity"? Nor do we really know what `QDefinition` is supposed to embody. So it is really hard to know of those instances should be there or not.

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
balacij: It would be nice to remove `getEqModQdsFromIm` and `qdEFromDD`. We could expose a type parameter that indicates what kind of model the IM is (e.g., `InstanceModel Definitional`), this way we could have a lens that operates on `X Definitional`s to get their `QDefinition`s.

Alternatively, we can really redesign what's going on here. It might be nice to see a 'system of equations' chunk that we gradually build by creating IMs.

code/drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
balacij: Yeah, `qtoc'` was a bit unnecessary. Fixed in b1dca9e50a046fce43c0806cf619c66a0054d6f0

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
balacij: No, it shouldn't. `coilHTCMin` is a `QDefinition` that deeply nests a `QuantityDict` within it, which isn't accessible elsewhere. I think we need to de-nest these things. Before I start that, @JacquesCarette, do you agree?

```haskell
coilHTCMin = mkQuantDef (unitary' "coilHTCMin"
  (nounPhraseSP $ "minimum convective heat " ++
  "transfer coefficient between coil and water")
  (staged (supMin (eqSymb coilHTC)) (subMin (eqSymb coilHTC))) 
  UT.heatTransferCoef Real) $ exactDbl 10
```

code/drasil-code/lib/Language/Drasil/CodeSpec.hs
JacquesCarette: > It would be nice to remove getEqModQdsFromIm and qdEFromDD

Why? I agree they shouldn't return chunks, but otherwise, this seems like a good way to extract information.

Yes, it would be "icing on the cake" if we could use Haskell's type system to prevent us from grabbing the wrong thing. But that's an optimization.

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
JacquesCarette: That all the information from a `QuantityDict` is *available* from a `QDefinition` seems quite reasonable to me. How that is encoded, that is a different question. So it depends on what we mean by a `QDefinition`, i.e. what is it meant to represent.

balacij:
> [I have no idea why you're talking about 'approximation' above.]

Trying to explain why I was talking about 'approximation' shows me that I was just confused. We can safely ignore that question.

> Whereas I suspect that the real problem is that we are not capturing the information that we need "the right way".

I certainly agree with that, too.

JacquesCarette:
'5 apples' is a quantity (of apples); '(n+k+3) apples' represents a quantity of apples. One is immediate knowledge, one is future knowledge. That's not the only difference, but it's a big one.

Whenever you're not sure between 'is' and 'represents', go stare at [This painting](https://en.wikipedia.org/wiki/The_Treachery_of_Images).

> I'm not sure if I like the existence of Quantity at all. Why shouldn't quantities (variables) always be represented by the same type (QuantityDict at the moment, but DefinedQuantityDict preferred because it has the prose descriptions as well)?

Because whatever a 'Quantity' is, it should be a promise of certain pieces of information being available. The exact representation should not matter. That a particular representation may in fact contain *additional* information most definitely should not matter.

Our important internal concepts should be 'information promises' and not concrete representations. 

In all cases, the most important thing (whether it's `QDefinition`, `DefinedQuantityDict` or `DataDefinition`) is "what information does this structure promise to provide". At use-site, we then can use exactly those items that contain at least all the information that is needed to function properly.

balacij:
We can safely ignore the above commentary now.

From one/two meetings ago, we agreed that:

1. `QDefinition` should instantiate the `Quantity` typeclass. It does represent a quantity.
2. Adding differentiation between definitional formulas (`::=`) and alternative formulas (`=`) might be fruitful.

I will leave this PR up for a little bit to see if there's any immediate work that can be done.

PR #4142, opened 2025-06-07
Avoid registering IdeaDicts extracted from theories in the ChunkDB.

balacij:
Contributes to #4126 

~~Builds on https://github.com/JacquesCarette/Drasil/pull/4139~~

What's odd to me about this is that only these examples had their theories registered in the `ChunkDB`! Whatever is going on here needs further investigation (at least before we merge this PR).

(I'm only marking this PR as "ready for review" to get the CI to run.)


JacquesCarette (commented):
I'm neither approving nor requesting changes per se. I think more thinking is needed to figure out what to do here.

balacij:
Sorry, my title might have been misleading. Theories are still registered in the ChunkDB. It's only the 'extracted/duplicated IdeaDicts' that this PR removes from registration -- #4139 makes it an unnecessary step to register these.

JacquesCarette:
Very misleading title indeed! Ok, I'll re-review with that in mind.

JacquesCarette (approved):

PR #4143, opened 2025-06-09
Only add needed concepts to GlassBR ChunkDB

jackwyand:
Contributes to #4036 

Resolves some duplicates that were reintroduced in #4141. It feels wrong to add them directly this way (rather we should be able to give the ChunkDB all of the concepts and not have to worry as mentioned in #2873 [giving the ChunkDB arbitrary data]) however the current design of 'projecting' `Ideas` into other chunks directly prevents this. For example one the duplicates removed by this, `stdOffDist`, is in `con` (the list removed) but is also used to define `sD`, which is in `terms`. Since they share a UID the warning is caused. Not to mention `sD` is then used for `standOffDist`, making it 3 chunks sharing a UID.

https://github.com/JacquesCarette/Drasil/blob/5df7609d06a6dd11ffb501ba21ab07314f641759/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L382-L384

https://github.com/JacquesCarette/Drasil/blob/5df7609d06a6dd11ffb501ba21ab07314f641759/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L110-L112


code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
```diff
@@ -138,7 +138,7 @@ ideaDicts =
   -- IdeaDicts
   [sciCompS, lateralLoad, materialProprty] ++ con' ++ doccon ++ educon ++ compcon ++
   -- CIs
-  nw progName : map nw doccon' ++ map nw mathcon' ++ map nw con ++
+  nw progName : nw iGlass : nw lGlass : map nw doccon' ++ map nw mathcon' ++
```
balacij: For the first 3, you should be able to write `map nw [progName, iGlass, lGlass]` if they're all CIs.

JacquesCarette (approved):
I'm fine with being incrementally better, even if some intermediate steps are sub-optimal. This lets us move forward while gaining a better understanding of the larger problem.

PR #4144, opened 2025-06-09
Remove hack for automatically extracting IdeaDicts from DataDefinitions.

balacij:
Contributes to #4126 

This hack was previously there because our data definitions have unique UIDs, unlike the majority of our chunks at the moment. So extracting `IdeaDict`s was necessary for the `termResolve` function to find an `IdeaDict`.

This change is possible because https://github.com/JacquesCarette/Drasil/commit/ac3ec2db97a96d39107849c89f7118715bcffecf makes the use of the "term map" really about "getting terms" rather than `IdeaDict`s.


JacquesCarette:
And the benefits of the slow and steady approach are making themselves clear.

JacquesCarette (approved):

PR #4154, opened 2025-06-10
Add only unique symbols from GlassBR `inputs` and `outputs` to ChunkDB

jackwyand:
Contributes to #4036

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
@@ -151,9 +151,8 @@ ideaDicts =
 tableOfAbbrvsIdeaDicts :: [IdeaDict]
 tableOfAbbrvsIdeaDicts =
   -- CIs
-  map nw acronyms ++
+  map nw acronyms
   -- QuantityDicts
-  map nw thisSymbols
 
 symbMap :: ChunkDB
 symbMap = cdb thisSymbols ideaDicts
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs
@@ -1,20 +1,23 @@
 module Drasil.GlassBR.Symbols where
 
-import Language.Drasil (QuantityDict, qw)
+import Language.Drasil (QuantityDict, qw, cnstrw)
 import Language.Drasil.Code (Mod(Mod), asVC)
 
 import Drasil.GlassBR.ModuleDefs (allMods, implVars)
-import Drasil.GlassBR.Unitals (inputs, outputs, specParamVals, symbols,
-  symbolsWithDefns, tmSymbols, interps, derivedInputDataConstraints, unitless)
+import Drasil.GlassBR.Unitals (specParamVals, symbols, symbolsWithDefns,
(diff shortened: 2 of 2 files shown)


code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
```diff
@@ -151,9 +151,8 @@ ideaDicts =
 tableOfAbbrvsIdeaDicts :: [IdeaDict]
 tableOfAbbrvsIdeaDicts =
   -- CIs
-  map nw acronyms ++
+  map nw acronyms
   -- QuantityDicts
```
JacquesCarette: Shouldn't you also remove the comment, as now it is dangling?

code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs
```diff
@@ -1,20 +1,23 @@
...
+import Drasil.GlassBR.Unitals (specParamVals, symbols, symbolsWithDefns,
+  tmSymbols, interps, derivedInputDataConstraints, unitless, probBr,
+  stressDistFac, nomThick, sdVector, inputsWUnitsUncrtn, inputsWUncrtn, glassTypeCon)
 
 import Data.List ((\\))
 
 symbolsForSymbolTable :: [QuantityDict]
 symbolsForSymbolTable = symbolsForTermTable ++ map qw symbols ++
-  unitless ++ map qw derivedInputDataConstraints
+  unitless ++ map qw [probBr, stressDistFac, nomThick, cnstrw glassTypeCon] ++
```
JacquesCarette, line 15: I'm a bit surprised: here you are adding more things to `symbolsForSymbolTable` than before -- how is this helping? [Genuine question!]

JacquesCarette (changes requested):

code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs
jackwyand: The reason I added things to `symbolsForSymbolTable` is because they needed to be included in the `SymbolMap` but _not_ the `TermMap`. Since some required symbols for the `SymbolMap` were being added from `symbolsForTermTable`, removing them from the `TermMap` meant they had to be added back to the `SymbolMap`.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
jackwyand: I only left it there in case something were to be added in the future but I guess that doesn't really follow how we have organized everything else so I will remove it.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs
jackwyand: Just to confirm however, they both need the removed symbols, it's just that the `TermMap` was already getting them from elsewhere so they were being duplicated. The `SymbolMap` however, was not, so they were added back as you can see.

JacquesCarette: Perfect answer, thank you.

JacquesCarette (approved):

PR #4156, opened 2025-06-11
Add documentation to explain `short` vs `getA` vs `abrv`

### LINKED ISSUE #4081: `getAcc` versus `short`. What is the difference?
While investigating #4008 I ran into an interesting issue. When creating the SRS documents we often use abbreviations/acronyms, however we seem to get these in _at least_ two different ways. First, we use `getAcc` which gets an "abbreviation in 'Sentence' form from a 'CI'." https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-lang/lib/Language/Drasil/Chunk/CommonIdea.hs#L51-L53 It does this through the `abrv` function. However, we also have `short` which uses the `Ch` constructor for a `Sentence`, and uses the `ShortStyle` style which matches in `spec` to run `lookupS`. https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs#L22-L24 https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-lang/lib/Language/Drasil/Sentence.hs#L100-L101 `lookupS` "Look[s] up the acronym/abbreviation of a term. Otherwise returns the singular form of a term. Takes a chunk database and a 'UID' associated with the term." https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-printers/lib/Language/Drasil/Printing/Import/Helpers.hs#L65-L68 To me it seems that they are very similar but only just because they accomplish seemingly the same thing, which I understand does not make them necessarily identical. I am wondering what the difference is (if any). The reason I ask this is because of my implementation for #4008 (see #4080) had an interesting bug where it would only work on some acronyms, and this seemed to be the root cause. Using `getAcc` does not create a `Sentence` that has `ShortStyles` associated with it, which is what I am using to add the tooltips.

I also did a bit of searching of past issues and it seems that `getAcc` is not the only function that may be affecting things, as similar ones such as `getA` or `getAccStr` also exist. I will need to do more investigation on these, however.

jackwyand:
Closes #4081 

Adds a bit of documentation explaining the discussion from #4081.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs
@@ -20,6 +20,10 @@ import qualified Language.Drasil.NounPhrase as NP
 import Drasil.Database.UID (HasUID(..))
 
 -- | Get short form (if it exists), else get term of an 'Idea'.
+-- Uses the UID of the 'Idea' in a 'Ch' Sentence constructor to get the short
+-- form using getA. getA may return Nothing, in which case lookupS uses the
+-- term, where lookupS is the main helper for getting the short form of a
+-- 'Ch' Sentence.
 short :: Idea c => c -> Sentence
 short c = sentenceShort (c ^. uid)
 
--- a/code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs
@@ -22,7 +22,7 @@ import Drasil.Database.UID (HasUID(..))
 -- | Get short form (if it exists), else get term of an 'Idea'.
 -- Uses the UID of the 'Idea' in a 'Ch' Sentence constructor to get the short
 -- form using getA. getA may return Nothing, in which case lookupS uses the
--- term, where lookupS is the main helper for getting the short form of a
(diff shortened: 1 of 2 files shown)


code/drasil-lang/lib/Language/Drasil/Development/Sentence.hs
```diff
@@ -20,6 +20,10 @@ import qualified Language.Drasil.NounPhrase as NP
 import Drasil.Database.UID (HasUID(..))
 
 -- | Get short form (if it exists), else get term of an 'Idea'.
+-- Uses the UID of the 'Idea' in a 'Ch' Sentence constructor to get the short
```
JacquesCarette, line 23: Nice improvement, and operationally correct.

@balacij you ought to be the next one editing this (later), to insert a more semantic description -- i.e. what exactly *is* a 'short form', and why is it acceptable to return the full term if there is no short form. [I do think such an explanation exists!]

JacquesCarette (approved):

Issue #4152, opened 2025-06-10
Game Physics: Extra period in Terminology and Definitions Section


smiths:
The [Terminology and Definitions section of the Game Physics example](https://jacquescarette.github.io/Drasil/examples/gamephysics/SRS/HTML/GamePhysics_SRS.html#Sec:TermDefs) has an extra period that should be removed at the end of the definition for a Right-handed coordinate system.

PR #4159, opened 2025-06-11
Removed extra period

sarrasoussia:
#4152: Removed at the end of the definition for a Right-handed coordinate system.


smiths:
@sarrasoussia, these changes will fail because you modified `stable`, but not `build`.   You need to change the Drasil code so that the generated documentation no longer has two periods.

JacquesCarette:
Right - no one should hand-edit generated files!

Issue #4151, opened 2025-06-10
Game Physics: Fix Figure 1


smiths:
Figure 1 in the Game Physics example shows the system as being named Chipmunk2D.  This is an old name.  We now call it GamePhysics.  The figure should be updated with the correct name.

PR #4160, opened 2025-06-11
Replaced Chipmunk2D to GamePhysics

sarrasoussia:
#4151: Replace every occurence of Chipmunk2D with GamePhysics.


sarrasoussia:
Oh yes, I think I totally missed the point when changing every single occurence of Chipmunk without paying attention to the actual game reference. I will definitely correct that.

PR #4165, opened 2025-06-12
Create basis ChunkDB that all new ChunkDBs are built upon

### LINKED ISSUE #4092: Create a `ChunkDB` that all case studies import and add to.
## Code Duplication

Look at these examples:

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L153-L163

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/hghc/lib/Drasil/HGHC/Body.hs#L63-L67

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs#L107-L116

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs#L144-L152

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs#L134-L143

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/sglpend/lib/Drasil/SglPend/Body.hs#L132-L140

https://github.com/JacquesCarette/Drasil/blob/196370e5a3ca95d9be69f9155f6c664dda897f5a/code/drasil-example/template/lib/Drasil/Template/Body.hs#L108-L114

## Goal

Each of our case studies have a basic set of chunks that we seem to always add to the `ChunkDB` for all of our case studies. The goal of this issue is to reduce the amount of duplication between `ChunkDB`s.

## Solution

We can do this in at least two ways:

1. Create a `ChunkDB` containing much of the stuff we import in at least 2 of our case studies, and use that as a basis for our other `cdb`s.
2. Create a `cdbWCommons` function that wraps `cdb` by injecting the same common imports to the relevant lists before also using `cdb` to create the `ChunkDB`.

jackwyand:
Closes #4092, #4168


code/drasil-database/package.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
 - drasil-lang
 - drasil-theory
 - drasil-utils
+- drasil-data
```
balacij: I think this is a sign that the basic `ChunkDB` might be best moved to `drasil-data`, or another package (seeing as our generators depend on it for even basic template generation).

jackwyand: The naming confuses me with this though. Doesn't it make sense for "Chunk Database" to be in `drasil-database`?

balacij: Yes, the `ChunkDB` type should definitely live in `-database`, but I think the instance of the `ChunkDB` that contains all the basic knowledge our software generators rely on is probably best living in `-data` or perhaps `-gen`.

I think the only `ChunkDB` instance that `-database` should provide is an empty one. The other packages can provide one that comes pre-filled with some chunks that the renderers depend on.

balacij:
Adding `siUnits` might be a good candidate to add to the `basicCDB` -- Perhaps after I'm done with #4173 and related PRs.

code/drasil-database/package.yaml
JacquesCarette: `drasil-metadata` would be a good place to move all the 'data' that Drasil depends on. `drasil-metadb` would be a good place for the database built from the metadata.  (I don't want `-data` to depend on `-database`).

I does indeed make sense to note have the database infrastructure mixed up with details of Drasil (beyond what's already there).

JacquesCarette: I seem to be missing something: how is `cdb'` different than `cdb`? More to the point: what is the function that uses `addCdb` over `basicCDB` ?

jackwyand: `cdb'` Creates a completely empty `ChunkDB`  to avoid the stable changes that happen when adding the `basisCDB` to `usedDB` in each of the examples. `addCdb` is only used in `cdb` at the moment to add the `basisCDB` to the newly created ones. Basically `addCdb` is just a helper function to concatenate two `ChunkDB`s together.

jackwyand: I like the `drasil-metadb` idea, I will try that.

balacij: Can you explain a bit more about what `-metadb` is for? Would `-metadb` contain a `ChunkDB` of the things from `-data` that `-gen` and its dependencies depend on? Or would it contain a `ChunkDB` of the things from `-metadata` that `-gen` and its dependencies depend on? Are we trying to have have two `ChunkDB`s to query from for 'data'? One for 'data' data and one for 'meta' data?

Aside: I also kind of want to avoid us having another package -- each one individually adds a few seconds to the build time, and the build time is a bit long as is. Haskell tooling just doesn't seem to work well for multi-root projects. If it's unavoidable to create a new package, that's fine, of course. I just want to double check first.

Issue #3567, opened 2023-07-26
Sort RefBy fields logically


samm82:
As noted in #3259 (see [this comment](https://github.com/JacquesCarette/Drasil/issues/3259#issuecomment-1650336592)), the output FR shows up first in the list in the RefBy field (the following screenshot is from [IM:calOfLandingTime from the Drasil website](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/srs/Projectile_SRS.html#IM:calOfLandingTime)):
![image](https://github.com/JacquesCarette/Drasil/assets/35857611/ab7753d9-00b8-4b23-9b93-ace6824ab65b)

I would think that the IMs would be first, then the FRs in the order they appear; the order of these should follow some logic. Maybe the ordering of the columns/rows in the traceability matrices could be used as a reference for how to sort them?

balacij:
Yep, it's just sorted by `UID` right now if I remember correctly. I think @JacquesCarette suggest sorting by visible label in #2468.

Now that I see it, the labelling/abbreviation convention doesn't seem uniform. IM labels follow camelCase while the (N)FRs use Upper-Kebab-Case.

JacquesCarette:
I definitely would not sort a `Sentence` but rather make sure our `Ref` data-structure is sortable by something that will correspond to visible label.

JacquesCarette:
We should take a look at what features of `Sentence` are actually used in the short names that appear in `Reference`. I am guessing very few. That would let us do the same kind of refactor I did with `NounPhrase`. Probably quite a bit easier, in fact.

PR #4170, opened 2025-06-13
Fix remaining projectile TermMap duplicates

jackwyand:
Contributes to #4036

### DIFF
--- a/code/drasil-example/projectile/lib/Drasil/Projectile/Unitals.hs
+++ b/code/drasil-example/projectile/lib/Drasil/Projectile/Unitals.hs
@@ -40,7 +40,7 @@ targPos   = constrained'    (uc       C.targPos   (subStr lP "target") Real metr
 ---
 -- The output contains a message, as a string, so it needs to be a quantity
 message :: QuantityDict
-message = vc "message" (nounPhraseSent (S "output message as a string")) lS String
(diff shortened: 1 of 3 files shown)


code/drasil-example/projectile/lib/Drasil/Projectile/Unitals.hs
```diff
@@ -40,7 +40,7 @@ targPos   = constrained'    (uc       C.targPos   (subStr lP "target") Real metr
 ---
 -- The output contains a message, as a string, so it needs to be a quantity
 message :: QuantityDict
-message = vc "message" (nounPhraseSent (S "output message as a string")) lS String
+message = vc "outputMessage" (nounPhraseSent (S "output message as a string")) lS String
```
jackwyand, line 43: It looks like changing the UID of this also changes the ordering of the "RefBy" note of one of the assumptions. I don't think this is a very impactful change to `/stable` so I kept it, however I am not sure why the ordering changed in the first place. I spent some time looking into it but I nothing really came of it.

balacij: I think that's #3567 / https://github.com/JacquesCarette/Drasil/issues/2468 . So, it's okay.

PR #4175, opened 2025-06-13
HGHC: Remove duplicate `UnitDefn`-related Terms

balacij:
Contributes to #4126


code/drasil-example/hghc/lib/Drasil/HGHC/Body.hs
```diff
@@ -70,9 +70,7 @@ ideaDicts =
   -- ConceptChunks
   map nw mathcon ++
   -- QuantityDicts
-  map nw symbols ++
-  -- UnitDefns
-  nw degree : map nw fundamentals ++ map nw derived
```
balacij, line 75: The `symbMap` (a `ChunkDB`) already had the correct `siUnits` list of `UnitDefn`s used (unlike the other examples)!

JacquesCarette (approved):

PR #4178, opened 2025-06-13
SSP: Remove duplicate `UnitDefn`-related Terms

balacij:
Contributes to #4126

### DIFF
--- a/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs
+++ b/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs
@@ -35,7 +35,7 @@ import Data.Drasil.Software.Products (prodtcon)
 import Data.Drasil.Theories.Physics (weightSrc, hsPressureSrc)
 
 import Data.Drasil.People (brooks, henryFrankis)
-import Data.Drasil.SI_Units (degree, metre, newton, pascal, kilogram, second, derived, fundamentals)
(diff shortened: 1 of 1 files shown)


code/drasil-example/ssp/lib/Drasil/SSP/Body.hs
```diff
@@ -160,13 +157,11 @@ ideaDicts =
...
   -- DefinedQuantityDicts
-  map nw symbols ++
-  -- UnitDefns
-  map nw derived ++ map nw fundamentals ++ map nw units
+  map nw symbols
 
 symbMap :: ChunkDB
 symbMap = cdb (map (^. output) iMods ++ map qw symbols) ideaDicts
-  (map cw iMods ++ map cw symbols ++ srsDomains) units dataDefs iMods
+  (map cw iMods ++ map cw symbols ++ srsDomains) (degree : siUnits) dataDefs iMods
```
balacij, line 164: `degree` is an interesting one. It's overloaded in two contexts: temperatures and angles. Locally, it can be a shorthand for "degrees Celsius," or it can refer to an angle of $X$ degrees.

JacquesCarette: which argues that, over time, we'll want to use some of these things with qualified names (inside the Haskell code). And they should most definitely have different UIDs.

PR #4189, opened 2025-06-16
SWHS: Remove undeniably duplicate `QuantityDict`-related `IdeaDict`s

balacij:
Contributes to #4126

This removes many, but not all.


balacij:
Force-pushed twice to fix minor issues.

PR #4194, opened 2025-06-17
Create `addCDB` function

jackwyand:
Contributes to #4092

### DIFF
--- a/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
+++ b/code/drasil-database/lib/Database/Drasil/ChunkDB.hs
@@ -25,7 +25,9 @@ module Database.Drasil.ChunkDB (
   -- ** Lenses
   unitTable, traceTable, refbyTable, citationTable,
   dataDefnTable, insmodelTable, gendefTable, theoryModelTable,
-  conceptinsTable, labelledcontentTable, refTable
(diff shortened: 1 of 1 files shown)


code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -246,6 +246,31 @@ cdb s t c u d ins gd tm ci lc r cits =
     _refTable = idMap "RefMap" r
   }
 
+
+addCdb :: ChunkDB -> ChunkDB -> ChunkDB
+addCdb cdb1 cdb2 =
```
balacij, line 253: I think you need to add this to the modules export list for it to pass the CI.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -246,6 +248,31 @@ cdb s t c u d ins gd tm ci lc r cits =
     _refTable = idMap "RefMap" r
   }
 
+
+addCdb :: ChunkDB -> ChunkDB -> ChunkDB
```
balacij, line 252: My only minor note, which I don't think should block the PR at all is to add a short comment such as: Union of two disjoint `ChunkDB`s

PR #4201, opened 2025-06-17
Projectile: Move weakly used `ConceptChunk`s (only used for their `IdeaDict`s) to `ConceptChunk`s table.

balacij:
Contributes to #4126


balacij:
@jackwyand and I checked out what was happening: The added ConceptChunk to the concept chunks table results in one symbol description being found, so it might be a "good" change.

smiths:
The description of the parameter `g` in the code has had the following text added to it: "the magnitude of the approximate acceleration due to gravity on Earth at sea level (m/s^2)".  I don't think we need to make our projectile motion program this specific.  Why can't we have a projectile at an elevation higher or lower than sea level?  Going further, why can't we have a projectile motion problem on another planet?  Since `g` is provided as a function parameter, this implies it can change.  I think this is the right approach.  If we say `g` is the acceleration at sea level, we should introduce a constant instead, since (forgetting about precision) the value of `g` at sea level is known.

balacij:
Just going to give a bit more context. We have the following definitions related to gravity defined in `drasil-data`:

```haskell
gravity = dcc "gravity" (cn "gravity")
  "the force that attracts one physical body with mass to another"
  
gravitationalAccel = dcc "gravitationalAccel" (cn "gravitational acceleration")
  "the approximate acceleration due to gravity on Earth at sea level"
  
gravitationalConst = dcc "gravitationalConst" (cn "gravitational constant")
  "the empirical physical constant used to show the force between two objects caused by gravity"
  
gravitationalMagnitude = dcc "gravitationalMagnitude" (cn "magnitude of gravitational acceleration")
  "the magnitude of the approximate acceleration due to gravity on Earth at sea level"
```

The "sea level" note is also included in the SRS of SSP.

So, it looks like our notions of gravitational acceleration and magnitude of gravitational acceleration are specialized to Earth. Now, there's a design issue here to build the generic notion of gravitational acceleration and magnitude of gravitational acceleration, and to build the specific Earth ones. For Projectile, since we're assuming $g=9.8$, is this okay as-is?

smiths:
No, I don't think this is okay "as is".  In all instances we are being prematurely concrete.  If we had a family approach, the family members on Earth at sea-level are what we have instantiated.  Since we don't have a family approach yet, we have skipped some levels of abstraction through implicit decisions.

I think the fix is fairly straight-forward.  For the current versions of our knowledge, we make it abstract.  When we later have the ability for refinement, we can modify the definitions.  Therefore, I suggest instead of "the approximate acceleration due to gravity on Earth at sea level" we use "the approximate acceleration due to gravity".  This is true for Earth at any elevation or on any planet.  It works for slope stability and for projectile.

balacij:
On that note, once https://github.com/JacquesCarette/Drasil/pull/4209 is merged, I will pull it into this branch and re-stabilize :)

balacij:
Ok, in that case, I think we can merge this one and fix the gravity-related issue later. This PR just happens to bring an error to light, which existed before this PR. It just wasn't visible because the description was not added to the list that it should have been added to.

smiths:
Your proposal to merge is fine with me, as long as we have an issue to remind us to provide a mechanism to refine the notion of gravity in the future.

balacij:
Thanks! I'll make sure to file one and deal with it soon. I just need a bit more clarification first (https://github.com/JacquesCarette/Drasil/pull/4209#issuecomment-2989114650).

PR #4202, opened 2025-06-17
SglPend: Move weakly used `ConceptChunk`s (only used for their `IdeaDict`s) to `ConceptChunk`s table.

balacij:
Contributes to #4126


jackwyand:
@balacij my hunch was correct about the ordering of querying the tables in the chunkDB, it looks like since `newtonSLRQD` shared a UID with `torque` it's description was being used instead since it was added in the term table (which is queried before the concept table). I think my most recent commit should fix the issue.

JacquesCarette (approved):

PR #4208, opened 2025-06-18
Move meta concepts to `drasil-metadata`

jackwyand:
Contributes to #4092 

@JacquesCarette are you able to check the changes I have made so far? I wanted to make sure that this is what you meant when we discussed separating the explanations and terms of the metadata being added to the basis ChunkDB. `Math.hs` contains the chunks that have been broken up so far.


JacquesCarette:
Good that you checked! This is probably technically correct, but in practice incorrect.

What I mean is that only the concepts that are used in Drasil itself should be split in this way. All the 'normal' stuff should stay as is in `drasil-data`.

JacquesCarette:
Any chunk that is referred to in `drasil-docLang`, for example. Probably any chunk that is referred to outside of `-example` (and `-data` itself) is a likely candidate.

balacij:
The diff is now +0/-0 and the PR is in draft. So, closing in favour of moving discussion to #4384.

PR #4209, opened 2025-06-18
Make "gravity" concepts more generic

balacij:
See https://github.com/JacquesCarette/Drasil/pull/4201#issuecomment-2985382469 for rationale.


JacquesCarette:
I don't think this change as-is is correct. `g` as a constant is not generic -- but this PR (indirectly?) changes the description of `g`.

What we need is as @smiths says: we need a proper refinement. We need both the generic and the specific version of gravity in our knowledge base. This non-fix makes a change instead of actually fixing the problem.

smiths:
If it is an option to "fake" refinement right now, I'm in favour of that.  If we can have a generic gravity and an Earth at sea-level gravity, that would be great.  I thought we had to pick one or the other (with me opting for the more abstract/general version).

balacij:
Before I write a ticket to fix this properly, to confirm: the goal would be to duplicate these 4 gravity-related concepts and make one set abstract, and the other specialized to Earth? I'm fine with doing that, but in that case, why is the change in #4209 bad? It uses Earth's common gravity constant assumption, so the description is fairly accurate, no? Are we hoping to have two symbols as well: $g$ (abstract) and $g_E$ (gravity constant assumption for Earth)?

JacquesCarette:
@smiths describes it well. We want to have abstract ones and then a *mechanism* to produce specializations, as many as we wish.

We can actually get quite close now (it will be somewhat hacky). So we should move in that direction.

The combinators would take in an existing chunk (understood to be more abstract), some delta information (potentially as functions that act via getters), and produce a **new** chunk. Hopefully the new one would refer to the old one (something we may not quite be fully able to do yet).

PR #4220, opened 2025-06-20
mdBook: Avoid building from source in CI, use binary releases

balacij:
This should fix the CI (https://github.com/JacquesCarette/Drasil/actions/runs/15780634999/job/44485655929?pr=4218).


balacij:
Before merging, I just want to double check that the `mdbook` executable will work. The CI doesnt build the mdbook code unless it has reason to:

```yaml
        if: ${{ steps.changes.outputs.md == 'true' || fromJSON(env.is_deployment) }}
```

balacij:
Fair enough, we can forego this. I assumed there was a problem with upstream packages, but whatever it was, looks fixed now.

Issue #4229, opened 2025-06-20
`moment` and `moment2D` share a UID causing duplicate warnings


jackwyand:
I have been looking at the duplicate UIDs in PDController and noticed 2 chunks that share a UID, `moment` and `moment2D`. A comment was left in the code stating "`moment2D` should eventually be a specialization of `moment`, not separately defined." I wanted to document this since it is clearly a problematic duplicate UID, however I am unsure what a "specialization" of another chunk would look like and how I should proceed with fixing this duplicate.

Here is the code in question: https://github.com/JacquesCarette/Drasil/blob/580d28b68f09fa37c8b30a51df1e4be384ede9aa/code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs#L72-L74

PR #4231, opened 2025-06-20
Remove `moment` duplicate

jackwyand:
Contributes to #4036 

This fixes a duplicate error but as mentioned in #4229 it doesn't feel like the correct way to fix this. However, we need to do some more design in order to fix it properly.

### DIFF
--- a/code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
+++ b/code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs
@@ -29,7 +29,7 @@ physicscon = [acceleration, angularAccel, angularDisplacement, angularVelocity,
   iyVel, kEnergy, linearAccel, linearDisplacement, linearVelocity, momentOfInertia, 
   position, potEnergy, pressure, scalarAccel, scalarPos, speed, time, torque, velocity,
   weight, xAccel, xConstAccel, xDist, xPos, xVel, yAccel, yConstAccel, yDist,
-  yPos, yVel,momentum, moment, moment2D, fOfGravity, positionVec, tension,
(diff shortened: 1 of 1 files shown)


JacquesCarette:
This really feels like too much of a hack. Would it be better to instead give `moment2D` a different `UID` instead of hiding the fact that it is malformed?

[And needs a merge]

jackwyand:
@JacquesCarette it won't let me request another review but I have made a new `"moment2D"` UID that is used instead, let me know if anything else should be changed.

code/drasil-data/lib/Data/Drasil/Concepts/Physics.hs
```diff
@@ -127,6 +127,8 @@ momentum = dccWDS "momentum" (cn "momentum")
   ( S "the quantity of motion" `S.of_` S "a moving body, measured as a product" `S.of_`
   phraseNP (QPP.mass `and_` velocity))
 moment = dccWDS "moment" (cn' "moment")
+  (S "A measure of the tendency of a body to rotate about a specific HELLO!!" +:+ phrase point `S.or_` phrase axis)
```
JacquesCarette: `HELLO!!` ???

JacquesCarette (changes requested):
Clearly this means that `CP.moment` is not used, but still, we shouldn't merge that as is.

code/drasil-data/lib/Data/Drasil/Concepts/Physics.hs
jackwyand: Haha, I must have been testing something earlier!

JacquesCarette (approved):

Issue #4207, opened 2025-06-17
Replace `usedinfodb` hack with a simpler hack


balacij:
See #1661 and #3864 for the complete context of why `usedinfodb` is necessary.

The major reason is for generation of our Table of Abbreviations and Acronyms (TAA). The TAA contains an enumeration of abbreviations and and descriptions from our `IdeaDict`s. Currently, to ensure that only things with abbreviations that are used in the generated SRS appear in the TAA, we manually write a list and embed it in `usedinfodb :: ChunkDB` (contained in `System`). This is a hack. A goal of #3864 is to enable automation of collecting these `IdeaDict`s and producing the TAA (#1661). However, this hack is more invasive than it needs to be. Currently, it involves creating a dummy `ChunkDB` that is mostly empty except for the list of `IdeaDict`s that we want to present in the TAA.

For example, SSP's `System`:

https://github.com/JacquesCarette/Drasil/blob/5564681ff6d3146c486eb99a83cba10a4fb346b3/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L70-L89

Note that it contains `_usedinfodb = usedDB`:

https://github.com/JacquesCarette/Drasil/blob/5564681ff6d3146c486eb99a83cba10a4fb346b3/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L174-L176

The goal of this ticket is to reduce the workload associated with #2873 -- avoiding creating extra `ChunkDB`s and its abuse. We should change how this information is funnelled into Drasil.

I propose that we remove these extra `ChunkDB`s (both the instance and it being held inside of `System`) in favour of it being immediately fed into `TAandA` in our `SRSDecl`s:

https://github.com/JacquesCarette/Drasil/blob/5564681ff6d3146c486eb99a83cba10a4fb346b3/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L94

I suggest the `SRSDecl`s because we already hack things in there, so localizing this kind of hack to it is probably best.

balacij:
@sarrasoussia Would you be interested in this ticket?

JacquesCarette:
Have to wonder: what if we scanned the SRS, recursing through its pieces, how close would we get to gathering everything?

Having those traversal functions is going to be needed, sooner or later, so it would be an interesting experiment to make.

balacij:
That's how we go about generating the table of symbols. So, I assume it would be fine.

[...]

PR #4232, opened 2025-06-20
Removing manual usedinfodb approach

sarrasoussia:
Working on #4207


JacquesCarette:
You need to have the build working before I review - the code doesn't compile! This will indeed be an adequate hack that replaces the even-worse hack of `usedinfodb`. Comments should be left in the code that this is (still) not satisfactory.

sarrasoussia:
I closed this PR because it introduced the `TAandA` section as a new `DocSection,` but after further review, I realized it made more sense to refactor it within `RefTab` to align better with the existing reference table structure. The new PR keeps the logic consistent, avoids unnecessary section duplication, and still removes the need for the `_usedinfodb` hack.

JacquesCarette:
Thanks for the meaningful comments.

Issue #3512, opened 2023-06-28
No modules, no functions generation


JacquesCarette:
We would like to generate code (like Projectile) with no functions at all (see #3247), just straight-line code. So we need a 'single function' option to the generator as a starting point.

As part of this task, the commit https://github.com/JacquesCarette/Drasil/commit/f2f547cc749e82e639b55167c713fb2509473c9c should be cherry-picked (and finished, i.e. changes to stable dealt with) along with modified to use that new option.

cd155:
As far as I know, this chart might be helpful in visualizing the current hierarchical structure of `Choices`

<img width="1200" src="https://user-images.githubusercontent.com/16947266/158629471-cc98127c-19cf-4a98-8d16-ecd26aff1230.png">

I guess the `modularity` part may need to be re-work, and adding a `single function` option under `unmodular`?

JacquesCarette:
Thanks Dong - yes, that should help a new person navigate the current choices!

Issue #3247, opened 2023-01-27
Projectile redux


JacquesCarette:
It would be "nice" to be able to also obtain this version of projectile:
```python
import math

v, t, p1 = 5.1, 43.0, 12.89
d = 2.0 * v ** 2.0 * math.sin(t) * math.cos(t) / 9.8 - p1
if (math.fabs(d / p1) < 2.0e-2) :
    return "The target was hit."
elif (d < 0.0) :
    return "The projectile fell short."
else :
    return "The projectile went long."
```

It is, on purpose, rather hideous. [I didn't take the values from the standard input file, I should probably do that.]  Nevertheless, we should be able to set options and write a recipe that produces this. (Though maybe those `return`s should be `print`s ?)

smiths:
Yes, I agree.  We should be able to produce a simple "bare bones" version of the program.  No separate files.  No classes.  No functions.  Just the requirements in the sequence they are listed (input, calculate, output).  Some people might say that this simple program should be written like you have it above.  The way we currently do it we make a simple program look much more complex than it has to.

balacij:
Duplicate of #1999, closing, but as always, feel free to re-open if you feel otherwise.

JacquesCarette:
Actually, this is related to, but different. That asks for unmodular, this asks for straight-line and with extra inlining.

PR #4233, opened 2025-06-22
Make unmodular code generation a bit more unmodular

balacij:
Contributes to #1999, #3247, #3512

Not quite ready to merge yet, but, this PR makes the `Unmodular` code generation choice a bit more unmodular by removing the generated function calls in favour of inlining them.

~~Three~~ ~~Four~~ Five major issues:

1. Code comments are completely removed from the unmodular option -- the comments that previously went above the function declarations.
2. Code needs to be tidied.
3. ODE-related things are untested.
4. `GProc`-generating code is untouched.
5. I have not tested generating derived inputs yet.

I'm also not quite sure yet how we should integrate this in the options. Should these changes go under Unmodular? SuperUnmodular? SpaghettiMonster? I currently have it editing `Unmodular`.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -78,6 +78,7 @@ instance RenderChoices Modularity where
 -- | Program implementation options.
 data ImplementationType = Library -- ^ Generated code does not include Controller.
                         | Program -- ^ Generated code includes Controller.
+  deriving Eq
 
 -- | Renders options for program implementation.
 instance RenderChoices ImplementationType where
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -78,7 +78,6 @@ instance RenderChoices Modularity where
 -- | Program implementation options.
 data ImplementationType = Library -- ^ Generated code does not include Controller.
                         | Program -- ^ Generated code includes Controller.
-  deriving Eq
 
 -- | Renders options for program implementation.
 instance RenderChoices ImplementationType where
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -209,6 +209,7 @@ makeDocConfig = DocConfig
 data Comments = CommentFunc -- ^ Function/method-level comments.
               | CommentClass -- ^ Class-level comments.
               | CommentMod -- ^ File/Module-level comments.
+              | CommentVarDecls -- ^ Variable declarations comments.
               deriving Eq
 
 -- | Renders options for implementation of comments.
@@ -216,6 +217,7 @@ instance RenderChoices Comments where
   showChs CommentFunc = S "CommentFunc"
   showChs CommentClass = S "CommentClass"
   showChs CommentMod = S "CommentMod"
+  showChs CommentVarDecls = S "CommentVarDecls"
 
 -- | Doxygen file verbosity options.
 data Verbosity = Verbose | Quiet
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
@@ -36,7 +36,7 @@ import Language.Drasil.CodeSpec (CodeSpec(..), HasOldCodeSpec(..), getODE)
 import Language.Drasil.Printers (SingleLine(OneLine), sentenceDoc)
 
 import Drasil.GOOL (OOProg, VisibilityTag(..),
-  ProgData(..), initialState)
+  ProgData(..), initialState, SMethod)
 import qualified Drasil.GOOL as OO (GSProgram, SFile, ProgramSym(..), unCI)
 import Drasil.GProc (ProcProg)
 import qualified Drasil.GProc as Proc (GSProgram, SFile, ProgramSym(..), unCI)
@@ -150,8 +150,8 @@ genPackage unRepr = do
       db = codeSpec g ^. systemdbO
       -- prps = show $ sentenceDoc db Implementation OneLine
       --   (foldlSent $ purpose $ codeSpec g)
-      prps = show $ sentenceDoc db Implementation OneLine 
-        (foldlSent $ codeSpec g ^. purpose)  
+      prps = show $ sentenceDoc db Implementation OneLine
+        (foldlSent $ codeSpec g ^. purpose)
       bckgrnd = show $ sentenceDoc db Implementation OneLine
         (foldlSent $ codeSpec g ^. background)
       mtvtn = show $ sentenceDoc db Implementation OneLine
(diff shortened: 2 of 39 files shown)


smiths:
@balacij, this change should be helpful when we return to the human-readable code paper.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -78,6 +78,7 @@ instance RenderChoices Modularity where
 -- | Program implementation options.
 data ImplementationType = Library -- ^ Generated code does not include Controller.
                         | Program -- ^ Generated code includes Controller.
+  deriving Eq
```
JacquesCarette: We've tended to avoid `deriving Eq` (even though in this case it is completely harmless) to emphasize a pattern-matching style over a "boolean" style. Why do you need `Eq`?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -183,34 +183,43 @@ genProgram :: (OOProg r) => GenState (OO.GSProgram r)
...
-        genInputConstraints Pub]) ++ [genOutputFormat]))
-    ([genInputClass Auxiliary, genConstClass Auxiliary]
-      ++ map (fmap Just) (concatMap genModClasses $ modules g))
+      clsyInputs :: OOProg r => [GenState (Maybe (SMethod r))]
+      clsyInputs
+        | any (`member` clsMap g) [giName, dvName, icName] = []
+        | otherwise = [genInputFormat Pub, genInputDerived Pub, genInputConstraints Pub]
+
+      modFuncs :: OOProg r => [GenState (Maybe (SMethod r))]
+      modFuncs = map (fmap Just) $ concatMap genModFuncs (modules g)
```
JacquesCarette, line 211: This is an odd piece of code -- it deserves a comment!

code/stable/projectile/projectile_u_p_l_b_b_c_d/src/cpp/Projectile.cpp
```diff
@@ -155,68 +155,6 @@ int main(int argc, const char *argv[]) {
     return 0;
 }
 
-double func_t_flight(InputParameters &inParams) {
```
JacquesCarette, line 158: while I'm completely fine with the logging of the inline version being much less verbose, we shouldn't lose quite so much logging. Issue six?

code/stable/projectile/projectile_u_p_l_b_b_c_d/src/cpp/Projectile.hpp
```diff
@@ -48,32 +48,6 @@ class Constants {
         
 };
 
-/** \brief Calculates flight duration: the time when the projectile lands (s)
```
JacquesCarette, line 51: When comments are turned on, I think a version of these comments should go above the inlined code.

code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/cpp/Projectile.cpp
```diff
@@ -33,39 +33,19 @@ int main(int argc, const char *argv[]) {
     double p_target;
     get_input(filename, v_launch, theta, p_target);
     input_constraints(v_launch, theta, p_target);
-    double t_flight = func_t_flight(v_launch, theta, g);
-    double p_land = func_p_land(v_launch, theta, g);
-    double d_offset = func_d_offset(p_target, p_land);
-    string s = func_s(p_target, epsilon, d_offset);
+    // flight duration: the time when the projectile lands (s)
```
JacquesCarette, line 36: Why have the comments started appearing here? [Don't get me wrong, this is a great *option* to have!] But we should have a version without them too.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -208,13 +209,15 @@ makeDocConfig = DocConfig
 data Comments = CommentFunc -- ^ Function/method-level comments.
               | CommentClass -- ^ Class-level comments.
               | CommentMod -- ^ File/Module-level comments.
+              | CommentVarDecls -- ^ Variable declarations comments.
```
JacquesCarette, line 211: Nice addition!

JacquesCarette (changes requested):

code/stable/projectile/projectile_u_p_l_b_b_c_d/src/cpp/Projectile.cpp
balacij: Issue six indeed!

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: I can actually remove this now -- I previously was avoiding a very small `case` statement in `genMainFunc` but I've since changed it back to a `case` statement.

code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/cpp/Projectile.cpp
balacij: > When comments are turned on, I think a version of these comments should go above the inlined code.

These are meant to be the inlined code comments. It's just a weaker variation of the ones that would go above the function-generating counterpart. I don't think we want the full commentary, do we?

JacquesCarette: Correct, not full commentary, thus my "a version of".

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: Fixed in 5fd7f57b2ba1d8e65ee7ad641f3d96b674cf0760

balacij:
Converting to Draft because it needs to:

1. Be more careful. It should continue to generate functions for function-typed things.
2. Be broken up.
3. Support `GProc`.
4. Be tested.

balacij:
@JacquesCarette and I discussed this in our meeting yesterday. Paraphrasing our discussion:

* We want to move away from the "modular"/"unmodular" designations (as it is ambiguous/fuzzy).
* We want more concrete terminology captured/used/invoked when describing what kind of architecture.
* For example [straight-line program](https://en.wikipedia.org/wiki/Straight-line_program) could be an example (informally: just assignments, no ifs/fors/whiles) of a program we generate.
* Rationale: The story we would tell (in the paper) starts with the most "unmodular" version and shows how we can generate "more modular" versions (embracing the fuzzy wording here) through knowledge capture.
  > Begin with the straight-line program, if we wanted to make it more readable, what would we do? Basically showing off Drasil's generation options and how we can gradually build "larger and larger" programs that describes more.
  > 
  > Least verbose projectile to Most verbose projectile
* Side effect of the discussion: The responsibility of depth of knowledge capture can vary about the desired artifacts. For example, not all components of the SRS are strictly necessary to describe a program -- everything can be "junk" -- or, an entirely different SRS format can be used (a heavily stripped down variant of the SmithEtAl template). How slim down can an SRS get and still generate (reasonable) runnable code?

balacij:
Regrettably, nothing in this PR is worth salvaging. #4804 explains how we should proceed.

PR #4240, opened 2025-06-23
Remove copies of data structure and data type chunks

jackwyand:
Contributes to #4036


JacquesCarette:
Where is the other copy?

JacquesCarette (approved):
Well, if no output is affected by this change, then clearly these were not used. Yes, these are probably 'conjugated' (I know, that's for verbs, not nouns) differently, but that seems like overkill.

PR #4243, opened 2025-06-23
GamePhysics: Move `defSymbols` to TermMap to avoid duplicates

jackwyand:
Contributes to #4036

After making this change I realized that it might only hurt with resolving #4126; maybe @balacij could take a look to see if this is a good change or not? I am not sure what is more of a priority at this point, having no duplicates within chunk tables or no duplicates across _all_ chunk tables.

### DIFF
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
@@ -139,16 +139,16 @@ ideaDicts =
   doccon ++ educon ++ prodtcon ++
   -- CIs
   map nw [progName, centreMass] ++ map nw doccon' ++ map nw CM.mathcon' ++
-  map nw CP.physicCon'
+  map nw CP.physicCon' ++
+  -- DefinedQuantityDicts
+  map nw defSymbols
   
 conceptChunks :: [ConceptChunk]
 conceptChunks = 
   -- ConceptChunks
   algorithm : softwarecon ++ CP.physicCon ++ CM.mathcon ++ srsDomains ++
   -- InstanceModels
-  map cw iMods ++
-  -- DefinedQuantityDicts
-  map cw defSymbols
+  map cw iMods
 
 symbMap :: ChunkDB
 symbMap = cdb symbolsAll ideaDicts conceptChunks


JacquesCarette:
If the within table that are major issues are under control, that is indeed different!

JacquesCarette:
Can you give me an explicit list of the chunks that were conflicting? Was it the same chunks but add to two tables or something else?

jackwyand:
Sure!

In the `conceptChunk` list that is added to the ChunkDB, there is a list called `defSymbols` (Ln 149)
https://github.com/JacquesCarette/Drasil/blob/def8f55cbdd4cd2b2cd857e8b6958009b4513c06/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L144-L153

Within this list, there are two other lists added, including `inputConstraints`:
https://github.com/JacquesCarette/Drasil/blob/def8f55cbdd4cd2b2cd857e8b6958009b4513c06/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs#L26-L27

`forceCons`, added in `inputConstraints`:
https://github.com/JacquesCarette/Drasil/blob/def8f55cbdd4cd2b2cd857e8b6958009b4513c06/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs#L317-L320

shares a UID with `force` since it uses that chunk in its creation
https://github.com/JacquesCarette/Drasil/blob/def8f55cbdd4cd2b2cd857e8b6958009b4513c06/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs#L334

`force` is then also added in `physicCon`, which contains all of the general physics concepts. `physicCon` is evidently also added in `conceptChunks`, causing the duplicate warning.

Hopefully this makes sense!

JacquesCarette:
Yes, it does.

I think the real problem here is that `forceCons` has the same `UID` as `force`. These are separate things. Unless `force` in GamePhysics is only ever "force of gravity"?

Feels to me that `constrained'` is a bad combinator.

balacij:
One key piece of information that @JacquesCarette mentioned about the "reasonable values" that the `UncertQ`s associate with `D-Q-D`s: "think about the order of magnitude of the numbers rather than the specific numbers." For example, in the context of a beam, a length of 50m is sane, a similar length 2 orders magnitude up or down is also fine. But a "beam" of length 1E-100m is not.

PR #4250, opened 2025-06-25
Create basis ChunkDB

jackwyand:
Contributes to #4092


code/drasil-database/lib/Database/Drasil.hs
```diff
@@ -2,10 +2,10 @@
 module Database.Drasil (
   -- * Chunk Database
   -- ** Types
-  ChunkDB(symbolTable, termTable, conceptChunkTable, CDB), RefbyMap, TraceMap, UMap
+  ChunkDB(..), RefbyMap, TraceMap, UMap
```
JacquesCarette: Please revert this. We want to always explicitly control what we expert, even if right now it is "everything".

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -8,14 +8,14 @@ module Database.Drasil.ChunkDB (
   -- * Types
   -- ** 'ChunkDB'
   -- | Main database type
-  ChunkDB(CDB, symbolTable, termTable, conceptChunkTable),
+  ChunkDB(..),
```
JacquesCarette: same here

code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs
```diff
@@ -0,0 +1,45 @@
...
+
+basisIdeaDicts :: [IdeaDict]
+basisIdeaDicts =
+  -- Actual IdeaDicts
+  doccon ++ prodtcon ++ educon ++ compcon ++
+  -- CIs
+  map nw doccon' ++
+  -- ConceptChunks
+  map nw [algorithm, errMsg, program] ++ map nw mathcon
+
```
JacquesCarette: Please insert some comments here describing what this is, what this is for.

code/drasil-metadb/package.yaml
```diff
@@ -0,0 +1,33 @@
+name:                drasil-metadb
+version:             0.1.1.0
+author:              "Dan Szymczak, Steven Palmer, Jacques Carette, Spencer Smith"
```
JacquesCarette: Remove Dan and Steven from this author list, since it is no longer true. Do feel free to add yourself.

JacquesCarette (changes requested):

code/drasil-database/lib/Database/Drasil.hs
jackwyand: Just did this in [b1229cb](https://github.com/JacquesCarette/Drasil/pull/4250/commits/b1229cb0defae268c2f7f340970d3d6a7606d245), should I also import everything individually or leave that as `(..)`?

https://github.com/JacquesCarette/Drasil/blob/b1229cb0defae268c2f7f340970d3d6a7606d245/code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs#L5

code/drasil-metadb/package.yaml
jackwyand: Fixed in [738cea2](https://github.com/JacquesCarette/Drasil/pull/4250/commits/738cea28d66ab144ba04620c347f9969863237bc)

code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs
jackwyand: Done in [a15d545](https://github.com/JacquesCarette/Drasil/pull/4250/commits/a15d54514ccbd7f56fda822e2ec15ca85f27e731)

code/drasil-database/lib/Database/Drasil.hs
JacquesCarette: We want explicit imports.

code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs
```diff
@@ -0,0 +1,52 @@
...
+import qualified Data.Map as Map (empty)
+import Data.Drasil.SI_Units (siUnits)
+
+-- | The basic idea dicts that are used to construct the basis chunk database.
+-- Every chunk added here is added to every new chunk database created that uses
+--  the cdb constructor. This ensures that the information in these idea dicts
+--  is always available in the chunk database.
+basisIdeaDicts :: [IdeaDict]
+basisIdeaDicts =
```
JacquesCarette: Sorry to be super-picky, but this is a "big" PR (in that it is a big conceptual change, not big code change). Could you justify, in comments here in the code, why every one of those `IdeaDicts` is included in `basis`?

A line about `srsDomains` and `siUnits` would be good too (probably inline below).

JacquesCarette (changes requested):

code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs
jackwyand: Sounds good. `srsDomains` was also expanded to multiple chunk lists in #4252, so I will add comments explaining those there.

balacij: One key requirement for `basisCDB` is that the boilerplate generated SRS document and the code generator rely on at least these chunks. "These chunks" preferably being the lowest upper bound.

The "Characteristics of the Intended Reader" section also bring about a lot of the commonly used concepts and ideas into scope, so it's nice to have them ready.

jackwyand: I added some justification for the chunks in [fa8dc6f](https://github.com/JacquesCarette/Drasil/pull/4250/commits/fa8dc6f2bdd09301763e2354fbd41e3898a976d5), [c5c1286](https://github.com/JacquesCarette/Drasil/pull/4250/commits/c5c12861988e7e8c05b510457d406e99b979ca51), [7dc8a5f](https://github.com/JacquesCarette/Drasil/pull/4252/commits/7dc8a5f41268d09511b83b4e228f12fb78a9de97), and [e789a4b](https://github.com/JacquesCarette/Drasil/pull/4252/commits/e789a4bff7a730327395b2b5926a3c74c02c79f9). I apologize for the messiness of the additions being between 2 PRs, I added more chunks to the `basis` while implementing it in #4252 and that seemed to cause some organizational problems. Everything should be added though!

code/drasil-database/lib/Database/Drasil.hs
jackwyand: Should be resolved with [3eaa4c4](https://github.com/JacquesCarette/Drasil/pull/4250/commits/3eaa4c4f8fb7ff20c34a7bec277828756471f7b7)

code/drasil-metadb/lib/MetaDatabase/Drasil/ChunkDB.hs
```diff
@@ -20,11 +23,31 @@ import Data.Drasil.SI_Units (siUnits)
 --  is always available in the chunk database.
 basisIdeaDicts :: [IdeaDict]
 basisIdeaDicts =
-  -- Actual IdeaDicts
+  -- | Actual IdeaDicts
```
JacquesCarette: I agree that these are useful to always have around -- but these could be in a "Base" ChunkDB that is in `data-drasil` (or similar), no?

What should be here are those chunks that Drasil itself uses internally.

jackwyand:
During the meeting on Tuesday we got on the same track about what the purpose of the basis ChunkDB is supposed to be. Since it is not for specifically metadata and more just chunks that are used across all case studies, I am thinking of moving it to `drasil-data`, or `drasil-gen` since @JacquesCarette mentioned that `drasil-data` should not be dependent on `drasil-database` ([comment](https://github.com/JacquesCarette/Drasil/pull/4165#discussion_r2150388102)). I want to confirm where it should stay and if there are any other changes to be made.

balacij:
Why would it be bad for `drasil-data` to depend on `drasil-database`? If "common undergraduate-level physics concepts"  are all defined in `drasil-data` and we import all of it into our case studies, I don't see why wrapping everything together in the same library under a single `ChunkDB` would be problematic? If some concepts however are only relevant to the SRS, those should probably go into `drasil-docLang` or something else.

It looks like we're on a slope to implementing something scope/domain-related (good).

JacquesCarette:
`drasil-database` is "database infrastructure" and `drasil-data` is, well, data. The data should not depend on the infrastructure!

So putting together all sorts of "basic data" is a different kind of activity than representing it.

I would be fine with `drasil-gen` as the place where this is all assembled.

JacquesCarette (approved):

PR #4253, opened 2025-06-26
DblPend: Remove unused physics quantities, only concepts are used

jackwyand:
Contributes to #4126

### DIFF
--- a/code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
+++ b/code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
@@ -25,7 +25,6 @@ import Data.Drasil.Concepts.Math (mathcon, cartesian, ode, mathcon', graph)
 import Data.Drasil.Concepts.Physics (gravity, physicCon, physicCon', pendulum, twoD, motion)
 import Data.Drasil.Concepts.PhysicalProperties (mass, physicalcon)
 import Data.Drasil.Concepts.Software (program, errMsg)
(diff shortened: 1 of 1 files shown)


code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
```diff
@@ -25,7 +25,6 @@ import Data.Drasil.Concepts.Math (mathcon, cartesian, ode, mathcon', graph)
 import Data.Drasil.Concepts.Physics (gravity, physicCon, physicCon', pendulum, twoD, motion)
 import Data.Drasil.Concepts.PhysicalProperties (mass, physicalcon)
 import Data.Drasil.Concepts.Software (program, errMsg)
-import Data.Drasil.Quantities.Physics (physicscon)
```
JacquesCarette, line 28: it feels odd that DblPend uses no concepts from physics!

jackwyand: Removing the quantities in `physicscon` didn't cause any changes (hence why I concluded it was unused) however `physicCon` (added in the concept chunk table) seems to be where DblPend gets the necessary physics information from. The issue was that `physicscon` and `physicCon` overlapped multiple UIDs across the chunk tables.

JacquesCarette: Fair enough!

JacquesCarette (approved):

PR #4254, opened 2025-06-26
GamePhysics: Instance models already added in separate table, remove from concept table

jackwyand:
Contributes to #4126


JacquesCarette (approved):
Interesting that this causes no failures. Good. But given what I have otherwise seen, still surprised.

PR #4262, opened 2025-07-03
PDController: Split term definitions from used concepts

jackwyand:
Contributes to #4126

Individual concepts added in `defs` have corresponding chunks with the same UID also added in the symbol or concept table, so only the `termDefs` are added to the term table to avoid the conflicting UIDs.

The stable changes seem to only be a reordering of some of the lists as it depended on the ordering in which the chunks were added.

~~Edit: It looks like some files were renamed for no reason, I will check to see what caused it.~~


JacquesCarette (approved):
Odd that the order changed... but that will likely change again, so not to worry right now.

PR #4264, opened 2025-07-04
Remove unused `codeVC` QuantityDict constructor

jackwyand:
Contributes to #4126

Cleans up an unused QuantityDict constructor working towards removing QuantityDict entirely.


JacquesCarette (approved):
Getting rid of these is an excellent move.

PR #4270, opened 2025-07-10
Remove vc'' QuantityDict constructor

### LINKED ISSUE #4267: Weakly used constructors related to QuantityDict creation
I have noticed a few weakly used and unused constructors, some of which are standalone QuantityDict constructors and others that use the `qw` constructor or similar ones that ultimately constructs / extracts a QuantityDict. They seem mostly straightforward to remove and will help clean up QuantityDict usage overall to align with the goal of removing it entirely.

### `vcUnit`:
**Constructor Definition**:
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-lang/lib/Language/Drasil/Chunk/Quantity.hs#L112-L114

**Usage**:
None.

### `vc''`:
**Constructor Definition**:
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-lang/lib/Language/Drasil/Chunk/Quantity.hs#L120-L122

**Usage**:
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L271

**Fix**:

Creating a ConceptChunk based on `loadDurFactor` and extracting a DefinedQuantityDict from it, similar to how `loadSF` is created, would be a good way to remove the need for this constructor. It will also make it easier when we move to only using DefinedQuantityDicts, as we will not need to convert this chunk later.

### `mkUnitary`:
**Constructor Definition**:
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-lang/lib/Language/Drasil/Chunk/Unitary.hs#L62-L64

**Usage**:
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L199-L200

**Fix**:
Instead of adding both `unitarySymbols` and `unitalSymbols` to one list, `symbols`, which is then added to `symbolsForSymbolTable` (ln 14)
https://github.com/JacquesCarette/Drasil/blob/91d692979cdfe4a4ebd893f50938e67e13feae0a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Symbols.hs#L13-L16
They could instead be added separately to `symbolsForSymbolTable` using `qw` (for now, will eventually be changed to a DefinedQuantityDict) to avoid having to use `mkUnitary`.

##

If any of these should actually be kept, let me know!

jackwyand:
Closes #4267 (once #4269 is merged)

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Assumptions.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Assumptions.hs
@@ -18,7 +18,7 @@ import Drasil.GlassBR.Concepts (beam, cantilever, edge, glaSlab, glass,
 import Drasil.GlassBR.MetaConcepts (progName)
 import Drasil.GlassBR.References (astm2009)
 import Drasil.GlassBR.Unitals (constantK, constantLoadDur, 
-  constantLoadSF, constantM, constantModElas, explosion, lateral, lDurFac,
+  constantLoadSF, constantM, constantModElas, explosion, lateral, loadDF,
   loadDur)
 
 assumptions :: [ConceptInstance]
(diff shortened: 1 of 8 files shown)


JacquesCarette (approved):

JacquesCarette:
There's still a conflict.

JacquesCarette (approved):

Issue #3314, opened 2023-03-22
`nc` / `mkIdea` conundrum


JacquesCarette:
So I was about to file an issue, based on the work at [Attributes and Chunks](https://github.com/JacquesCarette/Drasil/wiki/Attributes-and-Chunks) that went along the lines of "remove all `mkIdea` calls that use `Nothing` for the third argument". Because, basically, all those things are 'morally' `NamedChunk`, so why put them in an `IdeaDict` ?

This has been visited before, in [How should we handle abbreviations](https://github.com/JacquesCarette/Drasil/wiki/How-Should-We-Handle-Abbreviations?) and [When to use Maybe](https://github.com/JacquesCarette/Drasil/wiki/How-Should-We-Handle-Abbreviations?)

In this particular case, what it boils down to is: is there a true ontological difference between what we call `NamedChunk` (which is a bad name since it is assumed to have a `term`) and an `IdeaDict` which accepts the idea that some things **may** have an abbreviation? That seems kind of silly, doesn't it?

So I'm actually thinking the exact opposite now: merge `NamedChunk` and `IdeaDict` (leaving just `IdeaDict`). And yes, this goes in the opposite direction of what I was saying before, of trying to get rid of uses of `Maybe` !!!

Why the change of mind? In this *particular case*, I think that 'abbreviation' really is, and should be, an optional 'decoration' on something. But here's a way to change my mind again: are there situations where having an abbreviation would be **manditory**? I can't think of one, but that's far from proof.  Interestingly, Googling for "must have an abbreviation" (i.e. [this search](https://www.google.com/search?client=firefox-b-d&q=%22must+have+an+abbreviation%22) has 20 hits, one of which is to Drasil!!  So there are things out there that 'must have an abbreviation'. Ah, but we're not talking about concepts to be modeled by Drasil, but rather the meta-model. There would have to be a whole category of things that must have abbreviations, with that category being important enough to be *internalized* in Drasil, i.e. all the way down to Haskell code. And that, there isn't.

Upshot: there is no conceptual gain in keeping these separate. And there is no conceptual gain in having 'abbreviation' ever be manditory.

So this issue now says the opposite of what I started with:
1. merge `NamedChunk` and `IdeaDict` (call it `Idea`)
2. have the abbreviation remain optional, aka a `Maybe`.

samm82:
[...]

PR #4271, opened 2025-07-10
Create `DrasilConcept` typeclass that directs to a chunk that defines the meaning of the instantiated class' name.

balacij:
Contributes to #3314 

I'm a little bit confused now by what `DrasilConcept` should be doing, and I think we need to clarify the terminology. This PR should most likely not be merged as-is.

### Braindump

The 'Concept' typeclass provides information about the specific thing that an instance of a chunk (of type 'c') embodies. So, when I went to build `DrasilConcept`, I figured that it should point to the related `ConceptChunk` that explains what 'c' means (e.g., a `ConceptChunk` that defines what `Theory Model` means). And that's how I ended up with this implementation:

```haskell
-- | What does this chunk type encode?
-- 
-- All chunk types in Drasil should instantiate this class.
class DrasilConcept c where
  -- | Provide the 'UID' to the 'ConceptChunk' that defines the concept type.
  metaConcept :: Getter c UID
```

Now three questions immediately arose:

##### 1.1. Why to _another chunk_?

Well, that's just working on an assumption: that we want to capture "metadata" chunks as well, and keep them in close proximity to our "data" chunks, capturing information about their relationships.

##### 1.2. Why _only point_?

Because the other chunk should exist already, and duplicating the type class fields with "meta" versions is unnecessary.

#### 2. Why point to a `ConceptChunk` specifically?

In my head, this was just the most obvious thing to point to because we need to be able to point to something that defines our Drasil terminology, and using our current `ChunkDB`, we need to be able to resolve references to a single type. `ConceptChunk` is the most appropriate because it is the infimum of all chunks that instantiate `Concept`. But it might not be the right thing. As @JacquesCarette mentioned in https://github.com/JacquesCarette/Drasil/issues/3314#issuecomment-3053378405: 

> Concept is great because it defines things by "what information must be present" instead of by "a concrete representation".

So, an alternative design might look like the following, where it points towards another chunk that must have the necessary "conceptual" information:

```haskell
-- | What does this chunk type encode?
-- 
-- All chunk types in Drasil should instantiate this class.
class Concept cd => DrasilConcept c cd where
  -- | Provide the 'UID' to a chunk (of type 'cd') that explains what 'c' is.
  metaConcept :: Getter c cd
```

Or, there could be yet another alternative version that recreates the 5 requirements of `Concept` (i.e., term, maybe abbreviation, concept domain, definition, `UID`) a bit more 'directly' (immediately returning the values), perhaps without the `UID`.

#### 3. Hold on. Where do mandatory abbreviations fit in?

Recall: @JacquesCarette mentioned `DrasilConcept` in https://github.com/JacquesCarette/Drasil/issues/3314#issuecomment-2116307350:

> What's implicit in the above is that we need to split DrasilConcept out from Concept, i.e. those concepts that we use for ourselves can have much higher requirements, such as making the abbreviation manditory.

The context is `CI`, which is essentially an `IdeaDict` (recall: vocabulary: noun + maybe abbreviation) with the requirement that the abbreviation _must_ exist. Now, a `CI` could not fulfill the requirements of a `Concept` because it has no definition of the noun it declares.

##### 3.1. Should `CI` also carry a definition?

If some concepts so common/required to have an abbreviation, shouldn't they also be required to have a (prose) definition as well?

### Additional Notes

In the future, when we have typed `UID` references, the `UID` references should also be typed. This might be something good to consider in current design.

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -66,6 +66,7 @@ import Data.Maybe (maybeToList)
 import Drasil.Sections.ReferenceMaterial (emptySectSentPlu)
 
 -- * Main Function
+
 -- | Creates a document from a document description, a title combinator function, and system information.
 mkDoc :: SRSDecl -> (IdeaDict -> IdeaDict -> Sentence) -> System -> Document
 mkDoc dd comb si@SI {_sys = sys, _kind = kind, _authors = docauthors} =
--- a/code/drasil-lang/lib/Language/Drasil.hs
+++ b/code/drasil-lang/lib/Language/Drasil.hs
@@ -58,6 +58,7 @@ module Language.Drasil (
   , Definition(defn)
   , ConceptDomain(cdom)
   , Concept
+  , DrasilConcept(metaConcept)
   , HasSpace(typ)
   , HasUnitSymbol(usymb)
   , Quantity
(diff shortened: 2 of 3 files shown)


balacij:
The alternative design mentioned in (2) has one pro over the existing code in the PR: we can have two instances for any chunk. For example, one that directs to a ConceptChunk and one that directs to a CI (assuming we add a definition to the CI type as well). The major con is that it's no longer giving a `UID` but an entire. With typed UID references, we can have

```haskell
-- | What does this chunk type encode?
-- 
-- All chunk types in Drasil should instantiate this class.
class Concept cd => DrasilConcept c cd where
  -- | Provide the 'UID' to a chunk (of type 'cd') that explains what 'c' is.
  metaConcept :: Getter c (TUID cd)
```

balacij:
Examples that might help us:

Theories: 

```haskell
-- | These are internal-to-Drasil common ideas, and need to be defined at the 
-- same time as theories.
dataDefn, genDefn, inModel, thModel :: CI
-- | Data definition.
dataDefn = commonIdeaWithDict "dataDefn" (cn' "data definition")    "DD"  [softEng]
-- | General definition.
genDefn  = commonIdeaWithDict "genDefn"  (cn' "general definition") "GD"  [softEng]
-- | Instance model.
inModel  = commonIdeaWithDict "inModel"  (cn' "instance model")     "IM"  [softEng]
-- | Theoretical model.
thModel  = commonIdeaWithDict "thModel"  (cn' "theoretical model")  "TM"  [softEng]
```

^ We would want to "point" to these.

What about the various "atoms" we have? For example, `Expr`, `ModelKinds`, or the prospective `SystemKind`. An instance of `DrasilConcept` would likely point to a `ConceptChunk` as well that defines the type names. But each one would be a sum type -- how should we capture the meta-information on those constructors as well?

balacij:
While this is a very simple PR to revive, I'm going to just close it in favour of properly revisiting the work later (it will be just as simple to do 'correctly' later if we believe it to be simple now). I would prefer not to contribute work to an area of code I'm trying to avoid thinking about right now, and we are trying to close my old, outstanding PRs.

Issue #3260, opened 2023-02-02
Are we using too many typeclass constraints on `SystemInformation`? What is a `SystemInformation`?


balacij:
Related code:

https://github.com/JacquesCarette/Drasil/blob/e4b3354f1d8586ff25fec845ea9618af4bff25fe/code/drasil-sysinfo/lib/SysInfo/Drasil/SystemInformation.hs#L37-L66

It's very difficult to pin down a definition of what `SystemInformation` is, and I'm hoping this ticket will help us understand it a bit. Let's go through the fields:

```haskell
  { _sys         :: a
```

a "system" (?) has a common idea & idea,

continuing, an SI has:

```haskell
  , _kind        :: b
```

a kind/type, which is just an "idea",

```haskell
  , _authors     :: [c]
```

a list of authors, which should be extracted from some polymorphic type,

```haskell
  , _purpose     :: Purpose
```

a Purpose,

```haskell
  , _background  :: Background
```

a Background,

```haskell
  , _quants      :: [e]
```

quantity-like things that can be boiled down to quantities (`Quantity e, Eq e, MayHaveUnit e,`),

```haskell
  , _concepts    :: [f]
```

quantity-like things that can be boiled down to concepts?, (`Quantity f, MayHaveUnit f, Concept f, Eq f`)

```haskell
  , _instModels  :: [InstanceModel]
```

a list of instance models (this should likely be replaced with ChunkDB usage -- my problem! Oops!),

```haskell
  , _datadefs    :: [DataDefinition]
```

a list of data definitions (this should likely be replaced with ChunkDB usage too),

```haskell
  , _configFiles :: [String]
```

a list of configuration files, provided as Strings?,

```haskell
  , _inputs      :: [h]
```

some list of things that can be boiled down to quantities (`Quantity h, MayHaveUnit h`),

```haskell
  , _outputs     :: [i]
```

some list of things that can be boiled down to quantities (`Quantity i, MayHaveUnit i`)

```haskell
  , _defSequence :: [Block SimpleQDef]
```

I'm not fully sure what this is, but I would hope it is something that we could automatically pull from the composed instance models and data definitions?

```haskell
  , _constraints :: [j] --TODO: Add SymbolMap OR enough info to gen SymbolMap
```

a list of "constraints", a list of things that have a UID and are "constrained" (`HasUID j, Constrained j`)

(is this another thing that should be moved to chunkdb usage?)

```haskell
  , _constants   :: [ConstQDef]
```

[...]

PR #4274, opened 2025-07-10
Draft implementation of `SystemKind`.

balacij:
Contributes to #3260

Creating this PR as a draft because:

1. It's a draft implementation.
2. The CI will fail because I didn't propagate all `System`-related changes across all case studies -- only SSP.

### DIFF
--- a/code/drasil-system/lib/System/Drasil/System.hs
+++ b/code/drasil-system/lib/System/Drasil/System.hs
@@ -6,21 +6,24 @@
 -- https://github.com/JacquesCarette/Drasil/wiki/Creating-Your-Project-in-Drasil
 
 module System.Drasil.System (
-  -- * System Information
+  -- * System
   -- ** Types
-  System(..),
+  System(..), SystemKind(..),
   -- ** Lenses
   HasSystem(..),
+  -- ** Functions
+  whatsTheBigIdea,
   -- * Reference Database
   -- ** Types
   Purpose, Background, Scope, Motivation,
   ) where
 
-import Language.Drasil
+import Language.Drasil hiding (kind, Notebook)
 import Theory.Drasil
 import Database.Drasil (ChunkDB)
 
 import Control.Lens (makeClassy)
+import qualified Data.Drasil.Concepts.Documentation as Doc
 
 -- | Project Example purpose.
 type Purpose = [Sentence]
@@ -31,6 +34,19 @@ type Scope = [Sentence]
 -- | Project Example motivation.
 type Motivation = [Sentence]
 
+data SystemKind =
+    SRS
+  | Notebook
+  | Website
+
+whatsTheBigIdea :: System -> IdeaDict
+whatsTheBigIdea si = whatKind' (_kind si)
+  where
+    whatKind' :: SystemKind -> IdeaDict
+    whatKind' SRS = nw Doc.srs
+    whatKind' Notebook = nw Doc.notebook
+    whatKind' Website = mkIdea "website" (cn "website") (Just "web")
+
 -- | Data structure for holding all of the requisite information about a system
 -- to be used in artifact generation.
 data System where
@@ -39,13 +55,12 @@ data System where
 -- I'm thinking for getting concepts that are also quantities, we could
 -- use a lookup of some sort from their internal (Drasil) ids.
  SI :: (CommonIdea a, Idea a,
-  Idea b,
   Quantity e, Eq e, MayHaveUnit e,
   Quantity h, MayHaveUnit h,
   Quantity i, MayHaveUnit i,
   HasUID j, Constrained j) => 
   { _sys         :: a
-  , _kind        :: b
+  , _kind        :: SystemKind
   , _authors     :: People
   , _purpose     :: Purpose
   , _background  :: Background
--- a/code/drasil-system/package.yaml
+++ b/code/drasil-system/package.yaml
@@ -18,6 +18,7 @@ dependencies:
 - pretty
 - lens
 - containers
+- drasil-data
 - drasil-database
 - drasil-lang
 - drasil-theory
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -68,8 +68,8 @@ import Drasil.Sections.ReferenceMaterial (emptySectSentPlu)
 -- * Main Function
 -- | Creates a document from a document description, a title combinator function, and system information.
 mkDoc :: SRSDecl -> (IdeaDict -> IdeaDict -> Sentence) -> System -> Document
-mkDoc dd comb si@SI {_sys = sys, _kind = kind, _authors = docauthors} =
-  Document (nw kind `comb` nw sys) (foldlList Comma List $ map (S . name) docauthors) (findToC l) $
+mkDoc dd comb si@SI {_sys = sys, _authors = docauthors} =
+  Document (whatsTheBigIdea si `comb` nw sys) (foldlList Comma List $ map (S . name) docauthors) (findToC l) $
   mkSections fullSI l where
     fullSI = fillcdbSRS dd si
     l = mkDocDesc fullSI dd
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/DocumentLanguage.hs
@@ -7,7 +7,7 @@ import Drasil.DocumentLanguage.Notebook.Core (LsnDesc, LsnChapter(..),
 
 import Language.Drasil hiding (kind)
 
-import System.Drasil (System(SI), _authors, _kind, _sys)
+import System.Drasil (System(SI), _authors, _sys, whatsTheBigIdea)
 import Drasil.GetChunks (citeDB)
 
 import qualified Drasil.DocLang.Notebook as Lsn (intro, learnObj, caseProb, example, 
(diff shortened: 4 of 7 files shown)


code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -31,6 +34,19 @@ type Scope = [Sentence]
 -- | Project Example motivation.
 type Motivation = [Sentence]
 
+data SystemKind =
+    SRS
+  | Notebook
+  | Website
```
balacij: 3 kinds; 1 for each of our types of categories of `System`s created.

The ICO-related bits will go into the `SRS` constructor, and the `SRS` constructor will need to take on at least a bit more information: problem information (the ICO problem information) and solution (the theories, data definitions, constants, etc.).

I'm not sure what the other two will contain, but the goal will be to at least not have to instantiate a bunch of unused unused components of a `System` in the "notebook":

https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-example/projectile/lib/Drasil/Projectile/Lesson/Body.hs#L45-L64

And the website:

https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-website/lib/Drasil/Website/Body.hs#L72-L92

balacij: I'm following the discussion in https://github.com/JacquesCarette/Drasil/issues/3260 to create `SystemKind`, but an alternative implementation might avoid the sum type:

```haskell
data System sk where
  SI :: Concept sk => ...
    _kind :: sk,
...
  } -> System sk
```

How do we decide which is the better implementation? I'm not immediately sure, but https://github.com/JacquesCarette/Drasil/pull/4271#issuecomment-3058831231 is relevant ++ do we want `SystemKind` to be an open sum type? If so, this is the better implementation -- and we can also do some work to ensure that data placed in the hole have some sort of `kindOf` field set to "system kind."

code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -31,6 +34,19 @@ type Scope = [Sentence]
...
+
+whatsTheBigIdea :: System -> IdeaDict
+whatsTheBigIdea si = whatKind' (_kind si)
+  where
+    whatKind' :: SystemKind -> IdeaDict
+    whatKind' SRS = nw Doc.srs
+    whatKind' Notebook = nw Doc.notebook
+    whatKind' Website = mkIdea "website" (cn "website") (Just "web")
```
balacij: I wanted to make sure this stood out (because it's an oddity) -- hence: What's the big idea?!

code/drasil-system/package.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
 - pretty
 - lens
 - containers
+- drasil-data
```
balacij, line 21: I'm not sure how I feel about `drasil-system` being dependant on `drasil-data`, but it might mean that `drasil-data` could be split up.

code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -31,6 +34,19 @@ type Scope = [Sentence]
 -- | Project Example motivation.
 type Motivation = [Sentence]
 
+data SystemKind =
```
JacquesCarette, line 39: I'm thinking you should replace `SRS` by 2 things: `RunnableSoftware` and `Specification`.

code/drasil-system/lib/System/Drasil/System.hs
JacquesCarette: Having `System` be a data-structure that carries a payload is probably a good idea. It should likely be structured in such a way that the first piece of data is common to all kinds (as a bundle of data).

code/drasil-system/lib/System/Drasil/System.hs
JacquesCarette: It does stand out...

We shouldn't be making local `IdeaDict`s. If this concept is important, it should be in our database.

code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -39,13 +55,12 @@ data System where
...
 -- use a lookup of some sort from their internal (Drasil) ids.
  SI :: (CommonIdea a, Idea a,
-  Idea b,
   Quantity e, Eq e, MayHaveUnit e,
   Quantity h, MayHaveUnit h,
   Quantity i, MayHaveUnit i,
   HasUID j, Constrained j) => 
   { _sys         :: a
-  , _kind        :: b
+  , _kind        :: SystemKind
```
JacquesCarette, line 67: This does feel like an improvement! It adds more information into `_kind`.

code/drasil-system/package.yaml
JacquesCarette: This is why `drasil-metadata` needs to exist!!!

code/drasil-system/lib/System/Drasil/System.hs
balacij: Split in b78bd296db249b67f8f5e2ca286bbabb2a2a0203 -- Still a very simple implementation, I imagine we'll take multiple passes at moving things around. Also, I'm not using `RunnableSoftware` at all yet. Should it be used for the examples that currently support software generation? Or were you thinking that `System` could also replace `CodeSpec`?

code/drasil-system/package.yaml
balacij: Moved to `drasil-metadata` for the "new" concepts we need. I'll move the rest in a subsequent PR unless you prefer I do that in this one as well.

code/drasil-example/dblpend/package.yaml
```diff
@@ -30,6 +30,7 @@ dependencies:
 - drasil-lang
 - drasil-metadata
 - drasil-printers
+- drasil-system
```
balacij, line 33: I'm somewhat surprised that I need to add this to the other case studies. Things from `drasil-system` are being re-exported somewhere they shouldn't be.

code/drasil-metadata/lib/Drasil/Metadata.hs
```diff
@@ -0,0 +1,5 @@
+module Drasil.Metadata (
```
balacij, line 1: `drasil-metadata` didn't have a single module that re-exported things. So I created that here, and in a subsequent PR, will re-route the other exposed modules of `drasil-metadata` through this module.

code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -31,6 +36,21 @@ type Scope = [Sentence]
 -- | Project Example motivation.
 type Motivation = [Sentence]
 
+data SystemKind =
+    Specification
+  | RunnableSoftware
+  | Notebook
+  | Website
```
balacij, line 43: I won't add documentation to these until things are moved into their constructors // when we have good definitions.

code/drasil-website/lib/Drasil/Website/Body.hs
```diff
@@ -122,9 +122,8 @@ allRefs fl = [gitHubRef, wikiRef, infoEncodingWiki, chunksWiki, recipesWiki, pap
   ++ concatMap findAllRefs (sections fl)
 
 -- | Used for system name and kind inside of 'si'.
-webName, web :: CI
-webName = commonIdea "websiteName" (cn websiteTitle) "Drasil" []
-web = commonIdea "website" (cn "website") "web" []
+webName :: CI
+webName = commonIdea "websiteName" (cn websiteTitle) "Drasil" [] -- FIXME: Improper use of a `CI`.
```
balacij, line 126: Why improper use of a `CI`? Because it ignores what `CI`s are for. `CI`s are for declaring terminology with a known-existing abbreviation. Here, the noun phrase being declared is "Drasil - Generate all the things!" with the abbreviation "Drasil". The phrase is the title of Drasil's webpage. If anything, a more appropriate `_sys` would contain "Drasil's Website" or just "Drasil" (the fact that the renderer only pulls our a website is fine).

Also, it's not really used. It is only used in the `_sys` field of the `System` (see line no. 75), and if you change the content of `_sys` to `undefined :: CI`, the website still builds just fine, no diff.

code/drasil-system/lib/System/Drasil/System.hs
```diff
@@ -31,6 +36,21 @@ type Scope = [Sentence]
...
+
+whatsTheBigIdea :: System -> IdeaDict
+whatsTheBigIdea si = whatKind' (_kind si)
+  where
+    whatKind' :: SystemKind -> IdeaDict
+    whatKind' Specification = nw Doc.srs
+    whatKind' RunnableSoftware = runnableSoftware
+    whatKind' Notebook = nw Doc.notebook
+    whatKind' Website = website
```
balacij, line 52: I want to replace this with something else -- a typeclass instance perhaps, but I just don't know what yet.

`Idea` (term + abbreviation) might be appropriate. This would mean that the `System` is what the term/abbreviation "means." I think I'm fine with this for now.

Thoughts? Same PR or subsequent PR?

balacij (commented):
Quite a few subsequent PRs are needed! But that's good.

code/drasil-system/lib/System/Drasil/System.hs
balacij: Also, the `IdeaDict`s here are from `drasil-metadata` and from `drasil-data`. Hence why this PR still adds the dependancy to `drasil-system`. Moving `Doc.srs` and `Doc.notebook` (from `drasil-data` to `drasil-metadata`) will happen in a subsequent PR.

code/drasil-system/lib/System/Drasil/System.hs
JacquesCarette: My memory is that `CodeSpec` serves a different purpose. Yes, `RunnableSoftware` should be used for those examples.

code/drasil-system/package.yaml
JacquesCarette: incremental is fine.

code/drasil-system/lib/System/Drasil/System.hs
JacquesCarette: Can be subsequent PR.

JacquesCarette (approved):

Issue #4272, opened 2025-07-10
Replace `mkQuant` with variation that constructs DefinedQuantityDicts


jackwyand:
## `mkQuant`
One of the current QuantityDict constructors `mkQuant` should be replaced by an alternative that constructs DefinedQuantityDicts, working towards removing QuantityDict entirely. 
https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-lang/lib/Language/Drasil/Chunk/Quantity.hs#L64-L68
Currently, we have `dqd` which is one of the main constructors for DefinedQuantityDicts, and is the best candidate to replace `mkQuant` in my opinion.
https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-lang/lib/Language/Drasil/Chunk/DefinedQuantity.hs#L59-L61

However, `dqd` takes a ConceptChunk which wraps a lot of the information that `mkQuant` takes in separate arguments.

`mkQuant` takes a String for the UID, a NounPhrase, Symbol, Space, Maybe UnitDefn, and Maybe String for if it has an abbreviation, while `dqd` takes a ConceptChunk which would hold the UID, NounPhrase, and abbreviation, along with the added definition, and then the rest is analogous to `mkQuant`.

The main question I have about this ticket is if we need an entirely new function that acts similarly to `mkQuant` but for DefinedQuantityDicts, or if we should instead create the ConceptChunks separately for every DefinedQuantityDict and only use `dqd` (with it's variations depending on the use case).

## Fix

Either way, `mkQuant` will eventually need to be removed, so here is what should be done.

1. Fix any standalone uses of `mkQuant` where a new QuantityDict is constructed by giving it a definition
   - Only case of this is the following https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-example/pdcontroller/lib/Drasil/PDController/Unitals.hs#L179-L185

2. Fix constructors that have `mkQuant` nested in them by giving the chunks that they construct definitions as well
   - `unitary` makes use of `mkQuant` to construct a new UnitaryChunk https://github.com/JacquesCarette/Drasil/blob/141899aa79a44830c3b3bb28020ea93d1f0cf5e2/code/drasil-lang/lib/Language/Drasil/Chunk/Unitary.hs#L51-L55
[...]

PR #4275, opened 2025-07-11
Replace standalone usage of `mkQuant` with `dqd`

jackwyand:
Contributes to #4272

Since I replaced the term used for the original QuantityDict of the stiffness coefficient, the term for the existing concept is used instead. This omits the "of the spring" part, instead just stating the Stiffness Coefficient. This is what was already mainly used except for two use cases where the QuantityDict term was used instead. The alternative is making the concept term "Stiffness Coefficient of the spring" so that all uses of it include the "of the spring" part. This would have more changes, but may preferred over omitting it with less changes. Let me know what is better or if something else should be done instead.


smiths:
With respect to the changes to stable, it looks good to me.  The "of the spring" is not really necessary.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Unitals.hs
```diff
@@ -176,10 +176,8 @@ qdDampingCoeff
       symDampingCoeff
       Real
 
-qdStiffnessCoeff
-  = mkQuant "qdTimeConst"
-      (nounPhraseSent (S "Stiffness coefficient of the spring"))
+dqdStiffnessCoeff
+  = dqd ccStiffCoeff
```
JacquesCarette, line 182: I think we should leave a TODO comment to add back "of the spring" to the description. Hard to do right now, will become easier later.

JacquesCarette (changes requested):

code/drasil-example/pdcontroller/lib/Drasil/PDController/Unitals.hs
jackwyand: Done in [2a4598b](https://github.com/JacquesCarette/Drasil/pull/4275/commits/2a4598b62194d184335a194c79158c84a8d681ff).

Issue #4284, opened 2025-07-15
Replace `mkQuant'` QuantityDict constructor with DefinedQuantityDict constructor


jackwyand:
This issue is similar to #4272 in that we need to replace `mkQuant'` with a DefinedQuantityDict constructor variation.

### `implVar'`

The `implVar'` constructor nests `mkQuant'` to create a new QuantityDict variable.
https://github.com/JacquesCarette/Drasil/blob/7bb772731cdf3eb557fcccd19079e7feefcbf6ea/code/drasil-lang/lib/Language/Drasil/Chunk/Quantity.hs#L84-L90

In order to fix this, I propose passing in a Sentence to `implVar'` and constructing a new ConceptChunk, which would then be used to create a DefinedQuantityDict rather than a QuantityDict.

```haskell
implVar' :: String -> NP -> Sentence -> Maybe String -> Space -> Symbol -> 
  Maybe UnitDefn -> DefinedQuantityDict
implVar' s np d a t sym = dqd' (dccWDS s np d) f t 
  where f :: Stage -> Symbol
        f Implementation = sym
        f Equational = Empty
```

There is one issue with this however, the abbreviation is no longer captured since there is no ConceptChunk constructor that takes it. Creating a new constructor looks to be the easiest way to fix this as changing any of them would mean changing a lot of other code as well.

`implVar'` is used in only one location:
https://github.com/JacquesCarette/Drasil/blob/7bb772731cdf3eb557fcccd19079e7feefcbf6ea/code/drasil-lang/lib/Drasil/Code/CodeVar.hs#L121-L127

Changing `implVar'` to construct DefinedQuantityDicts will require changing the CodeChunk data type (constructed the value constructor `CodeC`). Currently CodeChunks take a QuantityDict, and it should be changed to use DefinedQuantityDicts. Doing so would require changing two other functions
https://github.com/JacquesCarette/Drasil/blob/7bb772731cdf3eb557fcccd19079e7feefcbf6ea/code/drasil-code/lib/Language/Drasil/Chunk/CodeBase.hs#L7-L13

I have not done as much investigation into how these would be changed, but I think changing the typeclass constraints to include Concept, and then using `dqdWr` (which has a similar structure to `qw`) to construct the DefinedQuantityDict is a step in the right direction. For example
```haskell
-- | Construct a 'CodeVarChunk' from a 'Quantity'.
quantvar :: (Quantity c, MayHaveUnit c, Concept c) => c -> CodeVarChunk
[...]

PR #4286, opened 2025-07-16
Use DefinedQuantityDicts in MultiDefns instead of QuantityDicts

jackwyand:
Contributes to #4284


jackwyand:
Notes:

- Force pushed since I wanted to add a new concept chunk constructor that also captures the abbreviations present in some of the QuantityDicts that were changed
     - After going through them I think they were all just `Nothing`, but it is good to have the new concept chunk constructor in case an abbreviation needs to be captured
- Since it is not being merged into main the CI did not run, however I tested it locally and everything seems fine

JacquesCarette (approved):

PR #4288, opened 2025-07-16
Refactor `drasil-code` to work with DefinedQuantityDicts

jackwyand:
Contributes to #4284 

Note: #4289 should be merged into this before this PR is merged, otherwise it will not compile. I just split up the changes to make it easier to understand and track all of them


jackwyand:
Should not be merged yet, still lots of work to do.

jackwyand:
I think that's okay, once #4289 is merged this should compile.

PR #4290, opened 2025-07-19
Remove `allInDB` function in favour routing TMs, GDs, etc. through `System`.

balacij:
Contributes to #2873 

Removes `allInDB`, a function previously used to gather all TMs, GDs, IMs, and DDs from the `ChunkDB` for rendering in the SRS.

Why? Because "which of these things to print" should not be based on what things are inside of the `ChunkDB`. We should be able to fairly arbitrarily throw more things in a `ChunkDB` without affecting what it generates.


balacij:
Looks good now, @JacquesCarette ! Thanks!

code/drasil-gen/lib/Language/Drasil/TypeCheck.hs
```diff
@@ -9,11 +9,11 @@ import Data.Either (isRight, rights)
 import Control.Lens ((^.))
 import Data.Bifunctor (second)
 import Data.List (partition)
-import System.Drasil (System(SI))
+import System.Drasil (System(SI, _instModels, _dataDefns, _systemdb))
```
JacquesCarette: We never export _ names. If they have an underscore in their name, it's because they are supposed to be internal. If you're going to use them directly as getters and not derive lenses for them, remove the _ .

JacquesCarette: Since I see that lenses are derived, this means you should use them, not the field accessors.

balacij: Fixed in 130e76ed4246989765cf1b7440ccadb7f226df2b

I originally avoided because I was expecting the `SystemKind` re-organizing to break our flow with lenses anyways, but I guess that's not really as problematic as I was thinking.

balacij:
Good for re-review now, @JacquesCarette !

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -97,26 +97,28 @@ mkSRS = [TableOfContents,
...
-  _constraints = inputConstraints,
-  _constants   = [],
-  _systemdb   = symbMap
+  _quants       = [] :: [QuantityDict],
+  _theoryModels = tMods,
+  _genDefns     = generalDefns,
+  _instModels   = iMods,
+  _dataDefns    = dataDefs,
+  _configFiles  = [],
+  _inputs       = inputSymbols,
```
balacij, line 117: In a follow-up PR, I will switch things to a smart constructor.

balacij:
If/when this is merged, I'll create a smart constructor for `System` as well. Too much re-merging will be necessary otherwise.

PR #4291, opened 2025-07-21
drasil-metadata: Rejig exports, organization, and clean up documentation

balacij:


balacij:
A little bit annoyed that I accidentally wrote one commit for the first, but I don't see a clean way to split it.

Force-pushed to fix the documentation once in the second commit, and update the commit message in the first.

PR #4297, opened 2025-07-23
Wiki: Explain how to acknowledge co-authors.

balacij:
I'm happy to write a bit more, but this should be good enough to get started.


smiths:
Good thinking to add this information @balacij.

Issue #4190, opened 2025-06-16
`symbols` vs `symbolsAll`: What?


balacij:
In many of our examples, we seem to have two lists: `symbols` and `symbolsAll` -- Why? What's the difference?

Related to #2873 

@jackwyand Since we're probably both going to reach this at some point, I'm assigning it to both of us.

JacquesCarette:
`git blame` ?

balacij:
For later:

With SWHS, this patch does not affect `stable/`, passes our tests, and does nothing but remove `symbols` for `symbolsAll` (which contains all of `symbols` and more).

```patch
diff --git a/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs b/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs
index c2de3dd752..fcb92fd751 100755
--- a/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs
+++ b/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs
@@ -52,7 +52,7 @@ import Drasil.SWHS.Requirements (funcReqs, inReqDesc, nfRequirements,
 import Drasil.SWHS.TMods (tMods)
 import Drasil.SWHS.Unitals (coilHTC, coilSA, consTol, constrained,
   htFluxC, htFluxP, inputs, inputConstraints, outputs, pcmE, pcmHTC, pcmSA,
-  simTime, specParamValList, symbols, symbolsAll, tempC, tempPCM,
+  simTime, specParamValList, symbolsAll, tempC, tempPCM,
   tempW, thickness, watE)
 
 import System.Drasil (SystemKind(Specification), mkSystem)
@@ -75,7 +75,7 @@ si :: System
 si = mkSystem
   progName' Specification [thulasi, brooks, spencerSmith]
   [purp] [] [scope] [motivation]
-  symbols
+  symbolsAll
   tMods genDefs SWHS.dataDefs iMods
   []
   inputs outputs constrained specParamValList
@@ -103,7 +103,7 @@ conceptChunks =
   algorithm : thermocon ++ softwarecon ++ physicCon ++ mathcon ++
   physicalcon ++ con ++ srsDomains ++
   -- DefinedQuantityDicts
-  map cw symbols ++
+  map cw symbolsAll ++
   -- ConstQDefs
   map cw specParamValList
 
@@ -116,7 +116,7 @@ abbreviationsList =
   -- CIs
[...]

PR #4298, opened 2025-07-24
Use `defResolve` to get code comment descriptions instead of manually searching concept chunk table

jackwyand:
Contributes to #4126

I am currently having a problem with this PR as it revealed an interesting issue. This function is where I think it is originating from: https://github.com/JacquesCarette/Drasil/blob/fb8a7fb07be04af1a49b3f1615b947028a2b787f/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs#L576-L580

Now that `defResolve` is being used to query the chunk tables for the code comments, it is giving an error because `diffCodeChunk` is creating new chunks with different UIDs (suffixing an existing one with "d"), but not adding them to any of the chunk tables. Therefore attempting to build any examples that use this return an error along the lines of "`Chunkd` not found in ConceptMap" (some existing chunk with "d" at the end). I need to somehow inject the newly created chunks from this function into the chunk tables, however I am unsure as to how I would do this.

For reference, `diffCodeChunk` is used in the ODE `ExternalLibraryCall`s in `/OdeLibraries.hs`, e.g., (Ln 479) https://github.com/JacquesCarette/Drasil/blob/fb8a7fb07be04af1a49b3f1615b947028a2b787f/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs#L460-L479

Maybe @JacquesCarette has any ideas? Let me know if more information is needed.


balacij:
Aside: It does look like "code chunks" are a good idea. I'm not quite a fan of them creating DQDs, they should probably have their own type and associated type system.

It would be nice if we could create fresh variables dynamically in GOOL, it is helpful. I'm just not immediately sure of how to do it without looking into GOOL. @sarrasoussia Did you happen to find any examples of these when you were looking into GOOL?

balacij:
Coincidentally, I was looking at #4190 yesterday and that was what I found as well (though I was only focused on `symbolsAll`). I think that approach is fine for now. If you could add explicit comments in the code that explains why the chunks are added and file a ticket to fix this the "right way" (i.e., where we don't assume a UID "+d suffix means that it represents a 'change of X' scheme), that would be best.

JacquesCarette:
I agree that this approach is fine for now (and that it's not how we want to do things longer term).

code/drasil-code/lib/Language/Drasil/Code/CodeQuantityDicts.hs
```diff
@@ -5,22 +5,26 @@ import Language.Drasil
...
   "structure holding the constant values"
   "structures holding the constant values")
-  (S "the structure holding the constant values"))
-  (label "consts") (Actor "Constants")
+  (S "the structure holding the constant values")
+  (Actor "Constants") (label "consts")
+
+
+codeDQDs :: [DefinedQuantityDict]
+codeDQDs = [inFileName, inParams, consts]
```
balacij, line 30: It might be good to add this to the basis CDB rather than each example individually?

code/stable/dblpend/src/cpp/Calculations.hpp
```diff
@@ -14,12 +14,12 @@
 
 using std::vector;
 
-/** \brief Calculates dependent variables (rad)
-    \param m_1 mass of the first object (kg)
-    \param m_2 mass of the second object (kg)
-    \param L_2 length of the second rod (m)
-    \param L_1 length of the first rod (m)
-    \return dependent variables (rad)
+/** \brief Calculates dependent variables: column vector of displacement of rods with its derivatives (rad)
```
balacij: Parsing this is difficult, but a tangential issue to this PR.

code/stable/dblpend/src/cpp/Calculations.hpp
```diff
@@ -14,12 +14,12 @@
...
+    \param m_1 mass of the first object: the mass of the first object (kg)
+    \param m_2 mass of the second object: the mass of the second object (kg)
+    \param L_2 length of the second rod: the length of the second rod (m)
+    \param L_1 length of the first rod: the length of the first rod (m)
+    \return dependent variables: column vector of displacement of rods with its derivatives (rad)
```
balacij: I never realized this information was repeated. Perhaps we can remove the description from the `\brief`? This would be a subsequent PR, and up to @JacquesCarette and @smiths .

code/stable/dblpend/src/cpp/Calculations.hpp
```diff
@@ -14,12 +14,12 @@
...
-    \param m_1 mass of the first object (kg)
-    \param m_2 mass of the second object (kg)
-    \param L_2 length of the second rod (m)
-    \param L_1 length of the first rod (m)
-    \return dependent variables (rad)
+/** \brief Calculates dependent variables: column vector of displacement of rods with its derivatives (rad)
+    \param m_1 mass of the first object: the mass of the first object (kg)
+    \param m_2 mass of the second object: the mass of the second object (kg)
+    \param L_2 length of the second rod: the length of the second rod (m)
+    \param L_1 length of the first rod: the length of the first rod (m)
```
balacij: Without thinking for too long:

1. Trying hard to force a diff.: `first object's mass: the mass of the first object` -- still verbose, but ever so slightly better.
2. One somewhat suggestion is to check if the term and the definitions are similar, and if so, to not show the definition.
3. Going back a step: Should all variables have an English definition? I still think so, and we previously discussed such as well. What this variable really is, is a (ontological) pointer, so maybe this is just the best we can do and this is what truly makes sense?

code/stable/dblpend/src/java/DblPend/InputParameters.java
```diff
@@ -13,12 +13,12 @@
...
+        \param filename name of the input file: the name of the input file
         \return array containing the following values:
-        \return length of the first rod (m)
-        \return length of the second rod (m)
-        \return mass of the first object (kg)
-        \return mass of the second object (kg)
+        \return length of the first rod: the length of the first rod (m)
+        \return length of the second rod: the length of the second rod (m)
+        \return mass of the first object: the mass of the first object (kg)
+        \return mass of the second object: the mass of the second object (kg)
```
balacij: This is not technically what's being returned. What's being returned is an `Object[]` which contains this information. Still technically distinct.

code/stable/dblpend/src/java/DblPend/InputParameters.java
```diff
@@ -13,12 +13,12 @@
 public class InputParameters {
     
     /** \brief Reads input from a file with the given file name
-        \param filename name of the input file
+        \param filename name of the input file: the name of the input file
```
balacij: "a filepath, absolute or relative, to the file containing the program's inputs"?

code/stable/glassbr/src/cpp/Calculations.hpp
```diff
@@ -14,77 +14,77 @@
...
-    \param inParams structure holding the input values
+    \param inParams structure holding the input values: the structure holding the input values
     \param NFL non-factored load: three second duration uniform load associated with a probability of breakage less than or equal to 8 lites per 1000 for monolithic AN glass (Pa)
     \return load resistance: the uniform lateral load that a glass construction can sustain based upon a given probability of breakage and load duration as defined in (pp. 1 and 53) Ref: astm2009 (Pa)
 */
 double func_LR(InputParameters &inParams, double NFL);
 
 /** \brief Calculates probability of breakage: the fraction of glass lites or plies that would break at the first occurrence of a specified load and duration, typically expressed in lites per 1000 (Ref: astm2016)
-    \param B risk of failure
+    \param B risk of failure: the risk of failure
```
balacij: "the percentage risk of glass slab failing to resist blast"?

Speaking of "failure", there's no definition of it in the [GlassBR SRS](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/HTML/GlassBR_SRS.html). We should probably have one that can be expanded here.

balacij (commented):
The code's looking great! It definitely brings up a lot of good issues.

code/stable/dblpend/src/cpp/Calculations.hpp
JacquesCarette: I agree, we should remove this second part from the `\brief`. I think we should do it for this PR.  I don't want to "bless" these quite awful descriptions as 'the right answer'.

code/stable/dblpend/src/cpp/Calculations.hpp
JacquesCarette: Tangential, but if we don't record it somewhere that this needs looked at, we won't! Looks like double pendulum needs a very serious review.

code/stable/dblpend/src/cpp/Calculations.hpp
JacquesCarette: The first bit should "what it is" (so 'mass of the first object' is fine) while the second should be a definition of that. That seems hard/pointless. In other words, it feels like having no definition is not so evil after all.

code/stable/glassbr/src/cpp/Calculations.hpp
```diff
@@ -14,77 +14,77 @@
 using std::ofstream;
 using std::string;
 
-/** \brief Calculates tolerable stress distribution factor
-    \param inParams structure holding the input values
-    \return tolerable stress distribution factor
+/** \brief Calculates tolerable stress distribution factor: the tolerable stress distribution factor
```
JacquesCarette: This all continues to be ridiculous, for all of `\brief`, `\param` and `\return`.  We should definitely not put that information there. [And the information that is there is quite bad.]

JacquesCarette (changes requested):
I did spot checks for a lot of files. While the code changes are fine, these stable changes are horrible.

We shouldn't put that information out. Let's see what that changes.

code/stable/dblpend/src/cpp/Calculations.hpp
jackwyand: I managed to do this in [331cc81](https://github.com/JacquesCarette/Drasil/pull/4298/commits/331cc8118a7990e8ea925f31aae5b99bcc6cdb65), let me know what you think.

code/stable/dblpend/src/cpp/Calculations.hpp
JacquesCarette: Oh, it looks like not just `\brief` is affected but `\param` too! These descriptions are also bad. Looks like the definition should not be added here either.

code/stable/dblpend/src/cpp/Calculations.hpp
```diff
@@ -15,11 +15,11 @@
...
-    \param m_1 mass of the first object (kg)
-    \param m_2 mass of the second object (kg)
-    \param L_2 length of the second rod (m)
-    \param L_1 length of the first rod (m)
-    \return dependent variables (rad)
+    \param m_1 mass of the first object: the mass of the first object (kg)
+    \param m_2 mass of the second object: the mass of the second object (kg)
+    \param L_2 length of the second rod: the length of the second rod (m)
+    \param L_1 length of the first rod: the length of the first rod (m)
+    \return dependent variables: column vector of displacement of rods with its derivatives (rad)
```
JacquesCarette: And indeed `\return` as well, though in this case it looks like it is the chunk itself that is odd (too generic?)

code/stable/dblpend/src/java/DblPend/InputParameters.java
JacquesCarette: Can you make a bug report about that? This is indeed an incorrect description.

JacquesCarette:
build failure - forgot to stabilize?

code/stable/dblpend/src/java/DblPend/InputParameters.java
balacij: Moved to https://github.com/JacquesCarette/Drasil/issues/4331 ! Sorry for the delay!

PR #4304, opened 2025-07-28
Switch "runnable case studies" to RunnableSoftware Systems rather than Specification systems

balacij:
Contributes to #3260

### DIFF
--- a/code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
+++ b/code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
@@ -8,6 +8,8 @@ import Language.Drasil.Chunk.Concept.NamedCombinators
 import qualified Language.Drasil.NounPhrase.Combinators as NP
 import qualified Language.Drasil.Sentence.Combinators as S
 import qualified Drasil.DocLang.SRS as SRS
+import Theory.Drasil (TheoryModel)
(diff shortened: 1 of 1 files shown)


balacij:
Thinking out loud a bit more regarding:
```haskell
data RequirementsSpecification = ?

data SolutionSpecification = ?
```

`RequirementsSpecification` would pull in things like IMs, DDs, GDs, etc. from `System`, and `SolutionSpecification` would pull in things like `configFiles` (a list of file names -- these are used in GlassBR to give it some usable run data, but I'm not entirely sure of all the specifics, so this is wishy-washy), `Choices`, etc.

JacquesCarette:
I agree that something along the lines of
```haskell
data SystemKind =
    Specification RequirementsSpecification
  | RunnableSoftware RequirementsSpecification SoftwareSpecification
  | Notebook
  | Website
```
is the way to go. Each system kind has its own set of choices -- I'm sure `Notebook` and `Website` will grow to have some too!

JacquesCarette (dismissed):
Definitely agree that those 5 examples should all be `RunnableSoftware`.

JacquesCarette (dismissed):
This now seems good.

JacquesCarette:
Well darn, I messed up the merge. I shouldn't use the web editor to do merges. Will fix.

PR #4305, opened 2025-07-29
Cleaning up type checking code a bit

balacij:
Contributes to #3788


code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -317,178 +326,182 @@ instance Typed Expr Space where
           = either
               (\_ -> all (\ r -> length r == columns && all (== expT) r) sss)
               (const False) expT
-        t = fromLeft (error "Infer on Matrix had a strong expectation of Left-valued data.") expT -- This error should never occur.
-
-  infer _ (Set s _) = Left s
```
balacij, line 322: This was obviously wrong, which is why `minThick` didn't have a typing issue. The type-checker finds the issue now. The solution is as simple as changing `Rational` in all of:

```haskell
nomThick = cuc' "nomThick"
  (nounPhraseSent $ S "nominal thickness t is in" +:+ eS (mkSet Rational (map dbl nominalThicknesses)))
  "the specified standard thickness of the glass plate" lT millimetre
  {-Discrete nominalThicknesses, but not implemented-} Rational
  [sfwrElem $ mkSet Rational (map dbl nominalThicknesses)] $ exactDbl 8 -- for testing
```

to `Real` (because part of `minThick`'s definition uses `nomThick`).

Subsequent PR or here?

balacij: Subsequent work should also involve checking why `Set` had a type to begin with.

code/drasil-lang/lib/Language/Drasil/Space.hs
```diff
@@ -107,3 +111,60 @@ isBasicNumSpace Array {}     = False
...
+assertRealVector (Vect Real) _   = Right ()
+assertRealVector t           msg = Left $ msg $ show t
+
+assertFunction :: Space -> (String -> String) -> Either String (NE.NonEmpty Primitive, Primitive)
+assertFunction (Function params out) _   = Right (params, out)
+assertFunction s                     msg = Left $ msg $ show s
+
+assertNonFunction :: Space -> (String -> String) -> Either String ()
+assertNonFunction f@Function{} msg = Left $ msg $ show f
+assertNonFunction _            _   = Right ()
```
balacij: For things with data to unpack, I unpacked in the `Right`, otherwise, `()`. I will make sure to add this note to the comments somewhere.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -232,82 +237,93 @@ instance LiteralC Expr where
...
+
+  lsp <- assertNumericVector lt $ msg "left"
+  rsp <- assertNumericVector rt $ msg "right"
+
+  if op == VSub then
+    assertNonNatNumeric lsp $ \sp ->
+      "Vector subtraction expects both operands to be vectors of non-natural numbers. Received `" ++ sp ++ "`."
+  else Right ()
+
+  lsp ~== rsp $ \lt' rt' -> "Vector " ++ pretty op ++ " expects both operands to be of the same numeric type. Received `" ++ lt' ++ "` and `" ++ rt' ++ "`."
```
balacij, line 256: I'm not sure if I'm a fan of (~==) -- from an "aesthetics" point of view, considering all other "asserts" are verbosely named.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -232,82 +237,93 @@ instance LiteralC Expr where
...
+  infer :: TypingContext Space -> Expr -> Either TypeError Space
   infer cxt (Lit lit) = infer cxt lit
 
-  infer cxt (AssocA _ (e:exs)) = 
+  infer cxt (AssocA _ (e:exs)) = do
+    et <- infer cxt e
+    assertNumeric et $
+      \sp -> "Associative arithmetic operation expects numeric operands, but found `" ++ sp ++ "`."
+    assertAllEq cxt exs et
+        "Associative arithmetic operation expects all operands to be of the same type."
```
balacij, line 272: I need to fix this error message.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -317,178 +333,183 @@ instance Typed Expr Space where
...
+    vet <- assertVector lt (\sp -> "List accessor expects a vector, but received `" ++ sp ++ "`.")
+
+    rt <- infer cxt n
+    assertIndexLike rt
+      (\sp -> "List accessor expects an index-like type (Integer or Natural), but received `" ++ sp ++ "`.")
+
+    pure vet
+
+  -- TODO: Why do we have *both* Index and IndexOf?
+  infer cxt (LABinaryOp IndexOf l n) = infer cxt (LABinaryOp Index l n)
```
balacij: A thing to follow-up on later.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -317,178 +333,183 @@ instance Typed Expr Space where
...
+    pure rt
+
+  infer cxt (ESBBinaryOp SContains l r) = do
+    lt <- infer cxt l
+    rt <- infer cxt r
+    set <- assertSet rt
+      (\sp -> "Set contains expects a set on the right-hand side, but found `" ++ sp ++ "`.")
+    assertEquivNumeric lt set
+      (\ls rs -> "Set contains should only be applied to Set of numeric type. Received `" ++ ls ++ "` / `" ++ rs ++ "`.")
+    pure S.Boolean
```
balacij, line 484: I don't actually see why for `Set`s in particular, its reasonable to assume they're always sets of numbers, but I followed logic. A subsequent PR can alter the rule.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -317,178 +333,183 @@ instance Typed Expr Space where
...
+    let msg dir sp = "'Big' operator range " ++ dir ++ " is not an index-like type (Integer or Natural), found: " ++ sp ++ "."
+
+    assertIndexLike botTy (msg "start")
+    assertIndexLike topTy (msg "stop")
+
+    assertEquivNumeric botTy topTy
+      (\ls rs -> "'Big' operator range expects start and stop to be of the same numeric type, but found `" ++ ls ++ "` and `" ++ rs ++ "`.")
+
+    -- FIXME: We have a `Symbol` in the `S.BoundedDD` but it's not used in type-checking.
+    pure bodyTy
```
balacij, line 503: `Operator` might need to be redesigned.

balacij (commented):
I think it's ready for your review, @JacquesCarette .

Though, if you end up being okay with the PR as-is, just don't merge it yet because I want to clean the git history and address my own comments.

balacij:
Finally... last comment before I get back to programming

### 3 (final)

```plaintext
161,162c156,161
< `theory:FS` OK!
< `theory:nrmShrForIM` OK!
---
> `theory:FS` exposes ill-typed expressions!
>   - ERROR: Associative arithmetic operation expects all operands to be of the same expected type.
>     Received:
>     - ERROR: 'Big' operator range body not Integer, found: Real.
> `theory:nrmShrForIM` exposes ill-typed expressions!
>   - ERROR: 'Big' operator range body not Integer, found: Real.
```

Relevant type-checking code (before):

https://github.com/JacquesCarette/Drasil/blob/2bde1fec7914e40d5b5620ea5e15c759d8348252/code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs#L460-L472

Relevant type-checking code (after):

https://github.com/JacquesCarette/Drasil/blob/e8048a81cc8ab97bb3b45ed7e91d007318e7d5c2/code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs#L478-L495

I just loosened the requirements on the range body. There's no real (ha ha) reason why the body should be `Integer`-typed.

Code related to `theory:FS`:

https://github.com/JacquesCarette/Drasil/blob/2bde1fec7914e40d5b5620ea5e15c759d8348252/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs#L53-L68

It looks fine to me.

Code related to `theory:nrmShrForIM`:

https://github.com/JacquesCarette/Drasil/blob/2bde1fec7914e40d5b5620ea5e15c759d8348252/code/drasil-example/ssp/lib/Drasil/SSP/IMods.hs#L374-L388

Again, it looks fine to me.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: I kind of like it!

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: To what? It's not a bad error, though it could be more specific.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: I'm fine either way wrt `Rational` -> `Real`.

We wanted 'typed sets' since these were meant to be more CS-y sets.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: I think one pulls out an index from an indexed expression, while the other is an index. We need better comments!

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
JacquesCarette: Because I think that's our only use case for now. It's not intrinsic.

balacij:
> In the log changes, the cosmetic:
>
> 12c11
> <   - ERROR: Associative arithmetic operation expects all operands to be of the same type.
> ---
> >   - ERROR: Associative arithmetic operation expects all operands to be of the same expected type.
>
> seems not so good, i.e. I don't see the addition of the word 'expected' helps the user in any ways. I did not immediately spot where this came from?

Fixed in 533cef7234eb5ec72b408ee1aca3ac356313d3cd now.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
balacij: Thanks! Fixed in 857a77af6461e85098be21d6ee57695855cd4012 and 67f235e9ff81b9ddd4c51ecfa05aad3e91f079a9 now.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
balacij: You're right, and after our latest discussion, we're going to need to rewrite this anyways. I also forgot that the types are dumped as well:

```haskell
assertAllEq :: Typed e t => TypingContext t -> [e] -> t -> TypeError -> Either TypeError ()
assertAllEq cxt es expect s
  | allTsAreSp = pure ()
  | otherwise  = Left $ temporaryIndent "  " (s ++ "\nReceived:\n" ++ dumpAllTs)
  where
    allTs = map (infer cxt) es
    allTsAreSp = all (\case
      Right  t -> t == expect
      Left _   -> False) allTs
    dumpAllTs = intercalate "\n" $ map (("- " ++) . either ("ERROR: " ++) show) allTs
```

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -232,82 +238,93 @@ instance LiteralC Expr where
...
 
-  infer cxt (AssocA _ (e:exs)) = 
+  infer cxt (AssocA _ (e:exs)) = do
+    et <- infer cxt e
+    assertNumeric et $
+      \sp -> "Associative arithmetic operation expects numeric operands, but found `" ++ sp ++ "`."
+    assertAllEq cxt exs et
+        "Associative arithmetic operation expects all operands to be of the same type."
+    pure et
+  infer _ (AssocA Add _) = Left "Associative addition requires at least one operand."
```
JacquesCarette, line 274: Actually, a 0-ary addition is '0', so this could be legal. Biggest issue is that we can't `infer` such a thing, we can only `check` it. Which makes me think: should `AssocA` really be an `infer` type?

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
```diff
@@ -232,82 +238,93 @@ instance LiteralC Expr where
...
+  infer _ (AssocA Mul _) = Left "Associative multiplication requires at least one operand."
```
JacquesCarette, line 275: And a 0-ary multiplication is '1'.

JacquesCarette (changes requested):
One nit-picky thing, otherwise this looks great.

code/drasil-lang/lib/Language/Drasil/Expr/Lang.hs
balacij: Yes, we should probably change this to a `check`, and I think we need to review all the type checking rules to really make use of a bidirectional type checker. We only `infer` right now, really.

balacij: Subsequent work, I assume? I will file a ticket and work on it.

JacquesCarette: Yes. So I'll remove my 'changes requested'.

JacquesCarette (approved):

Issue #4299, opened 2025-07-24
Automatically collect abbreviations for SRS


balacij:
Context: https://github.com/JacquesCarette/Drasil/issues/4207#issuecomment-2985573121

> Have to wonder: what if we scanned the SRS, recursing through its pieces, how close would we get to gathering everything?

Our generated SRSs contain a Table of Abbreviations and Acronyms. Currently, said tables rely on manually collected lists which are passed to the SRS format declaration:

https://github.com/JacquesCarette/Drasil/blob/fb8a7fb07be04af1a49b3f1615b947028a2b787f/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L160-L167

https://github.com/JacquesCarette/Drasil/blob/fb8a7fb07be04af1a49b3f1615b947028a2b787f/code/drasil-example/ssp/lib/Drasil/SSP/Body.hs#L84

The goal of this ticket is to remove these manually collected lists in favour of automatically collecting them from the generated SRS documents.

There are two ways we can do this:

1. (Simpler, less work) Traversing over generated SRS documents and scanning for acronyms placed. -- https://github.com/JacquesCarette/Drasil/issues/4207#issuecomment-2985573121
2. (Much more work) Switching the renderers to stateful ones that automatically collect a list of abbreviations and acronyms used in the document for later placing in the table of abbreviations and acronyms. -- https://github.com/JacquesCarette/Drasil/issues/4207#issuecomment-2991968653

(1) is best for now -- https://github.com/JacquesCarette/Drasil/issues/4207#issuecomment-2992367161 . However, if #3864 is completed before anyone attempts this ticket, (2) might not be too difficult either.

JacquesCarette:
I agree with everything above.

PR #4323, opened 2025-08-10
Auto-collect abbreviations for SRS Table of Abbreviations and Acronyms

sarrasoussia:
Replaces manual abbreviation lists with automatic extraction from document content. Removes abbreviationsList from DblPend example and implements extractAbbreviations().

Working on #4299


code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -193,6 +193,10 @@ fillReqs (_:xs) si = fillReqs xs si
 extractUnits :: DocDesc -> ChunkDB -> [UnitDefn]
 extractUnits dd cdb = collectUnits cdb $ ccss' (getDocDesc dd) (egetDocDesc dd) cdb
 
+-- | Extracts abbreviations/acronyms found in the document description ('DocDesc') from a database ('ChunkDB').
+extractAbbreviations :: DocDesc -> ChunkDB -> [IdeaDict]
+extractAbbreviations dd cdb = map nw $ collectAbbreviations $ ccss' (getDocDesc dd) (egetDocDesc dd) cdb
```
balacij: This looks like the right approach as per #4299. But, I think that `ccss'` only looks for DQDs:

```haskell
-- | Gets a list of defined quantities ('DefinedQuantityDict's) from 'Sentence's and expressions that are contained in the database ('ChunkDB').
ccss :: [Sentence] -> [ModelExpr] -> ChunkDB -> [DefinedQuantityDict]
ccss s e c = nub $ concatMap (`combine` c) s ++ concatMap (`combine'` c) e

-- | Gets a list of quantities ('DefinedQuantityDict's) from 'Sentence's and expressions that are contained in the database ('ChunkDB').
ccss' :: [Sentence] -> [ModelExpr] -> ChunkDB -> [DefinedQuantityDict]
ccss' s e c = nub $ concatMap (`vars'` c) s ++ concatMap (`vars` c) e
```

And `egetDocDesc` also only looks for expressions.

You'll need to reimplement something equivalent to these that looks for all `Ch` `Sentence` constructors and gathers the `UID`s referenced.

```haskell
data Sentence where
  -- This allows Sentences to hold plural forms of 'NounPhrase's and 'NamedIdea's.
  Ch    :: SentenceStyle -> TermCapitalization -> UID -> Sentence
```

sarrasoussia:
> @sarrasoussia your changes remove the current abbreviations and acronyms and replace them with "There are no abbreviations and acronyms.". We do not want to lose the previous information, like what SRS stands for etc.

I believe I understood the issue. In the original code, there was a manual acronyms list in `Unitals.hs` that contained the abbreviations like `twoD` or  `srs` but my new implementation doesn't use this, it tries to extract from the document itself.

I need to combine both approaches, keep the existing abbreviations that are defined in the chunks and add any additional abbreviations found through document traversal

sarrasoussia:
I'm currently resolving type-checking errors that emerged after migrating all examples from manual abbreviation lists to automatic abbreviation extraction (TAandA []).

- Type errors in swhs/swhsnopcm (vector/scalar confusion)
- Function application errors in pdcontroller
- Mathematical expression types in dblpend, ssp, gamephysics
- Symbol duplication warnings

sarrasoussia:
I completed the TAandA migration successfully (automatic abbreviation extraction now works), but some examples are showing type errors that seem unrelated to my changes. Should I fix these pre-existing type issues or focus only on the abbreviation work?

smiths:
@sarrasoussia you need a clean CI with your changes.  We don't need to look at existing problems that we considered as passing the CI in the past, but your changes should not introduce new errors.  It looks like your CI is failing because stable has not been updated.   However, you really shouldn't have changes in stable (except maybe minor ones) because the target is the same generated files as before; you are just getting to the target in a better way.  :smile:

balacij:
> Should I fix these pre-existing type issues or focus only on the abbreviation work?

Leaving this for another PR is probably best.

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -58,7 +58,7 @@ resourcePath = "../../../../datafiles/gamephysics/"
 
 mkSRS :: SRSDecl
 mkSRS = [TableOfContents,
-  RefSec $ RefProg intro [TUnits, tsymb tableOfSymbols, TAandA abbreviationsList],
+  RefSec $ RefProg intro [TUnits, tsymb tableOfSymbols, TAandA []],
```
balacij, line 61: I think you can remove the list paramater for `TAandA` now too!

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -142,13 +142,6 @@ symbMap :: ChunkDB
 symbMap = cdb symbolsAll ideaDicts conceptChunks
   ([] :: [UnitDefn]) dataDefs iMods generalDefns tMods concIns [] allRefs citations
 
-abbreviationsList :: [IdeaDict]
-abbreviationsList =
-  -- QuantityDicts
-  map nw symbolsAll ++
-  -- CIs
-  map nw acronyms
```
balacij, line 148: Nice! 😄

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -193,6 +193,36 @@ fillReqs (_:xs) si = fillReqs xs si
...
+getAllChunksFromDoc dd cdb = mapMaybe (lookupConceptChunk cdb) $ nub $ concatMap getSentenceUIDs (getDocDesc dd)
+
+-- | Helper to lookup a ConceptChunk by UID from the database.
+lookupConceptChunk :: ChunkDB -> UID -> Maybe ConceptChunk
+lookupConceptChunk cdb chunkUID = fst <$> Map.lookup chunkUID (conceptChunkTable cdb)
+
+-- | Extracts all UIDs from Ch constructors in a list of sentences.
+getSentenceUIDs :: Sentence -> [UID]
+getSentenceUIDs (Ch _ _ chunkUID) = [chunkUID]
+getSentenceUIDs (SyCh chunkUID) = [chunkUID]
```
balacij: I _think_ you can ignore the `UID` of `SyCh`, no? This is for symbols, right?

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -299,6 +299,13 @@ collectUnits :: Quantity c => ChunkDB -> [c] -> [UnitDefn]
 collectUnits m = map (unitWrapper . flip unitLookup (m ^. unitTable))
  . concatMap getUnits . mapMaybe (getUnitLup m)
 
+-- | Gets abbreviations/acronyms from a list of chunks that have abbreviations.
+collectAbbreviations :: Idea c => [c] -> [c]
+collectAbbreviations = filter hasAbbreviation
+  where hasAbbreviation c = case getA c of
+                              Nothing -> False
+                              Just _  -> True
```
balacij: I think this should be kept in `-docLang/.../DocumentLanguage.hs` next to your other changes. [`mapMaybe`](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Maybe.html#v:mapMaybe) might also benefit you if you intend to use `getA` right after anyways.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -193,6 +193,36 @@ fillReqs (_:xs) si = fillReqs xs si
 extractUnits :: DocDesc -> ChunkDB -> [UnitDefn]
 extractUnits dd cdb = collectUnits cdb $ ccss' (getDocDesc dd) (egetDocDesc dd) cdb
 
+-- | Extracts abbreviations/acronyms found in the document description ('DocDesc') from a database ('ChunkDB').
+extractAbbreviations :: DocDesc -> ChunkDB -> [IdeaDict]
+extractAbbreviations dd cdb = map nw $ collectAbbreviations $ getAllChunksFromDB cdb ++ getAllChunksFromDoc dd cdb
```
balacij: Instead of extracting `IdeaDict`s, it might be good to extract `TermAbbr`s.

balacij (changes requested):
Only a few minor notes :)

sarrasoussia:
> [740a7f1](https://github.com/JacquesCarette/Drasil/commit/740a7f1ec820fbe0402a4f7116a94a4681e61268) Your latest commit looks like it's causing a lot of unintended deletions. The diff in the PR is now `+144 −187,754` -- it looks like you accidentally deleted `stable/`? Is this happening through the `Makefile` out of curiosity? I feel like this has happened before -- are you stabilizing using `make stabilize`?

This is actually weird bc I did stabilize and I thought I checked the commit before. Ill recheck

balacij:
Did you use `make stabilize`? If so, can you show me what commands it executed?

balacij:
Ok, good. I was just worried that the `Makefile` was orchestrating some sort of dangerous actions that erased `stable/` because this PR erases `stable/` entirely and the same confusion with the Vectors PR.

JacquesCarette:
There is still something really weird going on here. It still looks like everything is being renamed (including all of `stable/`). 1009 files changed? Something is still wrong for sure.

balacij:
Unfortunately, it looks like the issue with the `main` merge has become tricky. Unravelling the git history is going to be very difficult. To get things back on track, I've created an alternative branch, [`tableAbbAccS`](https://github.com/JacquesCarette/Drasil/compare/tableAccAbbS). Rather than `git revert`ing the most recent problematic `main` merge (50c12475f12270d03ea798f2e4ee4573e776c260) as we tried, I've dropped that commit altogether in the new branch and then re-applied (via `cherry-pick`ing) your commits (which were good changes, by the way).

However, some code reversions from previous merges still exist in the new branch. The next step will be to manually fix these. The key with merge conflicts is to reconcile code changes, carefully mixing code from both branches, rather than accepting one side or the other. [Atlassian's](https://www.atlassian.com/git/tutorials/using-branches/git-merge) write-up about `git merge` is an excellent read on this topic. More generally, I've found their git tutorial to be of the best I've ever seen.

So, to proceed, I recommend:

1. Changing the source branch of this PR from `removeHackyUsedinfoDB` to `tableAbbAccS`, or creating a new PR targetting `main` from `tableAbbAccS`. Ignore the CI until step 3 is complete -- it will most likely fail, and that is okay.
2. Go through the code changes in `tableAbbAccS` and carefully look into which changes are "on topic" to the PR and log which ones aren't. You should do this by going through the "Files changed" tab and submitting a review on your own PR. Your goal here is to identify changes that are unrelated to your work on the abbreviations table.
3. For the ones that aren't "on topic," you should undo them manually. It will probably be a bit tedious, but it's the safest and simplest way to fix the PR now.
4. Inspect the CI -- if it's passing, we're okay. If it's not, we can look into why.

Let me know if you have any questions or get stuck at any step.

smiths:
Great advice @balacij.  Very helpful to provide so much detail and links to specific resources.

PR #4332, opened 2025-08-15
`drasil-system`: Create SmithEtAl-style software requirements specification encoding.

balacij:
Contributes to #3260 and #4304


JacquesCarette:
This definitely needs design first. I think the SmithEtAl format should be seen as exactly that, an output format.

PR #4333, opened 2025-08-15
Show an HTML-extracted copy of Jupyter-based SRS notebooks on the website.

balacij:
#4332 is a rabbit hole that lead to this. How? `drasil-{printers, docLang, gen}` do not play well with those changes, so I want to make sure I'm not causing any regressions on the Jupyter Notebook as I force my way through compilation errors.

I will test this branch on my repo and report back.


balacij:
It looks like it works just fine now -- you can see a rendered copy on my Drasil github.io page: https://balacij.github.io/Drasil/

I used `jupyter`'s `nbconvert` tool, and the HTML generated looks almost identical to what you would normally see in a Jupyter Notebook window, sans the cells, buttons, etc.

JacquesCarette (approved):

PR #4354, opened 2025-09-03
Makefile 'rm -rf's: Remove bad error suppression and add './' before …

balacij:
`-f` already ignores file not-exists errors, so the `-`s are a no-op in the most common case.

One extremely unlikely case: If `rm` somehow hits a file we don't have permission to delete, the error should be reported.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -317,7 +317,7 @@ $(filter %$(GEN_E_SUFFIX), $(GEN_EXAMPLES)): %$(GEN_E_SUFFIX):
 
 # Compiles GOOL examples to concrete code (Java, C++, etc.)
 $(GOOLTEST)$(GEN_E_SUFFIX):
-	- rm -rf "$(BUILD_FOLDER)$(GOOLTEST_DIR)"
+	rm -rf "./$(BUILD_FOLDER)$(GOOLTEST_DIR)"
 	stack build $(stackArgs) "drasil-code:exe:$(GOOLTEST_EXE)"
 	@mkdir -p "$(BUILD_FOLDER)$(GOOLTEST_DIR)"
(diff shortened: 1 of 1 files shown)


code/Makefile
```diff
@@ -317,7 +317,7 @@ $(filter %$(GEN_E_SUFFIX), $(GEN_EXAMPLES)): %$(GEN_E_SUFFIX):
 
 # Compiles GOOL examples to concrete code (Java, C++, etc.)
 $(GOOLTEST)$(GEN_E_SUFFIX):
-	- rm -rf "$(BUILD_FOLDER)$(GOOLTEST_DIR)"
+	rm -rf "./$(BUILD_FOLDER)$(GOOLTEST_DIR)"
```
JacquesCarette, line 320: Does `./` work on Windows?

balacij: I forgot about Windows, but we have other *nix file paths in the Makefile and we haven't heard any complaints.

Examples:

https://github.com/JacquesCarette/Drasil/blob/80ab0bbcf9128c21c282b24561bbc7fe7d1f9dc9/code/Makefile#L117

https://github.com/JacquesCarette/Drasil/blob/80ab0bbcf9128c21c282b24561bbc7fe7d1f9dc9/code/Makefile#L346-L348

https://github.com/JacquesCarette/Drasil/blob/80ab0bbcf9128c21c282b24561bbc7fe7d1f9dc9/code/Makefile#L448-L452

https://github.com/JacquesCarette/Drasil/blob/80ab0bbcf9128c21c282b24561bbc7fe7d1f9dc9/code/Makefile#L546-L551

To gain a bit more confidence without installing Windows in a VM myself, for our two recommended methods of setting up a Drasil work environment on Windows, cygwin seems to be fine with `/`s and *nix-like file paths (they show [examples of file paths](https://cygwin.com/cygwin-ug-net/using.html)), and WSL is of course fine with them.

JacquesCarette (approved):

PR #4355, opened 2025-09-03
Makefile: Let diff errors fail when run individually, but when run en masse, defer until `log_check.sh` is called.

### LINKED ISSUE #4340: CI not failing on bad `stable/` diff
Take the latest CI run for #4323, https://github.com/JacquesCarette/Drasil/actions/runs/17082365395/job/48439134765?pr=4323 :

Image

balacij:
Closes #4340 

With this PR:

1. `make x_diff` will no longer have their exit codes suppressed.
2. `make`/`make test` will build all code using `stack build` (rather than `stack build $X`, which only builds one package at a time), ignoring individual `x_diff` exit codes (i.e., of the individual example runs) in favour of using the log checking script to get the correct exit code and note all examples with log diffs.


JacquesCarette (commented):
Seems reasonable to me.

PR #4372, opened 2025-09-16
remove 'nounPhraseSent' from another place in Projectile. 

JacquesCarette:
Causes a minor change to the HTML that had a `<span>` in it around an acronym but no longer does (as expected.)


samm82 (dismissed):
Since the loss of the `<span>` is expected, this change looks good to me (especially since the old code is still there for future reference). My only thoughts are that we should maybe:

1. add additional comments explaining why we're using the new code but _also_ keeping the commented out code, and/or
2. make these kinds of changes in a new branch so that work can still be done until everything is done "correctly" and the generated code goes back to matching the `main` branch (or being better than it!).

I might not fully understand the vision here so these comments may not apply. 😁

code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs
```diff
@@ -34,9 +34,11 @@ rectVelGD = gd (equationalModel' rectVelQD) (getUnit projSpeed) (Just rectVelDer
...
 
 rectVelQD :: ModelQDef
-rectVelQD = mkQuantDef' projSpeed (nounPhraseSent $ foldlSent_ 
-            [atStart rectilinear, sParen $ short oneD, phrase velocity,
-             S "as a function" `S.of_` phraseNP (time `for` QP.constAccel)])
+rectVelQD = mkQuantDef' projSpeed 
+  (cn "rectilinear (1D) velocity as a function of time for constant acceleration")
+  -- (nounPhraseSent $ foldlSent_ 
+  --           [atStart rectilinear, sParen $ short oneD, phrase velocity,
+  --             S "as a function" `S.of_` phraseNP (time `for` QP.constAccel)])
```
balacij: Is 1D really an abbreviation for "rectilinear", or was the "(1D)" there to explain approximately what rectilinear means? If its the latter, perhaps we can remove it?

code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs
```diff
@@ -58,9 +60,11 @@ rectPosGD = gd (equationalModel' rectPosQD) (getUnit projPos) (Just rectPosDeriv
...
 
 rectPosQD :: ModelQDef
-rectPosQD = mkQuantDef' projPos (nounPhraseSent $ foldlSent_ 
-            [atStart rectilinear, sParen $ short oneD, phrase position,
-             S "as a function" `S.of_` phraseNP (time `for` QP.constAccel)])
+rectPosQD = mkQuantDef' projPos 
+  (cn "rectilinear (1D) position as a function of time for constant acceleration")
+  -- (nounPhraseSent $ foldlSent_ 
+  --       [atStart rectilinear, sParen $ short oneD, phrase position,
+  --        S "as a function" `S.of_` phraseNP (time `for` QP.constAccel)])
```
balacij: I'll wait for your response on the previous PR about keeping the commented out code.

balacij (dismissed):

code/drasil-example/projectile/lib/Drasil/Projectile/GenDefs.hs
JacquesCarette: I think the latter. I would be fine with removing it, but we need @smiths to agree to that.

smiths: Rectilinear is a special case of 1D motion.  Rectilinear means movement along a single straight path.  1D means movement along one spatial coordinate, which could be a curved coordinate.  The emphasis here is on rectilinear, so I would also be fine with removing the 1D part.

balacij: Removed in [c2e2804](https://github.com/JacquesCarette/Drasil/pull/4372/commits/c2e2804be38427bba9409e44a936a28d8fb34ca4)

samm82 (approved):
As discussed previously, these changes look good and as expected!

Issue #4084, opened 2025-05-29
Quasi-Types


balacij:
_Spin-off of Note 3 of https://github.com/JacquesCarette/Drasil/issues/4064#issuecomment-2887572656_

## Objective

In GlassBR, I believe there are two variables that involve quasi-types, without this being explicitly captured. The goal of this issue is to capture said information and use said information  to automatically produce constraints (hidden from the SRS) on quantities.

## Background

Pulling from the table of symbols from the [live, generated SRS](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/HTML/GlassBR_SRS.html) at time of writing:

* $t$:  Nominal thickness t is in {2.5,2.7,3.0,4.0,5.0,6.0,8.0,10.0,12.0,16.0,19.0,22.0}
* $g$: Glass type

Additionally noting point 11 from the "Terminology and Definitions" section:

> Glass type factor (GTF) - A multiplying factor for adjusting the LR of different glass type, that is, AN, FT, or HS, in monolithic glass, LG (Laminated Glass), or IG (Insulating Glass) constructions. 

Both $t$ and $GTF$ are examples of quantities that involve _quasi-types_.

## Quasi-Types

First, an example: $\mathbb{Z}^+$. We define $\mathbb{Z}^+$ as "any value of $\mathbb{Z}$ greater than 0," where $\mathbb{Z}$ is the type of integers. Strictly speaking, members of $\mathbb{Z}^+$ are not of type $\mathbb{Z}^+$, but of $\mathbb{Z}$. $\mathbb{Z}^+$ is only a set of values of type $\mathbb{Z}$. It is only by convention that we might say members of $\mathbb{Z}^+$ are of type $\mathbb{Z}^+$.

A quasi-type is a set of values that acts like a type, but is really a set of values of another type. This quasi-type can be defined by either an enumerated set of values of another type or by a predicate on values of another type. For example, we can define $\mathbb{Z}^{+, \le 10}$ with either $\set{1,2,3,4,5,6,7,8,9,10}$ or $\set{z | z~:~\mathbb{Z}, 0 \lt z \le 10}$. Either way, $\mathbb{Z}^{+, \le 10}$ is a set that acts like a subtype of $\mathbb{Z}$ and whose members are strictly of type $\mathbb{Z}$ but that we might claim to be of type $\mathbb{Z}^{+, \le 10}$.

## What does this mean for $t$ and $g$?

[...]

PR #4373, opened 2025-09-16
remove nounPhraseSent from GlassBR

JacquesCarette:
This one has larger repercussions on the output. Other than losing the `$x$` (and similar), it also changes "nominal thickness t is in ..." to just "nominal thickness" to be consistent with everything else. Having the discrete values in the 'name' of the concept was a hack. We should capture that information in other ways!


code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -103,7 +102,8 @@ standOffDist = uq (constrained' (uc sD (variable "SD") Real metre)
     sfwrRange $ Bounded (Inc, sy sdMin) (Inc, sy sdMax)] (exactDbl 45)) defaultUncrt
 
 nomThick = cuc' "nomThick"
-  (nounPhraseSent $ S "nominal thickness t is in" +:+ eS (mkSet Rational (map dbl nominalThicknesses)))
+  (cn "nominal thickness")
+  -- (nounPhraseSent $ S "nominal thickness t is in" +:+ eS (mkSet Rational (map dbl nominalThicknesses)))
```
balacij, line 106: This is related to https://github.com/JacquesCarette/Drasil/issues/4084 . 

I'm in favour of removing this, but I think our policy is to have a replacement in place before we do. Without this here, it won't appear anywhere else in the SRS, which is why we originally left it as-is (#3878).

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -236,13 +236,19 @@ loadDur     = uc' "loadDur"    (nounPhraseSP "duration of load")
   (S "the amount of time that a load is applied to the glass plate")
   (sub lT lDur) Real second
 
-sdx         = uc' "sdx" (nounPhraseSent $ phrase standOffDist +:+ sParen (phrase xComp))
+sdx         = uc' "sdx" 
+  (cn "stand off distance (x-component)")
+  -- (nounPhraseSent $ phrase standOffDist +:+ sParen (phrase xComp))
```
balacij, line 241: I think we can erase the commented out code? Or were you going to keep it for when we try to add back parentheses?

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
JacquesCarette: Good point. This is the wrong place, but the information should not completely disappear.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
JacquesCarette: The parentheses are still there, it is the math formatting that went away. I was sort of leaving the commented out code to help me re-find those spots. Maybe I should find a different way to do that?

samm82: I think a TODO or something similar would work; then even if you don't get to it, someone eventually will (and won't delete the useful commented-out code to "clean things up")

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
JacquesCarette: My plan is to fix that first. And quite likely, I will end up dropping this particular PR its current form, for a different way to get the same thing done.

JacquesCarette:
I'm closing this one - will redo differently.

PR #4374, opened 2025-09-17
Remove use of nounPhraseSent in SSP

JacquesCarette:
fix some of the nounPhraseSent, and move the problem to nounPhrase'' in other cases. But those can be properly fixed in a next wave.


JacquesCarette:
This one also is not enough of a step forward. Closing, will redo on a better basis.

PR #4375, opened 2025-09-17
Update stackage version to LTS 22.44 (GHC 9.6.7)

balacij:
https://www.stackage.org/lts-22.44

### DIFF
--- a/code/scripts/update_default_stackage.sh
+++ b/code/scripts/update_default_stackage.sh
@@ -8,7 +8,7 @@ update_files() {
     local replace="$4"
 
     find "$dir" -type f -name "$targets" | while read -r file; do
-        sed -i -E "s|$search|$replace|g" "$file"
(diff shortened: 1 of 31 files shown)


code/scripts/update_default_stackage.sh
```diff
@@ -8,7 +8,7 @@ update_files() {
     local replace="$4"
 
     find "$dir" -type f -name "$targets" | while read -r file; do
-        sed -i -E "s|$search|$replace|g" "$file"
+        sed -i '' -E "s|$search|$replace|g" "$file"
```
balacij, line 11: The default `sed` implementation on macOS seems different than that shipped with most Linux package managers. I've never had to pass '' in to `sed` before, but this was required now.

Relevant: https://unix.stackexchange.com/questions/401905/bsd-sed-vs-gnu-sed-and-i

Issue #1501, opened 2019-06-05
SWHS conservation of energy tolerance treated as both an input and a constant


bmaclach:
SWHS lists C<sub>tol</sub> as a required input, defined as the "relative tolerance for conservation of energy".

However, in the requirement related to verifying that the output follows the law of conservation of energy, the tolerance is hardcoded as 0.001%.
![image](https://user-images.githubusercontent.com/18634568/58970078-b6938780-8786-11e9-81cc-4f6ad990d1e8.png)

The 0.001% is also hardcoded in the "Properties of Correct Solution" section. 

We should decide whether this is an input variable or a constant, and update the example accordingly. Even if it is a constant, I think it makes sense to still use the C<sub>tol</sub> symbol and then just add the value of that constant to the "Values of Auxiliary Constants" section?

smiths:
@bmaclach good catch.  You are absolutely correct that the value 0.001% should not be hard-coded into the documentation.  I think we should make C_{tol} a constant, rather than an input.  When we get to thinking about program families, whether or not C_{tol} is compile time or run time could be a variability.

bmaclach:
Closing this as it will be covered by @samm82 as part of #1494.

samm82:
Should _A<sub>tol</sub>_ and _R<sub>tol</sub>_ also be constants (or variables in the future)? @bmaclach @smiths

bmaclach:
@samm82 Those are the absolute and relative tolerances for the ODE solver. They need to be defined, but whether as an input or constant does not matter. I would lean towards keeping them as inputs just so that we don't need to change anything, but @smiths may have a different preference.

Note that there's been some discussion about how these should not be introduced until the design document (#1498), but for now, since we don't have a design document, I think we should keep them in the SRS.

smiths:
I agree with @bmaclach.  The tolerances should be inputs, not constants.  We also should keep them in the SRS, because at the moment, we do not have any other options.  :smile:

Issue #1498, opened 2019-06-05
NoPCM SRS assumes an implementation choice?


bmaclach:
It's been mentioned in another issue (#492) how the differential equations in NoPCM have an analytical solution, and thus do not necessarily require an ODE solver. However, we still have the `abs_tol`, `rel_tol`, and `timeStep` inputs in the NoPCM SRS, which are only used as parameters for an ODE solver. By including these as inputs, are we not making a decision about the implementation (choosing to use an ODE solver)? If we removed these inputs, would we then be making the opposite decision (choosing to use the analytical solution)? It seems almost inevitable that we need to make this decision in the requirements stage. Should we add to the "System Constraints" section a note that the solution will use an ODE solver?

smiths:
Great observation @bmaclach.  You are absolutely correct that requiring, or not requiring, ODE solver parameters constrains the design.  Rather than constrain the design, the better course of action is to move these quantities to the design stage.  The SRS really should just be about the physics.  The design choices will add new requirements.  The problem for us is that we haven't yet figured out our design documentation.  If we make the design documentation the same style as the SRS, essentially requirements for the design, we could fit this in fairly naturally.  I think the best course of action is to really think about the Projectile example that @samm82 is working on.  This example seems like a good platform for experimentation.  It is complex enough, but still fairly simple.

If I've answered your question, you can close this issue.

JacquesCarette:
I don't like the phrasing "design choices will add new requirements". I agree that design choices will add new information that needs to be considered properly and appropriately - though I strongly hesitate to call them requirements [I can be convinced otherwise though]. "requirements for the design" is closer, but awkward.

smiths:
@JacquesCarette, I agree that saying "design choices add new requirements" isn't exactly what I wanted to say.  For all of SEs work on requirements and design, the terminology isn't quite as expressive as we want it to be.  :smile:  What I would like to see is the requirements document not mention anything about numerical parameters.  As soon as we do that, we constrain the design.  

When we make the design choices, within the design documentation, we have constrained the requirements.  We can maybe think of the design choices as software constraints.  That is the term I would use in the SRS if design decisions were externally imposed at the beginning of the process.

Somewhere we need to say that the choice to use an ODE solver has added a need for tolerances and a time step.  I think we say that in our design documentation (whatever that might look like) and then the code generator gets this information from here.  In my loose language above, I likened that to adding a "requirement."

JacquesCarette:
[...]

PR #4376, opened 2025-09-17
Remove table of symbols Table-based filtration

balacij:
Rendering the table of symbols relies on a table-based `UID` filtration to determine which `UID`s the document `UID` extraction tool found are actually quantities. This is odd because (1) it expected that all symbols to be displayed in the table of symbols to be in other tables and (2) all instances of quantities to appear in a column titled "Var". The new symbol dictionary table entries in this PR are all ones that do not appear in any other table or appear in a table where the symbol column is called "Symbol" rather than "Var".


code/stable/gamephysics/SRS/PDF/GamePhysics_SRS.tex
```diff
@@ -193,6 +193,8 @@ \subsection{Table of Symbols}
 \\
 $θ$ & Angular displacement & ${\text{rad}}$
 \\
+$π$ & Ratio of circumference to diameter for any circle & --
```
balacij, line 196: I'm somewhat surprised this wasn't already in the table of symbols. Is there any reason why it shouldn't be there? Or was this just a mistake?

code/stable/gamephysics/SRS/HTML/GamePhysics_SRS.html
```diff
@@ -498,6 +498,11 @@ <h2>Table of Symbols</h2>
                   <td>Angular displacement</td>
                   <td><em>rad</em></td>
                 </tr>
+                <tr>
+                  <td><em>π</em></td>
+                  <td>Ratio of circumference to diameter for any circle</td>
+                  <td>--</td>
+                </tr>
```
balacij, line 505: $\pi$ only appears in the input data constraints table, under a column entitled "Software Constraints":

Screenshot 2025-09-22 at 12 46 56 PM

code/stable/ssp/SRS/HTML/SSP_SRS.html
```diff
@@ -178,6 +178,13 @@ <h2>Table of Symbols</h2>
                   <th>Description</th>
                   <th>Units</th>
                 </tr>
+                <tr>
+                  <td><em>(x,y)</em></td>
+                  <td>
+                    Cartesian Position Coordinates: <em>y</em> is considered parallel to the direction of the force of gravity and <em>x</em> is considered perpendicular to <em>y</em>.
+                  </td>
+                  <td><em>m</em></td>
+                </tr>
```
balacij, line 187: This is just a bad encoding. $(x,y)$ is a bad symbol. However, it now appears because it is recorded as a symbol:

https://github.com/JacquesCarette/Drasil/blob/9f6dd893f5989f75f620e495398bf2a101ce0156/code/drasil-example/ssp/lib/Drasil/SSP/Unitals.hs#L190-L193

https://github.com/JacquesCarette/Drasil/blob/9f6dd893f5989f75f620e495398bf2a101ce0156/code/drasil-example/ssp/lib/Drasil/SSP/Unitals.hs#L524-L529

Note that $(x,y)$ only appears as an entry in the `Required Inputs` table with column title: Symbol.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -481,6 +486,11 @@ <h2>Table of Symbols</h2>
                   <td>Thermal flux vector</td>
                   <td><em>\(\frac{\text{W}}{\text{m}^{2}}\)</em></td>
                 </tr>
+                <tr>
+                  <td><em>R<sub>tol</sub></em></td>
+                  <td>Relative tolerance</td>
+                  <td>--</td>
+                </tr>
```
balacij, line 493: For both Absolute tolerance and Relative tolerance, the same issue. They only appear in the required inputs table.

balacij: These should have probably been referenced somewhere in the document as well. It looks like they were introduced as part of https://github.com/JacquesCarette/Drasil/issues/1501 .

code/stable/gamephysics/SRS/PDF/GamePhysics_SRS.tex
smiths: I can see an argument for including $\pi$ in our table of symbols, but I don't think it should be there.  I think we consider $\pi$ to be part of the background (context?) theory.  It is part of the language of mathematics that we just take for granted.  We have to stop at some level.  We can't feasibly provide a definition for every piece of information.  We have a symbol for addition, but we don't put it in the table of symbols.  We use functions like $\sin$ and $\cos$, but we don't put them in our table of symbols.  We have symbols like the alphabet and punctuation as part of communicating in written English, but we don't define those symbols.  

My point is that it isn't feasible to define everything down to an atomic level.  We need to have an assumed starting point, an assumed foundation.  Our mistake isn't not defining $\pi$.  Our mistake is not being explicit about what is in our foundation, and what isn't.  If we use something like Bill's little theories, we can be more explicit.

balacij: I see your point. Having background theories would be good here. We need to move this 'symbol introduction' knowledge into background theories, but we don't have the infrastructure for that yet. Those are all great examples, and I completely agree.

I think it would be best to still merge this PR, however, and to fix the real background theories issue later. What do you say?

JacquesCarette: I agree, merge now, fix these things later.

code/stable/ssp/SRS/HTML/SSP_SRS.html
JacquesCarette: In some sense, I like that this hack is now exposed. It will force us to fix it sooner rather than later.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
JacquesCarette: Yes, it does look like the document is missing something that mentions these.

JacquesCarette (approved):
I would be happy to merge as-is. But need a full review by @smiths . I think we want to first have a few issues opened to record the things we'd like to have removed, but then just go ahead with merging.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
smiths: This has come up before, like in #1498.  The reason we need numerical parameters is because we are doing a numerical solution, but an abstract requirements document should only talk about "what" the problem is, not "how" to solve it.  

Ideally we would have an explicit refinement stage between the requirements and the code.  We have design decisions to make that are about how to solve the problem.  For instance, deciding to use an ODE solver is a design decision that brings with it the need for numerical parameters as input.

If we had a theory of theories worked out, I think we could combine a mathematical theory for the ode with a numerical theory for solving odes.  We could calculate from this combination of theories what inputs are needed.

We don't have a theoretical framework for our theories, and we don't have anything implemented on that front; therefore, I recommend we address this by adding an accuracy requirement to the document.  We add a (nonfunctional) requirement that the relative error between the calculated and true solution be within tolerance.  We should probably add that this is only required if a solution can be calculated.

In general, it is very difficult to prove that a numerical solution is within the tolerance, but it can still give us an aim for the solution quality, and the parameter will now be available.

JacquesCarette:
Making sure this is still on your radar @smiths .

smiths (approved):
As suggested by @JacquesCarette, I created some issues (#4391 and #4390) to remind us of new problems that this PR highlights/introduces.  With those issues created, I'm fine with merging this PR.  I didn't create an issue for the definition of a Cartesian coordinate.  I don't mind this being in the table of symbols, since thinking about the position is natural for the reader.  If there is something off with the implementation, maybe @balacij can create an issue related to that.

PR #4363, opened 2025-09-07
Automatically collect abbreviations from generated `DocDesc`s

sarrasoussia:
Closes https://github.com/JacquesCarette/Drasil/issues/4299
Closes #1235 

Builds on #4323
Builds on #4695
Builds on #4843

PR #4379, opened 2025-09-19
PDController: Remove unused 'integral' CI.

balacij:
Contributes to #4363


smiths:
I had to look more closely at this one because the right solution may have been to introduce the use of I in the document.  PID controller uses the I, so if we talked about integral control at all, we would want it.  I searched the SRS and we use I once when we say "the Integral controller is beyond the scope of this project".  I'm glad to see that we make the scope explicit.  In the scope text, we could say Integral (I) controller, but there is no reason to introduce an abbreviation that we introduce and never use.

I support the removal of I from this example.

JacquesCarette (approved):

PR #4380, opened 2025-09-19
GlassBR: Remove unused 'not applicable' abbreviation from table of abb/accs.

balacij:
Contributes to #4363.


samm82 (approved):
Verified that "N/A" only shows (showed?) up in the table of abbreviations - looks good!

PR #4382, opened 2025-09-19
SWHSNoPCM: Remove duplicate (incorrect) copy of SWHSNoPCM CI.

balacij:
Contributes to #4363

This does change `stable/` slightly.


balacij:
Sorry, I was in a rabbit hole with this and didn't write a complete description. This is a duplicate chunk because we also have:

https://github.com/JacquesCarette/Drasil/blob/9f6dd893f5989f75f620e495398bf2a101ce0156/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/MetaConcepts.hs#L5-L10

The reason why I mention duplicate and incorrect is because the duplicated version (the one I removed with this PR) was also incorrect (because it was missing the important "NoPCM" long form part).

PR #4383, opened 2025-09-19
SWHS: Remove a duplicate, unused CI capturing SWHS.

balacij:
Contributes to #4363 

This one is fairly simple. This is just duplicate code I'm removing.

### DIFF
--- a/code/drasil-example/swhs/lib/Drasil/SWHS/Concepts.hs
+++ b/code/drasil-example/swhs/lib/Drasil/SWHS/Concepts.hs
@@ -68,14 +68,3 @@ tankPCM = dcc "tankPCM" (nounPhrase''
   (phrase sWHT +:+ S "incorporating" +:+ short phsChgMtrl)
   CapFirst CapWords)
   "solar water heating tank incorporating phase change material"
-
-swhsPCM :: CI
--- Nounphrase'' hack to get nounPhraseSP words to accept
--- nounPhrases instead of strings
--- Another capitalization hack.
-swhsPCM = commonIdea "swhsPCM" (nounPhrase''
-  (S "solar water heating systems" +:+ S "incorporating" +:+ short phsChgMtrl)
-  (S "solar water heating systems" +:+ S "incorporating" +:+ short phsChgMtrl)
-  CapFirst CapWords)
-  "SWHS"
-  []


code/drasil-example/swhs/lib/Drasil/SWHS/Concepts.hs
```diff
@@ -68,14 +68,3 @@ tankPCM = dcc "tankPCM" (nounPhrase''
...
-swhsPCM :: CI
--- Nounphrase'' hack to get nounPhraseSP words to accept
--- nounPhrases instead of strings
--- Another capitalization hack.
-swhsPCM = commonIdea "swhsPCM" (nounPhrase''
-  (S "solar water heating systems" +:+ S "incorporating" +:+ short phsChgMtrl)
-  (S "solar water heating systems" +:+ S "incorporating" +:+ short phsChgMtrl)
-  CapFirst CapWords)
-  "SWHS"
-  []
```
balacij, line 81: Duplicate of: https://github.com/JacquesCarette/Drasil/blob/9f6dd893f5989f75f620e495398bf2a101ce0156/code/drasil-example/swhs/lib/Drasil/SWHS/MetaConcepts.hs#L11-L20

balacij: Thinking about this again. Is this even necessary as a CI? It's obvious it was captured to avoid duplication of the sentence addition "incorporating PCM", but we still prefer to call SWHS "SWHS," so maybe this shouldn't have been captured as a CI.

Also, if the long-form includes "including phase change material", shouldn't the short-form also include that? e.g., SWHSIPCM?

smiths: I don't remember the use case for CI, so I can't comment on that.  I've used the short forms differently in the manual cases.  I called the one with PCM SWHS and the one without it, SWHS_NoPCM.  I'm not suggesting that we do that; I'm just illustrating that there aren't any fast rules for naming program family members.  For instance, Word on a PC and Word on a Mac are not the same program family member, but we call both of them Word.  Especially when we make minor variations to a program family member, we don't give it a new name.  For instance, we don't rename our case study examples if we turn logging on or off.

SWHS and SWHSIPCM are two related family members.  It is a choice of the designer of the family what to call them.  I don't think there is a universal naming policy we can adopt.  Personally, I think they should both be SWHS.  The name SWHSIPCM is also an ugly name.  :smile:  SWHS is bad enough, but adding the IPCM makes it unusable.

balacij: That example makes sense to me, and completely changed how I was seeing this. I forgot that more than one short-form may exist. I'm just not sure if/how we can reconcile this information with our existing chunk types.

That being said, this PR is still fine because it just removes one copy of this issue. We can figure out how to capture relationship properly later.

> The name SWHSIPCM is also an ugly name. 😄 SWHS is bad enough, but adding the IPCM makes it unusable.

I completely agree.

Now, how should this be dealt with in the table of abbreviations and acronyms? Do we just note that context-based, SWHS can mean either?

balacij: One of the nice things with admitting duplicate abbreviations right now is that #4080 lets us disambiguate them in the HTML easily :)

smiths: Do our abbreviations have to be unique between case studies?  It is a very high bar to never repeat abbreviations.  If we had scoping rules, we could allow abbreviations to be unique within their namespace.  This won't be the only time when this comes up.  For fun, I just looked up the acronym SRS.  The AI overview states:

SRS most commonly stands for Supplemental Restraint System, which is the system in a vehicle that includes airbags and seatbelts, designed to protect occupants during a collision. The term SRS can also refer to other things depending on the context, such as [Speech Recognition Score](https://www.google.com/search?sca_esv=44a7c7d5b1110c32&sxsrf=AE3TifN3D1mbrKNIWiryct_eC4Vy7uM4-A:1758333386156&q=Speech+Recognition+Score&sa=X&ved=2ahUKEwi1vYOnnuaPAxUQ48kDHcFYKg4QxccNegQIfRAB), [Spatial Reference System](https://www.google.com/search?sca_esv=44a7c7d5b1110c32&sxsrf=AE3TifN3D1mbrKNIWiryct_eC4Vy7uM4-A:1758333386156&q=Spatial+Reference+System&sa=X&ved=2ahUKEwi1vYOnnuaPAxUQ48kDHcFYKg4QxccNegQIfRAC), or [Sex Reassignment Surgery](https://www.google.com/search?sca_esv=44a7c7d5b1110c32&sxsrf=AE3TifN3D1mbrKNIWiryct_eC4Vy7uM4-A:1758333386156&q=Sex+Reassignment+Surgery&sa=X&ved=2ahUKEwi1vYOnnuaPAxUQ48kDHcFYKg4QxccNegQIfRAD) (now more commonly called [Gender Confirmation Surgery](https://www.google.com/search?sca_esv=44a7c7d5b1110c32&sxsrf=AE3TifN3D1mbrKNIWiryct_eC4Vy7uM4-A:1758333386156&q=Gender+Confirmation+Surgery&sa=X&ved=2ahUKEwi1vYOnnuaPAxUQ48kDHcFYKg4QxccNegQIfRAE)).It also stands for Supplemental Restraint System.

I don't think we should give a definition that includes information on how the context can be used to determine which definition applies.  If we aren't allowed to have the same acronym mean two different things in two different contexts, then I think we should have two different program names for now.  Maybe SWHS and SWHS+?

balacij: You're right -- abbreviations do not need to be unique, and we should be able to handle them. Across domains, I agree that alternative abbreviations should exist. I guess I never thought that we would ever pull in two domains where that ambiguity could exist, but we can think of "SWHS" (the family) as one and the specific instance "SWHSNoPCM" or "SWHSIPCM" locally with the more generic abbreviation unambiguously.

Should this change anything with #4382, where the title of the project is changed? I think it's really up to our style preference, I still tend towards the new title purely for disambiguating between the two projects by title.

JacquesCarette (approved):
Good discussion - but does not affect the PR directly, since it really is pure duplication.

PR #4396, opened 2025-10-07
SglPend: Broken Code Gen.

balacij:
This PR highlights some bizarre behaviour. This is not a PR that is intended to be merged anytime soon. It is a PR meant to highlight issues. Once SglPend's code generation works, then we can merge this PR (or a PR that replaces this one).


code/stable/sglpend/src/cpp/Control.cpp
```diff
@@ -0,0 +1,34 @@
...
+    double m;
+    double α;
+    double θ_p;
+    double θ_i;
+    double p_x^i;
+    double p_y^i;
+    get_input(filename, L_rod, m, α, θ_p, θ_i);
+    derived_values(L_rod, θ_i, p_x^i, p_y^i);
+    input_constraints(L_rod, θ_i);
+    write_output(θ_p);
```
balacij, line 31: 1. Invalid variable names. Conventional C++ does not use unicode characters. Additionally '^' is an operator not permitted in the name of variables.
2. `get_inputs` confused me for a second. We pass in variable names but they're automatically turned into references. I'm not sure if I'm a fan of not explicitly rendering something that indicates it is going to be passed as reference and could be mutated...
3. The only output is one of the inputs.

code/stable/sglpend/src/java/SglPend/Control.java
```diff
@@ -0,0 +1,38 @@
...
+    public static void main(String[] args) throws FileNotFoundException, IOException {
+        String filename = args[0];
+        double L_rod;
+        double m;
+        double α;
+        double θ_p;
+        double θ_i;
+        double p_x^i;
+        double p_y^i;
+        Object[] outputs = InputParameters.get_input(filename);
```
balacij, line 26: `outputs = Input...`

code/stable/sglpend/src/java/SglPend/Control.java
```diff
@@ -0,0 +1,38 @@
...
+        L_rod = (double)(outputs[0]);
+        m = (double)(outputs[1]);
+        α = (double)(outputs[2]);
+        θ_p = (double)(outputs[3]);
+        θ_i = (double)(outputs[4]);
+        outputs = InputParameters.derived_values(L_rod, θ_i);
+        p_x^i = (double)(outputs[0]);
+        p_y^i = (double)(outputs[1]);
+        InputParameters.input_constraints(L_rod, θ_i);
+        OutputFormat.write_output(θ_p);
```
balacij, line 36: Only `θ_p` is used, and it doesn't even have any constraints. We don't need to generate any of the other code.

code/stable/sglpend/src/java/SglPend/Control.java
balacij: Ah, this is making sense now:

```haskell
inputs :: [DefinedQuantityDict]
inputs = map dqdWr [lenRod, QPP.mass, QP.angularAccel, pendDisplacementAngle, initialPendAngle]

outputs :: [DefinedQuantityDict]
outputs = [dqdWr pendDisplacementAngle]
```

code/stable/sglpend/src/cpp/Control.cpp
JacquesCarette: That sure is weird - probably ought to create 3 issues for each of those.

JacquesCarette: For 1, the C++ generator ought to error out. And the solution is to have the names use the features of Drasil that let you have different renderings in code vs spec. Other examples use that already.

code/stable/sglpend/src/csharp/Control.cs
```diff
@@ -0,0 +1,26 @@
...
+        double m;
+        double α;
+        double θ_p;
+        double θ_i;
+        double p_x^i;
+        double p_y^i;
+        InputParameters.get_input(filename, out L_rod, out m, out α, out θ_p, out θ_i);
+        InputParameters.derived_values(L_rod, θ_i, out p_x^i, out p_y^i);
+        InputParameters.input_constraints(L_rod, θ_i);
+        OutputFormat.write_output(θ_p);
```
JacquesCarette, line 24: At least the csharp is consistent with the C++!

code/stable/sglpend/src/java/SglPend/Control.java
JacquesCarette: Hey, Drasil does *exactly* what it was asked to do! Fancy that!

Probably `pendDisplacementAngle` should not be an input.

balacij: Removing `pendDisplacementAngle` from the `inputs` list:

```console
sglpend: The following outputs cannot be computed: pendDisplacementAngle
Unused definitions are: pendDisplacementAngle, frequency, angularFrequency, period
Known values are: l_rod, mass, angularAcceleration, initialPendAngle, ixPos, iyPos
  error, called at lib/Language/Drasil/CodeSpec.hs:258:20 in drasil-code-0.1.9.0-CVigL3PxUMa78erpyMb2gY:Language.Drasil.CodeSpec
```

JacquesCarette: See my comments on the issue about the right way to fix this.

balacij: `initialPendAngle` (which corresponds to $\theta_i$) is already in the inputs list.

Note: I didn't look into this further yet.

smiths:
@balacij, don't investigate further.  You have enough on your plate right now.  :smile:  Thank you for starting the process of getting SglPend working properly.  This might be a good task for a future summer research assistant.  :smile:

code/stable/sglpend/src/cpp/Control.cpp
balacij: I've changed my opinion on (2) and filed issues (#5086 and #5087) for the other two issues.

balacij:
Closed in favour of #5086, #5087, #5088, and the eventual work on trying to get Drasil to generate code for `SglPend`.

Issue #4401, opened 2025-10-13
Wiki Cleanup


B-rando1:
As part of my RA work, I will be going through the Drasil Wiki and checking what needs to be updated, etc. I'll post my initial findings here.

To start off, I compared the contents of the `wiki` folder and the sidebar to see when there were missing or dead references.

#### Missing from Sidebar
I wasn't able to find any links in [/wiki/_Sidebar.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/_Sidebar.md) to any of the following pages in `/wiki/`. There's a chance they are in there, but I wasn't able to find them.
- [File-Input-Patterns-Design-Discussion.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/File-Input-Patterns-Design-Discussion.md)
- [GOOL-GProc Overview.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/GOOL-GProc%20Overview.md)
  - This was my fault, I forgot to add a link when I created the file in #3915. I will add it shortly.
- [Getting-To-Know-Potential-Users-of-Drasil.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/Getting-To-Know-Potential-Users-of-Drasil.md)
- [Nuggets.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/Nuggets.md)
- [Questions.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/Questions.md)
- [Reference-Design-and-Documentation.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/Reference-Design-and-Documentation.md)
- [What-If.md](https://github.com/JacquesCarette/Drasil/blob/main/wiki/What-If.md)

#### Dead Links in Sidebar
- [Converting GOOL Classes into Struct Libraries for Procedural Languages](https://github.com/JacquesCarette/Drasil/wiki/Converting-GOOL-Classes-into-Struct-Libraries-for-Procedural-Languages)
  - This was my fault, I forgot to remove the link when I removed the file in #3914. I will remove it shortly.

B-rando1:
This issue is a nice central link for all the wiki work (in case that becomes useful at some point 😄), but we may as well close it since it's not really an "issue".

PR #4403, opened 2025-10-13
Added gool overview to Wiki Sidebar

B-rando1:
I must've forgotten this in #3915. The placement may be up for discussion - under `folders` seems reasonable to me, but there may be other places for it.

@balacij since I believe you created the current wiki workflow, can you double-check the link structure I used? I tried to match the format of the other links, but I couldn't find a way to test it out on my branch.

Contributes to #4401


balacij:
Because I was specifically @'d, I believe `GOOL-GProc%20Overview` should have worked. Did you try that?

B-rando1:
Ah right, I noticed now that I used the full link instead of trying to follow the format, so my initial comment was inaccurate :)

I did preview it with VS Code, but iirc the internal links without the full path didn't work, since VS Code doesn't know it's a wiki.

B-rando1:
Turns out I did not remember correctly and it does work in VS Code :sweat_smile:

I'll keep that in mind, it should be useful for future wiki-related changes.

Issue #4398, opened 2025-10-09
Add LValues to Gool


TOTBWF:
Inside of [InterfaceCommon.hs](https://github.com/JacquesCarette/Drasil/blob/795898c6867dfe4f6bce8db710bf9c2f59125a0d/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L111C1-L113C47) we have the following very suspicious bit of code:

```haskell
class (VariableSym r) => VariableElim r where
  variableName :: r (Variable r) -> String
  variableType :: r (Variable r) -> r (Type r)
```

A quick search reveals that we only use `variableName` to write [this](https://github.com/JacquesCarette/Drasil/blob/795898c6867dfe4f6bce8db710bf9c2f59125a0d/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs#L209-L216) following cursed bit of code:

```haskell
arrayElem :: (OORenderSym r) => SValue r -> SVariable r -> SVariable r
arrayElem i' v' = do
  i <- IC.intToIndex i'
  v <- v'
  let vName = variableName v ++ "[" ++ render (RC.value i) ++ "]"
      vType = listInnerType $ return $ variableType v
      vRender = RC.variable v <> brackets (RC.value i)
  mkStateVar vName vType vRender
```

In short, we are using the `variableName` method to build a new "variable" name that looks like `x[e]` so that we can assign to array indices 😱 

We should fix this by introducing an explicit notion of l-value to gool, and making both variables and array indices lvalues. This issue is currently blocking the state refactor that @B-rando1 is working on, so we should fix it in the course of that work.

B-rando1:
I was reading up on lvalues, this [post](https://www.learncpp.com/cpp-tutorial/value-categories-lvalues-and-rvalues/) seems like a good reference.

B-rando1:
Some notes to summarize what I've found from looking at GOOL and talking with @TOTBWF and @JacquesCarette:
- `ValueSym`, which is extended by typeclasses such as `Literal` and `VariableValue` (dereferencing), corresponds to RValues.
[...]

PR #4404, opened 2025-10-13
Commented up some potential issues with GOOL

B-rando1:
In #4398 @TOTBWF hinted at some deeper grammatical and architectural issues with GOOL that may cause problems with #4360. In order to (hopefully) bring more clarity to this issue, I opted to add comments to some of the areas that I believe he was referring to.

The issues fall into a few categories:
- My singular comment in `LanguagePolymorphic` is representative of a lot of the backend of GOOL, where we often unwrap our functions down to `Doc` or `String`, make some changes, then wrap them back up into `Value`, `Variable`, etc. I get the sense this back-and-forth is potentially problematic in general, but it is especially so here because it assumes each renderer uses the same representation - an assumption that our refactor will violate. I think we _can_ still emulate that assumption, but it might get messy.
- Most of my comments with `YElim` classes in the `RendererClassesX` files are related to the above point. I'm not actually 100% sure if they're a bad thing as such; the bigger issue might be that they're used throughout the translation process instead of just at the end.
- Most of my other comments in `RendererClassesX` have to do with classes that seem to have almost exactly the same signature as classes in `InterfaceX`. My understanding is that they exist as simpler-to-implement methods that functions in `LanguagePolymorphic` can call to implement the `InterfaceX` methods. This _works_, but creates difficulties when debugging (since the actual implementations of many methods ends up being several calls deep). It also often relies on the representation of language constructs, as I mentioned in the `LanguagePolymorphic` point.
- There also seem to be a few classes in `RendererClassesX` that belong in `InterfaceX`, and vice-versa.

It seems like we need to parse through some of this in order to find our next steps. Ideally we could fix any blockers for the refactor in a separate PR, and then continue with the refactor on its branch. I'm still not entirely clear on which issues are actual blockers and which are simply code smells or even stylistic choices. If @TOTBWF and @JacquesCarette could take a look and give direction for a second pass of commenting, that might be a good place to start.

### DIFF
--- a/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs
+++ b/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs
@@ -206,6 +206,7 @@ objVar o' v' = do
         (variableType v) (R.objVar (RC.variable o) (RC.variable v))
   objVar' (variableBind v)
 
+-- This seems a bit sketchy - unwrapping everything to Doc, mixing it up, then wrapping it back to Variable
 arrayElem :: (OORenderSym r) => SValue r -> SVariable r -> SVariable r
 arrayElem i' v' = do
   i <- IC.intToIndex i'


code/drasil-gool/lib/Drasil/GOOL/InterfaceGOOL.hs
```diff
@@ -134,6 +134,7 @@ class (ValueSym r, OOTypeSym r) => OOValueSym r
 class (VariableSym r, OOTypeSym r) => OOVariableSym r where
   -- Bool: False for variable, True for constant.  Required by the Python renderer.
   staticVar'    :: Bool -> Label -> VSType r -> SVariable r
+  -- We should probably have a way to make this a constant that can't have an L-Value
```
JacquesCarette, line 137: Yes, a redesign of L-Values would be good.

code/drasil-gool/lib/Drasil/GOOL/RendererClassesOO.hs
```diff
@@ -40,10 +40,12 @@ class (BlockCommentSym r) => RenderFile r where
 
   fileFromData :: FilePath -> IG.FSModule r -> IG.SFile r
 
+-- This assumes a certain representation for Permanence
 class PermElim r where
   perm :: r (IG.Permanence r) -> Doc
   binding :: r (IG.Permanence r) -> Binding
 
+-- Why do we have these here on top of InterfaceCommon?
```
JacquesCarette, line 48: Search for the PRs that introduced them, I'm pretty sure there are comments there for why. There was a reason!

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs
```diff
@@ -206,6 +206,7 @@ objVar o' v' = do
         (variableType v) (R.objVar (RC.variable o) (RC.variable v))
   objVar' (variableBind v)
 
+-- This seems a bit sketchy - unwrapping everything to Doc, mixing it up, then wrapping it back to Variable
```
JacquesCarette, line 209: I think this is supposed to create something that points to an array element, and then will be used as if it were a variable.

This is likely only used on the LHS, so that a redesign of L-Values might get rid of this.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -89,6 +89,8 @@ class TypeSym r where
   funcType      :: [VSType r] -> VSType r -> VSType r
   void          :: VSType r
 
+-- We've established that getTypeString is a bad thing
+-- getType might be required, but without global state should it be a class?
```
JacquesCarette, line 93: `getType` is used to do a little bit of introspection. There might be other ways to do that.

It's a class because we should always be able to do that, and implementation can be per-language.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -108,6 +110,7 @@ class (TypeSym r) => VariableSym r where
   extVar    :: Library -> Label -> VSType r -> SVariable r
   arrayElem :: Integer -> SVariable r -> SVariable r
 
+-- Similar here - shouldn't we be handling this on a per-language basis?
```
JacquesCarette, line 113: classes exactly let you do things per-language. Now, if you mean that each language might want to vary this signature, that's different!

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -256,6 +260,8 @@ libFuncApp l n t vs = libFuncAppMixedArgs l n t vs []
 exists :: (ValueExpression r) => SValue r -> SValue r
 exists = notNull
 
+-- These all return SValue, but shouldn't they return MSStatement?
```
JacquesCarette, line 263: We're really hoping that all languages have `List` as values. Except that there might not be "lists" per se, just mis-named, they might be arrays (since `indexOf` is an evil thing to do to a list).

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -256,6 +260,8 @@ libFuncApp l n t vs = libFuncAppMixedArgs l n t vs []
...
+-- Also, they have the issue of not distinguishing Values and L-Values (See below)
```
JacquesCarette, line 264: Definitely.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -282,6 +288,7 @@ class (ValueSym r) => List r where
   --   Arguments are: List, Value
   indexOf :: SValue r -> SValue r -> SValue r
 
+-- Similar to List, shouldn't most of these be statements?
```
JacquesCarette, line 291: As with list, we're hoping not. But that might be a mistake.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -377,6 +384,8 @@ assignToListIndex :: (StatementSym r, VariableValue r, List r) => SVariable r
   -> SValue r -> SValue r -> MSStatement r
 assignToListIndex lst index v = valStmt $ listSet (valueOf lst) index v
 
+-- As Reed mentioned, these have a lack of distinction between values/variables
+-- and L-Values
```
JacquesCarette, line 388: Yes. But we need a reason to make that distinction. For L-Values, I think that's clear. Between values and variables, does our code actually need to know that? [The answer could be a resounding 'yes'; the point is to be explicit as to why.]

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -50,6 +50,7 @@ class (AssignStatement r, DeclStatement r, IOStatement r,
 
 -- Common Typeclasses --
 
+-- Couldn't this just be in InterfaceCommon?
```
JacquesCarette, line 53: Most likely.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -69,18 +70,23 @@ class BodyElim r where
 class RenderBlock r where
   multiBlock :: [MSBlock r] -> MSBlock r
 
+-- This assumes that the renderers encode Body as Doc
```
JacquesCarette, line 73: Right, this assumes that the only thing to ever be done with a `Block` will be to render it. This might not be a good assumption.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -69,18 +70,23 @@ class BodyElim r where
...
   typeFromData :: CodeType -> String -> Doc -> VSType r
 
+-- This is coupled to TypeData
 class InternalTypeElim r where
   type' :: r (Type r) -> Doc
 
 type VSUnOp a = VS (a (UnaryOp a))
 
+-- Why do we have these, when we also have NumericExpression in InterfaceCommon?
+-- It looks like the implementations for NumericExpression link here, but why?
```
JacquesCarette, line 89: I don't remember. Sleuth through the PRs that introduce this?

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -124,12 +130,16 @@ class OpElim r where
...
   scopeData :: r (Scope r) -> ScopeData
-  
+
+-- What exactly is this, since it adds a bunch of extra required data?
 class RenderVariable r where
   varFromData :: Binding -> String -> VSType r -> Doc -> SVariable r
-    
+
+-- This assumes variables are represented in a certain way, and
+-- its existence seems a bit sketchy
```
JacquesCarette, line 142: It is likely used for doing a bit of introspection. Might not be needed if the information is available by other means.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -157,6 +167,7 @@ class ValueElim r where
   valueInt :: r (Value r) -> Maybe Integer
   value :: r (Value r) -> Doc
 
+-- Again, wh are we recreating all these list functions?
```
JacquesCarette, line 170: There is likely a single language that needs a "middle layer" to intercept these, tweak, and pass on.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -198,13 +211,16 @@ class StatementElim r where
 
 class RenderVisibility r where
   visibilityFromData :: VisibilityTag -> Doc -> r (Visibility r)
-  
+
+-- This assumes a certain representation of visibility 
 class VisibilityElim r where
   visibility :: r (Visibility r) -> Doc
 
+-- Why do we need a Doc here?
```
JacquesCarette, line 219: Only the instance can tell!

JacquesCarette (commented):
Excellent questions. I commented as best I could.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
B-rando1: I think I was specifically referring to `listAdd`, `listAppend`, `listSet`. These usually mutate the list and return void.

JacquesCarette: Hmm, I think we should have two sets of "list interfaces", one that assumes pure, and one that assumes imperative.

code/drasil-gool/lib/Drasil/GOOL/RendererClassesOO.hs
B-rando1: I looked for `getFunc` from here and `getMethod` from InterfaceGOOL.

`getFunc` was renamed and moved around a few times, the earliest being when it was renamed from `get` to `getFunc` in [Created 'get' Value Function combination](commit/78eb0502ada96005935eab233af08a6345362526#diff-08b3c4d0291e235d6877ad838743580fd10df2e3405c21f808fc4566da184d42). Before then, it seems to appear from nowhere in [Finally Tagless GOOL](https://github.com/JacquesCarette/Drasil/commit/c5d7c424af850f95affff3e9b4468352fb347890?diff=unified#diff-4252893fa55097eed9030e01e08f9223e2da04bbed0f630161aae4eddfa72f58).

`getMethod` also seems to appear from nowhere in [Finally Tagless GOOL](https://github.com/JacquesCarette/Drasil/commit/c5d7c424af850f95affff3e9b4468352fb347890?diff=unified#diff-4252893fa55097eed9030e01e08f9223e2da04bbed0f630161aae4eddfa72f58).

I think the history is weird for that commit because it was merged from Brooks's fork. I'll have to go there and see if I can trace it back further when I have time.

PR #4223, opened 2025-06-20
Vector integration in the DblPend example

sarrasoussia:
This PR was originally intended to isolate the dblpend integration work from #4261, but a few small updates to other examples were necessary to ensure the entire project builds successfully.

PR #4407, opened 2025-10-14
Code generation after GA integration for the DblPend example

sarrasoussia:


JacquesCarette:
Perhaps it is the last 5 commits which are new, is that right?

balacij:
Similar to #4223's re-targetting, I re-targeted this PR to `vectorInteg` (i.e., #4223) so we can see the 'real' diff.

PR #4428, opened 2025-11-03
Remove `CodeBlock` constructor of `RawContent`

balacij:
Contributes to #2885 

Why?

1. `CodeBlock` was only used in Projectile's lesson plan to display 3 variable assignments.
2. The code related to `CodeBlock` rendering for other document types was incomplete.
3. It creates a dependancy chain between `drasil-code`, `drasil-docLang`, and `drasil-printers` that forces code from those packages to live in `drasil-lang`. At the very least, it contributes to keeping `CodeExpr` in `drasil-lang`.


code/stable/projectile/Lesson/Projectile_Lesson.ipynb
```diff
@@ -394,11 +394,10 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
```
JacquesCarette, line 397: I'm fine with removing `CodeBlock`, but I am not fine with "cell_type" changing from "code" to "markdown".

balacij: Note that this is the affected area:

Screenshot 2025-11-04 at 6 02 17 PM

With this PR, it becomes:

Screenshot 2025-11-04 at 5 51 27 PM

The underlying goal of this PR is to unblock moving the `CodeExpr`-related things into `drasil-code` and the display-related things into `drasil-docLang` or `drasil-printers` (undecided). Without this PR, since `drasil-code` depends on `drasil-lang` and `drasil-lang` depends on this code, we would have a cyclic dependency between `drasil-code` and `drasil-lang`.

I don't know how much we want to prioritize the lesson plan and Jupyter outputs right now, they haven't been touched in a while and without going into details, they look like they could use a lot of work (e.g., it is abusing the `$=` in `../Projectile/Lesson/Example.hs`). So, I see this PR as a minor setback for the lesson plan in favour of unblocking other (more important) work that could help this work again (and 'better') in the future.

Forcing this to stay as "code" is fair, but in that case, we would need to keep `CodeBlock`, changing its parameter from taking a `CodeExpr` to an `Expr` (which is also odd). I chose not to attempt this route originally because of said 'oddness', but I'm fine with trying it.

JacquesCarette: Let's keep thinking about other solutions to have certain things end up in the Jupyter document marked as 'code'. For example, a wrapper around `Expr` instead of being *inside* `Expr` might do. All we really need is enough information to emit a piece of code in the back-end.

smiths: I like the idea of trying to keep the Jupyter notebook having code, even if the solution isn't elegant.

balacij: > even if the solution isn't elegant

I think we can arrange that (without causing any further issue) 😄

balacij:
I'll leave this PR up as a reminder to myself, but I will close it on Friday at the latest if I haven't re-done the work.

PR #4430, opened 2025-11-03
Move `ExprC CodeExpr` typeclass instance to `Drasil.Code` namespace in `drasil-lang`

balacij:
Contributes to #2885


JacquesCarette (approved):
I was concerned about creating an orphan, but I see that this is rather the opposite, moving the instance closer to the declaration.

PR #4431, opened 2025-11-03
Install fresh JupyterLab copy to fix broken MathJax JupyterLab-based SRSs converted into HTML

### LINKED ISSUE #4339: Broken Jupyter TeX rendering in HTML extracted copy; MathJax version linked to CI-local Javascript file
In #4333, I added HTML-extracted copies of our Jupyter notebooks. See https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/Jupyter/DblPend_SRS.html .

The issue is that the LaTeX bits are not being rendered at all. When testing the CI on my fork and locally, I didn't realize that jupyter installations were broken -- so I had wrongly assumed that this was normal. 

The reason why DblPend's HTML-rendered Jupyter notebooks aren't loading the LaTeX correctly is because of the following line:

```html
<script src="file:///usr/share/javascript/mathjax/MathJax.js?config=TeX-AMS_CHTML-full,Safe"> </script>
```

Possible solutions:

1. Use a fresh copy of `Jupyter` using `pip` instead of the CIs shipped ones.
2. Pass a hardcoded `--mathjax='$FIXEDLINK'` to the `nbconvert` command. If using an outdated copy of Jupyter, this might not work however, as `--mathjax` was added in 2021 (jupyter/nbconvert # 1609) and our CI might use really old packages.

balacij:
Closes #4339

You can see a demo at https://balacij.github.io/Drasil/examples/swhsnopcm/SRS/Jupyter/SWHSNoPCM_SRS.html

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -129,7 +129,7 @@ jobs:
 
       - name: "Extract HTML copies of Jupyter Notebooks"
         run: |
-          sudo apt-get install -y --fix-missing jupyter
+          pip install jupyterlab
           make jupyter_html
         if: ${{ steps.changes.outputs.md == 'true' || fromJSON(env.is_deployment) }}
 


.github/workflows/Build.yaml
```diff
@@ -129,7 +129,7 @@ jobs:
 
       - name: "Extract HTML copies of Jupyter Notebooks"
         run: |
-          sudo apt-get install -y --fix-missing jupyter
+          pip install jupyterlab
```
balacij, line 132: Just one note: This does not cache jupyterlab' python package. I don't think that matters too much, however?

JacquesCarette: we'll need to redo a profile of our build CI (not for this PR). Things do seem to have drifted again, and I'm pretty sure we can make things more efficient.

PR #4433, opened 2025-11-04
removes extra calls to mkDocDesc

### LINKED ISSUE #4426: Broken control flow for dealing with SRS document
So I was trying to get rid of the "bad" automation that adds input requirements automatically, i.e.
https://github.com/JacquesCarette/Drasil/blob/75dbc3f07ec51c6f4abdefc400c5a0e055ad72c6/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs#L41-L45

But there's that call to `nub` there which is suspicious. It could be just "well, if someone adds the same requirement twice, don't worry about it". So I tried to take it out, just to see. What happens is that the *input requirements* get added twice!!!  So that `nub` is there to deal with that.

So I try something, out of curiousity: add a call to `trace` (from `Debug.Trace`) to let me see. And that's the big surprise: `fullReqs` does not get called twice, it gets called *four* times for all of SRS that use `FReqSub` !

JacquesCarette:
closes #4426 .

Some de-duplication between dplpend and sglpend snuck in as well.

### DIFF
--- a/code/drasil-example/template/lib/Drasil/Template/Body.hs
+++ b/code/drasil-example/template/lib/Drasil/Template/Body.hs
@@ -11,7 +11,7 @@ import Drasil.System (SystemKind(Specification), mkSystem, systemdb)
 import Drasil.Metadata
 import Language.Drasil
 import Drasil.SRSDocument
-import Drasil.DocLang (DocDesc)
+import Drasil.DocLang (DocDesc, tunitNone)
 import Drasil.Generator (cdb)
 import Theory.Drasil (DataDefinition, GenDefn, InstanceModel, TheoryModel)
 import qualified Language.Drasil.Sentence.Combinators as S
@@ -19,7 +19,6 @@ import qualified Language.Drasil.Sentence.Combinators as S
 import qualified Drasil.DocLang.SRS as SRS
 import Data.Drasil.Citations
 import Drasil.DocumentLanguage.TraceabilityGraph
-import Drasil.DocLang (tunitNone)
 
 sd  :: (System , DocDesc)
 sd = fillcdbSRS mkSRS si
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs
@@ -8,9 +8,8 @@ import qualified Language.Drasil.Development as D
 
 import Drasil.Metadata as M (dataDefn, inModel, thModel)
 import Drasil.SRSDocument
-import Drasil.DocLang (DocDesc)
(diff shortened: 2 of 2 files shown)


code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -43,11 +44,15 @@ import Drasil.PDController.ODEs (pidODEInfo)
...
 
-srs :: Document
-srs = mkDoc mkSRS (S.forGen titleize phrase) fullSI
+sd  :: (System , DocDesc)
+sd = fillcdbSRS mkSRS si
 
+-- sigh, this is used by others
 fullSI :: System
-fullSI = fillcdbSRS mkSRS si
+fullSI = fst sd
```
balacij, line 52: Because you mentioned this on Discord, we can do:

```haskell
fullSI :: System
sd :: (System, DocDesc)
sd@(fullSI, _) = fillcdbSRS mkSRS si
```

But I'm not exactly a fan of that. The `fst` is more readable IMO.

code/drasil-example/template/lib/Drasil/Template/Body.hs
```diff
@@ -20,22 +21,26 @@ import Data.Drasil.Citations
...
+sd  :: (System , DocDesc)
+sd = fillcdbSRS mkSRS si
 
+-- sigh, this is used by others
 fullSI :: System
-fullSI = fillcdbSRS mkSRS si
+fullSI = fst sd
+
+srs :: Document
+srs = mkDoc mkSRS (S.forGen titleize phrase) sd
```
balacij, line 31: In subsequent work, we should abstract all of this duplicated blocks of code somewhere/somehow.

balacij (dismissed):
LGTM!

balacij (approved):

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
JacquesCarette: Ah, that is the idiom that I was looking for! But happy enough not to use it.

code/drasil-example/template/lib/Drasil/Template/Body.hs
JacquesCarette: I agree - later.

JacquesCarette:
Hmm, now that you've made some changes @balacij , that means you can't be an approver anymore! Maybe we can get @smiths to?

smiths (approved):
I approve the changes based on the quality of the discussion between @JacquesCarette and @balacij, rather than on expert knowledge of the changes on my part.  :smile:

Issue #4349, opened 2025-08-28
Fill in `HasChunkRefs` lists


balacij:
With #4320, we introduced the following typeclass:

```haskell
-- | All chunks should expose what chunks they reference/rely on, so that we can
-- test 'ChunkDB's to ensure all presupposed chunks are already registered.
class HasChunkRefs a where
  chunkRefs :: a -> S.Set UID
```

Currently, all instances are are set to `mempty`. We should fill these in.

After this issue is resolved, we should:

1. Look in to what it would take to place chunks in the database "in order" (i.e., in a strict order about dependencies)
2. Look to rebuilding `ChunkDB`s `traceTable` and `refbyTable`.

Xinlu-Y:
> [@Xinlu-Y](https://github.com/Xinlu-Y) Are you still working on this? If so, have you made any progress on `ConceptChunk` and/or `ConceptInstance`? They seem to create a messy situation with `Citation`s that affects stable. If [#4700](https://github.com/JacquesCarette/Drasil/pull/4700) ends up being fruitful, then the `Citation`s issue might not be so bad, but I'm unsure right now.

yes, I’m still working on it, but I haven’t made much progress yet. I’ll take another look tomorrow and update you.

balacij:
Thank you, @Xinlu-Y, it looks like you got to all of the lists!

Are we okay to close this issue?

PR #4434, opened 2025-11-04
Fill chunk refs

Xinlu-Y:
#4349
- Implement real `HasChunkRefs` lookups for `ConceptChunk`, `ConceptInstance`, and `DefinedQuantityDict` so referenced UIDs flow up from definitions and short names

- Rework the base chunk DB to seed the few math citations up front, drop duplicate concept chunks pulled in via DQDs, and keep citation UIDs out of the reference table

- Add `inReqDesc` and thread it through every SRS so the input-values requirement now references the actual table; this includes injecting the table into each `LabelledContent` list and putting `SRS.sectionReferences` into `allRefs`

- Minor cleanups triggered by the new wiring, such as reordering `SI` derived units and adjusting assumption lists so the generated refs resolve


code/drasil-data/lib/Data/Drasil/SI_Units.hs
```diff
@@ -11,9 +11,10 @@ fundamentals :: [UnitDefn]
 fundamentals = [metre, kilogram, second, kelvin, mole, ampere, candela]
 
 derived :: [UnitDefn]
-derived = [becquerel, calorie, centigrade, coulomb, farad, gray, henry, hertz, joule,
-  katal, kilopascal, kilowatt, litre, lumen, lux,  millimetre, newton, ohm,
-  pascal, radian, siemens, sievert, steradian, tesla, volt, watt, weber]
+derived = [becquerel, centigrade, coulomb, hertz, joule, katal, litre,
```
JacquesCarette, line 18: Why have these changed order? We should make such changes unless there is a good reason to. Previous were in strict alphabetical order.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Assumptions.hs
```diff
@@ -43,35 +43,35 @@ assumpLDFC         = cic "assumpLDFC" (ldfConstantDesc loadDF)         "ldfConst
...
 
 standardValuesDesc :: UnitalChunk -> Sentence
-standardValuesDesc mainIdea = foldlSent [D.toSent $ atStartNP' (the value), S "provided in",
-  refS $ SRS.valsOfAuxCons ([]::[Contents]) ([]::[Section]), S "are assumed for the", phrase mainIdea,
-  sParen (ch mainIdea) `sC` S "and the", plural materialProprty `S.of_`
-  foldlList Comma List (map ch (take 3 assumptionConstants))]
+standardValuesDesc mainIdea = foldlSent [atStartNP' (the value), S "provided in",
+  refS $ SRS.valsOfAuxCons ([]::[Contents]) ([]::[Section]), S "are assumed for the", phrase mainIdea, 
+  sParen (ch mainIdea) `sC` S "and the", plural materialProprty `S.of_` 
+  foldlList Comma List (map (ch . view defLhs) assumptionConstants)]
```
JacquesCarette, line 72: `take 3` was a hack, but I'm concerned that this will actually change `stable`?

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -90,7 +92,7 @@ mkSRS
                  ShowDerivation,
                Constraints EmptyS inputsUC]],
 
-     ReqrmntSec $ ReqsProg [FReqsSub EmptyS [], NonFReqsSub], LCsSec,
+     ReqrmntSec $ ReqsProg [FReqsSub inputValuesDescription [], NonFReqsSub], LCsSec,
```
JacquesCarette: I'm sure this is a good change, but it is going to lead to changes in `stable`. So it is usually best to do such changes in isolation (even if it means five 1-line change PRs) instead of being bundled in a larger PR.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -146,13 +148,13 @@ symbMap = cdb (map dqdWr physicscon ++ symbolsAll ++ [dqdWr mass, dqdWr posInf,
...
   conceptInstances
-  labelledContent
+  labelledContentWithInputs
   allRefs
   citations
 
 -- | Holds all references and links used in the document.
 allRefs :: [Reference]
-allRefs = [externalLinkRef]
+allRefs = externalLinkRef : SRS.sectionReferences ++ map ref labelledContentWithInputs
```
JacquesCarette: Because this change, which is likely also good, will also change `stable`. We want to review these as separately as possible.

code/drasil-example/projectile/lib/Drasil/Projectile/Assumptions.hs
```diff
@@ -19,13 +18,13 @@ import Data.Drasil.Concepts.Physics (acceleration, collision, distance, gravity,
 import Drasil.Projectile.Concepts (launcher, projectile, target, projMotion)
 
 assumptions :: [ConceptInstance]
-assumptions = [twoDMotion, cartSyst, yAxisGravity, launchOrigin, targetXAxis,
-  posXDirection, constAccel, accelXZero, accelYGravity, neglectDrag, pointMass,
-  freeFlight, neglectCurv, timeStartZero, gravAccelValue]
+assumptions = [twoDMotion, neglectCurv, cartSyst, yAxisGravity, accelYGravity,
```
JacquesCarette, line 23: why the reorder?

code/drasil-example/swhs/lib/Drasil/SWHS/Assumptions.hs
```diff
@@ -32,7 +30,7 @@ import Drasil.SWHS.Unitals (wVol, volHtGen, tempC, tempInit, tempW,
 
 assumptions :: [ConceptInstance]
 assumptions = [assumpTEO, assumpHTCC, assumpCWTAT, assumpTPCAV, assumpDWPCoV, assumpSHECoV,
-  assumpLCCCW, assumpTHCCoT, assumpTHCCoL, assumpLCCWP, assumpCTNOD, assumpSITWP,
+  assumpLCCCW, assumpTHCCoT, assumpTHCCoL, assumpLCCWP, assumpSITWP, assumpCTNOD,
```
JacquesCarette, line 36: re-order?

code/drasil-example/swhs/lib/Drasil/SWHS/MetaConcepts.hs
```diff
@@ -2,14 +2,19 @@ module Drasil.SWHS.MetaConcepts where
...
   "solar water heating systems") "SWHS" [materialEng]
 
--- HACK: should re-decompose this noun phrase back into components!
 progName' :: CI
-progName' = commonIdea "swhsPCM" (nounPhraseSP "solar water heating systems incorporating PCM")
+-- Nounphrase'' hack to get nounPhraseSP words to accept
+-- nounPhrases instead of strings
+-- Another capitalization hack.
+progName' = commonIdea "swhsPCM" (nounPhrase''
+  (S "solar water heating systems incorporating PCM")
```
JacquesCarette: The first part (merging "incorporation") is fine, but inlining `PCM` is not a good change.

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
```diff
@@ -69,15 +73,19 @@ basisConceptChunks =
   --              in the basis.
   [algorithm, errMsg, program] ++ srsDomains ++ mathcon
 
+basisCitations :: [Citation]
+basisCitations = [cartesianWiki, lineSource, pointSource]
```
JacquesCarette, line 75: Why?

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
```diff
@@ -97,16 +105,23 @@ basisCDB =
...
-  $ insertAll tm
-  $ insertAll ci
-  $ insertAll cits
-  $ basisCDB { labelledcontentTable = idMap lc,
-               refTable = idMap r }
+cdb s t c u d ins gd tm ci lc r cits = finalDB
+  where
+    baseDB = basisCDB { labelledcontentTable = idMap lc,
+                        refTable = idMap r }
+    withIdeaDicts = insertAll t baseDB
```
JacquesCarette: I understand why you would find this more readable... if you are going to make this change, at least line up the `=` signs!

I don't actually think this is really an improvement though.

JacquesCarette (changes requested):
There seems to be multiple logically independent sets of changes in this PR. Why all in here? It feels like this should be broken up into smaller PRs.

code/drasil-data/lib/Data/Drasil/SI_Units.hs
Xinlu-Y: Because this change ensures every derived unit is inserted after the units it depends on. 
For example, `pascal` precedes `kilopascal`, `volt` precedes `farad/ohm`, and `weber` precedes `tesla/henry`. https://github.com/JacquesCarette/Drasil/blob/e36c1ec0f40bd2eb24941d826659ee830e9a88b4/code/drasil-database/lib/Database/Drasil/ChunkDB.hs#L175-L178 checks that all referenced units already exist when a new unit is inserted. The old order triggered missing-dependency errors.

code/drasil-example/projectile/lib/Drasil/Projectile/Assumptions.hs
Xinlu-Y: I reorder the assumptions so that each assumption appears after any other assumptions it refers to in its description.

In `Assumptions.hs`, some assumptions use `fromSource` or `fromSources` to refer to other assumptions.
These functions automatically create clickable links in the generated SRS document.
However, in the old order, some of those links pointed forward to assumptions that hadn’t been listed yet, so the links appeared but didn’t work when clicked.

The new order ensures that all referenced assumptions are introduced before they are linked.
For example:
- `neglectCurv` now comes before `cartSyst`
- `accelXZero`, `accelYGravity`, `neglectDrag`, and `freeFlight` now come before `constAccel`

code/drasil-example/swhs/lib/Drasil/SWHS/MetaConcepts.hs
Xinlu-Y: I’ll revert the inlined PCM.

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
Xinlu-Y: The previous form made it very hard to track the registration order of different chunk types.
even a slight ordering mistake could trigger the “Missing dependency” errors I kept running into.
During this round of fixes, I repeatedly ran into such issues:
- `GlassBR: assumpGC referenced astm2009, but the citations were inserted later.`
- `SWHS / SWHSNoPCM / PDController / SSP: functional requirements were registered before ReqInputs or the data-constraint tables they reference.`
- `Projectile: constAccel cited assumptions that weren’t yet in the database.`
- `DoublePendulum: inputValues referred to a table that hadn’t been inserted.`

To fix these missing-chunk errors, I think i had to keep verifying which kinds of chunks were inserted first.

Xinlu-Y:
I’m splitting this PR into smaller ones for easier review, stable updates will follow.

code/drasil-data/lib/Data/Drasil/SI_Units.hs
JacquesCarette: Excellent reason, thank you! You should put in a comment to that effect. I would make sure that the base units are in alphabetical order, and then by dependency (+alphabetical) order after that.

code/drasil-example/projectile/lib/Drasil/Projectile/Assumptions.hs
JacquesCarette: Great. Please add a comment to that effect. (And everywhere else you did this.)

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
JacquesCarette: Ah, interesting. This is likely to happen again.

I think the conclusion is that inserting into the chunk database by chunk type is very fragile. We're lucky it worked at all. Please create an issue for this: I think we should change our mechanism for insertion to be more dependency-aware. But that's a much bigger change, so should not be done now. So your changes here can stay, as they'll end up being moot when this gets changed.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
Xinlu-Y, line 1: Now that we use `insertAllOutOfOrder12`, the first call to `cdb … concIns labCon …` registers both the requirement (`instance:inputValues`) and its table (`doc:ReqInputs`) together, so the dependency is satisfied.
But when `fillReqs` runs later, it calls `insertAll` again for the concept instances only, without re-inserting their labelled content.
At that point `insertRefExpectingExistence` looks in the `chunkTable` for `doc:ReqInputs`, can’t find it, and throws error.

So here I just skip `fillReqs` when the system already provides its own requirements to avoid the duplicate insertion and keep the dependency check happy.

As a result, the requirement order in the chunk table becomes existing FRs first, followed by new NFRs

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
Xinlu-Y: Yeah, those citations are in `basisCitations` because the base math concepts: `cartesian`, `line`, and `point` already hard-code those references through `fromSource`.
So once any of those concepts are pulled into a `ChunkDB`, the matching citations have to be there too, otherwise the generator complains about missing references.

Putting them in `basisCitations` guarantees everything builds cleanly, but it’s a bit clunky. I have to manually update that global list. maybe there’s a better way to handle this. I’d love to hear from you.

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -261,8 +261,13 @@ insertAllOutOfOrder12 strtr as bs cs ds es fs gs hs is js lcs rs =
...
 
-    -- Calculate what chunks are depended on (i.e., UID -> Dependants)
-    chDpdts = invert $ M.fromList $ map (\c -> (c ^. uid, S.toList $ chunkRefs c)) calt
+    -- Calculate what chunks are depended on (i.e., UID -> Dependants), but only
+    -- keep dependencies that correspond to chunks we are actually inserting or
+    -- that already exist in the chunk table. References/LabelledContent live in
+    -- separate tables and are handled elsewhere.
+    allowedDeps = S.fromList (M.keys (chunkTable strtr) ++ map (^. uid) calt)
+    chDpdts = invert $ M.fromList $
+      map (\c -> (c ^. uid, S.toList $ S.filter (`S.member` allowedDeps) (chunkRefs c))) calt
```
balacij, line 270: I'm not quite sure I'm understanding this change. Does it not make it so that any chunk dependancy that is neither already in the `ChunkDB` nor in the list of chunks being added, is ignored? I think this would mean that we can insert chunks into the `ChunkDB` that have unsatisfied dependencies?

code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
```diff
@@ -150,7 +150,7 @@ symbMap = cdb (map (^. output) iMods ++ symbolsAll) ideaDicts conceptChunks []
 
 -- | Holds all references and links used in the document.
 allRefs :: [Reference]
-allRefs = [externalLinkRef]
+allRefs = externalLinkRef : SRS.sectionReferences ++ map ref (labelledContent ++ funcReqsTables)
```
balacij, line 153: Was this really needed? The `fillcdbSRS` function should be doing this already for any `LabelledContent` registered in the `ChunkDB`. (I'm not saying it should be doing that, only that it currently does that.)

code/drasil-lang/lib/Language/Drasil/Chunk/Concept/Core.hs
```diff
@@ -67,7 +76,11 @@ data ConceptInstance = ConInst { _ciuid :: UID
 makeLenses ''ConceptInstance
 
 instance HasChunkRefs ConceptInstance where
-  chunkRefs = const mempty -- FIXME: `chunkRefs` should actually collect the referenced chunks.
+  chunkRefs ci =
+    let conceptRefs    = chunkRefs (ci ^. cc)
+        shortNameRefs  = collectSentenceRefs (getSentSN (shortname ci))
+        domainRefs     = S.fromList (cdom ci)
+    in conceptRefs `S.union` shortNameRefs `S.union` domainRefs
```
balacij, line 83: This is really good! And your code now makes me think: Can we instantiate `HasChunkRefs` for `Sentence`? I think that would help simplify your code and make it so that you don't need to import anything here. It would only be union of the 3 lists of `chunkRefs` -- e.g., 
```
              chunkRefs (ci ^. cc)
    `S.union` (S.fromList (cdom ci))
    `S.union` chunkRefs (ci ^. defn)
    
...
(elsewhere)...

instance HasChunkRefs Sentence where
  chunkRefs s = ...
```
?

balacij: Now this makes me think: if all of this code is going to be more or less following the same structure, I think we can use deriving-via to auto-generate these instances.

code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
balacij: Ah, so they're requirements of other chunks already in the `basisChunkDB` then? That sounds sensible to me then. This one could probably be pulled out into its own PR.

balacij:
It looks like a considerable amount of this code was already merged into `main`.

Re: manually created instances of `HasChunkRefs`: it looks like we've got at least one mistake in `main` now, so I think we should go down the automated route to avoid any other issue while we're re-working our chunks, at least for now -- https://github.com/JacquesCarette/Drasil/issues/5159.

JacquesCarette:
So should I be ignoring this PR entirely, as it is / will be superseded by others? It is seriously bit-rotting!

Xinlu-Y:
> So should I be ignoring this PR entirely, as it is / will be superseded by others? It is seriously bit-rotting!

Yes, we can definitely close this PR.

PR #4435, opened 2025-11-04
Reduce whitespace-related PR clutter (for future PRs)

balacij:
This PR does two things:

1. It removes all trailing whitespace on all Haskell code lines.
2. It ensures all Haskell files have a trailing newline character.

The reason to do this is twofold (1) code style and (2) to avoid clutter of doing it slowly in each of our PRs.


balacij:
Ah, that's a good idea. Let's do that in a subsequent PR and issue.

PR #4445, opened 2025-11-11
Switch to out-of-order chunk insertion for the `ChunkDB`

### LINKED ISSUE #4441: ChunkDB insertion order is fragile: dependency checks fail if chunks are registered out of sequence
_Originally posted by @Xinlu-Y in https://github.com/JacquesCarette/Drasil/pull/4434#discussion_r2500567074_

balacij:
Builds on #4446 and #4454 .

Closes #4441.

This PR implements @smiths' suggestion for helping @Xinlu-Y with her `ChunkDB` work, i.e., to switch to out-of-order chunk insertion into the basis `ChunkDB`.

This code is delicate, so please take a bit more time than usual to double-check that I did everything correctly.


code/drasil-website/lib/Drasil/Website/Body.hs
```diff
@@ -94,9 +93,9 @@ symbMap :: FolderLocation -> ChunkDB
 symbMap fl = cdb ([] :: [DefinedQuantityDict]) (map nw [webName, phsChgMtrl, twoD] ++
   map getSysName allExampleSI ++ map nw [pendulum, motion, rigidBody, blast,
   heatTrans, sWHT, water, pidC, target, projectile, crtSlpSrf, shearForce,
-  normForce, slpSrf] ++ [nw $ fctSfty ^. defLhs] ++ [game, physics, condition, glaSlab, intrslce,
-  slope, safety, factor]) ([] :: [ConceptChunk]) ([] :: [UnitDefn]) [] [] [] []
```
balacij: These chunks need to be removed from the `ChunkDB` or else the website will crash with a `duplicate chunk found in mass insertion` error (for each of them individually).

code/drasil-lang/lib/Drasil/Database/Chunk.hs
```diff
@@ -39,6 +39,10 @@ instance HasUID Chunk where
   uid :: Getter Chunk UID
   uid = to (\(Chunk c) -> c ^. uid)
 
+instance HasChunkRefs Chunk where
+  chunkRefs :: Chunk -> S.Set UID
+  chunkRefs (Chunk c) = chunkRefs c
```
balacij, line 44: Surprised I didn't already write an instance for `HasChunkRefs Chunk`... but, it's needed. I'll push this into a separate PR.

balacij: Because commit splitting was a discussion today as well, I split off fab54bf7f92b70cfe4b648e45c4e436311f653ba with the following steps:

1. Beginning an interactive rebase: `git rebase -i HEAD~N`, where `N` is the number of commits I have on top of the base branch
2. Replace 'pick' with 'edit' for the commits I need to split.
3. When the `git rebase` stops on the commit, run `git reset HEAD^` (this is a _soft reset_ -- code is kept **unstaged**).
4. Selectively pick which code you want to commit, and commit it as normally. For example, `git add potato.hs && git commit && git add stew.hs && git commit`. Note: you **need** to have eventually have re-committed all changes you unstaged in each commit you wanted to split.
5. `git rebase --continue`.

I would suggest we add this to the wiki, but I don't think it's something anyone should be doing too often. It's just not fun in general when changes are heavily intertwined. We could add it regardless, though, just to have it for completeness sake.

balacij: (@Xinlu-Y , you might be interested in these steps if you want to look into breaking up your commits for your ChunkRefs-related PR)

smiths: Thank you for the advice on commit splitting @balacij.  I agree that it isn't something we would want to plan on doing, but sometimes it does come up.  I like the idea of having the instructions somewhere easier to find than in this discussion.  :smile:  Putting them on the wiki sounds good, but we might want to wait until @B-rando1 has completed the refactor.  :smile:

code/drasil-database/lib/Database/Drasil/ChunkDB.hs
```diff
@@ -229,6 +241,82 @@ insertAll as cdb = foldl' (flip insert) cdb as
...
+    chDeps = invert $ M.fromList $ map (\c -> (c ^. uid, S.toList $ chunkRefs c)) calt
+    hardLookup k m = fromMaybe (error "this situation will never happen") $ M.lookup k m
+
+    -- Create the chunk table for the incoming chunks
+    chDepsTab = M.fromList ((\c -> (c ^. uid, (c, hardLookup (c ^. uid) chDeps))) <$> calt)
+
+    -- Insert all incoming chunks with the existing chunk table, asserting that
+    -- none of the inserted chunks were already inserted.
+    chTab = M.unionWith (\(x, _) _ -> error $ "duplicate chunk found in mass insertion: " ++ show (x ^. uid))
+      (chunkTable strtr) (M.fromList $ map (\c -> (c ^. uid, (c, []))) calt)
```
balacij: To be perfectly clear, it _is necessary_ to add the new chunks in two phases: adding with an empty list of dependancies and then adding the list of dependencies because we need to make sure we're not overwriting any chunks of different types. Perhaps the code is just 'easier to write' this way and is not the 'most efficient' this way, however...

I think we would need 'deep equality' for the more memory-efficient version (i.e., the version that does not create two maps with more or less the same contents excluding the dependancies.

balacij:
Thank you, no rush, whenever you need this code really.

code/drasil-website/lib/Drasil/Website/Body.hs
balacij: Spun off into #4454.

JacquesCarette:
Merge conflict - and I don't know in which direction to go.

PR #4453, opened 2025-11-13
`drasil-gen` touch-ups

balacij:


JacquesCarette (dismissed):
This is a very nice set of abstractions. Clearly pays off in the examples.

PR #4465, opened 2025-11-17
Remove fillReqs, fullReqs and other similar functions that cause "Chunk overwrite" warnings w.r.t. the `ChunkDB`

### LINKED ISSUE #4345: `fillReqs`s is responsible for the remaining within-table `ChunkDB`-related chunk overwrites

### LINKED ISSUE #4450: Remaining 'chunk overwriting' warnings: Remove `fillReqs` & `fullReqs`
Relevant prior PR and discussion: #4351 and #4352 

The goal of this issue is to deal with the remaining 'chunk overwriting' errors that currently pop up when you run any of Drasil's case studies. For example, Projectile:

```console
WARNING! Overwriting `instance:verifyInVals` :: ConceptInstance
WARNING! Overwriting `instance:calcValues` :: ConceptInstance
WARNING! Overwriting `instance:outputValues` :: ConceptInstance
```

These occur because these chunks are already added to the `ChunkDB` but `fillReqs` adds them again:

https://github.com/JacquesCarette/Drasil/blob/9af23cb94ee638f3fc264e25ee73394de782d9ab/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L177-L185

This function is needed because for some of our examples, we generate the "input-values" functional requirement by passing a non-empty "input-values" FR description to `FReqsSub` in the `SRSDecl`:

https://github.com/JacquesCarette/Drasil/blob/9af23cb94ee638f3fc264e25ee73394de782d9ab/code/drasil-example/glassbr/lib/Drasil/GlassBR/Body.hs#L108-L111

https://github.com/JacquesCarette/Drasil/blob/9af23cb94ee638f3fc264e25ee73394de782d9ab/code/drasil-docLang/lib/Drasil/DocDecl.hs#L126-L129

https://github.com/JacquesCarette/Drasil/blob/9af23cb94ee638f3fc264e25ee73394de782d9ab/code/drasil-docLang/lib/Drasil/Sections/Requirements.hs#L40-L44

#### Objective

Generating input-values FRs is a good idea, but not in this route that heavily ties `drasil-docLang` to ICO problems. #4352 can be used for figuring out how to do this 'correctly', but, for now, the goal of this issue is just to undo some of this generation in favour of doing it 'better' later. It is more important to not have chunk overwrites.

#### Prospective Solution

1. Creating an "input-values" FR for each of the case studies. This doesn't mean copy/pasting the same code everywhere, it could just mean _importing_ the same common FR chunk from elsewhere.
2. Registering said chunk and disabling the automated input-values FR generation.
3. Removing `fillReqs` and `fullReqs`.

balacij:
Closes #4345
Closes #4450

Follows the solution mentioned in #4450.

Leaving this one as a draft because I need to format code and rename some variables for consistency. Overall, however, it is reviewable.

After this PR, we should be able to enable hard chunk-overwrite errors! 😄 

We should also file an issue to re-do the auto-generated I-C-O FRs 'correctly'.

### DIFF
--- a/code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
+++ b/code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
@@ -95,7 +95,7 @@ mkSRS
                  ShowDerivation,
                Constraints EmptyS inputsUC]],
 
-     ReqrmntSec $ ReqsProg [FReqsSub EmptyS funcReqsTables, NonFReqsSub], LCsSec,
+     ReqrmntSec $ ReqsProg [FReqsSub funcReqsTables, NonFReqsSub], LCsSec,
      TraceabilitySec $ TraceabilityProg $ traceMatStandard si, Bibliography]
 
 si :: System
--- a/code/drasil-docLang/lib/Drasil/DocDecl.hs
+++ b/code/drasil-docLang/lib/Drasil/DocDecl.hs
@@ -94,10 +94,8 @@ newtype ReqrmntSec = ReqsProg [ReqsSub]
 
 -- | Requirements subsections.
 data ReqsSub where
-  -- | Functional requirements. 'LabelledContent' for tables (includes input values).
-  FReqsSub    :: Sentence -> [LabelledContent] -> ReqsSub
-  -- | Functional requirements. 'LabelledContent' for tables (no input values).
-  FReqsSub'   :: [LabelledContent] -> ReqsSub
+  -- | Functional requirements. 'LabelledContent' for tables.
+  FReqsSub    :: [LabelledContent] -> ReqsSub
   -- | Non-Functional requirements.
   NonFReqsSub :: ReqsSub
 
@@ -123,8 +121,7 @@ mkDocDesc sys@SI{_systemdb = db} = map sec where
   sec (OffShelfSolnsSec e) = DL.OffShelfSolnsSec e
 
   reqSec :: ReqsSub -> DL.ReqsSub
-  reqSec (FReqsSub _ t) = DL.FReqsSub (fromConcInsDB funcReqDom) t
-  reqSec (FReqsSub' t) = DL.FReqsSub' (fromConcInsDB funcReqDom) t
+  reqSec (FReqsSub t) = DL.FReqsSub (fromConcInsDB funcReqDom) t
   reqSec NonFReqsSub = DL.NonFReqsSub $ fromConcInsDB nonFuncReqDom
 
   ssdSec :: SSDSub -> DL.SSDSub
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -355,7 +355,6 @@ mkReqrmntSec (ReqsProg l) = R.reqF $ map mkSubs l
   where
     mkSubs :: ReqsSub -> Section
     mkSubs (FReqsSub  frs tbs) = R.fReqF (mkEnumSimpleD frs ++ map LlC tbs)
-    mkSubs (FReqsSub' frs tbs) = R.fReqF (mkEnumSimpleD frs ++ map LlC tbs)
     mkSubs (NonFReqsSub nfrs) = R.nfReqF (mkEnumSimpleD nfrs)
 
 -- ** Likely Changes
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage/Core.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage/Core.hs
@@ -202,8 +202,6 @@ newtype ReqrmntSec = ReqsProg [ReqsSub]
 -- | Requirements subsections.
 data ReqsSub where
   -- | Functional requirements. LabelledContent needed for tables.
-  FReqsSub'   :: [ConceptInstance] -> [LabelledContent] -> ReqsSub
-  -- | Functional requirements. LabelledContent needed for tables.
   FReqsSub    :: [ConceptInstance] -> [LabelledContent] -> ReqsSub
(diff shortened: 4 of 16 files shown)


code/drasil-example/glassbr/lib/Drasil/GlassBR/Requirements.hs
```diff
@@ -57,6 +57,10 @@ inReqDesc = foldlList Comma List [D.toSent $ pluralNP (NP.the (combineNINI glass
 sysSetValsFollowingAssumpsDesc = foldlSent [D.toSent $ atStartNP (the system), S "shall set the known",
     plural value, S "as described in the table for", namedRef sysSetValsFollowingAssumpsTable (S "Required Assignments")]
 
+inputValues :: ConceptInstance
+inputValuesTable :: LabelledContent
+(inputValues, inputValuesTable) = inReqWTab (Just inReqDesc) inputs
```
balacij, line 62: I'm fairly certain this is a code smell. It was there before, too, however. Just masked a bit.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -152,7 +152,7 @@ symbMap = cdb (map dqdWr physicscon ++ symbolsAll ++ [dqdWr mass, dqdWr posInf,
   theoreticalModels
   conceptInstances
   citations
-  labelledContent
+  (labelledContent ++ funcReqTables)
```
balacij: I'm adding the functional requirements tables here because it seems we cause an import cycle otherwise, which makes sense situationally because knowledge is not created in the order of chunk types.

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
```diff
@@ -2728,42 +2728,42 @@ <h1>Traceability Matrices and Graphs</h1>
               <td></td>
             </tr>
             <tr>
-              <td><a href=#correct>NFR:Correctness</a></td>
+              <td><a href=#inputValues>FR:Input-Values</a></td>
```
balacij, line 2731: I'm not quite sure _why_ re-ordering happens, but I'll leave that issue as out-of-scope.

code/stable/dblpend/TraceyGraph/allvsall.dot
```diff
@@ -82,16 +82,16 @@ digraph allvsall {
...
 	instance:calcAng	[shape=box, color=black, style=filled, fillcolor=ivory, label="FR:calcAng"];
 	instance:outputValues	[shape=box, color=black, style=filled, fillcolor=ivory, label="FR:outputValues"];
+	instance:correct	[shape=box, color=black, style=filled, fillcolor=ivory, label="NFR:correct"];
+	instance:portable	[shape=box, color=black, style=filled, fillcolor=ivory, label="NFR:portable"];
 
-	subgraph NFR {
+	subgraph FR {
 	rank="same"
-	{instance:correct, instance:portable, instance:inputValues, instance:verifyInptVals, instance:calcAng, instance:outputValues}
+	{instance:inputValues, instance:verifyInptVals, instance:calcAng, instance:outputValues, instance:correct, instance:portable}
```
balacij, line 94: What's funny about this is that this subgraph is now correctly titled!

balacij: Correction: almost. It still contains NFRs. I'm not sure why there isn't a split.

code/stable/swhsnopcm/src/python/Control.py
```diff
@@ -10,9 +10,9 @@
 import OutputFormat
 
 filename = sys.argv[1]
-A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D, A_tol, R_tol, E_W = InputParameters.get_input(filename)
+E_W, A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D, A_tol, R_tol = InputParameters.get_input(filename)
 V_tank = InputParameters.derived_values(D, L)
-InputParameters.input_constraints(A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D, E_W)
+InputParameters.input_constraints(E_W, A_C, C_W, h_C, T_init, t_final, L, T_C, t_step, rho_W, D)
```
balacij: There is something wonky going on with SWHSNoPCM...

balacij: Re-ordering issue, nevermind.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Requirements.hs
balacij: And I think that's what the original code witnessed and tried to deal with as well.

The key background knowledge here is that for any "ICO problem" (which Drasil does not quite capture well enough yet) is that there should always be at least 3 _kinds_ of FRs (the I-C-O ones).

JacquesCarette: No, this is just top-level pattern-matching on a tuple, it's fine.

And yes, I'm sure the I-C-O "pattern" will emerge some more as things get refactored.

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
JacquesCarette: I think some lists are taken verbatim (instead of being sorted) and the changes to the code changed that. 

This is actually a good change in order though: FRs should come before NFRs.

PR #4476, opened 2025-11-20
`declareHasChunkRefs`: Generate instances of `HasChunkRefs` where possible

balacij:
Contributes to #4434 and #4349

A `Chunk` (in the code) is defined by the following constraint:

```haskell
-- | Constraint for anything that may be considered a valid chunk type.
type IsChunk a = (HasUID a, HasChunkRefs a, Typeable a)
```

Currently, none of our `HasChunkRefs` instances are meaningfully filled in (#4349). The goal of #4434 is to fill them in. The goal of this PR is to assist with #4434 by "automating the boring stuff." #4434 commonly contains code of the following form:

```haskell
    let ideaRefs   = chunkRefs (c ^. idea)
        defnRefs   = collectSentenceRefs (c ^. defn')
        domainRefs = S.fromList (cdom' c)
        ...        = ...
    in ideaRefs `S.union` defnRefs `S.union` domainRefs `S.union` ...
```

This PR introduces a bit of Generics- and TH-based automation to derive these 'boring' instances, for example those that merely take the `union` of the various atoms within a chunk.

### DIFF
--- a/code/drasil-database/lib/Drasil/Database/Chunk.hs
+++ b/code/drasil-database/lib/Drasil/Database/Chunk.hs
@@ -64,10 +64,10 @@ class HasChunkRefs a where
   chunkRefs :: a -> S.Set UID
 
 instance HasChunkRefs UID where
-  -- | UIDs are UIDs, not *UID references*, a TypedUIDRef is a *reference*.
-  -- Therefore, `UID` has no chunk references. They should only be used for the
-  -- UID of a thing being defined, *not as a reference (unless specifically
-  -- within the 'ChunkDB')*.
+  -- | 'UID's are meant to be "owned" (i.e., they are the unique identifier of
+  -- the chunk being defined), not *carried as references to other chunks*.
+  -- 'TypedUIDRef t' exists to be used as a *reference to another chunk of type
+  -- 't'*. Therefore, `UID` has no chunk references.
   chunkRefs _ = S.empty
   {-# INLINABLE chunkRefs #-}
 
--- a/code/drasil-database/lib/Drasil/Database/TypedUIDRef.hs
+++ b/code/drasil-database/lib/Drasil/Database/TypedUIDRef.hs
@@ -6,13 +6,19 @@ module Drasil.Database.TypedUIDRef (
 import Control.Lens ((^.))
 import Data.Maybe (fromMaybe)
 
-import Drasil.Database.Chunk (IsChunk)
+import Drasil.Database.Chunk (IsChunk, HasChunkRefs (..))
 import Drasil.Database.ChunkDB (ChunkDB, find)
 import Drasil.Database.UID (HasUID(..), UID)
+import qualified Data.Set as S (singleton)
 
 -- | 'UID' references that contain information about the type of data the 'UID'
 -- refers to, useful for type-safe dereferencing.
 newtype TypedUIDRef typ = TypedUIDRef UID
+
+instance HasChunkRefs (TypedUIDRef t) where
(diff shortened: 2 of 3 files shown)


balacij:
We should be able to remove https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-printers/lib/Language/Drasil/Debug/Print.hs in favour of a more generic approach similar to the one we take in this PR.

code/drasil-lang/package.yaml
```diff
@@ -10,6 +10,9 @@ description: Please see the README on GitHub at <https://github.com/JacquesCaret
 language: Haskell2010
 default-extensions:
 - StrictData
+- StandaloneDeriving
+- DerivingVia
+- DeriveGeneric
```
balacij, line 15: Are you okay with adding this as default extensions?

`DerivingVia` (and `StrictData`, but I added that a while ago) is the only one not included by default in [GHC2024](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/control.html#extension-GHC2024).

code/drasil-database/lib/Drasil/Database/TypedUIDRef.hs
```diff
@@ -6,14 +6,20 @@ module Drasil.Database.TypedUIDRef (
...
+import qualified Data.Set as S (singleton)
 
 -- | 'UID' references that contain information about the type of data the 'UID'
 -- refers to, useful for type-safe dereferencing.
 newtype TypedUIDRef typ = TypedUIDRef UID
 
+instance HasChunkRefs (TypedUIDRef t) where
+  -- | A 'TypedUIDRef t' carries a 'UID' referring to a chunk of type 't'.
+  chunkRefs (TypedUIDRef u) = S.singleton u
+  {-# INLINABLE chunkRefs #-}
```
balacij, line 21: Note: a `TypedUIDRef` is a valid reference to another chunk. However, we don't use them at all yet.

code/drasil-database/lib/Drasil/Database/Chunk.hs
```diff
@@ -56,3 +55,91 @@ unChunk (Chunk c) = cast c
...
+class HasChunkRefs a where
+  chunkRefs :: a -> S.Set UID
+
+instance HasChunkRefs UID where
+  -- | 'UID's are meant to be "owned" (i.e., they are the unique identifier of
+  -- the chunk being defined), not *carried as references to other chunks*.
+  -- 'TypedUIDRef t' exists to be used as a *reference to another chunk of type
+  -- 't'*. Therefore, `UID` has no chunk references.
+  chunkRefs _ = S.empty
+  {-# INLINABLE chunkRefs #-}
```
balacij, line 72: A `UID` is meant to be "owned," not carried as a reference to another chunk. However, we currently carry them as references to other chunks. We should change those references from `UID`s to `TypedUIDRef`s.

balacij (commented):
Once we convert external usage of `UID` to `TypedUIDRef` (excluding the times where a `UID` should actually indicate the `UID` of a chunk carrying said `UID`), we can also use `deriveHasChunkRefs` on our 'chunk atoms.' Furthermore, when this is done, we can also generate `HasUID`. Future work!

For now, I'm content with this PR.

Ready for review @JacquesCarette @Xinlu-Y ! Thank you!

code/drasil-lang/package.yaml
JacquesCarette: Yes, I'm fine with those.

JacquesCarette (approved):

PR #4480, opened 2025-11-22
`exportSmithEtAlSrs`: Do not return an altered `System`.

balacij:
`exportSmithEtAlSrs` involved `mkDoc`, which involves pre-generating the SRS and then generating it again correctly, where the difference between said generation is getting more data to inject into the `ChunkDB`. `GlassBR`s code generator previously relied on this because of a missing `astm2009` `Reference`. This PR creates the `astm2009` `Reference` before the `ChunkDB` is created so that we can avoid propagating the updated `System`.


JacquesCarette:
It looks like this merge broke the build! @balacij can you investigate please?

Issue #4344, opened 2025-08-28
Pending redesign, remove `LabelledContent` and `Reference` from `ChunkDB`


balacij:
> The `LabelledContent` and `Reference` chunks require some careful attention.

_Originally posted by @balacij in https://github.com/JacquesCarette/Drasil/pull/4320#discussion_r2271639147_
            
            
We should be able to move LC and R to `System`, kicking the "rebuilding them" rock down the road a bit. Doing this _should_ be fairly straightforward. This will:

1. Remove the `drasil-lang` dependency that `drasil-database` has
2. Remove the code that gives special treatment to LC and R in `drasil-database`

PR #4481, opened 2025-11-22
Remove special treatment of `LabelledContent` in `ChunkDB`.

balacij:
Contributes to #2885, #2873 and #4344

Builds on #4487 

Up to `Reference`s, they have a unique `UID`. I think it is okay to add them back into the normal chunk pool, unlike `Reference`s, because they are more "chunk-like" than `Reference`s. The entire `Reference` system needs to have its system re-designed because it currently relies on having a duplicate chunk with all other chunk types.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Document.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Document.hs
@@ -87,7 +87,7 @@ checkToC (Notebook t a sc) = Notebook t a sc
 
 -- | Smart constructor for labelled content chunks.
 llcc :: Reference -> RawContent -> LabelledContent
-llcc = LblC
+llcc r = LblC (r ^. uid) r
 
 -- | Smart constructor for unlabelled content chunks (no 'Reference').
 ulcc :: RawContent -> UnlabelledContent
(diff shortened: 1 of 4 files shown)


code/drasil-lang/lib/Language/Drasil/Document.hs
```diff
@@ -87,7 +87,7 @@ checkToC (Notebook t a sc) = Notebook t a sc
 
 -- | Smart constructor for labelled content chunks.
 llcc :: Reference -> RawContent -> LabelledContent
-llcc = LblC
+llcc r = LblC (r ^. uid) r
```
balacij, line 91: Note that here we are cloning the `UID` the inner-held `Reference` for its own `UID`. We can create new constructors for each LC type later, but I left this as-is for. In a subsequent PR, we can re-write the constructors for `L-C-`s so as to not "copy" the `UID` from the `Reference`, but to _copy its own `UID` to the `Reference`._

(I chose to not do that in this PR because my focus was on making #4489 happen, not cleaning up code related to `L-C-`s right now.)

JacquesCarette (approved):

PR #4488, opened 2025-11-23
Move `Reference` and "reference"-adjacent maps to `System` from `ChunkDB`

### LINKED ISSUE #4344: Pending redesign, remove `LabelledContent` and `Reference` from `ChunkDB`

balacij:
Contributes to #2885 and #2873

Builds on #4487 and #4478

Closes #4344 

In conjunction with #4481, this will allow us to make `drasil-database` a standalone package (I will do this in a follow-up PR).

This PR moves the `Reference` map and the `Doc`-based "referenced by" and "trace" maps from `ChunkDB` to `System`:
* `drasil-database`: Removed `refTable`, `traceTable`, and `refbyTable` from `ChunkDB`.
* `drasil-system`: Moved the above maps to `System`.
* `drasil-printers`: Refactored `PrintingInformation` to wrap `System` instead of `ChunkDB`. Printers now search the `System` for references (and the `ChunkDB` within the `System` for everything else).

I clean up a bit of code along the way as well, but a lot of them were necessary changes as a result of the above refactors.

Why else is this PR a good idea? Because all 3 of the moved tables are related to an SRS being rendered. They are not specifically relevant to code generation. If we had a `State`-like monad we used for our `drasil-printers`-based printers, these pieces of information would likely be held there. Our current next-best-thing is `PrintInformation` which acts somewhat like that, carrying a cache of known `Chunks` and other 'printing information'. Now, I wanted to move said 3 maps to `PrintingInformation` directly, but we can't because we need those 3 tables for the code generator as well (which does not use `PrintingInformation`). So, our next best thing is to move it into `drasil-system`. Why does the code generator use `Reference`s (primarily for display reasons)? Because we have a few references to citations in some quantity descriptions, but with the way the current `Reference` system works, both a `Citation` and a `Reference` is necessary.

This PR is a bit of a "side-step" in progress, moving one hack around. I think it is arguably a bit better because it tightened up a bit of code, and moves a hacky bit of code to an area of code that we already know is riddled with hacks and needs to be re-designed anyways. This PR also enables #4489 (it's nice to close out one small project -- that is, making `drasil-database` a standalone package).


JacquesCarette:
Can you deal with the conflicts @balacij ?

JacquesCarette:
And because I merged something else just before this, it needs another merge!

balacij:
Ok, re-merged again.

PR #4497, opened 2025-12-01
`drasil-lang`: Switch `RawContent`'s `CodeBlock` constructor from using `CodeExpr` to `Expr`.

balacij:
Contributes to #2885 

Succeeds #4428 

Why?

1. `CodeBlock` was only used in Projectile's lesson plan to display 3 variable assignments.
2. The code related to `CodeBlock` rendering for other document types was incomplete.
3. It ~~creates~~ contributes to a complex dependancy chain between `drasil-code`, `drasil-docLang`, and `drasil-printers` that forces code from those packages to live in `drasil-lang`. At the very least, it contributes to keeping `CodeExpr` in `drasil-lang`.

With this PR, we are a bit closer to being able to move `drasil-code`-related things in `drasil-lang` to `drasil-code`.

```
rg "import Drasil.Code" -ths drasil-lang/lib/Language
drasil-lang/lib/Language/Drasil.hs
300:import Drasil.Code.Classes (Callable, IsArgumentName)
301:import Drasil.Code.CodeVar (CodeIdea(..), CodeChunk(..),
304:import Drasil.Code.CodeExpr.Lang (CodeExpr)
305:import Drasil.Code.CodeExpr.Class (CodeExprC(..))
```

However, this PR alone is not enough because `drasil-code` relies on `drasil-printers` but `drasil-printers` relies on `CodeExpr`, which means if we were to move this code (the above listed `Drasil.Code.*` files) from `drasil-lang` to `drasil-code`, we could create a dependency cycle!

Subsequent PRs necessary to make that happen will involve splitting `Document`-related things from `drasil-printers` and `drasil-lang`.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Document/Contents.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Document/Contents.hs
@@ -10,24 +10,19 @@ module Language.Drasil.Document.Contents (
   unlbldCode
 ) where
 
-import Drasil.Code.CodeExpr.Lang (CodeExpr)
+import Control.Lens ((^.))
+
 import Language.Drasil.Classes (Definition(..))
 import Language.Drasil.ShortName (HasShortName(..), getSentSN)
 import Language.Drasil.Document (llcc, ulcc)
-import Language.Drasil.Document.Core
-    (LabelledContent,
-     RawContent(Enumeration, EqnBlock, CodeBlock),
-     Contents(UlC),
-     ListTuple,
-     ItemType(Flat),
-     ListType(Simple))
+import Language.Drasil.Document.Combinators (bulletFlat, mkEnumAbbrevList)
+import Language.Drasil.Document.Core (LabelledContent, RawContent(Enumeration,
+  EqnBlock, CodeBlock), Contents(UlC), ListTuple, ItemType(Flat), ListType(Simple))
+import Language.Drasil.Expr.Lang (Expr)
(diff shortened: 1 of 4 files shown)


code/drasil-lang/lib/Language/Drasil/Document/Contents.hs
```diff
@@ -38,7 +33,7 @@ unlbldExpr :: ModelExpr -> Contents
 unlbldExpr c = UlC $ ulcc $ EqnBlock c
 
 -- | Unlabelled code expression
-unlbldCode :: CodeExpr -> Contents
+unlbldCode :: Expr -> Contents
```
JacquesCarette, line 36: It's definitely a side effect of our lack of tests that you can get away with this! If this were not a partial change towards an improved state, I'd say 'no', But I understand that this is temporary.

JacquesCarette (approved):

code/drasil-lang/lib/Language/Drasil/Document/Contents.hs
balacij: Both were the wrong choice, really. It's just that we can't import `drasil-gool` in `drasil-lang` without creating a cyclic dependency. I think this code wants a constrained polymorphic input type that is anything 'renderable'.

JacquesCarette: Ah yes, that would be a better choice indeed. That kind of polymorphism would probably be good in quite a few of our structures!

PR #4502, opened 2025-12-02
Handful of `drasil-code` touch-ups

balacij:
A series of small changes aimed at re-organizing code in `drasil-code`. No structure/design changes!

~~Leaving this PR as a draft because it is a WIP!~~

### DIFF
--- a/code/drasil-code/stack.yaml
+++ b/code/drasil-code/stack.yaml
@@ -31,7 +31,6 @@ resolver: lts-22.44
 packages:
 - .
 - ../drasil-build
-- ../drasil-codeLang
 - ../drasil-data
 - ../drasil-database
 - ../drasil-gool


code/drasil-code/lib/Language/Drasil/Code/Code.hs
```diff
@@ -1,19 +1,44 @@
...
+createCodeFiles (Code cs) = mapM_ createCodeFile cs
+
+-- | Helper that uses pairs of 'Code' to create a file written with the given
+-- document at the given 'FilePath'.
+createCodeFile :: (FilePath, Doc) -> IO ()
+createCodeFile (path, code) = do
+  createDirIfMissing True (takeDirectory path)
+  h <- openFile path WriteMode
+  hPutStrLn h (render code)
+  hClose h
```
balacij, line 37: With #4610, this will be moved again (out of scope, however).

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -167,19 +168,38 @@ matchConcepts = Map.fromList . map (\(cnc,cdc) -> (cnc ^. uid, cdc))
...
+spaceToCodeType S.String         = [C.String]
+spaceToCodeType (S.Vect s)       = map C.List (spaceToCodeType s)
+spaceToCodeType (S.Matrix _ _ s) = map (C.List . C.List) (spaceToCodeType s)
+spaceToCodeType (S.Set s)        = map C.List (spaceToCodeType s)
+spaceToCodeType (S.Array s)      = map C.Array (spaceToCodeType s)
+spaceToCodeType (S.Actor s)      = [C.Object s]
+spaceToCodeType S.Void           = [C.Void]
+spaceToCodeType (S.Function i t) = [C.Func is ts | is <- ins, ts <- trgs]
+    where trgs = spaceToCodeType t
+          ins  = map spaceToCodeType (toList i)
```
balacij, line 225: I think the fact that this returns a list of preferentially-ordered `CodeType`s indicates that `spaceToCodeType` needs to be tied to the language we want to generate artifacts in. This is a shorthand for rolling up those choices together.

code/drasil-code/stack.yaml
```diff
@@ -31,6 +31,7 @@ resolver: lts-22.44
 packages:
 - .
 - ../drasil-build
+- ../drasil-codeLang
```
balacij: This actually needs to be removed now!

code/drasil-code/lib/Language/Drasil/ICOSolutionSearch.hs
```diff
@@ -17,22 +19,42 @@ type Known = CodeVarChunk
...
+    requiredInputs = inputs \\ [quantvar def]
+
+prettyError :: [Def] -> [Known] -> [Need] -> a
+prettyError defs knowns needs = error $
+  "The following outputs cannot be computed: " ++ lm needs ++ "\n" ++
+  "Unused definitions are: " ++ lm defs ++ "\n" ++
+  "Known values are: " ++ lm knowns
+  where
+    lm :: HasUID c => [c] -> String
+    lm = intercalate ", " . map showUID
```
balacij, line 60: While longer, I hope you agree that this is more readable!

code/drasil-code/stack.yaml
balacij: Fixed in 9aec3d0cc9af5d1500aeef12b7a88d86bac55c6f

code/drasil-code/lib/Language/Drasil/Choices.hs
JacquesCarette: Why do you think this is language-dependent?

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -164,22 +171,59 @@ instance RenderChoices CodeConcept where
...
 matchSpace s ts sm = \sp -> if sp == s then ts else sm sp
 
 -- | Builds a 'SpaceMatch' from an association list of 'Spaces' and 'CodeTypes'.
-matchSpaces :: [(Space, [CodeType])] -> SpaceMatch
+matchSpaces :: [(Space, [C.CodeType])] -> SpaceMatch
 matchSpaces spMtchs = matchSpaces' spMtchs spaceToCodeType
   where matchSpaces' ((s,ct):sms) sm = matchSpaces' sms $ matchSpace s ct sm
         matchSpaces' [] sm = sm
 
+-- | Default mapping between 'Space' and 'CodeType'.
```
JacquesCarette, line 208: Why is this a better home than where it was before? I was looking over this whole `Choices` file with Brandon earlier today, and I was thinking that it's a horrible mash-up of stuff and should likely be broken up into a bunch of smaller modules -- quite the opposite of adding even more to it.

code/drasil-code/lib/Language/Drasil/ICOSolutionSearch.hs
```diff
@@ -17,22 +19,42 @@ type Known = CodeVarChunk
...
-                       ++ intercalate ", " (map showUID defs') ++ "\n"
-                     ++ "Known values are: "
-                       ++ intercalate ", " (map showUID k))
-              else getExecOrder' (ord ++ new) (defs' \\ new) kNew nNew
+-- | Topologically sort a list of 'Def's to form a path from 'Known' values to
+-- values that 'Need' to be calculated.
+solveExecOrder :: [Def] -> [Known] -> [Need] -> ChunkDB -> [Def]
+solveExecOrder allDefs knowns needs = solve [] allDefs knowns (needs \\ knowns)
+
+solve :: [Def] -> [Def] -> [Known] -> [Need] -> ChunkDB -> [Def]
```
JacquesCarette, line 27: why call it `solve` and not `topologicalSort` ?

code/drasil-code/lib/Language/Drasil/ICOSolutionSearch.hs
JacquesCarette: Yes!

JacquesCarette (changes requested):
Way, way too many changes in a single PR! I think some really ought to have had a proper discussion, but it is really hard to do when they are all mixed together. Please, don't do this in the future.

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: Context:

https://github.com/JacquesCarette/Drasil/blob/413ad75eb030c45ed1e0d999431666813e8f0e53/code/drasil-code/lib/Language/Drasil/Choices.hs#L168-L171

In the pipeline, we have 3 levels to think about here:

1. The program specification.
2. The abstract program.
3. The concrete program.

These above options map data from the program specification level to a mixture of the other two. For example, with the following:

```haskell
spaceToCodeType S.Real           = [C.Double, C.Float]
```

We are saying "if the language you are trying to generate does not support `Double`, then just use `Float`." We're not making a specific choice here, we're just throwing a set of options at Drasil and seeing what it knows will work (if Drasil can find a `Space -> CodeType` map combination that works for a given language).

I think this really wants to be something closer to a mapping of `Space` to _sensible_ `CodeType`s at stage (2). This would be something we use to "check" someones interpretation of an SRS. And then we would need another mapping from `CodeType` to the final type language of the individual languages we want to generate. Two deliberate choices.

Note that there will be significant duplication between the (2) and (3)'s type languages, but I don't think this is necessarily a bad thing. For example, because (3) will contain custom type solutions for a given language (e.g., `numpy.float32`, `std::ratio`) that are not contained in (2). We shouldn't be adding `numpy.float32` to the general `CodeType` type if no other language will support it.

Regarding the greater design, there is something to be said about `CodeType` being an ADT. Users should be able to declare new types. For example, if I had a new library I wanted to encode an interface to for Drasil, adding to a single ADT is not practical.

code/drasil-code/lib/Language/Drasil/ICOSolutionSearch.hs
balacij: It was originally `solveExecOrder0`, but I renamed it to `solve` because it was just an internal function with the extra parameter for carrying the path and the original name was too long (and I originally referenced `solve` a few more times).

I suppose `topologicalSort` is more appropriate, however! I'll move this work to a new PR.

PR #4503, opened 2025-12-03
Remove `drasil-code`-related re-exports from `drasil-lang`.

balacij:
For the remaining `drasil-code`-related code that lives in `drasil-lang`, this stops `-lang`'s re-exporting their contents in `Language.Drasil`.

This is a follow-up to #4497.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Chunk/Code.hs
+++ b/code/drasil-code/lib/Language/Drasil/Chunk/Code.hs
@@ -3,7 +3,7 @@
 module Language.Drasil.Chunk.Code (
   CodeIdea(..), CodeChunk(..), CodeVarChunk(..), CodeFuncChunk(..),
   VarOrFunc(..), obv, quantvar, quantfunc, ccObjVar, codevars, codevars',
-  funcResolve, varResolve, listToArray, programName, funcPrefix,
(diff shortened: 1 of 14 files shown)


code/drasil-docLang/package.yaml
```diff
@@ -20,14 +20,15 @@ dependencies:
 - pretty
 - transformers
 - multiplate
-- drasil-lang
+- drasil-code
```
balacij, line 23: This is needed exclusively because of `programName`!!!!

```haskell
-- | Convert an abbreviation into one deemed 'code-friendly', removing spaces,
--   and replacing special characters with underscores.
--
--   FIXME: This should NOT be treated as a 'getter', but something we cache
--   local to something that has a 'program name'.
programName :: CommonIdea c => c -> String
programName = toPlainName . filter (not . isSpace) . abrv
```

balacij: What it is really used for, however, is for the "abbreviation" of a system! In a subsequent PR, this dependency should be removed.

balacij (commented):
As a reminder, the goal of this kind of work is to move `drasil-code`-related code from `drasill-lang` to where it should belong.

PR #4504, opened 2025-12-04
`System`: Replace polymorphic variable with `CI` for "system name" and add "program name" `String` field based on system name.

balacij:
This work is intended supplement #4503 so that `drasil-docLang` does not have a direct dependency on `drasil-code`, but it also contributes to generally cleaning up code.

### DIFF
--- a/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
+++ b/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
@@ -11,10 +11,13 @@ module Data.Drasil.ExternalLibraries.ODELibraries (
   odeInfoChunks
 ) where
 
-import Language.Drasil (HasSymbol(symbol), HasUID(uid), MayHaveUnit(getUnit),
+import Control.Lens ((^.), _1, _2, over)
+
+import Drasil.Database (HasUID(..), (+++))
+import Language.Drasil (HasSymbol(symbol), MayHaveUnit(getUnit),
   HasSpace(typ), Space (Actor, Natural, Real, Void, Boolean, String, Array, Vect), implVar, implVar',
   compoundPhrase, nounPhrase, nounPhraseSP, label, sub, Idea(getA), NamedIdea(term), Stage(..),
-  (+++), Definition (defn), (+:+), Sentence (S), DefinedQuantityDict, dqdWr, implVarAU')
+  Definition (defn), (+:+), Sentence (S), DefinedQuantityDict, dqdWr, implVarAU')
 import Language.Drasil.Display (Symbol(Label, Concat))
 
 import Language.Drasil.Code (Lang(..), ExternalLibrary, Step, Argument,
@@ -42,8 +45,6 @@ import Language.Drasil.Code (Lang(..), ExternalLibrary, Step, Argument,
 
 import Drasil.Code.CodeExpr
 import Drasil.Code.CodeExpr.Development
-
-import Control.Lens ((^.), _1, _2, over)
 
 -- SciPy Library (Python)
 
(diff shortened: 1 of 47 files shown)


code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -91,23 +93,23 @@ fullExList codePth srsDoxPth = Enumeration $ Bullet $ map (, Nothing) (allExampl
...
 -- | Display the points for generated documents and call 'versionList' to display the code.
 individualExList :: Example -> [ItemType]
 -- No choices mean no generated code, so we do not need to display generated code and thus do not call versionList.
-individualExList ex@E{systemE = SI{_sys = sys}, choicesE = [], codePath = srsP} =
+individualExList ex@E{choicesE = [], codePath = srsP} =
   [Flat $ namedRef (buildDrasilExSrcRef ex) (S "Drasil Source Code"),
-  Flat $ S "SRS:" +:+ namedRef (getSRSRef srsP HTML $ programName sys) (S "[HTML]")
-  +:+ namedRef (getSRSRef srsP TeX $ programName sys) (S "[PDF]")
-  +:+ namedRef (getSRSRef srsP MDBook $ programName sys) (S "[mdBook]")]
+  Flat $ S "SRS:" +:+ namedRef (getSRSRef srsP HTML $ getAbrv ex) (S "[HTML]")
```
JacquesCarette: Isn't this a step backwards instead of sideways? We don't want program names to be abbreviations!

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -133,9 +135,9 @@ versionList getRef ex@E{systemE = SI{_sys = sys}, choicesE = chcs} =
     -- Determine the version name based on the system name and if there is more than one set of choices.
     verName chc = case chcs of
       -- If there is one set of choices, then the program does not have multiple versions.
-      [_] -> programName sys
+      [_] -> getAbrv ex
```
JacquesCarette: Here and below as well, this seems like a step backwards.

code/drasil-website/lib/Drasil/Website/Example.hs
```diff
@@ -187,44 +189,44 @@ getCodeRef :: Example -> Lang -> String -> Reference
 -- since that was checked in an earlier function.
 --
 -- Pattern matches so that examples that only have a single set of choices will be referenced one way.
-getCodeRef ex@E{systemE=SI{_sys = sys}, choicesE = chcs} l verName =
+getCodeRef ex@E{choicesE = chcs} l verName =
   makeURI refUID refURI refShortNm
   where
     -- Append system name and program language to ensure a unique id for each.
-    refUID = "codeRef" ++ sysName ++ programLang
+    refUID = "codeRef" ++ sysName' ++ programLang
```
JacquesCarette: was it needed here (and below) to change the name? Also: step backwards?

code/drasil-website/lib/Drasil/Website/Example.hs
balacij: Unfortunately, that's how they currently are captured. We should fix it. That being said, this `getAbrv` is a local function, not the same kind of "abbreviation getter" we have in `drasil-lang`. This work de-couples a bit, but there's much code that still assumes that the `CI` we use in `System` contains the codename/program name. I just changed the name in ce326d7149da079cda7485f489d3c5e1f2928114 .

code/drasil-website/lib/Drasil/Website/Example.hs
balacij: Changed again in 5adced5cc1ee998a0c2b5f75dfbaa23a6d958450 . Yes, it was necessary because `sysName` is now a lens that `Drasil.System` exposes (and we use in this file).

JacquesCarette (approved):

PR #4509, opened 2025-12-06
Need help! Removing `codeExprDoc` from `drasil-printers`?

balacij:
I will annotate this PR for discussion, but this PR and branch are expected to be deleted in the end.

The goal of my work is to move the `Drasil.Code.*` modules from `drasil-lang` into `drasil-code`. However, that work is currently impossible because it would create a cyclic dependency between `drasil-printers` and `drasil-code`:

* `drasil-printers` relies on `CodeExpr` and the other modules in `Drasil.Code` (namely code variables) for describing how `CodeExpr` can be rendered into `Doc`s for human-readable, non-code purposes (I suppose ASCII variants).
* `drasil-code` relies on `drasil-printers` for said rendering function and more (e.g., rendering functions for `ModelExpr`, `Sentence`, `Symbol`, etc.).


code/drasil-printers/lib/Language/Drasil/Plain/Print.hs
```diff
@@ -30,10 +29,6 @@ data SingleLine = OneLine | MultiLine
 exprDoc :: PrintingInformation -> SingleLine -> L.Expr -> Doc
 exprDoc pinfo sl e = pExprDoc sl (expr e pinfo)
 
--- | Create code expressions for a document in 'Doc' format.
-codeExprDoc :: PrintingInformation -> SingleLine -> C.CodeExpr -> Doc
-codeExprDoc pinfo sl e = pExprDoc sl (codeExpr e pinfo)
```
balacij, line 35: This function is used by `drasil-code` for a single purpose: rendering `CodeExpr`-based quantity constraints into human-readable text for outputting input value constraint violation messages (see `stable/` changes).

code/drasil-printers/lib/Language/Drasil/Printing/Import/CodeExpr.hs
```diff
@@ -1,223 +0,0 @@
-{-# LANGUAGE GADTs #-}
-
--- | Defines functions to render 'CodeExpr's as printable 'P.Expr's.
-module Language.Drasil.Printing.Import.CodeExpr (codeExpr) where
```
balacij, line 4: The reason why I tried to delete this file at all is because, well, it is odd. `CodeExpr` is intended for rendering into GOOL (solely, to my knowledge), not necessarily rendering comments. Considering it is only used for a single purpose (the input constraint violation error messages), I'm not sure if the approach taken to outputting messages is a good one.

code/stable/glassbr/src/cpp/InputParameters.cpp
```diff
@@ -205,10 +205,8 @@ void InputParameters::input_constraints() {
         std::cout << ", but is expected to be ";
         std::cout << "between ";
         std::cout << 0.1;
-        std::cout << " (d_min)";
```
balacij, line 208: While here we output a variable name in "implementation" `Stage`, in other areas, we have the flexibility of printing any general `CodeExpr`.

code/stable/projectile/projectile_m_l_nol_u_u_v_f/src/julia/InputParameters.jl
```diff
@@ -49,7 +49,6 @@ function input_constraints(v_launch::Float32, theta::Float32, p_target::Float32)
         print(Float32(0.0))
         print(" and ")
         print(pi / Float32(2.0))
-        print(" ((pi)/(2))")
```
balacij, line 52: For example, $\frac{\pi}{2}$.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -419,15 +419,15 @@ printConstraint v c = do
...
+          [printStr " and "] ++ printExpr ub db ++ [printStrLn "."]
       printConstraint' _ (Range _ (UpTo (_, e))) = do
         ub <- convExpr e
-        return $ [printStr "below ", print ub] ++ printExpr e db ++
+        return $ [printStr "below "] ++ printExpr ub db ++
           [printStrLn "."]
       printConstraint' _ (Range _ (UpFrom (_, e))) = do
         lb <- convExpr e
-        return $ [printStr "above ", print lb] ++ printExpr e db ++ [printStrLn "."]
+        return $ [printStr "above "] ++ printExpr lb db ++ [printStrLn "."]
```
balacij, line 430: One question to ask ourselves first is: What do we want the format of the constraint violation messages to be?

I somewhat like these, but I would also like to see `CON$X$` (where `CON$X$` is a label we defined in the SRS for a constraint) added as a prefix.

code/drasil-printers/lib/Language/Drasil/Printing/Import/CodeExpr.hs
balacij: I could be wrong here as well!

But I neglected to also mention the _initial_ reason I was looking at this code at all: `drasil-printers` and `drasil-code` have a cyclic dependency that we are currently avoiding through pushing some code from `drasil-code` into `drasil-lang`. This includes the "code variable chunks" and `CodeExpr`. My goal is to be able to move the `drasil-code`-related code from `drasil-lang` into `drasil-code`.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
JacquesCarette: I agree that more traceability in constraint violation messages would be even better.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -63,7 +63,7 @@ import Drasil.GOOL (MSBody, MSBlock, SVariable, SValue, MSStatement,
   VisibilitySym(..), MethodSym(..), StateVarSym(..), pubDVar, convType,
     convTypeOO, VisibilityTag(..))
 
-import qualified Drasil.GOOL as OO (SFile)
+import qualified Drasil.GOOL as OO (SFile, SValue)
```
JacquesCarette, line 66: These imports (of `SFile` and `SValue), as well as `ProcProg` all seem to point to something being odd here.

code/drasil-printers/lib/Language/Drasil/Plain/Print.hs
JacquesCarette: This seems like reasonable functionality - where does it belong? And maybe it should be in two stages (i.e. `codeExprDoc` would get fed the results of `codeExpr e pinfo` so that `codeExpr` would not be needed here).

code/drasil-printers/lib/Language/Drasil/Printing/Import/CodeExpr.hs
JacquesCarette: I certainly agree with that goal.

code/stable/glassbr/src/cpp/InputParameters.cpp
JacquesCarette: Indeed. But removing this output points to a real problem, i.e. these changes should not do that!

balacij:
None of this code was ever meant to be merged. We can close it without worry. I'm working on `drasil-code` elsewhere now (no new issues to file/refer to yet, but also nothing to salvage here that isn't already in #4989 or #4974).

PR #4514, opened 2025-12-09
Haschunkrefs sentence

Xinlu-Y:
Contributes to #4476 

- Move the `sdep/shortdep/lnames` helpers into `Language.Drasil.Sentence`, derive a real `HasChunkRefs` instance, and export `sentenceRefs` so any chunk field of type `Sentence` reports its `UIDs`.
- Leave `Language.Drasil.Sentence.Extract` as a re-export shim so existing imports still build.

This prepares the automation work in #4434 by turning `Sentence` into a proper “chunk atom” for `declareHasChunkRefs`.


JacquesCarette (dismissed):
This makes sense to me, but I also want @balacij 's review, to make sure this is going in the same direction that he's working on.

code/drasil-lang/lib/Language/Drasil/Sentence.hs
```diff
@@ -149,3 +154,81 @@ capSent x          = x
...
+lnames' = concatMap lnames
+{-# INLINE lnames' #-}
+
+sentenceRefs :: Sentence -> Set.Set UID
+sentenceRefs sent = Set.fromList (lnames sent ++ sdep sent ++ shortdep sent)
+{-# INLINE sentenceRefs #-}
+
+instance HasChunkRefs Sentence where
+  chunkRefs = sentenceRefs
+  {-# INLINABLE chunkRefs #-}
```
balacij, line 248: This is great! Now that we have this, do we still need to expose`sdep`, `shortdep`, etc.?

code/drasil-lang/lib/Language/Drasil/Sentence.hs
```diff
@@ -149,3 +154,81 @@ capSent x          = x
...
+lnames' :: [Sentence] -> [UID]
+lnames' = concatMap lnames
+{-# INLINE lnames' #-}
+
+sentenceRefs :: Sentence -> Set.Set UID
+sentenceRefs sent = Set.fromList (lnames sent ++ sdep sent ++ shortdep sent)
+{-# INLINE sentenceRefs #-}
+
+instance HasChunkRefs Sentence where
+  chunkRefs = sentenceRefs
```
balacij: Can't we use `getUIDs` instead of `sentenceRefs`? This way we only need to traverse a `Sentence` once to get the same information?

balacij: If not, seeing as `sentenceRefs` isn't exposed by the module, can we just inline it into the definition of `chunkRefs`?

code/drasil-lang/lib/Language/Drasil/Sentence.hs
balacij: We're also now using a `Set` for `chunkRefs`. Should we change those functions to use sets, too? (This would need to be a separate PR if so)

balacij (changes requested):

JacquesCarette:
When are you going to be able to get back to this @Xinlu-Y ?

Xinlu-Y:
> When are you going to be able to get back to this @Xinlu-Y ?

I’m working on it today

code/drasil-lang/lib/Language/Drasil/Sentence.hs
Xinlu-Y: > This is great! Now that we have this, do we still need to expose`sdep`, `shortdep`, etc.?

Probably keep them for now. They’re still used outside this module (TraceTable, GetChunks, and re-exported via Development/Sentence.Extract).

code/drasil-lang/lib/Language/Drasil/Sentence.hs
Xinlu-Y: > Can't we use `getUIDs` instead of `sentenceRefs`? This way we only need to traverse a `Sentence` once to get the same information?

`getUIDs` only collects symbolic UIDs (and from expressions). it ignores `Ref` and `Ch ShortStyle`. `sentenceRefs` was combining symbol + short + ref. So `getUIDs` isn’t equivalent. I added `getAllUIDs` that collects all three in one pass

code/drasil-lang/lib/Language/Drasil/Sentence.hs
Xinlu-Y: > We're also now using a `Set` for `chunkRefs`. Should we change those functions to use sets, too? (This would need to be a separate PR if so)

`sdep/lnames'` currently return `[UID]` and are used as lists in `GetChunks.hs` and `TraceTable.hs`.
https://github.com/JacquesCarette/Drasil/blob/c3655dd09dc1b3a60259a9eea2a153055ba82f26/code/drasil-docLang/lib/Drasil/GetChunks.hs#L19-L21 https://github.com/JacquesCarette/Drasil/blob/c3655dd09dc1b3a60259a9eea2a153055ba82f26/code/drasil-docLang/lib/Drasil/GetChunks.hs#L39-L41 https://github.com/JacquesCarette/Drasil/blob/c3655dd09dc1b3a60259a9eea2a153055ba82f26/code/drasil-docLang/lib/Drasil/TraceTable.hs#L37-L38
If `sdep/lnames'` change to `Set UID`, those call sites would need to convert back to a list (e.g., `Set.toAscList`), right? Does the order matter for outputs?

code/drasil-lang/lib/Language/Drasil/Sentence.hs
```diff
@@ -54,9 +59,9 @@ infixr 5 :+:
 data Sentence where
   -- | Ch looks up the term for a given 'UID' and displays the term with a given 'SentenceStyle' and 'CapitalizationRule'.
   -- This allows Sentences to hold plural forms of 'NamedIdea's.
-  Ch    :: SentenceStyle -> TermCapitalization -> UID -> Sentence
+  Ch    :: SentenceStyle -> TermCapitalization -> UIDRef typ -> Sentence
   -- | A branch of Ch dedicated to SymbolStyle only.
-  SyCh  :: UID -> Sentence
+  SyCh  :: UIDRef typ -> Sentence
```
balacij, line 64: This particular one was done in #4810.

It would be good to add constraints on what we expect from `typ` for the other things, but those should probably be split off into other PRs.

code/drasil-lang/lib/Language/Drasil/Sentence.hs
balacij: I think this code was all overwritten in another one of your PRs? This PR might be outdated now.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
```diff
@@ -26,6 +26,14 @@ instance HasChunkRefs (UIDRef t) where
 hide :: IsChunk t => t -> UIDRef t
 hide = UIDRef . (^. uid)
 
+-- | Create a 'UIDRef' from a raw 'UID'.
+uidRef :: UID -> UIDRef t
+uidRef = UIDRef
+
+-- | Extract the 'UID' from a 'UIDRef'.
+unRef :: UIDRef t -> UID
+unRef (UIDRef u) = u
```
balacij, line 35: This was pulled into #4808 as `raw` and `rawUni`.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
```diff
@@ -26,6 +26,14 @@ instance HasChunkRefs (UIDRef t) where
 hide :: IsChunk t => t -> UIDRef t
 hide = UIDRef . (^. uid)
 
+-- | Create a 'UIDRef' from a raw 'UID'.
+uidRef :: UID -> UIDRef t
+uidRef = UIDRef
```
balacij, line 31: I think we should try to avoid 'upgrading' `UID` references. I think `hide` is a bit better because it means that the specific chunk (and correct type) was in scope at some point.

code/drasil-printers/lib/Language/Drasil/Printing/Import/Sentence.hs
```diff
@@ -29,16 +30,16 @@ spec _  (Sy s)                  = P.E $ pUnit s
 spec sm (NP np)                 = spec sm (toSent $ phraseNP np)
 spec _  Percent                 = P.E $ P.MO P.Perc
 spec _  (P s)                   = P.E $ symbol s
-spec sm (SyCh s)                = P.E $ symbol $ lookupC' sm s
+spec sm (SyCh s)                = P.E $ symbol $ lookupC' sm (unRef s)
```
balacij, line 33: Similarly, at 'get chunk' sites, we should be try to be using 'unhide' as much as possible, which knows about the correct type to be looking for from the `ChunkDB`. That being said, for this in particular, this is difficult. I had to resort to a fairly major hack in #4810.

JacquesCarette:
What do we need to do to get this PR back into shape to be merged? (Other than a merge, of course)

Xinlu-Y:
> What do we need to do to get this PR back into shape to be merged? (Other than a merge, of course)

I believe all the work in this PR has been merged by subsequent PRs, so we can go ahead and close it.

PR #4515, opened 2025-12-09
Haschunkrefs np

Xinlu-Y:
Contributes to #4476 and #4434.

- Moved NP data types into `NounPhrase.Types` to remove the old NP ↔ Sentence circular dependency.

- Added a real `HasChunkRefs` instance for NP using the same logic from `fill-chunk-refs`. NP now reports any referenced UIDs.

- Exported `sentenceRefs` so NP (and others later) can reuse the same reference-extraction code.

This sets up the next steps where `Reference`, `NamedIdea`, etc. will switch to automatic `HasChunkRefs`.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/NounPhrase/Core.hs
+++ b/code/drasil-lang/lib/Language/Drasil/NounPhrase/Core.hs
@@ -4,46 +4,12 @@ module Language.Drasil.NounPhrase.Core (
   module Language.Drasil.NounPhrase.Types
 ) where
 
-import Drasil.Database (HasChunkRefs(..), UID)
+import Drasil.Database (HasChunkRefs(..))
 
 import Language.Drasil.NounPhrase.Types
-import qualified Language.Drasil.Sentence as Sent
-import Language.Drasil.Sentence (Sentence, sentenceRefs)
-
-import qualified Data.Set as Set
 
 -- | Gather the chunk references mentioned within an 'NP'.
 instance HasChunkRefs NP where
-    chunkRefs = npRefs
+    -- NPStruct only contains Strings and Symbols, so it cannot embed UID refs.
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
Again, makes sense to me, but I want @balacij 's review as well.

code/drasil-lang/lib/Language/Drasil/NounPhrase/Core.hs
```diff
@@ -1,49 +1,49 @@
...
+npStructRefs :: NPStruct -> Set.Set UID
+npStructRefs = sentenceRefs . npStructToSentence
+{-# INLINABLE npStructRefs #-}
+
+-- | Translate 'NPStruct' into a 'Sentence' so that sentence helpers can traverse it.
+npStructToSentence :: NPStruct -> Sentence
+npStructToSentence (S s)       = Sent.S s
+npStructToSentence (a :-: b)   = npStructToSentence a Sent.:+: npStructToSentence b
+npStructToSentence (a :+: b)   = npStructToSentence a Sent.+:+ npStructToSentence b
+npStructToSentence (P sym)     = Sent.P sym
```
balacij: What inside of an `NPStruct` can contain a `UID` reference? It looks like it would always be empty, no?

JacquesCarette: Indeed, `NPStruct` can never contain UIDs.

balacij: In that case, I think we can short-circuit the definition of `HasChunkRefs NP` to `mempty` with an explanation of why `mempty` is appropriate.

JacquesCarette:
Not quite sure where this stands? Could I get an update @balacij @Xinlu-Y ?

PR #4534, opened 2025-12-11
Updated links for Grounded Theory wiki page

### LINKED ISSUE #4528: Wiki: Grounded Theory
I noticed a couple of broken links in the [Grounded Theory](https://github.com/JacquesCarette/Drasil/wiki/Grounded-Theory) page:
- The [Resources page](http://www.groundedtheoryonline.com/bibliography-and-references/) link give Page not found. I think the correct link is [Resources](https://www.groundedtheoryonline.com/resources/).
- The [Grounded Theory for Geeks](https://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=C3AE411B048E7229DE69A3ECD243EA92?doi=10.1.1.589.4145&rep=rep1&type=pdf) gives 502. Is [this](https://dl.acm.org/doi/10.1145/2578903.2579162) the same paper?

Out of curiosity, why do we have this page? It doesn't seem very Drasil-like. Is it helpful for the process of developing Drasil itself?

Contributes to #4401

B-rando1:
The links haven't been confirmed to be the correct replacements, so we need to double-check before merging.

Closes #4528


B-rando1:
I added those changes, do they look good?

Issue #4541, opened 2025-12-12
Move `UID`-based chunk references to `TypedUIDRef`-based chunk references


balacij:
Consider the following:

```haskell
data ThisChunk = TC {
    _tcUid :: UID
  , _tcOtherThings :: [UID]
  , ...
}

data SomeADT = Constructor1 UID
             | Constructor2 UID
             | ...
```

This is a fair representation of how `UID`s might appear in in our chunk types (`ThisChunk`) and our chunk 'atoms' (`SomeADT`). `UID`s are used to create references to other chunks and to indicate the `UID` of a chunk itself.

With #4476, we diverge: we solidified that `UID`s are meant to be carried solely within chunk types (e.g., `ThisChunk`), indicating that the chunk's `UID` should be that `UID`. In other words, it "owns"/is uniquely identified by that `UID`. However, as in the above, we don't use `UID`s like this at the moment.

With #4320, we introduced `TypedUIDRef` to build type-safe `UID` references.

https://github.com/JacquesCarette/Drasil/blob/dd015080662abaa73079a675e1e41e045a178e31/code/drasil-database/lib/Drasil/Database/TypedUIDRef.hs#L14-L16

It happens that this is more useful as well to indicate that the carried `UID` is meant to be a _reference_ to another chunk. We should replace all _references_ to other chunks with either `TypedUIDRef` or a `newtype UIDRef = UIDRef UID` for disambiguating between the `UID` that a chunk is related to and the `UID`s that it is related to.

With this information around, we should be able to further propagate the automation #4476 introduced to chunk atoms (e.g., as @Xinlu-Y had to manually do in #4515 and #4514 ).

For anyone attempting this, for example, you can begin with:

https://github.com/JacquesCarette/Drasil/blob/dd015080662abaa73079a675e1e41e045a178e31/code/drasil-lang/lib/Language/Drasil/Sentence.hs#L45-L59

You should be able to search for `UID` in the codebase to see all other areas that this needs to be propagated as well.

PR #4555, opened 2025-12-13
`drasil-database`: Create `UIDRef`, an untyped version of `TypedUIDRef`.

balacij:
Contributes to #4541

Renamed `TypedUIDRef.hs` to `UIDRef.hs` and created `UIDRef` (another `newtype` wrapper around `UID` to capture that a `UID` is meant to be a reference).

I would like a bit of feedback regarding code style and naming. I will review my own PR.

### DIFF
--- a/code/drasil-database/lib/Drasil/Database/README.md
+++ b/code/drasil-database/lib/Drasil/Database/README.md
@@ -7,4 +7,4 @@ _Last updated: Dec. 4th, 2025_
 * `Dump.hs`: Simple tool for dumping all known chunks in a database (grouped by type).
 * `UID.hs`: Defines the structure of universally unique identifiers we use for our chunk database.
 * `TH.hs`: Defines an automation tool for declaring that a 'chunk type' has 'chunk references'.
-* `TypedUIDRef.hs`: `UID`s are great! But _untyped._ Creates a chunk reference data type that carries type information at the Haskell-type-level for type-safe `UID` references!
+* `UIDRef.hs`: `UID`s are great! But _untyped._ Creates a chunk reference data type that carries type information at the Haskell-type-level for type-safe `UID` references! Also carries a data type for treating chunks as being _unityped_.


code/drasil-database/lib/Drasil/Database/UIDRef.hs
```diff
@@ -0,0 +1,58 @@
...
+
+import Drasil.Database.Chunk (IsChunk, HasChunkRefs (..))
+import Drasil.Database.ChunkDB (ChunkDB, find)
+import Drasil.Database.UID (HasUID(..), UID)
+import qualified Data.Set as S (singleton)
+
+-- | A reference to another chunk through its 'UID', with a type variable to
+-- retain information about the original chunk's type, useful for type-safe
+-- dereferencing.
+newtype TypedUIDRef typ = TypedUIDRef UID
```
balacij: Should we keep both versions of `UIDRef` in the same file?

code/drasil-database/lib/Drasil/Database/UIDRef.hs
```diff
@@ -0,0 +1,58 @@
...
+typedDeref :: IsChunk t => TypedUIDRef t -> ChunkDB -> Maybe t
+typedDeref (TypedUIDRef u) = find u
+
+-- | Find a chunk by a 'TypedUIDRef', erroring if not found.
+typedDerefOrErr :: IsChunk t => TypedUIDRef t -> ChunkDB -> t
+typedDerefOrErr tu cdb = fromMaybe (error "Typed UID dereference failed.") (typedDeref tu cdb)
+
+-- | A variant of 'TypedUIDRef' without type information about the chunk being
+-- referred to.
+newtype UIDRef = UIDRef UID
```
balacij: I somewhat prefer `UIDRef` to be called `UnsafeUIDRef` and `TypedUIDRef` to be called `UIDRef`, but 'safe' is not quite accurate because we don't have a type-level guarantee that a chunk is registered in the database.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
```diff
@@ -0,0 +1,58 @@
...
+
+instance HasChunkRefs UIDRef where
+  -- | A 'UIDRef t' carries a 'UID' referring to another chunk.
+  chunkRefs (UIDRef u) = S.singleton u
+  {-# INLINABLE chunkRefs #-}
+
+-- | Create a 'UIDRef' to a chunk.
+mkRef :: IsChunk t => t -> UIDRef
+mkRef = UIDRef . (^. uid)
```
balacij: `ref` is fairly common in the codebase, so I went with `mkRef`, but I know @JacquesCarette prefers no `mk` prefixes. Any suggestions?

code/drasil-database/lib/Drasil/Database/UIDRef.hs
JacquesCarette: There are conceptually very similar, so I'm fine having them in the same file.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
JacquesCarette: Maybe `UnitypedUIDRef` rather than `Unsafe` then? But I'm happy with making `Typed` the default by eliding that from its name.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
JacquesCarette: Yes, but `Ref`is for `Reference` which means something specific. Here it is a rather different notion of 'reference', so that's actually confusing!

I think a good name for this function is actually `hide`: it takes a thing and obscures it by a level of indirection.

code/drasil-database/lib/Drasil/Database/UIDRef.hs
balacij: `UnitypedUIDRef` is nice!

code/drasil-database/lib/Drasil/Database/UIDRef.hs
balacij: `hide`/`unhide` works for me.

balacij:
Pending review, I would like to see if Copilot can handle converting `UID` references to `UnitypedUIDRef`s or perhaps even `UIDRef`!

@JacquesCarette This is ready for review again.

PR #4567, opened 2025-12-16
Updated What-If and Removed Miscellaneous Ideas

### LINKED ISSUE #4538: Wiki: Miscellaneous Ideas
The [Miscelaneous Ideas](https://github.com/JacquesCarette/Drasil/wiki/Miscellaneous-ideas) page doesn't have anything _wrong_ with it, and we can leave it as-is if we want. With that said, it only contains one point, and that point has been expressed to some degree in other places (e.g. [What If](https://github.com/JacquesCarette/Drasil/wiki/What-If) mentions the idea of a different development lifecycle). If it is helpful to have that thought in its "bare" form on this separate page, then that's fine; but if it's no longer needed, we could get rid of it.

Contributes to #4401

B-rando1:
I added a paragraph to What If to more explicitly state the ideas of Miscellaneous Ideas, that Drasil offers a different development curve with higher upfront costs but lower long-term costs. It was tricky to find the right place to add that paragraph, and I think I added a bit of redundant information to that section, so let me know if it needs another pass.

Closes #4538

### DIFF
--- a/wiki/What-If.md
+++ b/wiki/What-If.md
@@ -516,7 +516,7 @@ not?
 
 **S** Ah yes, the hunt for the 'silver bullet', which doesn't exist.
 
-The tradeoff of ITD is that it transfers the effort of development to be earlier in the process. In traditional software development, the upfront cost of creating a functioning piece of software is relatively small, with relatively large costs for maintanence over the lifespan of the software. The ITD requires much higher upfront costs, but allows the generation of fully-documented families of programs at an incrementally timy cost: a long-term payoff for a high initial investment. This long-term payoff is worth it for some applications, but not for others.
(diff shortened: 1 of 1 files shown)


wiki/What-If.md
```diff
@@ -516,10 +516,12 @@ not?
 
 **S** Ah yes, the hunt for the 'silver bullet', which doesn't exist.
 
-If you look at the developer steps and the activities, a lot of them
+The tradeoff of ITD is that it transfers the effort of development to be earlier in the process. In traditional software development, the upfront cost of creating a functioning piece of software is relatively small, with relatively large costs for maintanence over the lifespan of the software. The ITD requires much higher upfront costs, but allows the generation of fully-documented families of programs at an incrementally timy cost: a long-term payoff for a high initial investment. This long-term payoff is worth it for some applications, but not for others.
```
JacquesCarette:
```suggestion
The tradeoff of ITD is that it transfers the effort of development to be earlier in the process. In traditional software development, the upfront cost of creating a functioning piece of software is relatively small, with relatively large costs for maintanence over the lifespan of the software. The ITD requires much higher upfront costs, but allows the generation of fully-documented families of programs at an incrementally timy cost: a long-term payoff for a high initial investment. This long-term payoff is worth it for some applications, but not for others. It is expected that this up-front cost will be *amortized* over many projects; many have already noticed that fundamental theory applies widely and changes slowly.
```

JacquesCarette (approved):

PR #4573, opened 2025-12-18
Made README Quick Start point to New Workspace Setup

B-rando1:
They were conflicting, and we decided in the issue that we should de-duplicate the instructions by linking from one to the other.

While I was editing that section, I also added a note on the error messages (e.g. type errors between `Real` and `Vect Real`) that pop up when Drasil is building the examples. I remember when I first tried setting up Drasil on my own a bit before my summer research started, I was confused by those messages and didn't realize that it was working as expected.

Closes #3362

PR #4576, opened 2025-12-18
Updated New Workspace Setup

### LINKED ISSUE #4455: Wiki: New Workspace Setup
There are a few things that could be updated in [New Workspace Setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup).
- [ ] The introduction mentions WSL2 as the best way to use Drasil on Windows, but the actual Windows section only mentions Git Bash and Cygwin. It could be easy for someone to skip the introduction, so we might want to mention it in the Windows section too.
  - While we're talking about it, do we even need to talk about Git Bash and Cygwin at this point? I never had issues with WSL2 and Drasil when I was using Windows, so it seems like the only method we need to talk about. If @samm82 or @jackwyand have opinions on this (I think you both use Windows?), feel free to let me know!
- [ ] The page recommends installing Haskell from Stack. I believe GHCup is the preferred option these days, as it installs not only Stack but several other helpful utilities.
- [ ] In the Stack install instructions, the Mac section has a debugging note for Linux users that doesn't show up in the Linux section.
- [ ] I think it would be helpful if, for each package, we make it more clear which users of Drasil need to install what. E.g. as a GOOL person I haven't had to install Graphviz.

I also have a comment on the level of detail in this page. For a lot of the packages we have very in-depth, platform-specific instructions. This can be quite helpful, but for many of the packages I'm guessing we're just summarizing what's on their own install pages. Would it not be more sustainable to simply link to those pages so that we don't have to worry about our instructions going out of date? We could simply point people to the official guides and mention any subtleties not mentioned there. I don't think it's worth undoing what we have as long as it is up-to-date, but as we add or update the instructions it would be worth keeping in mind.

Contributes to #4401.

B-rando1:
I'm leaving this as draft until #4573 is merged, since I'll need to update some of the links added there.

- Made WSL2 more prominent as the recommended option for Windows
- Switched the Stack install instructions to install it via GHCup
- Added a note that everything after VS Code is technically optional, and that users should read the sections to see if they need them for their purposes.
- Removed an instance "in order to" :smile: 

Closes #4455

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ We wrote a [position paper](https://github.com/JacquesCarette/Drasil/blob/main/P
 
 ## Quick Start
 
-Start by installing the build tools as directed in [New Workspace Setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup). For a minimal setup, only the [System Prerequisites](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#system-prerequisites), [Basic Development Tools](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#basic-development-tools), [Stack](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#stack), and [VS Code](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#vscode) sections are required.
+Start by installing the build tools as directed in [New Workspace Setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup). For a minimal setup, follow everything up to [Optional Utilities](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup#optional-utilities).
 
 After this, you can:
 


wiki/New-Workspace-Setup.md
```diff
@@ -137,80 +139,27 @@ To copy/clone `Drasil`'s software on your machine, please open up your terminal
...
-</details>
-
-<details>
-
-<summary><h4>Mac</h4></summary>
-
-For Mac users, open up your terminal (Apple logo + Space, then search "Terminal"), and run:
-```
-xcode-select --install
-```
```
balacij, line 170: This is still required for almost all kinds of development on MacOS. We should probably keep it somewhere. Maybe we can have a special section at the top for MacOS?

wiki/New-Workspace-Setup.md
```diff
@@ -36,7 +36,9 @@ Please follow the instructions related to your operating system.
 
 <summary><h4>Windows</h4></summary>
 
-If you're on Windows, you will need some sort of bash shell, a git client, and common development tools (e.g., GNU Make, rm, sort, rev, etc).
+If you're on Windows, there are two options.
+- The simpler and recommended option is to [install WSL](https://learn.microsoft.com/en-us/windows/wsl/setup/environment), the Windows Subsystem for Linux. With this option, you can safely ignore all Windows installation notes (except for that regarding Unicode support), and follow the instructions for Linux machines with the *apt*-package manager (e.g., Debian, Ubuntu, etc). From experience, installations with Windows WSL over native installations provide better tools, quicker installs, and an overall smoother experience.
+- You can also use a bash shell such as Git Bash or Cygwin. You will then need, to install a git client and common development tools (e.g., GNU Make, rm, sort, rev, etc). Instructions for these are below.
```
balacij: While WSL can be simpler once it starts working, it is slower/heavier than a native Haskell/GHC install. It's difficult to say which path is better for Drasil because we deal with more tools after the Haskell code is built.

We might just need to learn from experience in the upcoming summer semester.

B-rando1: My personal experience with WSL2 was very positive. I was using Windows until a bit over a year ago, and WSL2 handled everything I needed it for, including the summer of Drasil. The only time I noticed slow behaviour was when I was using it for a personal project I had created in the Windows file system (which isn't how you would do Drasil in WSL2 anyway).

Since it sounds like some people had a different experience than I did, I can make the language more neutral between the two options and add more information on the tradeoffs involved.

PR #4581, opened 2025-12-26
Refactors to be done

JacquesCarette:
Start a list of refactors that should be done to clean a lot of things up. Didn't think an issue would be the best way to track these, as 1) the list is going to grow over time, and 2) some may take a while to get done.


B-rando1 (approved):
This seems like a good idea to me. I can add a section for GOOL in a future PR, which should help keep that refactor organized.

PR #4598, opened 2025-12-30
Language.Drasil.Sentence.Fold should not depend on Language.Drasil.Document

JacquesCarette:
Move the 3 folds that produce `Content` from `drasil-lang` to `drasil-docLang`'s `Drasil.Document.Contents` and adjust accordingly.


samm82 (dismissed):
Other than the linting alerts, this looks good to me! (I've been trying to upgrade my Haskell, etc. installations so I haven't been able to confirm the code works but with such a simple change I'm assuming it would. 😅)

PR #4599, opened 2025-12-30
drasil-printers no longer depends on drasil-system.

JacquesCarette:
Small change (be more precise in what gets passed in, i.e. mainly a `ChunkDB` instead of a whole `System`) in `PrintingInformation`, and adjust a few debug routines to take some arguments explicitly.

But the result is going to be a big change in our package dependency graph! Hopefully that will be one small step to enable further refactorings in a similar vein.


balacij (approved):
I'm very happy to see this one.

PR #4602, opened 2025-12-31
Move a few definitions from drasil-data to drasil-metadata

JacquesCarette:
They allow `drasil-system` to not depend on `drasil-data` anymore, only on `drasil-metadata`. Only `drasil-example` should depend on `drasil-data` (in the long run).

### DIFF
--- a/code/drasil-metadata/lib/Drasil/Metadata/Documentation.hs
+++ b/code/drasil-metadata/lib/Drasil/Metadata/Documentation.hs
@@ -16,7 +16,6 @@ specification   = nc "specification"  (cn'    "specification"      )
 softwareReq :: IdeaDict
 softwareReq                  = compoundNCPP software requirement
 
-
 srs, notebook, requirement :: CI
 notebook    = commonIdeaWithDict "notebook"    (cn' "notebook")                                      "NB"      [softEng]
 requirement = commonIdeaWithDict "requirement" (cn' "requirement")                                   "R"       [softEng]
--- a/code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
+++ b/code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
@@ -1,6 +1,5 @@
 module Drasil.DblPend.Unitals where
 
-
 import Language.Drasil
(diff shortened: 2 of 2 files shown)


code/drasil-metadata/lib/Drasil/Metadata/Documentation.hs
```diff
@@ -0,0 +1,23 @@
...
+software        = nc "software"       (cn     "software"           )
+specification   = nc "specification"  (cn'    "specification"      )
+
+softwareReq :: IdeaDict
+softwareReq                  = compoundNCPP software requirement
+
+srs, notebook, requirement :: CI
+notebook    = commonIdeaWithDict "notebook"    (cn' "notebook")                                      "NB"      [softEng]
+requirement = commonIdeaWithDict "requirement" (cn' "requirement")                                   "R"       [softEng]
+srs         = commonIdeaWithDict "srs"         softReqSpec                                           "SRS"     [softEng]
```
samm82, line 22: Unless we're future-proofing (i.e., assuming that we'll be adding more content here from `drasil-data` that would benefit from similar spacing), removing this extra spacing makes this clearer.

```suggestion
software, specification :: IdeaDict
software      = nc "software"      (cn  "software"     )
specification = nc "specification" (cn' "specification")

softwareReq :: IdeaDict
softwareReq = compoundNCPP software requirement

srs, notebook, requirement :: CI
notebook    = commonIdeaWithDict "notebook"    (cn' "notebook")    "NB"  [softEng]
requirement = commonIdeaWithDict "requirement" (cn' "requirement") "R"   [softEng]
srs         = commonIdeaWithDict "srs"         softReqSpec         "SRS" [softEng]
```

samm82 (approved):
Other than my minor spacing comment, this looks good!

code/drasil-metadata/lib/Drasil/Metadata/Documentation.hs
JacquesCarette: The spacing comes from the source file where I grabbed these from. I'm pretty sure that more things will need to be moved in this way, and then likely the spacing will need adjusted each time. So I'm going to keep it as is for now, until the dust settles.

PR #4613, opened 2026-01-06
Use `DefinedQuantityDict` as type of `System`'s `_quants` list rather than polymorphic type

balacij:
This lets `makeClassy` generate lenses for the `_quants` field.

This PR is one step towards removing the export of `System`s `SI` constructor.

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
@@ -35,7 +35,7 @@ import Drasil.Database (findOrErr, ChunkDB, insertAll, UID, HasUID(..), invert)
 import Drasil.Database.SearchTools (findAllDataDefns, findAllGenDefns,
   findAllInstMods, findAllTheoryMods, findAllConcInsts, findAllLabelledContent)
 
-import Drasil.System (System(SI), whatsTheBigIdea, _systemdb, _quants,
+import Drasil.System (System(SI), whatsTheBigIdea, _systemdb, HasSystem (..),
   _authors, refTable, refbyTable, traceTable, systemdb, sysName, programName)
 import Drasil.GetChunks (ccss, ccss', citeDB)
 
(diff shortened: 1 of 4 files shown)


code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
```diff
@@ -220,19 +220,12 @@ mkRefSec si dd (RefProg c l) = SRS.refMat [c] (map (mkSubRef si) l)
...
-    -- error out because some of the symbols in tables are only `QuantityDict`s, and thus
-    -- missing a `Concept`.
-    mkSubRef SI {_quants = v, _systemdb = cdb} (TSymb con) =
+    mkSubRef sys@SI {_systemdb = cdb} (TSymb con) =
       SRS.tOfSymb
       [tsIntro con,
                 LlC $ table Equational (sortBySymbol
                 $ filter (`hasStageSymbol` Equational)
-                (nub $ map dqdWr v ++ ccss' (getDocDesc dd) (egetDocDesc dd) cdb))
+                (nub $ sys ^. quants ++ ccss' (getDocDesc dd) (egetDocDesc dd) cdb))
```
balacij, line 228: The comments talk about removing `_quants` if we have document traversal, and it looks like we do! `++ ccss' ...` does that. It's not pretty, but it works. We can do it 'better' later if needed.

code/drasil-system/lib/Drasil/System.hs
```diff
@@ -74,7 +73,8 @@ data System where
   , _background   :: Background
   , _scope        :: Scope
   , _motivation   :: Motivation
-  , _quants       :: [e]
+  , _quants       :: [DefinedQuantityDict]
+  -- ^ The list of quantities to be shown in generated SRS documents.
```
balacij, line 77: I'm also going to look into if we can remove this list altogether.

balacij: (in subsequent work.)

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
JacquesCarette: but this doesn't change anything - before there was a binding of `v` to get the `_quants` field, but now it uses the `quants` lens instead by doing a self-reference! So this does not address the comment at all, it is just a programming change.

code/drasil-system/lib/Drasil/System.hs
JacquesCarette: The whole point of the comments was indeed to remove that list entirely. I think you're removing the comments in various files prematurely.

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
balacij: Yes, but using `dqdWr` there was redundant because it rebuilds the `DefinedQuantityDict`s that already exist from their contents. So, it just creates duplication.

Note that before this PR that the `quants` lens did not exist. `makeClassy` and `makeLenses` only make lenses for monomorphic types, which `_quants` was not.

The way that this PR makes a small step towards removing the exported `SI` `System` constructor (implicitly I included accessors here...) is by removing reliance on the internal record field accessor (which we should not be exporting).

code/drasil-system/lib/Drasil/System.hs
balacij: What do you think about filing an issue instead? FIXMEs plastered in the codebase are easily forgotten and often don't contain enough information for them to be actionable by anyone but the author of the comment.

(I'm also actively working on this issue.)

code/drasil-docLang/lib/Drasil/DocumentLanguage.hs
JacquesCarette: Ah, ok, that makes sense.

code/drasil-system/lib/Drasil/System.hs
JacquesCarette: Yes, an issue is better than code comments.

PR #4614, opened 2026-01-06
`SglPend`: Remove unused `DefinedQuantityDict` from scope.

balacij:
The unit vector DQD was not used anywhere in the document. Proof: It's not referenced anywhere in any of the code belonging to `SglPend`. It is only imported for render in the SRS.


JacquesCarette (approved):
I have grepped through stable to verify that, in the rendered document, the unit vector did not in fact appear (outside to table). So this is a good change.

If the unit vector should appear, then it's a document problem, not a table-of-symbols problem.

PR #4616, opened 2026-01-06
`DblPend`: Remove variables not immediately used in the SRS.

balacij:
Proof for `unitVect` and `pi_`: We could remove the imports. Nothing relied on them.

For tension, something did rely on them, however. But in a way that immediately uses the `NP` of `QP.tension`. This should be filed as work under a new ticket -- but this will require design work. Specifically, it will require moving `UID`-based references to other things with `NP`s.


JacquesCarette (approved):
Removal from table is good.

Using a chunk just for its `NP` is *good*, so I don't think we'll want to change anything wrt `tension`.

PR #4617, opened 2026-01-06
`DblPend`: Align SRS-designated system outputs with generated code

balacij:
Builds on #4616 

There is only one output to the system: `theta`. See:

https://github.com/JacquesCarette/Drasil/blob/d9c2775ffa0f988d8cd1b06779534bd77e42b9be/code/stable/dblpend/src/python/OutputFormat.py#L6-L12

Yes, this does make the SRS a bit 'ugly' but it removes a hack that made it 'pretty'.


balacij:
Good catch! It doesn't strictly need to build on #4616.

But, keeping it building on that PR avoids the eventual merge conflict (seeing as you've already approved of #4616).

Sorry!

PR #4623, opened 2026-01-07
`System`: Switch the type of the inputs and outputs from polymorphic `Quantity a`s to `DefinedQuantityDict`.

balacij:
If I had to give only one reason for this PR, it would be that it shows us how our chunk types are used. Having multiple kinds of "quantities" in `drasil-lang` with no clear rationale behind which one to choose when creating a variable is confusing and unhelpful.

However, the original bigger picture goal behind this PR was to avoid exporting `System`'s constructor and field accessors in favour of external modules interacting with the existing `System` through the lenses created by `makeClassy ''System`. However, like `makeLenses`, `makeClassy` does not generate lenses for polymorphic variables. This PR makes those remaining types monomorphic using the "smallest" chunks (in the original nesting-based sense of chunks) possible.


code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
```diff
@@ -80,7 +80,7 @@ mkSRS = [TableOfContents, -- This creates the Table of Contents
         , GDs [] ([Label, Units] ++ stdFields) ShowDerivation
         , DDs [] ([Label, Symbol, Units] ++ stdFields) ShowDerivation
         , IMs [] ([Label, Input, Output, InConstraints, OutConstraints] ++ stdFields) ShowDerivation
-        , Constraints EmptyS inConstraints
+        , Constraints EmptyS (map (`uq` defaultUncrt) inConstraints)
```
balacij, line 83: It's odd that "default uncertainty" margins of 10% are added to the input constraints here, but it wasn't any better the previous way. The code smell was just masked.

code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
```diff
@@ -165,16 +165,16 @@ lenRodCon_1, lenRodCon_2, pendDisAngleCon_1, pendDisAngleCon_2, massCon_1, massC
   :: ConstrConcept
 lenRodCon_1       = constrained' lenRod_1 [gtZeroConstr] (dbl 1)
 lenRodCon_2       = constrained' lenRod_2 [gtZeroConstr] (dbl 1)
-pendDisAngleCon_1 = constrained' pendDisAngle_1 [gtZeroConstr] (dbl 30)
+pendDisAngleCon_1 = constrained' pendDisAngle_1 [gtZeroConstr] (dbl 30) -- FIXME: These are "reasonable values," but they're not used in the SRS?
 pendDisAngleCon_2 = constrained' pendDisAngle_2 [gtZeroConstr] (dbl 30)
```
balacij, line 169: Note that these are supposed to be output variables.

code/drasil-example/sglpend/lib/Drasil/SglPend/Body.hs
```diff
@@ -73,7 +73,7 @@ mkSRS = [TableOfContents, -- This creates the Table of Contents
         , GDs [] ([Label, Units] ++ stdFields) ShowDerivation
         , DDs [] ([Label, Symbol, Units] ++ stdFields) ShowDerivation
         , IMs [] ([Label, Input, Output, InConstraints, OutConstraints] ++ stdFields) ShowDerivation
-        , Constraints EmptyS inConstraints
+        , Constraints EmptyS (map (`uq` defaultUncrt) inConstraints) -- FIXME: Why do all values in this table need a reasonable value?
```
balacij, line 76: @smiths Is it strictly necessary to have reasonable values for all inputs? Or is this data optional? Also, where does this information come from? Should we expect that a user include this from an IM/TM/GD/DD? Or should we expect that a user create reasonable values whenever creating "problem-concrete input variables"?

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
```diff
@@ -236,7 +236,7 @@ aspectRatioMax = dqd' (dcc "aspectRatioMax"
 -----------------
 
 constrained :: [ConstrConcept]
-constrained = map cnstrw' inputConstraints ++ map cnstrw' outputs
+constrained = map cnstrw' inputConstraints ++ outputs
```
balacij, line 239: The extra `cnstrw'` was redundant -- it was extracting `ConstrConcept`s from  existing `ConstrConcept`s.

code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
balacij: Why is it odd that it's added here? Because this is the `SRSDecl` (the data type that says how to arrange the content of the generate SRS -- a kind of "choice" which is not a preferred spot for creating knowledge).

balacij:
Working on the assumption that `System` is staying, this is now ready for review, @JacquesCarette. I also updated the OP with more rationale.

code/drasil-example/sglpend/lib/Drasil/SglPend/Body.hs
smiths: I don't think it is strictly necessary to have reasonable values for all inputs.  People have been creating scientific software outside of Drasil without this information for decades.  :smile:  However, I think we should keep the expectation of reasonable values.  From a requirements point of view, this is valuable information to capture.  What would be even better would be a range of reasonable values, with those values checked to make sure they satisfy the input assumptions.  

In our projectile motion example, we make the assumption that the curvature of the earth can be ignored.  This assumption is only valid if the speed of the projectiles isn't too fast.  At the speed that people can throw a baseball, or hit a golf ball, we don't have to worry about the curvature of the earth.  However, if our software is for ballistics, the speeds are high enough that the curvature matters.  The inputs and the assumptions are tied together.

I'd like to keep reasonable values, to continue to remind us about this valuable information.

The information comes indirectly from the scope decisions.  Once we decide the scope of the problem (projectile motion for sports, versus projectile motion for ballistics), we have information on reasonable values.  Within that scope, it is the domain experts that would say what values are reasonable.  We only need reasonable values for the inputs.  The inputs only come up for the instance models.  Therefore, we need reasonable values for the inputs given in the instance models.  These same inputs will appear in one of the requirements.  The reasonable values are not problem-specific.  The actual inputs are problem-specific.  :smile:

code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
JacquesCarette: Agree that this is odd - open an issue?

code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
JacquesCarette: And why is this worthy of note?

code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
```diff
@@ -165,16 +165,16 @@ lenRodCon_1, lenRodCon_2, pendDisAngleCon_1, pendDisAngleCon_2, massCon_1, massC
...
 massCon_1         = constrained' massObj_1 [gtZeroConstr] (dbl 0.5)
 massCon_2         = constrained' massObj_2 [gtZeroConstr] (dbl 0.5)
 
-inConstraints :: [UncertQ]
-inConstraints = map (`uq` defaultUncrt) [lenRodCon_1, lenRodCon_2, massCon_1, massCon_2]
+inConstraints :: [ConstrConcept]
+inConstraints = [lenRodCon_1, lenRodCon_2, massCon_1, massCon_2]
```
JacquesCarette, line 174: why are you removing the uncertainty? You moved it to a much worse place!

code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -70,7 +70,7 @@ mkSRS = [TableOfContents,
         , GDs [] ([Label, Units] ++ stdFields) ShowDerivation
         , DDs [] ([Label, Symbol, Units] ++ stdFields) ShowDerivation
         , IMs [instModIntro] ([Label, Input, Output, InConstraints, OutConstraints] ++ stdFields) ShowDerivation
-        , Constraints EmptyS inputConstraints
+        , Constraints EmptyS (map (`uq` defaultUncrt) inputConstraints)
```
JacquesCarette, line 73: Adding uncertainty should never be done in the document-creation phase, only in the creation phase of `inputConstraints`.

code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -128,7 +128,7 @@ si = mkSystem progName Specification
   [purp] [background] [scope] [motivation]
   tMods genDefns dataDefs iMods
   []
-  inputs outputs (map cnstrw' constrained) constants
```
JacquesCarette, line 131: why?

code/drasil-system/lib/Drasil/System.hs
```diff
@@ -78,25 +76,24 @@ data System where
   , _dataDefns    :: [DataDefinition]
   , _instModels   :: [InstanceModel]
   , _configFiles  :: [String]
-  , _inputs       :: [h]
-  , _outputs      :: [i]
-  , _constraints  :: [j] --TODO: Add SymbolMap OR enough info to gen SymbolMap
+  , _inputs       :: [DefinedQuantityDict]
+  , _outputs      :: [DefinedQuantityDict]
+  , _constraints  :: [ConstrConcept] --TODO: Add SymbolMap OR enough info to gen SymbolMap
```
JacquesCarette, line 81: This is a bad change. This is a code-driven change rather than a conceptual change, trying to get around a Haskell weakness.

The polymorphic inputs were better.

code/drasil-example/sglpend/lib/Drasil/SglPend/Body.hs
balacij: Thank you! This is very informative.

> I'd like to keep reasonable values, to continue to remind us about this valuable information.

To be clear, I wasn't suggesting removing them. I'm primarily concerned with "where they come from," and your response answers this perfectly.

With this in mind, the concrete types I chose are incorrect! But I don't think we have types to express this knowledge.

I chose the following types for the 3 fields I changed:

```
  , _inputs       :: [DefinedQuantityDict]
  , _outputs      :: [DefinedQuantityDict]
  , _constraints  :: [ConstrConcept]
```

In e30463ba411e01a9c42f7ef9103b57e8f344e2aa, I wrote:

```
`System`: Make `_constraints` field monormophic; `ConstrConcept`.
`ConstrConcept` is chosen over `UncertQ` (the only two options that
satisfy the requirement of defining a quantity that has constraints)
because the "uncertainty" values/percentages are not used anywhere but
the outputted SRS document. Are these numbers just "for show"? Or are
they numbers we can use somehow?
```

Your answers tell me:

1. The reasonable values are numbers we should be able to use (and so choosing `ConstrConcept` over `UncertQ` was the wrong choice).
2. These are numbers, while technically optional, are sensible to require here instead of at the `SRSDecl` level (see https://github.com/JacquesCarette/Drasil/pull/4623#discussion_r2691129784 ). And if there is a situation where choosing reasonable values is not sensible, it might be that producing an SRS is also not sensible.
3. The types we have here are still "wrong." `_inputs` and `_outputs` should also carry information about their reasonable values at least. Why reasonable values? Because we require them for building a human-readable abstract model of an ICO problem (which I'm going to assume is `System`, and which we can generate an SRS using). `_inputs` and `_outputs` should also likely carry information about the constraints imposed on them (physical+software constraints). This isn't to say that we merge `ConstrConcept` and `UncertQ` into `DefinedQuantityDict`, but that we create a new chunk type(s) (e.g., `InputVar` and `OutputVar`) that add extra metadata about the `DefinedQuantityDict`s.

code/drasil-example/dblpend/lib/Drasil/DblPend/Unitals.hs
balacij: Both spots look random to me. I think https://github.com/JacquesCarette/Drasil/pull/4623#discussion_r2691205378 sheds a bit of light (https://github.com/JacquesCarette/Drasil/pull/4623#discussion_r2691472866).

code/drasil-system/lib/Drasil/System.hs
balacij: Using polymorphic variables confuses what `System` is (at least to me). Using concrete types shows intention much clearer. This isn't really purely a code-driven change (which I neglected to mention in the OP...) because I also was looking to understand what `System` really is. I believe this step back might have been helpful if anything comes from https://github.com/JacquesCarette/Drasil/pull/4623#discussion_r2691472866 , but I'll wait for your thoughts on that thread as well.

code/drasil-example/sglpend/lib/Drasil/SglPend/Body.hs
smiths: Thank you for the follow-up @balacij.  Another point about reasonable values, which I didn't mention previously, we can use them to build a test case.  If I remember correctly @samm82 did some preliminary work on that before we pivoted that topic of his thesis research.

code/drasil-system/lib/Drasil/System.hs
JacquesCarette: I think `System` *really is nothing*, i.e. it is a flawed concept.

We may need representations of:
1. an ICO architecture (that would have inputs and outputs)
2. a "software intensive solution" (which would have authors, for example, and a description of a piece of software that would solve the solution; that would have an architecture)
3. a specification of what properties our solution should have (that is often done via an SRS)
and most likely more things that I can't quite think of right now.

What `System` *is* right now is largely uninteresting (for what purpose we're (ab)using it is somewhat interesting). What we need for an explainable Drasil is much more important.

Issue #4610, opened 2026-01-02
There should be no output `IO ()` outside of `drasil-gen`


JacquesCarette:
What I really mean is that `drasil-*/lib` outside of `drasil-gen` should not do output IO. (There's some input IO for meta-programming reasons, that's ok).

Right now, `drasil-code`, `drasil-utils` and `drasil-printers` fail this. All should be move.

[The "failure" in `drasil-utils` is sort of benign, and is a fine factoring out of common code. It still does not belong there.]

PR #4634, opened 2026-01-13
`drasil-code`: Merge `Language.Drasil.Code.CodeGeneration` into `Language.Drasil.Code.Code`.

balacij:
Although this work will be temporary and will be superseded by work for #4610 , it was already done and I believe is worth merging.

The reason for the merge is because of how heavily related they are. They rely on the same "secrets."


JacquesCarette:
also needs a merge

balacij:
Should be good for re-review again, @JacquesCarette

PR #4635, opened 2026-01-13
Move `isConstructor` closer to the smart constructors for `MethodInfo`.

balacij:
Reason: It should either be deleted or moved closer to `MethodInfo` because it relies on `MethodInfo` secrets.


JacquesCarette:
Thanks for the much better description.

PR #4645, opened 2026-01-19
Remove a lot of unnecessary duplication

JacquesCarette:
`Expr`, `ModelExpr` and `CodeExpr` had a lot of duplication that was not needed at all. This removes all of that.

What is left is a clearer picture of what makes them different from each other (generally: just a handful of constructors). Makes further refactoring easier.


balacij (approved):
Very straightforward.

I look forward to re-designing `(Model|Code)?Expr` after our latest discussion about `sin`. If done right, that might even merge them back together (maybe a dream).

PR #4650, opened 2026-01-20
drasil-gool should not depend on drasil-metadata

JacquesCarette:
It's partially okay to depend on drasil-utils, but otherwise drasil-gool should not depend on the rest of Drasil.

A dependence had snuck in via `watermark`. This had been implemented incorrectly by using it directly; the correct implementation is to pass it in as yet another argument, as is now done. `drasil-code` can depend on `drasil-metadata` so that's okay.


JacquesCarette:
Ok, I've fixed the argument order, tests pass now. Review when you can @balacij and/or @B-rando1 .

Issue #4653, opened 2026-01-20
New Name for "Auxiliary Files"


B-rando1:
Just branching off our discussion from #4625. We want to stop using the term "Auxiliary Files" because it implies they are less important than code - which goes against Drasil's core ideas. @smiths's [comment](https://github.com/JacquesCarette/Drasil/issues/4625#issuecomment-37528929650) has a few good ideas for new names. "Project Scaffolding" is good, but we agreed that "Software Dossier" is probably the best.

There are a few things we can do now that we have this name:
- Rename everything that is (or was until recently) called "Auxiliary"
- Start working towards a `software-dossier` package.
  - Would it make sense to start by making a `software-dossier` sub-package within `drasil-code`?
- Add this to the wiki somewhere. I don't think we currently have a page on Drasil's terminology, but I think we need one. There are places all over the codebase where we mean a very specific thing when we use a word, and that specific meaning is largely implicit. A wiki page would help.

smiths:
Thank you for the reminder about the Coding Style section of the Guide.  This is a good foundation.

JacquesCarette:
Starting by isolating all the `software-dossier` modules currently in `drasil-code` is a good idea. I don't think it needs to be a sub-package (in the technical, cabal sense). Certainly the 'code' parts of `drasil-code` should not depend on anything in `software-dossier`. The other way around is probably fine, since a 'software dossier' will indeed contain some code as well as many other artifacts.

B-rando1:
List of files that we can move into a new `SoftwareDossier` folder:
- Most of `FileData.hs`
  - We might want to move `PackageData` out before or after
- `FileNames.hs`
- `PackageFiles.hs`
  - First we need to split of the `spaceToCodeType` function, since it is unrelated to the rest of the file
- Parts of `GenerateGOOL.hs`
- Most of `ClassInterface.hs`
  - We might want to move the `package` function somewhere else
- Everything in `LanguageRenderer` is mostly SoftwareDossier-related too. Some parts of it might need to be redesigned first, though. I'll need to take a better look.

PR #4654, opened 2026-01-21
Added "Artifacts Terminology" to wiki

B-rando1:
Contributes to #4653

It might be good to double-check my definitions for "artifact" and "software dossier" to make sure I'm not missing anything.

I put examples for all of the types of artifacts I could think of, but let me know if I missed anything important there.

### DIFF
--- a/wiki/Terminology.md
+++ b/wiki/Terminology.md
@@ -10,11 +10,12 @@ whenever we use them in the context of Drasil.
 - "Artifact": any and all files associated with a software project. These can
 include the code, SRS, makefile, README, Doxygen config files, sample inputs,
 dot files, etc. Implicitly, they are anything that Drasil can (or should be able
-to) generate.
+to) generate. Note that the 'file' part is not actually important, it merely
+corresponds to the most common representation of artifacts in current use.
 - "Software Dossier": artifacts directly related to the code, that are not
 themselves code. Examples include Doxygen config files, READMEs, sample inputs,
 and makefiles.
-- "Code": What it sounds like - executable code. Currently we support C++, C#,
+- "Code": What it sounds like - source code. Currently we support C++, C#,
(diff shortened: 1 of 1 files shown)


wiki/Terminology.md
```diff
@@ -0,0 +1,21 @@
+## Introduction
+
+The purpose of this page is to explain some of the words we use to mean specific
+Drasil-related words. These all have a general meaning that hints at how we use
+them in Drsail, but we often have a more specific meaning in mind
```
JacquesCarette:
```suggestion
them in Drasil, but we often have a more specific meaning in mind
```

wiki/Terminology.md
```diff
@@ -0,0 +1,21 @@
...
+whenever we use them in the context of Drasil.
+
+## Artifacts
+
+- "Artifact": any and all files associated with a software project. These can
+include the code, SRS, makefile, README, Doxygen config files, sample inputs,
+dot files, etc. Implicitly, they are anything that Drasil can (or should be able
+to) generate.
```
JacquesCarette:
```suggestion
to) generate. Note that the 'file' part is not actually important, it merely corresponds to the most common representation of artifacts in current use.
```

wiki/Terminology.md
```diff
@@ -0,0 +1,21 @@
...
+- "Software Dossier": artifacts directly related to the code, that are not
+themselves code. Examples include Doxygen config files, READMEs, sample inputs,
+and makefiles.
+- "Code": What it sounds like - executable code. Currently we support C++, C#,
```
JacquesCarette:
```suggestion
- "Code": What it sounds like - source code. Currently we support C++, C#,
```

JacquesCarette (changes requested):

PR #4655, opened 2026-01-22
GlassBR: Collapse symbols lists.

balacij:


samm82 (approved):
If there's no longer a reason to keep these "groups" of symbols separate, this change looks good to me!

JacquesCarette (approved):
I agree that the separate file is no longer the best way to handle this, and was going to be a hindrance going forward.

PR #4659, opened 2026-01-22
`drasil-code`: Auto-register variables defined in `Mod`s.

balacij:
This work only affects GlassBR.

This won't pass the CI yet because two _code-only_ variables (interpY and interpZ) appear in the SRS, but this code removes them from the scope of the SRS-generator's `ChunkDB`! Work will be necessary before this can be merged that: (a) removes them from the SRS and (b) hooks them through the `Choices` about how the ICO problem definition should be changed. This way, the two _code-only_ variables only appear in the `ChunkDB` associated with the code-generator.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/CodeSpec.hs
+++ b/code/drasil-code/lib/Language/Drasil/CodeSpec.hs
@@ -174,7 +174,11 @@ codeSpec si chs ms = CS {
     els = extLibs chs
     libReqs = concatMap odeLibReqs els
     infoReqs = concatMap odeInfoReqs els
-    db' = insertAll (libReqs ++ infoReqs) $ si ^. systemdb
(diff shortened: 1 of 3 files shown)


code/drasil-code/lib/Language/Drasil/CodeSpec.hs
```diff
@@ -174,7 +174,11 @@ codeSpec si chs ms = CS {
     els = extLibs chs
     libReqs = concatMap odeLibReqs els
     infoReqs = concatMap odeInfoReqs els
-    db' = insertAll (libReqs ++ infoReqs) $ si ^. systemdb
+    msReqs = map asVC $ concatMap (\(Mod _ _ _ _ l) -> l) ms
```
balacij, line 177: On an unrelated note, `asVC` is a hack that we should be able to remove merely by making `Func`s chunks. They're treated as chunks anyways!

Issue #3355, opened 2023-04-27
How can we remove the need for GlassBR's `ModuleDefs.hs` code?


balacij:
`ModuleDefs.hs` hand codes some parsing/interpretation code needed for parsing the inputs of GlassBR. How can we avoid the 'manual' part of this and have it generated instead?

Inspired by discussion in #2281

PR #4664, opened 2026-01-23
GlassBR: Move code-oriented QDefs forced into SRS to Choices as hand-wired QDefs.

balacij:
The IM and DD involving interpY and interpZ appear to be forced into the SRS as a result of not having any way to read in the quantities directly (#3355). We should be able to force them out of the SRS by moving to a list of hand-wired QDefs provided through `Choices`.

This PR is slightly dubious and I was a bit rough with the removal of the DD and the IM, which is why I'm leaving this PR as a draft. But feedback would be appreciated on the rationale @JacquesCarette @smiths . In the future, these models (DD/IM) can be moved towards an SDS _if_ we don't already have a better mechanism in place for input parameters with default values AND support for matrices/lists.

### DIFF
--- a/code/drasil-example/glassbr/lib/Drasil/GlassBR/Choices.hs
+++ b/code/drasil-example/glassbr/lib/Drasil/GlassBR/Choices.hs
@@ -24,15 +24,15 @@ choices = defaultChoices {
   srsConstraints = makeConstraints Exception Exception,
   defaultConfigFiles = configFp,
   extraMods = allMods,
-  handWiredDefs = [strDisFacQD, calofDemandQD, tolPreQD]
+  handWiredDefs = [strDisFacQD, demandQD, tolPreQD]
 }
 
 strDisFacQD :: SimpleQDef
 strDisFacQD = mkQuantDef stressDistFac
   $ apply interpZ [str "SDF.txt", sy aspectRatio, sy dimlessLoad]
 
-calofDemandQD :: SimpleQDef
-calofDemandQD = mkQuantDef demand
+demandQD :: SimpleQDef
+demandQD = mkQuantDef demand
   $ apply interpY [str "TSD.txt", sy standOffDist, sy eqTNTWeight]
 
 tolPreQD :: SimpleQDef


code/stable/glassbr/SRS/PDF/GlassBR_SRS.tex
```diff
@@ -93,8 +93,6 @@ \subsection{Table of Symbols}
 \\
 $\mathit{interpY}$ & InterpY & --
```
JacquesCarette: Something is weird here: `interpY` is still in the table of symbols, but `interpZ` is not. This seems inconsistent: either both should be, or none should be.

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
```diff
@@ -1237,9 +1232,7 @@ <h3>Data Definitions</h3>
                     </tr>
                     <tr>
                       <th>RefBy</th>
```
JacquesCarette, line 1229: This RefBy change seems not good? The "calculation of demand" is still a thing, but with this change, it has gone?

code/stable/glassbr/SRS/HTML/GlassBR_SRS.html
```diff
@@ -1341,68 +1332,7 @@ <h3>Data Definitions</h3>
...
-                      <td>
```
JacquesCarette, line 1344: And here, even more content has disappeared. This content might have been "hooked in" in a sub-optimal way before -- but it shouldn't entirely disappear. It needs to been hooked in some other way.

code/stable/glassbr/SRS/mdBook/src/SecFRs.md
```diff
@@ -70,7 +70,6 @@ Output-Values: Output the values from the table for [Required Outputs](./SecFRs.
 |\\(h\\)                      |Minimum thickness                                     |[DD:minThick](./SecDDs.md#DD:minThick)          |\\({\text{m}}\\) |
 |\\(\mathit{isSafeLR}\\)      |3 second load equivalent resistance safety requirement|[IM:isSafeLR](./SecIMs.md#IM:isSafeLR)          |--               |
 |\\(\mathit{isSafePb}\\)      |Probability of glass breakage safety requirement      |[IM:isSafePb](./SecIMs.md#IM:isSafePb)          |--               |
-|\\(J\\)                      |Stress distribution factor (Function)                 |[IM:stressDistFac](./SecIMs.md#IM:stressDistFac)|--               |
```
JacquesCarette, line 73: Was J a hack?

code/stable/glassbr/SRS/mdBook/src/SecIMs.md
```diff
@@ -19,31 +19,12 @@ The goal [GS:Predict-Glass-Withstands-Explosion](./SecGoalStmt.md#willBreakGS) i
...
 |Description       |<ul><li>\\(B\\) is the risk of failure (Unitless)</li><li>\\(k\\) is the surface flaw parameter (\\(\frac{\text{m}^{12}}{\text{N}^{7}}\\))</li><li>\\(a\\) is the plate length (long dimension) (\\({\text{m}}\\))</li><li>\\(b\\) is the plate width (short dimension) (\\({\text{m}}\\))</li><li>\\(m\\) is the surface flaw parameter (\\(\frac{\text{m}^{12}}{\text{N}^{7}}\\))</li><li>\\(E\\) is the modulus of elasticity of glass (\\({\text{Pa}}\\))</li><li>\\(h\\) is the minimum thickness (\\({\text{m}}\\))</li><li>\\(\mathit{LDF}\\) is the load duration factor (Unitless)</li><li>\\(J\\) is the stress distribution factor (Function) (Unitless)</li></ul>|
```
JacquesCarette, line 21: Look here, J is in the description, but no longer in the notes nor does it appear as an IM. This leaves the SRS in an inconsistent state.

code/stable/glassbr/SRS/mdBook/src/SecIMs.md
```diff
@@ -76,9 +57,9 @@ The goal [GS:Predict-Glass-Withstands-Explosion](./SecGoalStmt.md#willBreakGS) i
...
 |Description       |<ul><li>\\(\hat{q}\\) is the dimensionless load (Unitless)</li><li>\\(q\\) is the applied load (demand) (\\({\text{Pa}}\\))</li><li>\\(a\\) is the plate length (long dimension) (\\({\text{m}}\\))</li><li>\\(b\\) is the plate width (short dimension) (\\({\text{m}}\\))</li><li>\\(E\\) is the modulus of elasticity of glass (\\({\text{Pa}}\\))</li><li>\\(h\\) is the minimum thickness (\\({\text{m}}\\))</li><li>\\(\mathit{GTF}\\) is the glass type factor (Unitless)</li></ul>|
-|Notes             |<ul><li>\\(q\\) is the 3 second duration equivalent pressure, as given in [DD:calofDemand](./SecDDs.md#DD:calofDemand).</li><li>\\(a\\) and \\(b\\) are the dimensions of the plate, where (\\(a\geq{}b\\)).</li><li>\\(E\\) comes from [A:standardValues](./SecAssumps.md#assumpSV).</li><li>\\(h\\) is defined in [DD:minThick](./SecDDs.md#DD:minThick) and is based on the nominal thicknesses.</li><li>\\(\mathit{GTF}\\) is defined in [DD:gTF](./SecDDs.md#DD:gTF).</li></ul>      |
```
JacquesCarette, line 79: Same here: q is used in the Description above, but now has gone from here.

code/stable/glassbr/SRS/PDF/GlassBR_SRS.tex
balacij: This is now dealt with in 195f1bbbfd91f3683b5fd839f4f3d804a9f59ed9. I missed an IM that also referenced `interpY`.

code/stable/glassbr/SRS/mdBook/src/SecFRs.md
balacij: I'm not entirely sure yet, but it is definitely not of type: Function!

This is the code we generate for it.

```csharp
    /** \brief Calculates stress distribution factor (Function)
        \param inParams structure holding the input values
        \param q_hat dimensionless load
        \return stress distribution factor (Function)
    */
    public static double func_J(InputParameters inParams, double q_hat) {
        StreamWriter outfile;
        outfile = new StreamWriter("log.txt", true);
        outfile.WriteLine("function func_J called with inputs: {");
        outfile.Write("  inParams = ");
        outfile.Write("Instance of InputParameters object");
        outfile.WriteLine(", ");
        outfile.Write("  q_hat = ");
        outfile.WriteLine(q_hat);
        outfile.WriteLine("  }");
        outfile.Close();
        
        return Interpolation.interpZ("SDF.txt", inParams.AR, q_hat);
    }
```

balacij:
I'm not going to close this PR yet because I think I can make a case for it again. However, I'll need to revisit it properly and figure out how to rectify the issues brought up.

PR #4665, opened 2026-01-23
`Choices` configuration files: Require they be valid, relative file paths with file extensions.

### LINKED ISSUE #2193: Prevent access to external files not in Config Files
Configuration files are files that must be in same directory as executable in order for the executable to run. An example would be GlassBR that uses "SDF.txt" and "TSD.txt" in DataDefs.hs. While adding readme's to drasil in #2123 , config files were not stored anywhere, so they were added as a field in system information. They were added as a list of strings that the user must manually enter when the system information is defined. That leads to the following inconsistency:

Chunks can access files even if they are not mentioned in the config files field.

There are a few options
a) a warning is given if a file is accessed outside of the config files
b) an error   is given if a file is accessed outside of the config files
c) The config files field is automatically updated whenever the user accesses an external file.

balacij:
Closes #2193

Stumbled upon the issue and figured this would be a simple enough solution for now at least.

I'm not quite sure how we feel about `drasil-utils` anymore, but `RelativeFile` is probably a type that belongs there. What do we think?

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Choices.hs
+++ b/code/drasil-code/lib/Language/Drasil/Choices.hs
@@ -9,18 +9,17 @@ module Language.Drasil.Choices (
   CodeConcept(..), matchConcepts, SpaceMatch, matchSpaces, ImplementationType(..),
   ConstraintBehaviour(..), Comments(..), Verbosity(..), Visibility(..),
   Logging(..), AuxFile(..), getSampleData, hasSampleInput, defaultChoices,
-  choicesSent, showChs, InternalConcept(..),
-  RelativeFile, relativeFile, relFileToStr
+  choicesSent, showChs, InternalConcept(..)
 ) where
 
 import Control.Lens ((^.))
 import Data.Map (Map)
 import qualified Data.Map as Map
-import System.FilePath (isAbsolute, isValid, hasExtension)
 
 import Drasil.Database (UID, HasUID (..))
 import Drasil.GOOL (CodeType)
 import Language.Drasil hiding (None)
+import Utils.Drasil (RelativeFile)
 
 import Data.Drasil.ExternalLibraries.ODELibraries (odeInfoChunks)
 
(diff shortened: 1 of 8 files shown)


code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -406,3 +409,15 @@ defaultICName InputFormat         = "InputFormat"
...
+-- | A valid, relative file path with an extension.
+newtype RelativeFile = RF { relFileToStr :: String }
+  deriving Eq
+
+-- | Create a 'RelativeFile' given a raw 'String'.
+relativeFile :: String -> RelativeFile
+relativeFile fp
+  | not $ isValid fp = error $ "`" ++ fp ++ "` is not a valid file path."
+  | not $ hasExtension fp = error $ "`" ++ fp ++ "` does not contain a file extension."
+  | isAbsolute fp = error $ "`" ++ fp ++ "` is not a relative file path."
```
samm82: This error message could potentially be more descriptive by emphasizing that it being absolute is the problem; something like "`fp` is an absolute file path, but a relative file path was expected."

Nothing major and I don't mean to bikeshed; everything else looks good!

JacquesCarette: Agree that this would be a better error message.

code/drasil-code/lib/Language/Drasil/Choices.hs
```diff
@@ -406,3 +409,15 @@ defaultICName InputFormat         = "InputFormat"
 defaultICName OutputFormat        = "OutputFormat"
 defaultICName Calculations        = "Calculations"
 defaultICName Constants           = "Constants"
+
+-- | A valid, relative file path with an extension.
+newtype RelativeFile = RF { relFileToStr :: String }
```
JacquesCarette: I'm okay with `drasil-utils`. And thus this (and `relativeFile` below) belong there rather than here.

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: Thanks! I just updated this in 8622e26e2ab8e9116185081111475cea2d0c2bf9

code/drasil-code/lib/Language/Drasil/Choices.hs
balacij: Thanks! Just moved this in f25edd7bc09ea411fb0ae42d6f61423f03d5c990

Issue #4652, opened 2026-01-20
Rename & Move `Language.Drasil.Code.Imperative.GOOL.ClassInterface`


B-rando1:
Similar to #4642 and #4646, [ClassInterface](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs) has nothing to do with "Imperative" or "GOOL", so it should be moved out of there. As with #4642, this will require us to parameterize `ProgData` and `GOOLState` to remove the `GOOL` imports. From a quick look around,  I don't think that will be difficult.

The name "ClassInterface" is also vague and potentially misleading (Haskell classes, not OO classes). Yet again, we don't have a good name for the types of files this file deals with. `AuxiliaryFileTypeclasses` is probably not right... maybe `FileTypeclasses` if we can't think of anything better.

JacquesCarette:
We should also think about whether we even want these classes. Rule of thumb: unless you're doing fancy type-level programming, if a class has a single instance, then that class does not serve any real purpose. So these might also be up for deletion, rather than renaming+moving !

B-rando1:
Just summarizing #4670 here so that I can close it.

- @balacij mentioned that the classes might be overkill for what we need. I don't understand the alternative. In #4660 I did as much as I could easily do to shrink the classes down to the parts that actually differ between languages. It seems to me that typeclasses are the most elegant way of managing the remaining parts that have different implementations between languages. If there are alternatives, I'd be happy to talk about them!
- @balacij also mentioned that many of the things currently imported from `drasil-gool` are just version strings, language names, etc. We could move them to `drasil-metadata` to improve the import graph.
- @JacquesCarette suggested defining new types in `drasil-code` for the subset of each type from `drasil-gool` that is actually used in `drasil-code`. In #4678 I successfully applied this to `GOOLState`. with `ProgData`, however, all of it is used by `drasil-code`, so it feels a bit off to redefine the same type in `drasil-code` just to remove some imports. I'm still trying to think if there's some way we can move things around to prevent this, but nothing comes to mind as of yet.

PR #4670, opened 2026-01-24
<Help Needed> Trying to remove GOOL imports from ClassInterface

B-rando1:
I've been trying to remove the GOOL imports from ClassInterface as discussed in #4652, but I keep running into the same problems. The implementations of the methods have to be generic, which causes issues when they need to, for example, access a record field of `ProgData`. Specifically, `LanguagePolymorphic` and `Import` have errors.

I'm a bit stumped with this one. Maybe the deeper issue is that we need to know all about the program in order to run the `doxConfig` and `makefile` functions. Either way, I think we'll need to either decide not to remove the GOOL import or move some other stuff around first.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
@@ -7,7 +7,6 @@ module Language.Drasil.Code.Imperative.GOOL.ClassInterface (
 
 import Text.PrettyPrint.HughesPJ (Doc)
 
-import Drasil.GOOL (GOOLState)
 import Language.Drasil.Printers (PrintingInformation)
 
 import Language.Drasil (Expr)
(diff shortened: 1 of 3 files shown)


code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs
```diff
@@ -23,18 +22,18 @@ import Language.Drasil.Code.Imperative.README (ReadMeInfo(..))
 -- omptimize doxygen document, information necessary for a makefile,
 -- auxiliary helper documents
 class AuxiliarySym r where
```
balacij, line 24: This class seems like overkill for what we need? Could re-building this help simplify things?

balacij: It also looks like all of the other modules in `Language.Drasil.Code.Imperative.GOOL` only rely on GOOL for language names and "version" strings.

It sounds like these are pieces of metadata that need to be moved towards `drasil-metadata`, whether it be as chunks (a task for another day) or the same raw strings. Am I correct in thinking that this code would then no longer need to rely on `drasil-gool`?

JacquesCarette: I think this is likely the way to go: don't pass `GOOLState` at all, but a new, focused data-structure that only contains the actually needed information. Then try something similar with `ProgData`.

B-rando1: From #4678, it looks like we should be able to get rid of some of these big GOOL-related structures by defining new types with only the data that we need.

For removing the typeclasses completely, it might be possible but it would need a bigger redesign. In #4660 I removed everything from the typeclasses that was implemented identically between renderers, so everything left needs some way of changing its behaviour based on the target. Currently we do this using typeclasses, but we could also do it by passing functions around, for example. I'm not exactly sure what that would look like or if the end result would be better, but it might be worth trying if I get a chance.

B-rando1:
I put a summary in #4652. Closing now.

Issue #4675, opened 2026-01-26
Move `d-printers`/`Language.Drasil.Format` to `drasil-gen`


balacij:
```haskell
-- | Possible formats for printer output.
module Language.Drasil.Format where

-- | Document types include Software Requirements Specification and Website.
-- Choosing SRS will generate both TeX and HTML files, while Website generates only as HTML.
-- This also determines what folders the generated files will be placed into.
data DocType = SRS | Website | Lesson

-- | Possible formats for printer output.
data Format = TeX | Plain | HTML | Jupyter | MDBook

instance Show Format where
  show TeX     = "PDF"
  show Plain   = "Plain"
  show HTML    = "HTML"
  show Jupyter = "Jupyter"
  show MDBook  = "mdBook"

-- | Shows the different types of documents.
instance Show DocType where
  show Lesson  = "Lesson"
  show SRS     = "SRS"
  show Website = "Website"
```

None of these things need to be known by `drasil-printers`. These are kinds of _choices_ for the SRS/website/lesson plan generation.

PR #4677, opened 2026-01-26
Clean up "Lesson Plan"-related generation code.

balacij:
Contributes to #4680

The goal of this PR is to unblock #4675, but it also somewhat cleans up the code, if only by breaking off a piece of `Drasil/Generator/Generate.hs` into another file. I will add a comment in a review regarding how this unblocks #4675.

This PR does two things:

1. It moves said code to a new file within `drasil-gen`
2. It simplifies the generation-related code by bypassing an unnecessary chain of calls, bypassing the `DocType` `Lesson` option.

Realistically, all code related to JSON, Jupyter, and Markdown should be re-evaluated at some point.

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/Format.hs
+++ b/code/drasil-printers/lib/Language/Drasil/Format.hs
@@ -4,7 +4,7 @@ module Language.Drasil.Format where
 -- | Document types include Software Requirements Specification and Website.
 -- Choosing SRS will generate both TeX and HTML files, while Website generates only as HTML.
 -- This also determines what folders the generated files will be placed into.
-data DocType = SRS | Website | Lesson
+data DocType = SRS | Website
 
 -- | Possible formats for printer output.
 data Format = TeX | Plain | HTML | Jupyter | MDBook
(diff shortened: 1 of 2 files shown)


code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
```diff
@@ -48,8 +41,9 @@ build (Document t a c) =
   makeMetadata $$
   text "}"
 
-build' :: Document -> Doc
-build' (Document t a c) =
+-- | Build an SRS document in JSON format.
+genJupyterSRS :: Document -> Doc
+genJupyterSRS (Document t a c) =
```
balacij, line 46: These functions are so tightly coupled with how they're used that their names are a bit inappropriate. Shifting the responsibility of deciding which one to use (`build` or `build'`) from `genJupyter` to the user of `genJupyter` is marginally better because it will let us move `Language.Drasil.Format` (which contains 'document output format' options for the 'document generator') from `drasil-printers` to `drasil-gen` (where the option definition should belong).

balacij: It's also better because it shows us that `drasil-printers` knows too much about some things it shouldn't (namely the lesson plan and the SRS). Between `genJupyterLessonPlan` and `genJupyterSRS` should truly be `genJupyter` that they rely on. `drasil-printers` might then be the appropriate spot for that, but the two specialized versions are likely best in `drasil-gen`.

PR #4678, opened 2026-01-26
Switched drasil-code from GOOLState to SoftwareDossierState

B-rando1:
As discussed in #4670, `GOOLState` is much more broad than what any of `drasil-code` actually needs. I created `SoftwareDossierState` to be the subset of `GOOLState` that we actually need, and switched everything over to it. (I was actually surprised, there's no imports of `GOOLState` left in `drasil-code` now!)

Contributes to #4652

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
@@ -42,7 +42,7 @@ import Language.Drasil.Code.Imperative.Modules (genInputMod, genInputModProc,
 import Language.Drasil.Code.Imperative.DrasilState (GenState, DrasilState(..),
   ScopeType(..), designLog, modExportMap, clsDefMap, genICName)
 import Language.Drasil.Code.Imperative.GOOL.ClassInterface (
-  SoftwareDossierState(..), AuxiliarySym(..), package)
+  makeSds, AuxiliarySym(..), package)
 import Language.Drasil.Code.Imperative.README (ReadMeInfo(..))
 import Language.Drasil.Code.FileData (PackageData(..), fileAndContents)
 import Language.Drasil.Code.FileNames(sampleInputName)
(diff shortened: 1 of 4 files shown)


code/drasil-code/test/Main.hs
```diff
@@ -73,7 +75,12 @@ classes :: (OOProg r, AuxiliarySym r', Monad r') => (r (OO.Program r) -> ProgDat
 classes unRepr unRepr' = zipWith
   (\p gs -> let (p',gs') = runState p gs
                 pd = unRepr p'
-  in unRepr' $ package pd [makefile [] Program [] gs' pd])
+                fileInfoState = Sds {
+                  _headers = gs' ^. headers,
```
JacquesCarette: if a record has fields with _ names, the convention is that we should not use these outside of the one file that defines that record. Two alternatives:
1. use ```Sds (gs' ^. headers) (gs' ^. sources) (gs' ^. mainMod)``` instead
2. define a constructor for `Sds` and use that.

There might also be a lens-y way of doing that too, but I forget what it is.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -147,8 +147,13 @@ genPackage unRepr = do
   p <- genProgram
   let info = OO.unCI $ evalState ci initialState
       (reprPD, s) = runState p info
+      fileInfoState = Sds {
```
JacquesCarette: See comment in `test/Main.hs`

code/drasil-code/test/Main.hs
```diff
@@ -84,7 +91,12 @@ jlClasses :: (ProcProg r, AuxiliarySym r', Monad r') => (r (Proc.Program r) -> P
 jlClasses unRepr unRepr' = zipWith
   (\p gs -> let (p',gs') = runState p gs
                 pd = unRepr p'
-  in unRepr' $ package pd [makefile [] Program [] gs' pd])
+                fileInfoState = Sds {
```
JacquesCarette: same comment here

PR #4691, opened 2026-01-27
Change `shortdep` to return `Set UID`

Xinlu-Y:
This PR is based on the lnames-set-uid branch #4690 

- `shortdep` now returns `Set UID`
- `sentenceRefs` updated to union `shortdep` with `sdep/lnames`


JacquesCarette:
I have, at least partially, already disabled the "dismiss stale reviews" option. But I guess I could more aggressively do so.

The other option is for me to do a whole batch of reviews, and only afterwards do a whole bunch of merges. The issue arises when I interleave these steps.

Opinions?

PR #4695, opened 2026-01-28
`introduceAbb`: Only titleize nouns on abbreviation/acronym introduction for proper nouns.

balacij:
Follows @smiths' comment on #4363: https://github.com/JacquesCarette/Drasil/pull/4363#discussion_r2505757980


samm82 (commented):
Based on the changes to stable, it looks like we only use `introduceAbb` for SRS, which is really surprising to me! Should we look for other places where we could use it for consistency, probably as part of another PR?

EDIT: I see related discussion being held on #4363. 😅

balacij:
> which is really surprising to me!

Yeah, me too. It's a very helpful function. We definitely should be using it more.

Issue #4671, opened 2026-01-26
`Drasil.Database.SearchTools`


JacquesCarette:
First, I dislike that it is in `drasil-printers`. It most definitely does not belong there. [As a different hack, I think I'll move it to `drasil-theory`. It is not a good home either, but at least introduces no new dependencies.]

Second, I really dislike this code:
```haskell
-- | Search for a chunk that is an instance of 'Idea' and return its "term" and
-- abbreviation, erroring out if it doesn't exist.
termResolve :: (NP -> Maybe String -> c) -> ChunkDB -> UID -> c
termResolve f db trg
  | (Just c) <- find trg db :: Maybe IdeaDict            = go f c
  | (Just c) <- find trg db :: Maybe DefinedQuantityDict = go f c
  | (Just c) <- find trg db :: Maybe ConceptChunk        = go f c
  | (Just c) <- find trg db :: Maybe UnitDefn            = go f c
  | (Just c) <- find trg db :: Maybe DataDefinition      = go f c
  | (Just c) <- find trg db :: Maybe InstanceModel       = go f c
  | (Just c) <- find trg db :: Maybe GenDefn             = go f c
  | (Just c) <- find trg db :: Maybe TheoryModel         = go f c
  | (Just c) <- find trg db :: Maybe ConceptInstance     = go f c
  | otherwise = error $ "Term: " ++ show trg ++ " not found in TermMap"
  where
    go :: Idea t => (NP -> Maybe String -> c) -> t -> c
    go f' c = f' (c ^. term) (getA c)
```
This tightly couples what chunks with have with this code. Turns out there are at least 4 such routines in Drasil's code base (and only two in this particular file.) These are basically impossible to find. What I mean is that if you add a new Chunk, you really ought to look for all similar code that does a variation of `find` on a `ChunkDB` and adjust it.

It's easy to blame the architecture of `ChunkDB` for this, i.e. the use of `Typeable`. That is part of it. But the old `ChunkDB` had a similar problem of having to know about all chunk kinds. So this just moves the problem around.

The better solution is to know, a priori, what kind of chunk one is looking for. So if `UID`s were typed, that would be possible [but we should test that to make sure!]

Issue #4300, opened 2025-07-27
Examine all "grab all X from the ChunkDB"


balacij:
```shell
rg "((symbol|term|conceptChunk)|\_?(unit|dataDefn|insmodel|gendef|theoryModel|conceptins|citation|labelledcontent|ref|trace|refby))Table" -ths --sort-files
```

^ This will find all exported record fields and lenses.

We should make sure that all usage of these are for debugging purposes only. All other uses should be re-routed through something else.

Why? Because a `ChunkDB` should not need to be carefully curated to ensure that only $X$ things are used. Re-routing through another chunk will show us why all $X$ chunks should be used together.

PR #4698, opened 2026-01-28
Where `X` is an SRS "Model" type, remove all `findAllX` functions.

balacij:
Contributes to #4300 and #4671

These functions were used exclusively for grabbing all `X`s from the `ChunkDB` when they should have been grabbing them from the `System`. In the future, this code will be refactored again when we work on theories.


JacquesCarette (approved):
Deleting code, my favourite kind of PR.

PR #4704, opened 2026-01-29
Refine more imports related to `Language.Drasil.NounPhrase.Core`.

balacij:
It looks like that module was just re-exporting things from `.Types`.


balacij:
Fixed the conflict.

I'm not against `.Core` nor `.Types`. But I think the Haskell community tends towards `.Internal` for the 'core' modules that expose more than they would want an external user to use normally.

balacij:
Let's continue with `.Core` then.

PR #4706, opened 2026-01-29
Instantiate `HasChunkRefs` for `Citation`, `CitationField` and `CitationKind`.

balacij:
Contributes to https://github.com/JacquesCarette/Drasil/issues/4349.
Builds on https://github.com/JacquesCarette/Drasil/pull/4705.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Chunk/Citation.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Chunk/Citation.hs
@@ -16,7 +16,8 @@ module Language.Drasil.Chunk.Citation (
 
 import Control.Lens (makeLenses, Lens')
 
-import Drasil.Database (HasChunkRefs(..), UID, HasUID(..), showUID, mkUid)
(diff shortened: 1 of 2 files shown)


code/drasil-lang/lib/Language/Drasil/Chunk/Citation.hs
```diff
@@ -43,16 +44,14 @@ data Citation = Cite
   , _citeID   :: UID
   ,  sn       :: ShortName
   }
+declareHasChunkRefs ''Citation
```
balacij, line 47: So, realistically, @JacquesCarette, we can plaster this on all 'atoms' and chunks alike, and it should always keep an up-to-date set of referenced `UID`s, holding true even as developers add more possibilities for chunk references to chunks.

I only hesitate to plaster this everywhere because it is TemplateHaskell, but this might be a pre-mature consideration. Do you have any thought on this?

JacquesCarette: If it becomes a big deal, we can consolidate Chunk declarations into fewer files.

PR #4707, opened 2026-01-29
Moved `spaceToCodeType` into Choices.hs

B-rando1:
Contributes to #4653 

As noted in the issue, the `spaceToCodetype` function seemed quite out-of-place in `PackageFiles.hs`. `Choices.hs` is the only place in `drasil-code` where it is used (it's also used in the examples), so I put it into that file.


JacquesCarette (approved):
Agree it's a good change.

PR #4709, opened 2026-01-29
SWHS&NoPCM: Fix reference for "Heat flux into the water from the coil" GD.

balacij:
(ignore the accidentally highlighted "qc")
Screenshot 2026-01-29 at 11 13 09 AM

As per the original artifacts:

1. SWHS: https://github.com/smiths/swhs/blob/94e2aa9a53b46f4b874fc8812f4be37d9e909c11/docs/SRS/PCM_SRS.tex#L899-L943
2. NoPCM: https://github.com/smiths/caseStudies/blob/b28ba8e6343fb78c7c062bf6faaec39f5aa20b39/CaseStudies/noPCM/docs/SRS/SRS.tex#L874-L901

The correct reference for this GD is `lightstone2012`.


smiths (approved):
I agree with this change.  The previous reference was wrong.  (The Koothoor reference is about fuel pins in reactors, not about SWHS.)  The change is great.

PR #4710, opened 2026-01-29
Use `hibbeler2004` as source for `TM:NewtonSecLawMot` and `TM:NewtonSecLawRotMot`

balacij:
Reason why "both" at the same time: SglPend inherits its references from DblPend.

If we instead prefer to add a citation somewhere, I can do that too, of course. I'm just taking the "easiest" path to dealing with this.


smiths (changes requested):
Rather than lose the Hibbeler citations, I'd like to add the citation as a source for 
TM:NewtonSecLawMot and TM:NewtonSecLawRotMot.  Both of these models are so well known that we didn't include sources  previously, but if someone wants extra information the Hibbeler text is a great source.  (It is the book I used when I was an undergrad and it is still used today.)

balacij:
That sounds like a much better solution, thank you @smiths!

This is now ready for review again.

smiths (approved):
The changes look great.

Issue #4714, opened 2026-01-30
FileData, FileAndContents, PackageFiles


B-rando1:
Today I noticed something interesting.

https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-code/lib/Language/Drasil/Code/FileData.hs#L9
- `FileAndContents` is a path and a doc

---

https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-gool/lib/Drasil/Shared/AST.hs#L59
- `FileData` is a path and a `ModData`
- https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-gool/lib/Drasil/Shared/AST.hs#L75
  - `ModData` is a name and a doc
- So `FileData` is a subtype of `FileAndContents`, modulo some isomorphism

---

https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-code/lib/Language/Drasil/Code/PackageFiles.hs#L21
- `PackageFiles` is a list of pairs of path and doc
- So essentially a list of `FileAndContents`, modulo some isomorphism

---

I feel like at least one of these types is unnecessary! From an in-person chat with @JacquesCarette, `PackageFiles` is on the chopping block. I think it still makes sense to keep `FileAndContents` as the type containing the subset of `FileData` that `drasil-code` needs.

Beyond that, [consolidatePackageFiles](https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-code/lib/Language/Drasil/Code/PackageFiles.hs#L26) is doing some silly things and can be removed or at least heavily reduced. Most of the other functions in that file are likely unnecessary as well.

JacquesCarette:
It might make sense to separate path from file name (and indeed, file name + extension) in our 'raw data'.

We need to better understand this to come up with a reasonable design.

B-rando1:
In my previous comment, I mentioned a discussion where we looked at this line:

https://github.com/JacquesCarette/Drasil/blob/092f539dc74c7b30b47c219264268dc1ddf928fe/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs#L133

[...]

Issue #4625, opened 2026-01-07
Refactoring `Language.Drasil.Code.Imperative.GOOL.Data`


B-rando1:
To make other refactors feasible, I am trying to reorganize some parts of `drasil-code` and `drasil-gool` to get a simpler dependency graph.

From talking with @JacquesCarette yesterday, [Language.Drasil.Code.Imperative.GOOL.Data](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/Data.hs) might be a good place to start. From the [diagram](https://jacquescarette.github.io/Drasil/graphs/drasil-code.pdf) on the website, you can see that it has two categories of modules that use it: all of the renderers, and `Language.Drasil.Code.CodeGeneration`. The hope is that we can move things around so that there aren't two categories of modules using this module.

`Data` exports two things:
- `AuxData`, which holds information about auxiliary files
  - It contains `auxFilePath` to access a file's path and `auxDoc` to access its doc representation.
  - `ad` a constructor for `AuxData`
- `PackData`, which holds information about packages
  - It contains `packProg`, which holds information (from GOOL) about the program; and `packAux`, which is a list of `AuxData` for the package.
  - `packD` is a constructor for `PackData`.

Each of the renderers (e.g. [Python](https://github.com/JacquesCarette/Drasil/blob/cfdeee9df0166fc2b5e99243addea9e6996659af/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/PythonRenderer.hs#L43)) use `AuxData` and `PackData` for implementing their typeclasses.

[CodeGeneration](https://github.com/JacquesCarette/Drasil) uses `AuxData` in the `makeCode` function, where it converts a list of `FileData` and a list of `AuxData` to a `Code` object.

`Generator`'s [generateCode](https://github.com/JacquesCarette/Drasil/blob/cfdeee9df0166fc2b5e99243addea9e6996659af/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs#L117) function uses `PackData` and is also the only function that uses `makeCode`. `generateCode` in turn is called by [genCode](https://github.com/JacquesCarette/Drasil/blob/cfdeee9df0166fc2b5e99243addea9e6996659af/code/drasil-gen/lib/Drasil/Generator/Generate.hs#L203) in `Generate` from `drasil-gen`.

[...]

PR #4721, opened 2026-01-30
Removed PackageFiles.hs

B-rando1:
Contributes to #4714 

I didn't realize the whole file would disappear as a result of these changes!

One small note: I had to add a `drasil-gool` import back into `FileData.hs`, which is sort of a regression from #4643. The reason is that, as noted in #4625, `FileAndContents` _previously_ was only ever used to store SoftwareDossier files. This PR pushes its use higher so that it's used for all kinds of files, including files from GOOL. The import of GOOL is necessary to convert from `drasil-gool`'s representation to `drasil-code`'s, which I think is warranted. Currently that conversion function is only used in one place, but `Main.hs` (the test file) will need it when I clean it up.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/FileData.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/FileData.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE FlexibleContexts #-}
 -- | Defines the underlying data types used in the package extension.
 module Language.Drasil.Code.FileData (FileAndContents(filePath, fileDoc),
   fileAndContents, hasPathAndDocToFileAndContents, PackageData(packageProg,
@@ -14,11 +16,11 @@ data FileAndContents = FileAndContents {filePath :: FilePath, fileDoc :: Doc}
 fileAndContents :: FilePath -> Doc -> FileAndContents
 fileAndContents = FileAndContents
 
-instance HasPathAndDoc FileAndContents where
+instance HasPathAndDoc FileAndContents Doc where
   getPath = filePath
   getDoc = fileDoc
 
-hasPathAndDocToFileAndContents :: (HasPathAndDoc a) => a -> FileAndContents
+hasPathAndDocToFileAndContents :: (HasPathAndDoc a Doc) => a -> FileAndContents
 hasPathAndDocToFileAndContents file = fileAndContents (getPath file) (getDoc file)
 
 -- | The underlying data type for packages in all renderers.
--- a/code/drasil-utils/lib/Utils/Drasil/TypeClasses.hs
+++ b/code/drasil-utils/lib/Utils/Drasil/TypeClasses.hs
@@ -1,9 +1,8 @@
+{-# LANGUAGE FunctionalDependencies #-}
 module Utils.Drasil.TypeClasses (
   HasPathAndDoc(..)
 ) where
 
-import Text.PrettyPrint.HughesPJ (Doc)
-
-class HasPathAndDoc a where
+class HasPathAndDoc a b | a -> b where
   getPath :: a -> FilePath
-  getDoc :: a -> Doc
+  getDoc :: a -> b
--- a/code/drasil-gool/lib/Drasil/Shared/AST.hs
+++ b/code/drasil-gool/lib/Drasil/Shared/AST.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
 module Drasil.Shared.AST (Terminator(..), VisibilityTag(..), ScopeTag(..),
   ScopeData(..), sd, QualifiedName, qualName, FileType(..), isSource,
   Binding(..), onBinding, BindData(bind, bindDoc), bd, FileData(filePath,
@@ -62,7 +63,7 @@ data FileData = FileD {filePath :: FilePath, fileMod :: ModData}
 fileD :: FilePath -> ModData -> FileData
 fileD = FileD
 
-instance HasPathAndDoc FileData where
+instance HasPathAndDoc FileData Doc where
   getPath = filePath
   getDoc = modDoc . fileMod
 


code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -335,6 +338,17 @@ genModulesProc = do
   if con then error "genModulesProc: Procedural renderers do not support classes"
   else return $ mn : inp ++ cal : out ++ moddef
 
+-- | Helper to convert a FileAndContents into a real file with the given document
```
JacquesCarette: This code should be split into
```haskell
createFile :: FilePath -> String -> IO ()
```
that can go into `drasil-utils`, and then the "breaking up" of `FileAndContents` and `render` should happen at the use-site of `createFile`.

There should be no `IO` functions in `drasil-code`.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -128,8 +131,8 @@ generateCode l unReprProg unReprPack g = do
       aux
         | l == Python = fileAndContents "__init__.py" mempty : baseAux
         | otherwise   = baseAux
-      packageFiles = consolidatePackageFiles (progMods $ packageProg $ unReprPack pckg) aux
-  createPackageFiles packageFiles
+      packageFiles = map fileDataToFileAndContents (progMods $ packageProg $ unReprPack pckg) ++ aux
```
JacquesCarette: I think we should keep digging: where is `progMods` defined, and why is it returning what it returns?

code/drasil-code/lib/Language/Drasil/Code/FileData.hs
```diff
@@ -12,6 +14,9 @@ data FileAndContents = FileAndContents {filePath :: FilePath, fileDoc :: Doc}
 fileAndContents :: FilePath -> Doc -> FileAndContents
 fileAndContents = FileAndContents
 
+fileDataToFileAndContents :: G.FileData -> FileAndContents
```
JacquesCarette: We could consider going 'polymorphic' here, by:
1. creating `HasPath` and `HasContents` getters (which would have instances for `G.FileData`, and indeed `FileAndContents` too)
2. making this function take any structure that has those two getters.

That would break the dependency.

code/drasil-utils/lib/Utils/Drasil/FileIO.hs
```diff
@@ -0,0 +1,15 @@
+module Utils.Drasil.FileIO (createFile) where
+
+import Utils.Drasil.Directory (createDirIfMissing)
+import Text.PrettyPrint.HughesPJ (Doc, render)
+import System.FilePath.Posix (takeDirectory)
```
balacij:
```suggestion
import System.FilePath (takeDirectory)
```

[`System.FilePath`](https://hackage-content.haskell.org/package/filepath-1.5.5.0/docs/System-FilePath.html) is more portable cross Posix and Windows.

code/drasil-utils/lib/Utils/Drasil/FileIO.hs
```diff
@@ -0,0 +1,15 @@
...
+import System.IO (hPutStrLn, hClose, openFile, IOMode(WriteMode))
+
+-- | Helper to convert a FileAndContents into a real file with the given document
+-- at the given FilePath
```
balacij:
```suggestion
-- | Write a 'Doc' with default rendering options to a given 'FilePath'.
```

balacij (changes requested):

code/drasil-code/lib/Language/Drasil/Code/FileData.hs
B-rando1: After trying this, I think I like the look of it. See [the new commit](https://github.com/JacquesCarette/Drasil/pull/4721/commits/64bc86ad2ddc3c2892f526142e9f383c6cb17b91).

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
B-rando1: `progMods` is a field of [ProgData](https://github.com/JacquesCarette/Drasil/blob/df48cffb270734451665dad575419f8f42993621/code/drasil-gool/lib/Drasil/Shared/AST.hs#L111), which stores a program's name (`progName :: String`), purpose (`progPurp :: String`), and modules (`progMods :: [FileData]`).

B-rando1: So looking at `pckg`, its type is `packRepr (PackageData ProgData)` (where `packRepr` has some constraints including `Monad`). Passing it to `unReprPack` gives us `PackageData ProgData`, then we project the `packageProg` field of that to get `ProgData`, then we project the `progMods` field of that to get the `[FileData]` that we want to `map` over with our conversion function to `FileAndContents`.

B-rando1: I'm not quite sure what direction you're thinking, but we could try making a `HasFiles` class for everything with a list of files. It would slightly shorten this line and remove some more GOOL imports.

JacquesCarette: That's a good idea.

code/drasil-utils/lib/Utils/Drasil/FileIO.hs
```diff
@@ -0,0 +1,14 @@
...
+import Text.PrettyPrint.HughesPJ (Doc, render)
+import System.FilePath (takeDirectory)
+import System.IO (hPutStrLn, hClose, openFile, IOMode(WriteMode))
+
+-- | Write a 'Doc' with default rendering options to a given 'FilePath'.
+createFile :: FilePath -> Doc -> IO ()
+createFile path contents = do
+  createDirIfMissing True (takeDirectory path)
+  h <- openFile path WriteMode
+  hPutStrLn h (render contents)
```
JacquesCarette: The second argument should be a `String`, and it should be the caller's job to call `render`.

code/drasil-utils/lib/Utils/Drasil/TypeClasses.hs
```diff
@@ -0,0 +1,9 @@
+module Utils.Drasil.TypeClasses (
+  HasPathAndDoc(..)
+) where
+
+import Text.PrettyPrint.HughesPJ (Doc)
+
+class HasPathAndDoc a where
+  getPath :: a -> FilePath
+  getDoc :: a -> Doc
```
JacquesCarette: I'd prefer `getContents :: a -> b` where `HasPathAndDoc` would have 2 parameters.

B-rando1: I made that change now. Looks like I finally got a reason to use those Functional Dependencies that Reed keeps talking about!

JacquesCarette (approved):

PR #4722, opened 2026-01-30
Cleaned up GOOLTest File

B-rando1:
Builds on #4721 (was necessary for some of the functions I used). Might be good to review that one first.

There were some similar improvements in that PR that I wanted to apply to `Main.hs`, so I did them here. Thanks to @TOTBWF for helping me untangle `genCode`!

We also enabled a bunch of language extensions along the way :sweat_smile: hopefully those are good changes as well.

### DIFF
--- a/code/drasil-code/test/Main.hs
+++ b/code/drasil-code/test/Main.hs
@@ -104,5 +104,5 @@ jlClasses unRepr unRepr' = zipWith
 -- | Creates the requested 'Code' by producing files.
 createCodeFiles :: [(Label, D.FileAndContents)] -> IO ()
 createCodeFiles = traverse_ $ \(name, file) -> do
-  let path = name </> D.filePath file
+  let path = name </> D.filePath file -- FIXME [Brandon Bosman, Feb. 10, 2026]: make GOOL allow us to add name to path internally
   createFile path (render $ D.fileDoc file)


code/drasil-code/test/Main.hs
```diff
@@ -62,14 +65,13 @@ main = do
 
 -- | Gathers all information needed to generate code, sorts it, and calls the renderers.
 genCode :: [PackageData ProgData] -> IO()
-genCode files = createCodeFiles (concatMap (\p -> replicate (length (progMods
-  (packageProg p)) + length (packageAux p)) (progName $ packageProg p)) files) $
-    makeCode (map (progMods . packageProg) files) (map packageAux files)
+genCode files =
```
JacquesCarette, line 67: much nicer!

code/drasil-code/test/Main.hs
```diff
@@ -95,31 +96,12 @@ jlClasses unRepr unRepr' = zipWith
...
-    createDirIfMissing False n
-    setCurrentDirectory n
-    createDirIfMissing True (takeDirectory path)
-    h <- openFile path WriteMode
-    hPutStrLn h (render code)
-    hClose h
-    setCurrentDirectory ".."
+createCodeFiles :: [(Label, D.FileAndContents)] -> IO ()
+createCodeFiles = traverse_ $ \(name, file) -> do
+  let path = name ++ "/" ++ D.filePath file
```
B-rando1: Something seems suspicious here. It seems like we should be adding the name to the path much earlier in the pipeline. If you don't mind, I'd like to see if it's a quick fix before we merge.

JacquesCarette: Yes, that is odd. And we should likely use path-manipulation instead of `String`-manipulation (and explicit "/") to do so.

B-rando1: Yes, using `</>` is cleaner and probably safer than `++ "/" ++` :smile: 

Interestingly, `rg '"/"' -ths` shows that we manually insert slashes to separate directories in a good dozen other places already. Might warrant a separate issue...

B-rando1: image

From looking around, it looks like GOOL uses its internal state to create the pathname and doesn't let you customize the directory structure (e.g. by passing an optional intermediate directory, which is what we'd need here). Because the `gooltest` directory has an extra layer of nesting compared to the other examples (see screenshot), `Main.hs` needs to modify the paths after they're generated.

The two important functions seem to be [fileDoc](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/GOOL/InterfaceGOOL.hs#L48) ([Python implementation](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/PythonRenderer.hs#L137-L139) with [polymorphic helper](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs#L546-L554)) and [fileFromData](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/GOOL/RendererClassesOO.hs#L41) ([Python implementation](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/PythonRenderer.hs#L149) with [polymorphic helper](https://github.com/JacquesCarette/Drasil/blob/f3a015a2900334eb6f5f8dc84e83442da8ac819f/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LanguagePolymorphic.hs#L569-L581)). They both just take the name of the module and use that to create the path.

This might be worth looking at later, but it'll require some thinking and restructuring. The current state of this PR seems to be a local maxima in terms of adjusting the pathname in a clean way, so we should be good to merge if there's nothing else that needs to be changed.

TOTBWF: Yeah, after taking a peek myself I think that `PackageData` should probably be storing a data structure that encodes a directory tree rather than a list of paths. I've found myself in a similar situation in `panbench` for writing benchmarks that test multiple files.

JacquesCarette (approved):

PR #4716, opened 2026-01-30
Demo: Switch `Sentence`'s `SyCh` constructor from `UID` to `UIDRef`s.

balacij:
Contributes to #4541.

This PR already needs to be broken up, but it's helpful to have it live as a PR that I can potentially re-purpose.

PR #4725, opened 2026-01-31
Switch `ch` `Sentence` constructor from using `HasUID` to `IsChunk`.

balacij:
Builds on #4723 and #4724

This will simplify work necessary for re-writing https://github.com/JacquesCarette/Drasil/pull/4716.

As a result of this, it was sensible to also replace `HasUID` usage in
`NamedIdea` to `IsChunk` as well (this results in a smaller diff, and is
sensible; `UID`s should be best kept as 'secret' as possible, even
though this doesn't really do that).

### DIFF
--- a/code/drasil-docLang/lib/Drasil/Sentence/Combinators.hs
+++ b/code/drasil-docLang/lib/Drasil/Sentence/Combinators.hs
@@ -26,11 +26,12 @@ import Data.List (transpose)
 import Drasil.Database (IsChunk)
 import Language.Drasil (ConceptChunk, DefinesQuantity(defLhs) , UnitDefn, MayHaveUnit(..)
   , UnitalChunk , HasUnitSymbol(usymb), Quantity, Concept, Definition(defn), NamedIdea(..)
+  , HasSymbol
   , HasShortName(..) , short, atStart, titleize, phrase, plural , Section , ItemType(..), ListType(Bullet)
   , ModelExpr , refS, namedRef
   , Sentence(S, Percent, (:+:), Sy, EmptyS), eS
   , ch, sParen, sDash, (+:+), sC, (+:+.), (!.), (+:), capSent, fromSource, fterms
-  , foldlList, SepType(Comma), FoldType(List), foldlSent , Referable, HasSymbol)
+  , foldlList, SepType(Comma), FoldType(List), foldlSent , Referable)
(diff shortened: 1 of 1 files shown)


samm82 (approved):
Looks good to me!

code/drasil-docLang/lib/Drasil/Sentence/Combinators.hs
```diff
@@ -66,8 +63,9 @@ definedIn' q info = ch (q ^. defLhs) `S.is` S "defined" `S.in_` refS q +:+. info
 definedIn'' :: (Referable r, HasShortName r) => r -> Sentence
 definedIn'' q =  S "defined" `S.in_` refS q
 
--- | Takes a 'Symbol' and its 'Reference' (does not append a period at the end!). Outputs as "@symbol@ is defined in @source@".
-definedIn''' :: (HasSymbol q, HasUID q, Referable r, HasShortName r) => q -> r -> Sentence
+-- | Takes a 'Symbol' and its 'Reference' (does not append a period at the
+-- end!). Outputs as "@symbol@ is defined in @source@".
+definedIn''' :: (Quantity q, Referable r, HasShortName r) => q -> r -> Sentence
```
JacquesCarette: I don't mind not using `HasUID` directly in most code, but replacing `HasSymbol` with `Quantity`, even though it appears that things still work, seem like too radical a narrowing. Doesn't `Quantity` imply quite a few more things, none of which are actually needed?

Why not just change `HasUID` to `IsChunk` and leave the `HasSymbol` alone?

balacij: Looks like I was rushing and didn't think this one through fully -- it can have a broader scope. d620ad4300 fixes this.

JacquesCarette: Yes, but you only did it in this one spot, while the same narrow scope occurred in most of the other files you edited.

balacij: Ah, thank you. [0a46379](https://github.com/JacquesCarette/Drasil/pull/4725/commits/0a4637954e1ec2fa759833b9ef98d96632584016) resolves the remaining ones now.

JacquesCarette (approved):

PR #4727, opened 2026-02-03
Seed basis citations and update case-study references

Xinlu-Y:
- Seed common citations (`cartesianWiki, lineSource, pointSource`) into `basisCDB` so all case studies have them by default.
- Regenerate stable outputs.

Solution for this [comment](https://github.com/JacquesCarette/Drasil/issues/4349#issuecomment-3818534339)


smiths:
> This makes me think of the [Benefits of Drasil](https://github.com/JacquesCarette/Drasil/wiki/Benefits-of-Drasil) page. Is it worth a mention?

I think the problem with using the current example is that @balacij is doing work that could undo this example.  We don't want to draw attention to an example that isn't actually part of our Drasil code.

@B-rando1 thank you for the reminder that we have a list for tracking cool changes that we make in Drasil.  Small changes in the Drasil code that lead to large changes in the generated code.  We should keep looking for cool examples.

PR #4746, opened 2026-02-09
Remove global basis citations and register math-source refs per case

Xinlu-Y:
Previously, #4727 added `cartesianWiki`, `lineSource`, and `pointSource` globally for all cases to avoid missing refs. That fixed missing-reference failures short term, but over-included bibliography entries in unrelated cases.

This pr makes citation inclusion explicit at case level.


code/drasil-gen/lib/Drasil/Generator/BaseChunkDB.hs
```diff
@@ -69,10 +68,11 @@ basisConceptChunks =
...
   --              in the basis.
-  [algorithm, errMsg, program] ++ srsDomains ++ mathcon
-
-basisCitations :: [Citation]
-basisCitations = [cartesianWiki, lineSource, pointSource]
+  [algorithm, errMsg, program] ++ srsDomains ++ basisMathConcepts
+  where
+    -- These concepts carry citation refs in their definitions and should be
+    -- provided explicitly by case studies that use them.
+    basisMathConcepts = filter (`notElem` [cartesian, line, point]) mathcon
```
balacij, line 75: If/when #4700 is merged, I think it would be better to keep `basisCitations`.

We should think of `BaseChunkDB` as a part of a swath of knowledge we "import" in all of our case studies and selectively use. Like importing `numpy` but only using it for `sin`.

In this sense, we really ought to rename `BaseChunkDB` and `basisCDB` to something more like `commonUndergradScienceKnowledge`. `srsDomains` should also be moved to something else, e.g., `smithEtAlBackgroundKnowledge` (I'm not really sure what's in `srsDomains`, so this is taking throwing a shot in the dark based on the name). And then projects should import `commonDrasilCaseStudyBackgroundKnowledge` which merges `commonUndergradScienceKnowledge` with `smithEtAlBackgroundKnowledge`.

JacquesCarette: `commonKnowledge` would be more future-proof.

PR #4750, opened 2026-02-14
Moved `PackageData` and `package` to new file

### LINKED ISSUE #4708: Extract "Package"-related functionality into a common file
As I noted in my [comment](https://github.com/JacquesCarette/Drasil/issues/4653#issuecomment-3818187080) in another issue, a few files are mostly software-dossier related but have a few functions/types related to packages. [FileData](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/lib/Language/Drasil/Code/FileData.hs) and [ClassInterface](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/ClassInterface.hs) are the main examples. There are likely other files in `drasil-code` with bits of package-related functionality as well. I think it makes sense to extract these functions/types into a common file to help separate their purposes.

The only existing file that we could move these into is [PackageFiles](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/lib/Language/Drasil/Code/PackageFiles.hs). It's a bit of a specific part of "package"-related stuff, so I'm torn on whether we should move more stuff here or just create a new file or new files. I'll look at it a bit more and add a comment with the direction I end up leaning.

B-rando1:
Closes #4708 

I _think_ I've convinced myself that this is the right change to make. The resulting `PackageData` file looks a bit odd to me, I think because putting the `PackageData` type in the same file as `package` illuminates how the abstract representation of `PackageData` is only ever concretized in a single way. But maybe it's good that it sticks out, it forces us to consider if this is the best way to express that data structure.

At any rate, this will help with #4653, since we can't separate the `SoftwareDossier`-related functionality while `Package`-related functionality is in the same files.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/PackageData.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/PackageData.hs
@@ -1,5 +1,6 @@
+{-# LANGUAGE PatternSynonyms #-}
 module Language.Drasil.Code.PackageData (PackageData(packageProg, packageAux),
-  package
+  pattern PackageData, package
 ) where
 
 import Language.Drasil.Code.FileData (FileAndContents(..))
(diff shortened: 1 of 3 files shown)


code/drasil-code/lib/Language/Drasil/Code/PackageData.hs
```diff
@@ -0,0 +1,17 @@
+module Language.Drasil.Code.PackageData (PackageData(packageProg, packageAux),
+  package
+) where
+
+import Language.Drasil.Code.FileData (FileAndContents(..))
+import Text.PrettyPrint.HughesPJ (isEmpty)
+import Drasil.GOOL (ProgData, onCodeList)
+
+-- | The underlying data type for packages in all renderers.
+data PackageData a = PackD {packageProg :: a, packageAux :: [FileAndContents]}
```
B-rando1: I know we only just added the type parameter in #4643 (to remove an import of `ProgData` from `Drasil.GOOL`), but since we need `ProgData` anyway for `package`, it might make sense to take it out and have `packageProg :: ProgData`. I tried this locally; it was very easy and didn't affect the imports at all.

JacquesCarette: Sure.

PR #4751, opened 2026-02-16
Implement `HasChunkRefs` for `ConceptInstance`

Xinlu-Y:
contributes to #4349
This PR wires up `HasChunkRefs` for `ConceptInstance` (it was mempty before). Now `ConceptInstance` reports refs from:
- its underlying `ConceptChunk`
- its `ShortName`
- its concept domain (`cdom`)

### DIFF
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
@@ -130,7 +130,7 @@ symbMap = cdbWithRefs cdbRefs
   tMods concIns citations labelledContent
 
 cdbRefs :: [Reference]
-cdbRefs = SRS.sectionReferences ++ [pymunk]
+cdbRefs = pymunk : SRS.sectionReferences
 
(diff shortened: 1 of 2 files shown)


code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs
```diff
@@ -125,9 +125,13 @@ conceptChunks =
   [cw surface]
 
 symbMap :: ChunkDB
-symbMap = cdb symbols ideaDicts conceptChunks [] dataDefs iMods generalDefns
+symbMap = cdbWithRefs cdbRefs
+  symbols ideaDicts conceptChunks [] dataDefs iMods generalDefns
   tMods concIns citations labelledContent
 
+cdbRefs :: [Reference]
+cdbRefs = SRS.sectionReferences ++ [pymunk]
```
JacquesCarette: It's inefficient to add to the tail of a list; if it is not necessary (and it shouldn't be!), then please add to the head, here and in other files.

PR #4756, opened 2026-02-19
Moved `ClassInterface.hs` to `SoftwareDossier` Folder

B-rando1:
Contributes to #4653 

I also renamed a bunch of the contents of the file while I was at it. Most of the renames are pretty straightforward; the only one I wasn't sure about was `auxHelperDoc` → `unReprDoc`. I _think_ the name is accurate, because in every case, `auxHelperDoc` is implemented as the un-representation function for its language. I think it's just a helper function for unpacking data.


code/drasil-code/lib/Language/Drasil/SoftwareDossier/SoftwareDossierSym.hs
```diff
@@ -2,13 +2,13 @@
 
 -- | Defines a package extension for GOOL, with functions for pairing a GOOL
```
JacquesCarette: Comment as to what this file is about should be adapted too.

PR #4757, opened 2026-02-21
Moved FileData into Utils.Drasil

B-rando1:
If we decide this is the right direction to go in, this will contribute to #4653.

To quote my comment from that issue:
> I'm realizing that the use of FileData.hs (and FileAndContents, which it defines) has changed in the three weeks since I put it on the list of files to move into a SoftwareDossier folder. At that point it was generic enough to be used for any file, but since it was only actually used for Software-Dossier files, I was planning on moving it there.
> 
> Now (mainly because of https://github.com/JacquesCarette/Drasil/pull/4721 I believe) it's being used to represent all sorts of files, both Software-Dossier and Code. I think this is a good thing (why have two isomorphic types when you can just have one?), but it also raises the question of where to put this file. It doesn't really belong to either Software-Dossier or Code; it's shared between them (and could potentially be used by other packages). I think drasil-utils might be the package that makes the most sense.

This PR shows what that would look like if we move `FileAndContents` into `drasil-utils`. The diff looks harmless enough. 

Now that I'm looking at it, though, I'm not sure if putting `FileData` in "utils" makes complete sense either. `FileAndContents` has somewhat broad applicability, but I'm not sure if it's broad enough to qualify for "utils". But there are problems with its current location as well.

I can think of a few options:
- Keep `FileData` in the "Code" part of `drasil-code`, and have the "Software-Dossier" part of `drasil-code` import it from there.
- Create another subsection of `drasil-code` for anything shared between the "Code" and "Software-Dossier" parts.
- Move `FileData` into `drasil-utils`, as this PR does. We would then have the option to use it in more packages, which _might_ be useful.
- Create another sub-package of Drasil for "Things that are shared between multiple packages but don't belong in `drasil-utils`" (when I say it like that, this sounds like a bad idea idea :sweat_smile:)

I might be overthinking this :thinking: I guess I'm just trying to get a cleaner mental model of where we should put everything.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code.hs
@@ -46,7 +46,7 @@ module Language.Drasil.Code (
 import Prelude hiding (break, print, return, log, exp)
 
 import Drasil.Code.CodeExpr (field)
-import Language.Drasil.Code.Imperative.GOOL.ClassInterface (
+import Language.Drasil.SoftwareDossier.SoftwareDossierSym (
   SoftwareDossierState, makeSds)
 import Language.Drasil.Code.Imperative.Generator (generator, generateCode,
   generateCodeProc)
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Build/Import.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Build/Import.hs
@@ -12,7 +12,7 @@ import Utils.Drasil (capitalize)
 import Language.Drasil.Code.Imperative.Build.AST (asFragment, DocConfig(..),
   BuildConfig(BuildConfig), BuildDependencies(..), Ext(..), includeExt,
   NameOpts, nameOpts, packSep, Runnable(Runnable), BuildName(..), RunType(..))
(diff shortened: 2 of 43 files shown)


code/drasil-utils/lib/Utils/Drasil.hs
```diff
@@ -29,7 +29,10 @@ module Utils.Drasil (
   toPlainName, repUnd,
 
   -- ** TypeClasses
-  HasPathAndDoc(..)
+  HasPathAndDoc(..),
+  
+  -- ** File representations
+  FileAndContents(..), fileAndContents, hasPathAndDocToFileAndContents
```
JacquesCarette: Other packages should have a single interface, and everything should be imported through that. (We currently cheat a bit, we ought to move towards fixing that.)

`Utils` is different. We should *not* have a `Utils.Drasil` file through which we funnel everything, but get at the pieces directly!  There are no commonalities (or even shared secrets) in `Utils.Drasil` that makes sense to hide -- only within each module, not at the package level.

Upshot: `Utils.Drasil.FileData` should be exported and used directly, and not re-exported in `Utils.Drasil`.

JacquesCarette (approved):

PR #4761, opened 2026-02-23
Implement `HasChunkRefs` for drasil-theory models

Xinlu-Y:
contributes to #4349
I put these in one PR because they’re all part of the same drasil-theory dependency-tracking path.

### DIFF
--- a/code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
+++ b/code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
@@ -29,7 +29,8 @@ data ConstraintSet e = CL {
 makeLenses ''ConstraintSet
 
 instance HasChunkRefs (ConstraintSet e) where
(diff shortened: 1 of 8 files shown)


JacquesCarette (approved):

code/drasil-example/dblpend/lib/Drasil/DblPend/Body.hs
```diff
@@ -134,9 +134,12 @@ conceptChunks =
   [cw len]
 
 symbMap :: ChunkDB
-symbMap = cdb symbols ideaDicts conceptChunks []
+symbMap = cdbWithRefs cdbRefs symbols ideaDicts conceptChunks []
   dataDefs iMods genDefns tMods concIns citations (labelledContent ++ funcReqsTables)
 
+cdbRefs :: [Reference]
+cdbRefs = SRS.sectionReferences
```
balacij, line 141: Do you remember which theory chunks depended on `SRS.sectionReferences` out of curiosity?

This seems like an odd requirement to me.

PR #4766, opened 2026-02-27
SWHSNoPCM: Remove unused citation/bibliography entry.

balacij:
Contributes to #4700

`lightstone2012` is only referenced in SWHS, not SWHSNoPCM. It looks like this was accidentally kept when SWHSNoPCM was created (assuming it was created by copy/pasting SWHS!).


smiths (approved):
Yes, I agree that noPCM should not cite Lightstone.

Issue #4775, opened 2026-03-02
`drasil-lang/../Sentence.hs` contains `headSent`, a Markdown header sentence constructor


balacij:
https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-lang/lib/Language/Drasil/Sentence.hs#L151-L153

## Objective

Jumping to the end, the goal of this issue is to delete `headSent` altogether. It is a duplicate of existing tools in `drasil-printers` and should not be needed by its current use sites.

## Background of `headSent`

It is used in exactly two locations within Projectile's Lesson Plan sub-project:

1. https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/projectile/lib/Drasil/Projectile/Lesson/CaseProb.hs#L30-L34
2. https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/projectile/lib/Drasil/Projectile/Lesson/Review.hs#L15

For both locations, `headSent` is used for its intended purpose, to create section headers. But why is this necessary at all? For our documents, we already have `Section`, for which we can use to create sections within documents (with correctly constructed internal `Reference`s. labels, etc.).

So why don't we use them? Because none of the constructors for the Lesson Plan permit using them.

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/Core.hs#L8-L42

This is further purported through how Jupyter notebooks are created:

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-printers/lib/Language/Drasil/Printing/Import/Document.hs#L22-L27

Note that `makeDocument` uses `createLayout'` (not `createLayout`):

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-printers/lib/Language/Drasil/Printing/Import/Document.hs#L76-L81

`createLayout'` relies on `cel` (not `sec`):

[...]

PR #4776, opened 2026-03-02
Lesson Plan: Permit passing in arbitrary `Section`s through the document layout description.

balacij:
Contributes to #4775

This PR lets us pass in sub-`Section`s into each hardcoded section block of the lesson plan document layout description language.

Recall that said description language was immature as-is:

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/Core.hs#L21

As a result of this PR, we can begin replacing `headSent` usage in raw `Content`s in the lesson plan with `Section`s that automatically create properly formatted headers (i.e., without `headSent`).

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/Core.hs
+++ b/code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/Core.hs
@@ -21,10 +21,10 @@ data LsnChapter = Intro Intro
 -- TODO: Work on detail structure of Lesson Plan
 
 -- ** Introduction
-data Intro = IntrodProg [Contents] [Section]
+newtype Intro = IntrodProg [Contents]
 
 -- ** Learning Objectives
-data LearnObj = LrnObjProg [Contents] [Section]
+newtype LearnObj = LrnObjProg [Contents]
 
 -- ** Review Chapter
 data Review = ReviewProg [Contents] [Section]
@@ -33,13 +33,13 @@ data Review = ReviewProg [Contents] [Section]
 data CaseProb = CaseProbProg [Contents] [Section]
 
 -- ** Examples of the lesson
-data Example = ExampleProg [Contents] [Section]
+newtype Example = ExampleProg [Contents]
 
 -- ** Summary
-data Smmry = SmmryProg [Contents] [Section]
+newtype Smmry = SmmryProg [Contents]
 
 -- ** Appendix
-data Apndx = ApndxProg [Contents] [Section]
+newtype Apndx = ApndxProg [Contents]
 
 -- * Multiplate Definition and Type
 
@@ -65,12 +65,12 @@ instance Multiplate DLPlate where
     lc (Apndx x) = Apndx <$> apndx p x
     lc BibSec = pure BibSec
 
-    introd (IntrodProg con secs) = pure $ IntrodProg con secs
-    lrnObj (LrnObjProg con secs) = pure $ LrnObjProg con secs
+    introd (IntrodProg con) = pure $ IntrodProg con
+    lrnObj (LrnObjProg con) = pure $ LrnObjProg con
     rvw (ReviewProg con secs) = pure $ ReviewProg con secs
     csProb (CaseProbProg con secs) = pure $ CaseProbProg con secs
-    exmp (ExampleProg con secs) = pure $ ExampleProg con secs
-    smry (SmmryProg con secs) = pure $ SmmryProg con secs
-    aps (ApndxProg con secs) = pure $ ApndxProg con secs
+    exmp (ExampleProg con) = pure $ ExampleProg con
+    smry (SmmryProg con) = pure $ SmmryProg con
+    aps (ApndxProg con) = pure $ ApndxProg con
   mkPlate b = DLPlate (b lsnChap) (b intro) (b learnObj) (b review)
     (b caseProb) (b example) (b smmry) (b apndx)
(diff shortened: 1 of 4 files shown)


code/drasil-docLang/lib/Drasil/DocumentLanguage/Notebook/Core.hs
```diff
@@ -21,25 +21,25 @@ data LsnChapter = Intro Intro
 -- TODO: Work on detail structure of Lesson Plan
 
 -- ** Introduction
-newtype Intro = IntrodProg [Contents]
+data Intro = IntrodProg [Contents] [Section]
```
JacquesCarette: I don't see why this is a 'good' solution though: why *add* a `[Section]` ?

balacij: I added it to each "high-level section" scaffold to add arbitrary subsections.

This definitely isn't the optimal solution. It's just a sensible one given that we don't know what types of subsections we might see in each. (Well, we can probably figure that out, but that's a project unto itself that I don't want to start right now.)

Adding in `[Section]` here does defeat the purpose of `SRSDecl` and `LsnDesc` (to capture more semantic structure from the documents) and doing this is only marginally better than writing the Jupyter code immediately ourselves -- better in that we might be able to render in different formats.

If, later, we have more use of the lesson plan structure, we can figure out how to flesh this out properly, with data types "semantically richer" than `[Section]`.

This is also similar to what we do in `SRSDecl` for two areas:

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-docLang/lib/Drasil/DocumentLanguage/Core.hs#L141-L142

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-docLang/lib/Drasil/DocumentLanguage/Core.hs#L156-L159

JacquesCarette: Well, if those are meant to be a sort of 'prelude' paragraph followed by sections, that's not so bad. But `PDProg` seems weird, because should the `[PDSub]` be as documented, *sub-sections*?

I still don't think that deleting the `headSent` hack is outweighed by this hack.

balacij: > Well, if those are meant to be a sort of 'prelude' paragraph followed by sections, that's not so bad.

Yes. And that's what I'm doing in this PR.

> But PDProg seems weird, because should the [PDSub] be as documented, sub-sections?

Yes, `PDSub` captures the 3 major kinds of subsections of the "problem description" subsection of the specific system description (SSD).

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-docLang/lib/Drasil/DocumentLanguage/Core.hs#L161-L168

What's odd here is how we use the list of sections in `PDProg`. We only use it in GlassBR to build an alternative version of the `TermsAndDefs` `PDSub`. I'm not quite sure why this is necessary, but we can file an issue to look into this. Regardless, the `[Section]` allows adding in arbitrary sections to the generated problem descriptions.

> I still don't think that deleting the headSent hack is outweighed by this hack.

I guess I don't see this as a hack at all. Consider `PDSub` here, it is an enumeration of all possible subsections of the "Problem Description" subsection of the "Specific System Description" (SSD). However, we use `[Section]` as a means of allowing us to add arbitrary sections to an SSD. Similarly, with `[Section]`s added in this PR, we allow creating arbitrary sections.  However, the long-term hope here would be that the list of `Section`s would be replaced with something like `[PDSub]`.

It is a hack in a sense that we jump to writing `Section`s rather than filling in boxes (like we currently do with the SRS). But this can/will be fixed when we have a better idea of the kinds of subsections that could appear in section.

Also, using `headSent` forces us into Markdown formatting. We could easily write a generator that exports mdBook (using `Section`s to capture how to break up an mdBook project into multiple files) or LaTeX (again, using `Section`s to capture what content belongs to which [sub]section).

EDIT: One more thing about using `headSent`: Our internal ID/reference system assumes that we placed specific labels. `headSent` creates markdown headers (where there are automatically assigned labels in md->html conversion that are usually of the form "the-header-text-with-dashes", i.e., ones we do not use).

JacquesCarette: I do see `headSent` as a montrous hack and should go.

What I'm not seeing is enough *design* in this PR. It feels like a too much of a "code fix" and not enough of a "let's (re)design `PDSub` and `LsnDesc` and whatever else so we don't need such hacks"

balacij: I definitely agree with about the lack of design in this PR. As the case study is, we just don't have enough information about the sections to know what to put in place of arbitrary `Section`s lists yet.

If we want to be more conservative, we can restrict adding arbitrary `Section`s to only that absolutely necessary (i.e., `CaseProb` and `Review`) for https://github.com/JacquesCarette/Drasil/pull/4778 and https://github.com/JacquesCarette/Drasil/pull/4779 to be merged. However, I added the list of `Section`s to each 'section' constructor so we can gather information for proper design later.

JacquesCarette: Let's indeed go minimalistic.

balacij: Sounds good to me. I just updated this. 

I was off by one 'section'. It was `CaseProb` _and_ `Review`.

PR #4778, opened 2026-03-02
Projectile Lesson Plan: Use `Section`s instead of `headSent`-created headers to create pseudo-sections.

balacij:
Contributes to #4775 

Builds on #4776 (hard requirement) and #4777 (nice-to-have to clean up export list/clean up code).

This PR removes all `headSent` usage in Drasil, which is all currently confined to Projectile's lesson plan.

Issue #3877, opened 2024-07-24
De-duplicate and fix rendering of Jupyter artifacts


BilalM04:
The current Jupyter printer ([`Language.Drasil.JSON.Print`](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-printers/lib/Language/Drasil/JSON/Print.hs)) is heavily duplicated, most of it being a direct copy paste of functions from [`Language.Drasil.HTML.Print`](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-printers/lib/Language/Drasil/HTML/Print.hs). Duplicated functions should be imported rather than copied.

The Jupyter rendering also uses a lot of Hacky HTML rather than pure markdown, specifically tables. With the new Markdown printer, this hacky HTML should be removed by leveraging the new functions in the Markdown printer and helper. Additionally, several components in the Jupyter artifacts are just broken. For example, list indentation, expressions, and captions. 

**Goal**: Fix broken components and reduce code duplication.

JacquesCarette:
Agreed!

PR #4779, opened 2026-03-02
Delete `headSent`.

balacij:
Closes https://github.com/JacquesCarette/Drasil/issues/4775

Builds on #4778 

As of https://github.com/JacquesCarette/Drasil/pull/4778, `headSent` is unused and hence can be deleted, as this does.


code/drasil-example/projectile/lib/Drasil/Projectile/Lesson/Review.hs
```diff
@@ -9,10 +11,18 @@ import Drasil.Document.Contents (foldlSP, foldlSP_)
 import qualified Language.Drasil.Sentence.Combinators as S
 
 reviewContent :: [Contents]
-reviewContent = [reviewHead, reviewContextP1, LlC E.lcrectVel, LlC E.lcrectPos, LlC E.lcrectNoTime, reviewEqns, reviewContextP2]
+reviewContent = []
```
JacquesCarette, line 14: If `reviewContent` is now empty, why is it even needed?

code/stable/projectile/Lesson/Projectile_Lesson.ipynb
```diff
@@ -46,7 +46,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Rectilinear Kinematics: Continuous Motion\n",    "\n"
+    "## Rectilinear Kinematics: Continuous Motion\n",
+                                                                      "<a id=\"Sec:RectKin\"></a>\n",    "\n"
```
JacquesCarette, line 50: So I'm not keen on this change:
1. why is it out-dented?
2. why is it empty?

This is a symptom that something in this fix is a bit too much of a hack, and shows up in this way.

code/drasil-example/projectile/lib/Drasil/Projectile/Lesson/Review.hs
balacij: This was technically code that is already in `main`. It was shown in this PR review because the HEAD was not updated. It's updated now.

However, this comment stands regardless, so I filed #4853.

code/stable/projectile/Lesson/Projectile_Lesson.ipynb
balacij: Unfortunately, all of the formatting in the `.ipynb` file is bad. It can use a review.

As per my previous comment, this is code already in `main`. So, this is out of scope for this PR. I _believe_ this is in scope for the greater review of the `.ipynb` file generator: #3877.

Issue #4368, opened 2025-09-16
Projectile's SRS: Mentions LCs and ULCs, has neither


balacij:
See the [SRS](https://jacquescarette.github.io/Drasil/examples/projectile/SRS/HTML/Projectile_SRS.html).

If there are no (un)likely changes, I don't think we should be discussing them in the SRS. @smiths What do you think?

PR #4783, opened 2026-03-03
remove unneeded unlikely changes

JoeZZG:
#4368
removed `unlikely changes` when unlikely section does not exist

### DIFF
--- a/code/drasil-docLang/lib/Drasil/DocDecl.hs
+++ b/code/drasil-docLang/lib/Drasil/DocDecl.hs
@@ -113,11 +113,13 @@ mkDocDesc sys@SI{_systemdb = db} = map sec where
   sec (GSDSec g) = DL.GSDSec g
   sec (SSDSec (SSDProg s)) = DL.SSDSec $ DL.SSDProg $ map ssdSec s
   sec (ReqrmntSec (ReqsProg r)) = DL.ReqrmntSec $ DL.ReqsProg $ map reqSec r
+  -- sec LCsSec = DL.LCsSec $ DL.LCsProg $ fromConcInsDB likeChgDom
   sec LCsSec = case fromConcInsDB likeChgDom of
-    [] -> error "Missing likely changes"
+    [] -> error "LCsSec was included in the SRS declaration but no likely changes were defined"
     xs -> DL.LCsSec $ DL.LCsProg $ fromList xs
+  -- sec UCsSec = DL.UCsSec $ DL.UCsProg $ fromConcInsDB unlikeChgDom
   sec UCsSec = case fromConcInsDB unlikeChgDom of
-    [] -> error "Missing unlikely changes"
+    [] -> error "UCsSec was included in the SRS declaration but no unlikely changes were defined"
     xs -> DL.UCsSec $ DL.UCsProg $ fromList xs
   sec (TraceabilitySec t) = DL.TraceabilitySec t
   sec (AuxConstntSec a) = DL.AuxConstntSec a


code/drasil-docLang/lib/Drasil/Sections/TableOfContents.hs
```diff
@@ -60,21 +60,24 @@ off the shelf solutions
...
-toToC (ReqrmntSec r)       = mktReqrmntSec r
-toToC (LCsSec lc)          = mktLCsSec lc
-toToC (UCsSec ulcs)        = mktUCsSec ulcs
-toToC (TraceabilitySec t)  = mktTraceabilitySec t
-toToC (AuxConstntSec acs)  = mktAuxConsSec acs
-toToC Bibliography         = mktBib
-toToC (AppndxSec a)        = mktAppndxSec a
-toToC (OffShelfSolnsSec o) = mktOffShelfSolnSec o
+-- Returns 'Nothing' for sections that should be omitted (e.g., empty changes sections).
+toToC :: DocSection -> Maybe ItemType
```
JacquesCarette: While this technically works (same with some code further below), it's too surface a fix.

The correct fix would be to replace the `List` in `LCsProg` and `UCsProg` with NEList from `Data.List.NonEmpty`. That would push the responsibility to the SRS creator to not create an `LCsSec` or `UCsSec` at all if it is empty, instead of patching things up later.

code/drasil-docLang/lib/Drasil/DocDecl.hs
```diff
@@ -103,22 +105,22 @@ data ReqsSub where
...
-  sec (SSDSec (SSDProg s)) = DL.SSDSec $ DL.SSDProg $ map ssdSec s
-  sec (ReqrmntSec (ReqsProg r)) = DL.ReqrmntSec $ DL.ReqsProg $ map reqSec r
-  sec LCsSec = DL.LCsSec $ DL.LCsProg $ fromConcInsDB likeChgDom
-  sec UCsSec = DL.UCsSec $ DL.UCsProg $ fromConcInsDB unlikeChgDom
-  sec (TraceabilitySec t) = DL.TraceabilitySec t
-  sec (AuxConstntSec a) = DL.AuxConstntSec a
-  sec Bibliography = DL.Bibliography
-  sec (AppndxSec a) = DL.AppndxSec a
-  sec (OffShelfSolnsSec e) = DL.OffShelfSolnsSec e
+mkDocDesc sys@SI{_systemdb = db} = mapMaybe sec where
```
JacquesCarette: This should have changed too, i.e. to not need the `Maybe` at all.

JacquesCarette (changes requested):

code/drasil-docLang/lib/Drasil/DocDecl.hs
```diff
@@ -103,7 +104,12 @@ data ReqsSub where
 
 -- | Creates the document description (translates 'SRSDecl' into a more usable form for generating documents).
 mkDocDesc :: System -> SRSDecl -> DocDesc
-mkDocDesc sys@SI{_systemdb = db} = map sec where
+mkDocDesc sys@SI{_systemdb = db} = map sec . filter validSec where
+  validSec :: DocSection -> Bool
+  validSec LCsSec = not $ null $ fromConcInsDB likeChgDom
```
JacquesCarette: Hmm, I think this is a user error, not something that should be done silently, i.e. a user has asked for a 'likely changes' section to be included in the document, but there are none.  Thus I think this should instead throw an error.

JoeZZG: I will make the error messages longer

JoeZZG: I will also comment out the old code instead of deleting them

PR #4788, opened 2026-03-03
Move `MVar` field accessors, constructors, destructors closer to `MVar`'s definition.

balacij:
Moving from a "`Helpers`" file to `Build.Drasil.Make.MakeString`, which is where `MVar` is defined.

### DIFF
--- a/code/drasil-build/lib/Build/Drasil/Make/MakeString.hs
+++ b/code/drasil-build/lib/Build/Drasil/Make/MakeString.hs
@@ -63,3 +63,29 @@ mkImplicitVar = Mv . Implicit
 -- | Constructor for 'Free' variables.
 mkFreeVar :: VarName -> MakeString
 mkFreeVar = Mv . Free
+
+-- | Extracts the variable name from a Makefile variable.
+varName :: MVar -> String
+varName (Free s) = s
+varName (Implicit s) = s
+varName (Os s _ _ _) = s
+
+-- | Extracts information for Windows OS from a variable.
+win :: MVar -> String
+win (Os _ w _ _) = w
+win _ = error "Expected Os Variable"
+
+-- | Extracts information for Mac OS from a variable.
+mac :: MVar -> String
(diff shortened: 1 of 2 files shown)


code/drasil-build/lib/Build/Drasil/Make/MakeString.hs
```diff
@@ -63,3 +63,29 @@ mkImplicitVar = Mv . Implicit
...
 mkFreeVar :: VarName -> MakeString
 mkFreeVar = Mv . Free
+
+-- | Extracts the variable name from a Makefile variable.
+varName :: MVar -> String
+varName (Free s) = s
+varName (Implicit s) = s
+varName (Os s _ _ _) = s
+
+-- | Extracts information for Windows OS from a variable.
```
JacquesCarette, line 73: The 3 functions `win`, `mac`, `linux` seems like a mis-design: shouldn't `Os` have 2 arguments, a `String` and a triple with "constructors" something like `Windows`, `MacOS` and `Linux` ? Seems odd to require so much information that will be used in an OS-specific manner.

But maybe `MVar` is used in other ways as well, so all that information is needed. Nevertheless, this design seems really odd - and since you're editing anyways, may as well look at that too!

balacij: I think `Os` should be generalized to a function that performs different actions based on the host OS.

In general, the `Makefile` AST captured in `drasil-build` is weak and needs to be reviewed. I will write more on it soon ™️.

This PR was just a very easy one to split off from any larger work.

JacquesCarette: If it is an important enabler for more work, rather than just shuffling things around, I'm willing to accept it.

balacij: Yes, it's just for maintainability.

PR #4790, opened 2026-03-04
Fix ambiguous performance NFR wording in GamePhysics example

Purav-Kanda:
Fixes ambiguous wording in the performance non-functional requirement for the GamePhysics example. The previous text used the vague term "comparable". This change replaces it with a parameterized requirement specifying execution time within (target)% of a competitor.

### DIFF
--- a/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
+++ b/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Requirements.hs
@@ -11,6 +11,8 @@ import Drasil.DocLang (mkMaintainableNFR)
 import Data.Drasil.Concepts.Documentation as Doc (body, funcReqDom, input_,
   nonFuncReqDom, output_, physicalConstraint, physicalSim, property, solutionCharSpec)
 
+import Drasil.Sentence.Combinators (addPercent)
+
 import qualified Data.Drasil.Concepts.Physics as CP (collision, elasticity,
   friction, rigidBody, space)
 import qualified Data.Drasil.Concepts.Math as CM (surface)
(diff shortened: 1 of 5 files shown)


smiths:
@Purav-Kanda thank you for your pull request.  I'm glad you are looking to get involved.  :smile:  

I've enabled the workflow for your pull request, but I already know it will fail.  I think it might be helpful for you to see the reason that it fails.  Your changes have modified the generated files for game physics, but you haven't told Drasil about these changes, so the test will fail when it compares to the stable version of our documentation.  Have a look at the Contributor's Guide:

https://github.com/JacquesCarette/Drasil/wiki/Contributor's-Guide

smiths:
@Purav-Kanda thank you for your continued work on this PR.  I've started the workflow to check the changes.  You have updated stable, so I'm optimistic that this one will pass.  (I'm writing this while the checks are running, so I don't know that for sure yet.)  However, even if the checks pass, we won't be able to accept this PR.  You have used the generic wording from the original issue, but "target" and "competitor" are parameters of the generic specification that should be filled in.  Since this is only the game physics example that is effected, we could possibly get away with specific values just for this example.  However, the real goal is to make the generic pattern for the performance requirements part of Drasil's knowledge so that the performance requirements for other examples and future cases studies can benefit from this pattern.  The other examples could just provide values for the parameters of target percentage and competitor (ideally with a reference or a link) and then the corresponding requirement would be generated.  Since you are just starting out, the generic version may be too much of a leap at first.

I'm not sure how @JacquesCarette will feel about fixing one specific example, but why don't you give it a try, and we'll see what he says.  :smile:  For target, let's say 1%.  For the competitor, you'll need to find a game physics library that we could compare to.  There may already be some examples in the game physics generated SRS.  If not, google or ChatGPT should help you.

Purav-Kanda:
Purav Kanda reacted to your message:

smiths:
@Purav-Kanda your changes to stable are too extensive.  There should only be changes limited to the scope of the small part you are changing.  There is a command `make stableize` that could be helpful.

Please mark your PR as draft until you are ready for it to be reviewed.

balacij:
Your latest change looks better now, @Purav-Kanda ! Now you need to run what @smiths was mentioning: `make stabilize`. This will update the `stable/` folder with the updated artifacts. Then, commit and push the results, and the CI looks like it will pass 😄

Issue #4791, opened 2026-03-04
`drasil-theory`: Remove `Theory` typeclass


balacij:
https://github.com/JacquesCarette/Drasil/blob/d9c5a7d1e961b17d9840b79b934842085b2e9e2a/code/drasil-theory/lib/Theory/Drasil/Theory.hs#L19-L28

Removing the above typeclass is the first step towards re-implementing theories in Drasil.

PR #4793, opened 2026-03-04
`*Model` rendering: use `mexpress` to gather expressions for rendering to support multi-expression rendering.

balacij:
Contributes to #4791

Builds on #4795

The goal here is to remove reliance on `invariants` (a function belonging to the `Theory` typeclass that we are trying to remove as per #4791) that carries a list of invariants the theory 'defines'/'associates'/?.

~~Now, as a result of removing reliance on `invariants` in the rendering pipeline, we have one small change: constraints/invariants are rendered with `\land` between all invariants rather than using a stacked view of each individual constriant.~~

~~This should eventually get done properly again. So, this is a side-step, or perhaps a step backwards so we can get rid of `Theory` and begin rebuilding.~~


code/drasil-theory/lib/Theory/Drasil/Theory.hs
```diff
@@ -112,6 +112,7 @@ instance HasShortName       TheoryModel where shortname = lb
 instance HasRefAddress      TheoryModel where getRefAdd l = RP (prepend $ abrv l) (ra l)
 -- | Finds the idea of a 'TheoryModel' (abbreviation).
 instance CommonIdea         TheoryModel where abrv _ = abrv thModel
+instance Express            TheoryModel where express = express . (^. mk)
```
balacij: This follows the other kinds of models we have. Reminder: we will be rebuilding this. This is PR a side-step.

JacquesCarette: But can you explain why `express` is different than `tmDispExprs` ? This is indeed a side-step, and I'm not against, I just want a better understanding first.

balacij: Recall that `express` is a function that shows how arbitrary things (satisfying its typeclass) can be expressed mathematically in a single expression:

https://github.com/JacquesCarette/Drasil/blob/ed53a9db79f039dace418d9155b90f03442baee7/code/drasil-lang/lib/Language/Drasil/ExprClasses.hs#L9

So, with this PR, we are asking the internal `ModelKinds` of a `TheoryModel` to be `express`ed as a single expression.

On the other hand, `tmDispExprs` displays any/all found expressions (which can be >1!) belonging to a `TheoryModel` in a list, which we then vertically stack when rendering.

`tmDispExprs` also pulls directly from two individual theory components, namely: `defined_quantities` (poorly named, it really means all formulas declared in the theory) and `invariants`.

It happens that we only have divergence for a single theory model with this change, and that divergence has to do with an equational constraints model ([TM:equilibrium](https://jacquescarette.github.io/Drasil/examples/ssp/SRS/HTML/SSP_SRS.html#TM:equilibrium)). Internally, said theory is captured with an `EquationalConstraints` model kind, which involves declaring a `ConstraintSet`, which is `express`ed as the following:

```haskell
instance Express e => Express (ConstraintSet e) where
  express = foldr1 ($&&) . map express . NE.toList . (^. invs)
```

i.e., the conjunction of all 'constraints' to form a single constraint.

balacij: To sum up, two differences the `express` route and that previous:

1. Where they pull from (internal `ModelKinds` or directly through `Theory` accessors).
2. How many expressions they're allowed to display (1 or $n$).

JacquesCarette: So the better fix would be to add a method `mexpress` or `expressions` to `Express` that would return the non-empty list.

balacij: Sounds good. Just re-purposed this PR.

balacij:
Ready for review again, @JacquesCarette

PR #4798, opened 2026-03-05
Remove now-unused `invariants` accessor from `Theory`.

balacij:
Contributes to https://github.com/JacquesCarette/Drasil/issues/4791

Builds on https://github.com/JacquesCarette/Drasil/pull/4796 and https://github.com/JacquesCarette/Drasil/pull/4794

With the above PRs, `invariants` is unused and can be removed. Hence, we can remove the `invs` field of `TheoryModel` too.

You might wonder: why not remove all of `Theory` at once? Because the fields used to fill in the accessors of `Theory` are all fields of the `TheoryModel` record. We need to go through each field individually to remove all usage first.


JacquesCarette:
Needs a merge

Issue #4764, opened 2026-02-25
Splitting up DrasilState


B-rando1:
As discussed in @JacquesCarette's [comment](https://github.com/JacquesCarette/Drasil/issues/4653#issuecomment-3948262705) in #4653, we are looking to split up [DrasilState](https://github.com/JacquesCarette/Drasil/blob/0bd4b9e03b67c0eef639845419dd49b329dccccd/code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs#L53) so that we can more cleanly operate on pieces of it without dragging the whole state into every corner of `drasil-code`. We can continue talking about it here.

B-rando1:
Notes from meeting:
- See if VS Code has features to speed up tracking backward and forward references
- Talk to Jason, since he's doing choices-related work as well
- Start by putting all of Choices in a sub-record
- Check out https://github.com/DrasilOrg/Notes/blob/ghrcCodeVariants/projects/Re-architecting%20drasil-code%20for%20new%20generation%20options%202025/notes/VARIABILITIES.md

B-rando1:
I'm still in the process of loading this discussion back into RAM, but I do have a thought to consider.

Iirc before we were mainly talking about splitting up DrasilState so that each piece of it belongs only to the parts of the code that need it. I believe this is a part of it, but we might also want to think of it in terms of different "phases" of choices. This is how @balacij has been framing it (e.g. see [VARIABILITIES](https://github.com/DrasilOrg/Notes/blob/ghrcCodeVariants/projects/Re-architecting%20drasil-code%20for%20new%20generation%20options%202025/notes/VARIABILITIES.md)).

This might help us understand why some choices fit cleanly into the "Code" box or the "Software Dossier" box, while others appear necessarily tied to both. For example, something like [Architecture](https://github.com/JacquesCarette/Drasil/blob/727ddc825b21b218c690417403c7e6247f19f66e/code/drasil-code/lib/Language/Drasil/Choices.hs#L79-L85) seems to only show up in the "Code" section because it's a lower-level choice that only applies to code, whereas something like [genDataDesc](https://github.com/JacquesCarette/Drasil/blob/727ddc825b21b218c690417403c7e6247f19f66e/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs#L467-L474) (which doesn't currently involve choices, but it has a comment that it should eventually) is a higher-level choice that affects both code and software-dossier.

[...]

PR #4802, opened 2026-03-06
DrasilState: Move Choices to sub-record

B-rando1:
Contributes to #4764.

Today I learned a bit about declaration groups in Haskell. Because there are cyclic dependencies in the types involved with `DrasilState`,  it matters where in the code we call `makeLenses`. If we call it between two type declarations, it splits the file into multiple declaration groups, and types from one declaration group can't refer to types from another one. Long story short, that's why I moved all of the `makeLenses` commands beneath the declaration of `DrasilState`.

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Descriptions.hs
@@ -42,8 +42,8 @@ unmodularDesc = do
   g <- get
   let implTypeStr Program = "program"
       implTypeStr Library = "library"
-  return $ show $ sentenceDoc OneLine $ spec (printfo g) $ capSent $ foldlSent 
-    ([S "a", S (implTypeStr (g ^. implType)), S "to"] ++ codeSpec g ^. purpose)
+  return $ show $ sentenceDoc OneLine $ spec (printfo g) $ capSent $ foldlSent
+      ([S "a", S (implTypeStr (g ^. implType)), S "to"] ++ codeSpec g ^. purpose)
 
 -- | Returns description of what is contained in the Input Parameters module.
 -- If user chooses the 'Bundled' input parameter, this module will include the structure for holding the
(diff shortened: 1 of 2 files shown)


code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
```diff
@@ -38,11 +41,32 @@ data SoftwareDossierInfo = SoftwareDossierInfo {
...
+  _date :: String,
+  _logName :: String,
+  _logKind :: [Logging],
+  _dsICNames :: InternalConcept -> Name
+}
+
+makeChoicesInfo :: Modularity -> ImplementationType -> Structure -> ConstantStructure -> ConstantRepr -> MatchedConceptMap
+  -> MatchedSpaces -> ConstraintBehaviour -> ConstraintBehaviour -> [Comments] -> String
+  -> String -> [Logging] -> (InternalConcept -> Name) -> ChoicesInfo
+makeChoicesInfo = ChoicesInfo
```
balacij, line 66: Sorry, I'm a little bit out of the loop. Why are we nesting these options in another record?

JacquesCarette: Because we want to be able to factor `DrasilState` into independent pieces. This is a programming-level means to do so.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
```diff
@@ -38,11 +41,32 @@ data SoftwareDossierInfo = SoftwareDossierInfo {
   _softwareDossierFiles :: [SoftwareDossierFile],
   _sampleData :: [Expr]
 }
-makeLenses ''SoftwareDossierInfo
 
 makeSoftwareDossierInfo :: Verbosity -> [SoftwareDossierFile] -> [Expr] -> SoftwareDossierInfo
 makeSoftwareDossierInfo = SoftwareDossierInfo
 
+data ChoicesInfo = ChoicesInfo {
+  _modular :: Modularity,
```
JacquesCarette: Hmm, I prefer `modular`, `implType` etc to `getModular` and so on as the external interface.  Plus we should be using *classy lenses* here, not just lenses.

To accomplish that:
1. rename the fields (they can even be `_getModular` !)
2. create a class `Choices` with lenses `modular`, etc
3. create an instance for both `ChoicesInfo` and `DrasilState`
4. no need to change the client code!

B-rando1: I switched the code over, so future changes should be easier. The client code is still affected, because we weren't using lenses before for the `Choices`-related fields and now we are.

PR #4807, opened 2026-03-08
Data drasil split

JacquesCarette:
Don't import anything from `drasil-metadata` from the examples, go through `drasil-data` instead. Also adjusts the cabal and stack files accordingly.


code/drasil-metadata/package.yaml
```diff
@@ -28,6 +28,11 @@ library:
   source-dirs: lib
   exposed-modules:
   - Drasil.Metadata
+  - Drasil.Metadata.Documentation
+  - Drasil.Metadata.Domains
+  - Drasil.Metadata.Drasil
+  - Drasil.Metadata.SupportedSoftware
+  - Drasil.Metadata.TheoryConcepts
```
balacij, line 35: I'm perfectly happy with diverging from the single-module-reexporting-everything modules, but I'm wondering what your rationale was. Should we be doing this for more packages, too?

code/drasil-example/glassbr/package.yaml
```diff
@@ -28,7 +28,6 @@ dependencies:
 - drasil-gen
 - drasil-gool
 - drasil-lang
-- drasil-metadata
```
balacij, line 31: Nice!

code/drasil-data/lib/Data/Drasil/Domains.hs
```diff
@@ -0,0 +1,7 @@
+-- | Defines Domains where concepts live
+module Data.Drasil.Domains (
+  civilEng, materialEng, mathematics, physics
+  ) where
+
+-- re-export from Drasil.Metadata
+import Drasil.Metadata.Domains (mathematics, materialEng, physics, civilEng)
```
balacij, line 7: Is there anything from `Drasil.Metadata.Domains` that you wouldn't want re-exported? Or can we export the entire module?

code/drasil-metadata/package.yaml
JacquesCarette: There two kinds of packages that don't need such a single module: 1. 'data' modules, and 2. 'utilities' modules. Modules that provide functionality should do so via as narrow an API as possible.

code/drasil-data/lib/Data/Drasil/Domains.hs
JacquesCarette: Right now, it should re-export everything. But in the future, who knows, we might want to build some stuff first to create a domain, and we won't want to export all that working code. Plus I like the documentation aspect.

Issue #4806, opened 2026-03-07
`drasil-utils` shallow file analysis


balacij:
# Per-File Shallow Analysis

Issues are noted with checkmark lists.

### `Utils.Drasil`

Re-exports all exports from all other modules in `drasil-utils`.

### `Utils.Drasil.Directory`

Contains a single function: `createDirIfMessing`. This is used to guard against errors thrown in debugging logs when we run `make debug`.

### `Utils.Drasil.Document`

Contains `Doc`-related utilities, with 2 particularly dubious ones:

* [ ] `filterEmpty`: Removing "empty" `Doc`s from a list.
* [ ] `listToDoc`: Converts a list of `String`s into a `Doc` of form: `listToDoc [a,b,c,...] == a, b, c, ...`. Why dubious? Because it assumes a specific format, and should likely need an "and" before the last entry. This smells like it should be specialized to `Sentence` (and the [natural] language being generated).

### `Utils.Drasil.English`

Contains two functions:

1. `capitalize :: String -> String`: Capitalizes first character and puts the remaining characters in lowercase.
2. `stringList :: [String] -> String`: Merges a list of items into a comma separated list with an "and" before the last item.

1. [ ] `stringList` could be named better. `commaList`? `commaSepList`? `inlineList`?

### `Utils.Drasil.FileData`

See `.TypeClasses` as well.

Contains a single datatype definition:

```haskell
-- | The underlying data type for auxiliary files in all renderers.
data FileAndContents = FileAndContents {filePath :: FilePath, fileDoc :: Doc}
```

What this looks like is an entry for file generation. What it looks like it really should be: `Map RelativeFilePath Doc`. I'm sure there's more that can be done here, too.

### `Utils.Drasil.FileIO`

Contains a single function: `createFile` that accepts a filepath and `String`, and writes the `String` to said filepath. `String` is likely not the type we want here. We use a lot of `render`s (`Doc -> String`) around our `createFile` usage sites. Perhaps having the type signature as `FilePath -> Doc -> IO ()` would be better.

### `Utils.Drasil.FilePath`

Contains an encoding for a `RelativeFile`. This type is good and should be used more throughout Drasil for file generation.

This file along with `FileData`, `FileIO`, `TypeClasses`, and `Document` should be moved to a new package that all other packages import for anything related to artifact generation. e.g., `drasil-artifacts`.

### `Utils.Drasil.Lists`

Contains a number of functions that operate on lists:

[...]

PR #4812, opened 2026-03-08
Update export lists in `drasil-utils`

balacij:
Similar to #4811, very minor, but done while I was working on #4806


code/drasil-utils/lib/Utils/Drasil.hs
```diff
@@ -1,41 +1,19 @@
--- | Gather Drasil's utility functions and re-export for easy use.
+-- | Re-exports all utilities.
 module Utils.Drasil (
```
JacquesCarette, line 2: We should, in time, get rid of this module entirely.

JacquesCarette (approved):

Issue #4214, opened 2025-06-19
Split compiling generated artifacts into new CI Workflow


balacij:
We should move compiling the generated artifacts into a separate CI workflow that only runs if there are changes to `code/stable/X/src/` folders. This would let us shave off some time on the current workflow (e.g., https://github.com/JacquesCarette/Drasil/actions/runs/15766034428/job/44442573869 could have its  "Install System Requirements" step reduced in time and "Compile generated software artifacts" moved to being performed only when necessary).

A [Workflows matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) might be helpful -- it could show compiling artifacts of each case study individually.

balacij:
They're not a big deal right now -- I had no intention of working on this right now, it was more really just to remember the idea.

Issue #3258, opened 2023-02-01
Save CI minutes: only run Build & Test CI if `code/` has edits in commits


balacij:
Since our “Build & Test” CI workflow only involves work with the `code/` folder, we should only run it in the instances where the `code/` folder is edited.

We might be able to use the [`paths-filter`](https://github.com/dorny/paths-filter) to get it going, since GitHub doesn't have this feature natively.

smiths:
Good idea @balacij.

djarabek:
This was done a while ago in https://github.com/JacquesCarette/Drasil/commit/64e3402960a988270c5a6331115e6a1066feb41d

PR #4816, opened 2026-03-13
Split Build and Deploy CI workflows

Xinlu-Y:
The old single workflow ran everything on every PR, including stuff like Doxygen, 
module graphs, and code analysis that nobody looks at until deployment. So, I Separated the old `Build.yaml` into two workflows: one for PR validation, one for deployment.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -40,7 +40,7 @@ jobs:
       - name: "Install TeX requirements"
         if: ${{ steps.changes.outputs.tex == 'true' }}
         run: |
-          sudo apt-get install -y --fix-missing fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-xetex texlive-luatex
+          sudo apt-get install -y --fix-missing fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-luatex
 
       - name: "Install mdBook"
         if: ${{ steps.changes.outputs.md == 'true' }}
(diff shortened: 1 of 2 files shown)


smiths:
This seems like a great idea to me.

.github/workflows/Build.yaml
```diff
@@ -46,16 +29,30 @@ jobs:
...
+          sudo apt-get update
+
+      - name: "Install core system requirements"
+        run: |
+          sudo apt-get install -y --fix-missing libgmp-dev python3 graphviz g++ default-jdk mono-devel
+
+      - name: "Install TeX requirements"
+        if: ${{ steps.changes.outputs.tex == 'true' }}
+        run: |
+          sudo apt-get install -y --fix-missing fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-xetex texlive-luatex
```
JacquesCarette: why are we installing both xetex and luatex? Do we use both?

Xinlu-Y: > why are we installing both xetex and luatex? Do we use both?

Both xetex and luatex were already in the old YAML, and I carried them over without verifying.
oh..I just checked the codebase and it looks like only lualatex is used.

balacij (approved):
This is a very good change for the CI!

As a next step, we can also split the `Build` workflow into a few more according to what needs to be built: https://github.com/JacquesCarette/Drasil/issues/4214 & https://github.com/JacquesCarette/Drasil/issues/3258

e.g., build software artifacts if there's a change to any `src/` folder in `stable/`, build tex if there's a change to any `.tex` files in `stable/`, etc.

PR #4829, opened 2026-03-18
demo: Switch `Sentence`'s `Ch` constructor's `UID`-based chunk references towards `UIDRef`-based ones.

balacij:
This will remove one of the major uses of `termResolve'` (the infamous issue mentioned in #4671).

However, this work needs to be broken up and better understood. Why are all of these changes necessary? I will attach a review.

(Yes, this fails the CI, I'm aware, that's not the point of this PR.)


code/drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
```diff
@@ -68,7 +69,7 @@ qtoc :: (Quantity (q Expr), MayHaveUnit (q Expr), DefiningExpr q, Concept (q Exp
 qtoc q = CD (codeChunk $ quantfunc q) (expr $ q ^. defnExpr) [] Definition
 
 -- | Constructs a 'CodeDefinition' where the underlying 'CodeChunk' is for a variable.
-qtov :: CanGenCode e => QDefinition e -> CodeDefinition
+qtov :: (Typeable e, CanGenCode e) => QDefinition e -> CodeDefinition
```
balacij, line 72: These `Typeable` constraints are a bit ugly to plaster everywhere. I'm not immediately sure how to deal with them, or if I'm just missing something obvious.

code/drasil-lang/lib/Language/Drasil/Sentence.hs
```diff
@@ -59,7 +59,7 @@ infixr 5 :+:
 data Sentence where
   -- | Ch looks up the term for a given 'UID' and displays the term with a given 'SentenceStyle' and 'CapitalizationRule'.
   -- This allows Sentences to hold plural forms of 'NamedIdea's.
-  Ch    :: SentenceStyle -> TermCapitalization -> UID -> Sentence
+  Ch    :: (TypeableChunk t, Idea t) => SentenceStyle -> TermCapitalization -> UIDRef t -> Sentence
```
balacij, line 62: Again, `Typeable` here is ugly. Need to figure out how to deal with this nicely.

`SentenceStyle` can take on 3 values: Plural, Term, and Short. Plural corresponds to plural-full. We don't support plural-short. This needs to be split into two separate options.

Now, it's because this has the option to output short-form nouns that we put the constraint `Idea` on the chunk (`Idea = NamedIdea + getA`, where `getA` is an accessor for the short-form of a chunk's noun phrase). (explaining this is odd, we need to fix the descriptions). As a result, a lot of functions that previously formed `Sentence`s using this constructor that only required `NamedIdea` constrained needed to be weakened to `Idea`.

code/drasil-printers/lib/Language/Drasil/Printing/Import/Helpers.hs
```diff
@@ -79,15 +79,28 @@ lookupSymb pinfo u = sytyF (pinfo ^. stg) (findOrErr (raw u) (pinfo ^. sysdb) ::
...
+  where c = unhideOrErr ur (sm ^. sysdb)
+
 -- | Look up the plural form of a term given a chunk database and a 'UID' associated with the term.
 lookupP :: PrintingInformation -> UID -> TermCapitalization -> Sentence
 lookupP sm c pCap = resolveCapP pCap $ longForm $ termResolve' (sm ^. sysdb) c
 
+-- | Look up the plural form of a term given a chunk database and a 'UID' associated with the term.
+lookupP' :: (TypeableChunk t, Idea t) => PrintingInformation -> UIDRef t -> TermCapitalization -> Sentence
+lookupP' sm ur pCap = resolveCapP pCap $ c ^. term
+  where c = unhideOrErr ur (sm ^. sysdb)
```
balacij, line 109: On the bright side of this PR, we get to use `unhideOrErr` (which looks for a single chunk `UID` of a specific type) rather than `termResolve'` which tries to cast a chunk to $n$ different chunk types that have terms.

code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs
```diff
@@ -35,9 +36,9 @@ instance HasChunkRefs (ConstraintSet e) where
 -- | Finds the 'UID' of the 'ConstraintSet'.
 instance HasUID        (ConstraintSet e) where uid  = con . uid
 -- | Finds the term ('NP') of the 'ConstraintSet'.
-instance NamedIdea     (ConstraintSet e) where term = con . term
+instance Typeable e => NamedIdea     (ConstraintSet e) where term = con . term
```
balacij, line 39: Again, a wart. What's odd about this is that all types are supposed to have `Typeable` predefined for them by GHC. But GHC yells at us about unresolved constraints if these aren't explicitly here as well.

Also, HLS warns that this is a redundant constraint! However, GHC does not!

code/drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs
```diff
@@ -14,10 +14,11 @@ import Control.Lens ((^.), makeLenses, Lens')
 import Drasil.Database (mkUid, UID, HasUID(..), declareHasChunkRefs,
   Generically(..), IsChunk)
 import Language.Drasil.NaturalLanguage.English.NounPhrase.Core (NP)
+import Data.Typeable (Typeable)
 
 -- | A NamedIdea is a 'term' that we've identified (has a 'UID') as being worthy
 -- of naming.
-class IsChunk c => NamedIdea c where
+class (Typeable c, IsChunk c) => NamedIdea c where
```
JacquesCarette, line 21: This is the oddest `Typeable`, which is definitely going to propagate downstream. Why is this one needed?

code/drasil-lang/lib/Language/Drasil/Sentence.hs
JacquesCarette: This one is more than ugly - if it is needed, it points to a bad design. Maybe in `UIDRef t`?

JacquesCarette (commented):
I'm guessing the problem is deep in the database code.

Rather than `Typeable` down there, I would be tempted to use either `Proxy` or an explicit kind, or something like it. We want to *not forget* the type. `Typeable` lets us forget and then recover later. That's too 'dynamic'.

JacquesCarette:
As I suspected, this is all due to using `Typeable` in `ChunkDB`. As [others suggest](https://stackoverflow.com/questions/27044209/haskell-why-use-proxy) either `Proxy` or `Tagged` might be better.

I'm thinking maybe `Tagged` with DataKind which enumerates the known chunk types that go into ChunkDB. That way you maintain a single table, but with enough information retained about what type went in that you can restore it when you take it out -- without `Typeable`.

balacij:
Now the next issue: we associate multiple chunk types to the same `UID`, but only register one in the `ChunkDB`, but reference each one. Take the latest CI failure for example:

```console
λ ~/Programming/Drasil/Drasil/code/ mvpTypedUIDRefSentenceChs rg "aR\b" -ths drasil-example/glassbr
drasil-example/glassbr/lib/Drasil/GlassBR/LabelledContent.hs
16:import Drasil.GlassBR.Concepts (aR, stdOffDist)
41:  `S.versus` titleize aspectRatio +:+ sParen (short aR)

drasil-example/glassbr/lib/Drasil/GlassBR/Concepts.hs
3:  blastRisk, glaPlane, aR, stdOffDist, idglass,
19:annealed, aR, fullyT, glassTypeFac, heatS, lDurFac, iGlass, lGlass,
25:aR            = commonIdeaWithDict "aR"            (nounPhraseSP "aspect ratio")            "AR"       [idglass]

drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
22:import Drasil.GlassBR.Concepts (aR, annealed, fullyT, glaPlane, glassTypeFac,
316:aspectRatioCon   = cc' aR
```

```console
λ ~/Programming/Drasil/Drasil/code/ mvpTypedUIDRefSentenceChs rg "aspectRatio\b" -ths
drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
199:            aspectRatio, aspectRatioMin, aspectRatioMax]
201:eta, meltFrac, fracMin, consTol, aspectRatio, aspectRatioMin, aspectRatioMax :: DefinedQuantityDict
223:aspectRatio = dqd' (dcc "aspectRatio"
230:   (const $ subMin (eqSymb aspectRatio)) Real Nothing
234:   (const $ subMax (eqSymb aspectRatio)) Real Nothing

drasil-example/swhs/lib/Drasil/SWHS/DataDefs.hs
23:import Drasil.SWHS.Unitals (aspectRatio, coilHTC, coilSA, diam, eta, htCapLP,
174:aspRatQD = mkQuantDef aspectRatio aspRatEq
180:aspRat = ddENoRefs aspRatQD Nothing "aspectRatio" []

drasil-example/glassbr/lib/Drasil/GlassBR/IMods.hs
56:  (qwC aspectRatio aspectRatioConstraint : [qwUC dimlessLoad]) (dqdWr stressDistFac)
66:--   [sy dimlessLoad, sy aspectRatio]
67:strDisFacEq = apply interpZ [str "SDF.txt", sy aspectRatio, sy dimlessLoad]
103:  [qwC aspectRatio aspectRatioConstraint, qwUC $ tolStrDisFac ^. output] tolLoad []
109:tolPreEq = apply interpY [str "SDF.txt", sy aspectRatio, sy sdfTol]

drasil-example/glassbr/lib/Drasil/GlassBR/LabelledContent.hs
17:import Drasil.GlassBR.Unitals (aspectRatio, charWeight, demand, demandq,
41:  `S.versus` titleize aspectRatio +:+ sParen (short aR)

drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
41:plateLen, plateWidth, aspectRatio, charWeight, standOffDist :: UncertQ
67:derivedInsWUncrtn = [aspectRatio]
90:aspectRatio = uq (constrained' (dqdNoUnit aspectRatioCon (variable "AR") Real)

drasil-example/glassbr/lib/Drasil/GlassBR/DataDefs.hs
89:aspRatQD = mkQuantDef aspectRatio aspRatEq
92:aspRat = ddE aspRatQD [dRef astm2009] Nothing "aspectRatio" [aGrtrThanB]
```

Let's zoom in on this bit:

```console
drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
90:aspectRatio = uq (constrained' (dqdNoUnit aspectRatioCon (variable "AR") Real)
```

`aspectRatio` is references in a `Sentence` twice:

```console
drasil-example/glassbr/lib/Drasil/GlassBR/LabelledContent.hs
41:  `S.versus` titleize aspectRatio +:+ sParen (short aR)
```

1. `titleize aspectRatio`
2. `short aR`

Creating `UIDRef`s to each type leads to two `UIDRef`s to different types but the same `UID`:

```console
drasil-example/glassbr/lib/Drasil/GlassBR/Concepts.hs
..., aR, ... :: CI
25:aR            = commonIdeaWithDict "aR"            (nounPhraseSP "aspect ratio")            "AR"       [idglass]
...
drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
..., aspectRatio, ... :: UncertQ
90:aspectRatio = uq (constrained' (dqdNoUnit aspectRatioCon (variable "AR") Real)
```

In the first, `aR` is of type `CI` while the second involves an `UncertQ`. However, they both share a `UID` through their _third_ (!) variant:

```console
aspectRatioCon   = cc' aR
  (S $ "the ratio of the long dimension of the glass to the short dimension of " ++
    "the glass. For glass supported on four sides, the aspect ratio is " ++
    "always equal to or greater than 1.0. For glass supported on three " ++
    "sides, the ratio of the length of one of the supported edges " ++
    "perpendicular to the free edge, to the length of the free edge, is " ++
    "equal to or greater than 0.5")
```

However, only one is ever registered into the `ChunkDB`, but **the one inserted isn't even one of the 3 types aforementioned**, but a `DefinedQuantityDict` (!):

```haskell
--derived inputs with uncertainties and no units
derivedInsWUncrtn :: [UncertQ]
derivedInsWUncrtn = [aspectRatio]

...

derivedInputDataConstraints :: [UncertQ]
derivedInputDataConstraints = derivedInsWUnitsUncrtn
  ++ derivedInsWUncrtn

...

symbols :: [DefinedQuantityDict]
symbols = NE.toList inputs ++ tmSymbols ++ map dqdWr specParamVals ++
  ...
  map dqdWr derivedInputDataConstraints ++
  ...

...

symbolsWCodeSymbols :: [DefinedQuantityDict]
symbolsWCodeSymbols = map asVC (concatMap (\(Mod _ _ _ _ l) -> l) allMods)
  ++ implVars ++ symbols

...

symbMap :: ChunkDB
symbMap = withCommonKnowledge [] symbolsWCodeSymbols ideaDicts cis conceptChunks []
  GB.dataDefs iMods [] tMods concIns citations labCon
```


We need to be able to flatten these 3 pieces of data into just `aspectRatio` so that we always just reference `aspectRatio` of a single type, and then we need to get rid of chunk casting functions so that we never run into these issues where we have a single `UID` refer to 4 (!!!!) different chunk types, but only one inserted into the `ChunkDB`.

Issue #4831, opened 2026-03-23
Proposal: Split `drasil-printers` into `drasil-printing` (core) and `drasil-printers` (format printers)


JoeZZG:
From last group meeting, I was suggested by Dr. Carette to split `drasil-printers` in order to complete `move Language.Drasil.Document and .* to drasil-docLang`  in [refactor wiki](https://github.com/JacquesCarette/Drasil/wiki/Refactors).

I am working on this part of code and plan to split `drasil-printers` into:

- `drasil-printing` — shared infrastructure only (AST, Import/*, LayoutObj, PrintingInformation, Plain printer)
- `drasil-printers` — format printers only (HTML, TeX, JSON, Markdown, Debug)

All modules names stay the same, but the downstream packages only need to update dependency and no import changes.

After the proposed split, `drasil-docLang` depends on `drasil-printing` without pulling in all the format printers, so that I can do my job in refactors

Does this proposal look good? Any advice? Thank you!

PR #4836, opened 2026-03-24
Split drasil printers

JoeZZG:
#4831
`drasil-printing` is a new package, which contains the shared printing infrasturecture:
`Printing.AST`, `Printing.LayoutObj`, `Printing.PrintingInformation`, `Printing.Import.*` (all expression/document converters), `Printing.Citation`, `Printing.Helpers`, `Plain.Print`, `Config`

`drasil-printers` is now smaller and only contains format-specific printers (HTML, TeX, JSON, Markdown and Debug), and depends on `drasil-printing`

No other things changed

### DIFF
--- a/code/drasil-printing/package.yaml
+++ b/code/drasil-printing/package.yaml
@@ -41,7 +41,6 @@ library:
   - Language.Drasil.Printing.Citation
   - Language.Drasil.Printing.Helpers
   - Language.Drasil.Printing.Import
-  - Language.Drasil.Printing.Import.Sentence
   - Language.Drasil.Printing.LayoutObj
   - Language.Drasil.Printing.PrintingInformation
   other-modules:
@@ -51,6 +50,7 @@ library:
   - Language.Drasil.Printing.Import.Helpers
   - Language.Drasil.Printing.Import.Literal
   - Language.Drasil.Printing.Import.ModelExpr
+  - Language.Drasil.Printing.Import.Sentence
   - Language.Drasil.Printing.Import.Space
   - Language.Drasil.Printing.Import.Symbol
   when:
--- a/code/drasil-printers/lib/Language/Drasil/Debug/Print.hs
+++ b/code/drasil-printers/lib/Language/Drasil/Debug/Print.hs
@@ -16,7 +16,7 @@ import Language.Drasil
 import Theory.Drasil
 
 import Language.Drasil.Plain.Print
-import Language.Drasil.Printing.Import.Sentence (spec)
+import Language.Drasil.Printing.Import (spec)
 import Language.Drasil.Printing.PrintingInformation
 
 type UIDMap = Map.Map UID [UID]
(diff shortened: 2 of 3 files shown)


code/drasil-printers/package.yaml
```diff
@@ -36,7 +37,22 @@ library:
...
+  - Language.Drasil.JSON.Helpers
+  - Language.Drasil.JSON.Print
+  - Language.Drasil.Markdown.Config
+  - Language.Drasil.Markdown.CreateMd
+  - Language.Drasil.Markdown.Helpers
+  - Language.Drasil.Markdown.Print
+  - Language.Drasil.TeX.Helpers
+  - Language.Drasil.TeX.Monad
+  - Language.Drasil.TeX.Preamble
+  - Language.Drasil.TeX.Print
```
balacij, line 55: Did these need to be explicitly listed?

code/drasil-printing/package.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - Language.Drasil.Printing.Import.Document
+  - Language.Drasil.Printing.Import.Expr
+  - Language.Drasil.Printing.Import.Helpers
+  - Language.Drasil.Printing.Import.Literal
+  - Language.Drasil.Printing.Import.ModelExpr
+  - Language.Drasil.Printing.Import.Sentence
+  - Language.Drasil.Printing.Import.Space
+  - Language.Drasil.Printing.Import.Symbol
+  - Language.Drasil.Printing.LayoutObj
+  - Language.Drasil.Printing.PrintingInformation
```
balacij: Should `drasil-printing` be doing this much? I thought that these modules were  meant to be kept in `drasil-printers`?

code/drasil-printers/package.yaml
JacquesCarette: They don't, but I do prefer being explicit (as with lists of exports)

code/drasil-printing/package.yaml
JacquesCarette: What I think this does, in part, is to separate the parts that are about specific output formats (kept in `drasil-printers`) from "the rest". So that `drasil-printing` becomes the 'awfully messy' package and `drasil-printers` may be regarded as fairly clear.

Of course, the new `drasil-printing` should not have so many exposed modules! It would be best to see how few of these modules need to be exposed.

balacij: Ah, ok. In that case, should `LayoutObj`, `PrintingInformation`, and `Language.Drasil.Printing.AST` stay in `drasil-printers`?

code/drasil-printers/package.yaml
balacij: I agree with lists of exports (there is a lot more attention/care needed here), but this one is a bit cumbersome if/when we write new modules or erase old. Stack/hpack generates this list for us (in the generated `.cabal` file used to orchestrate compilation).

JacquesCarette: Sure, we can rely on that.

code/drasil-printing/package.yaml
JacquesCarette: Once @JoeZZG trims down the exposed modules, we'll see. I guess one could see now, by staring at the dependency graph of the unsplit version.

Issue #4475, opened 2025-11-20
Switch `Doc` backend: `pretty` -> `prettyprinter`


balacij:
(A suggestion by @JacquesCarette)

We should switch to [`prettyprinter`](https://hackage.haskell.org/package/prettyprinter) (from our current [`pretty`](https://hackage-content.haskell.org/package/pretty)). The replacement looks like it will be fairly 'drop-in'.

It looks like `prettyprinter` has a number of extra features, but one thing in particular stands out: it uses `Text` instead of `String`!

balacij:
Superseded by #4989 .

balacij:
I was wrong -- this was not superseded by #4989 because we have `pretty` usage in more packages than `drasil-printers`.

PR #4847, opened 2026-03-27
init. `drasil-build-artifacts`: a package for building artifacts

balacij:
Contributes to #4475 , #3864 , and #4806

Currently, only contains files moved into it from `drasil-utils`.

Eventually, we should:

1. Move away from `pretty` towards `prettyprinter`, perhaps even re-exporting that entire package through `drasil-artifacts` -- #4475.
2. Create an "artifacts writer" monad that all other renderers use -- #3864.


code/drasil-example/dblpend/stack.yaml
```diff
@@ -10,11 +11,15 @@ packages:
 - ../../drasil-gen
 - ../../drasil-gool
 - ../../drasil-lang
+- ../../drasil-metadata
```
JacquesCarette, line 14: why do we add all transitive dependencies in manually? wouldn't it be better not to and let stack take care of that?

balacij: Stack has no way of locating these packages is the issue (because they are not listed on stackage). Since they are local packages, we have no choice but to add a location for which to find them.

Note that the main Makefile-based build is okay without these changes, but these are necessary if/when we `cd` into a specific drasil sub-project and `stack build` it.

JacquesCarette: Ah ok.

PR #4857, opened 2026-03-31
Fix ODE codegen to output full state vector

Xinlu-Y:
addressed [this issue](https://github.com/Xinlu-Y/Drasil/issues/2)

- Because ODE code generator previously extracted only the first element (`r.y[0]`) of the state vector at each timestep, discarding all other components. 
- Root cause: `appendCurrSolFS` in `ExternalLibrary.hs` hardcoded `idx cs (int 0)`, and the solution list variable reused the dependent variable's type (`Vect Real`) instead of wrapping it as `Vect (Vect Real)`.
- I fixed this by removing the hardcoded index-0 extraction and introducing `solListVar` to give the solution list the correct nested type, so statically-typed languages generate `vector<vector<double>>` / `ArrayList<ArrayList<Double>>` instead of flat types.


code/drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
```diff
@@ -74,8 +74,12 @@ qtov q = CD (codeChunk $ quantvar q) (toCodeExpr $ q ^. defnExpr) [] Definition
...
 odeDef info = CD
-  (codeChunk $ quantfunc $ depVar info)
+  (codeChunk $ quantfunc odeSolList)
   (matrix [odeSyst info])
   (matrix [initVal info]:
     map ($ info) [tInit, tFinal, absTol . odeOpts, relTol . odeOpts, stepSize . odeOpts])
   ODE
+  where
+    dv = depVar info
+    odeSolList = implVarAU' (show $ dv ^. uid) (dv ^. term) (dv ^. defn)
```
JacquesCarette: This `show` is very suspicious. What is going on here? What is the first argument to `implVarAU'`, i.e. what does it mean?

code/stable/glassbr/src/csharp/OutputFormat.cs
```diff
@@ -36,13 +36,17 @@ public static void write_output(Boolean isSafePb, Boolean isSafeLR, double P_b,
         StreamWriter outputfile;
         outputfile = new StreamWriter("output.txt", false);
         outputfile.Write("isSafePb = ");
-        outputfile.WriteLine(isSafePb);
+        outputfile.Write(isSafePb);
+        outputfile.WriteLine("");
```
JacquesCarette: Why are we getting these extra, empty lines?

code/stable/glassbr/src/cpp/OutputFormat.cpp
```diff
@@ -28,12 +28,16 @@ void write_output(bool isSafePb, bool isSafeLR, double P_b, double J) {
     ofstream outputfile;
     outputfile.open("output.txt", std::fstream::out);
     outputfile << "isSafePb = ";
-    outputfile << isSafePb << std::endl;
+    outputfile << isSafePb;
+    outputfile << "" << std::endl;
```
JacquesCarette: why the extra empty string?

code/drasil-code/lib/Language/Drasil/Chunk/CodeDefinition.hs
Xinlu-Y: The firs argument is the chunk/UID name, not an arbitrary display string. I agree using the generic `Show` rendering of a UID here was incorrect. I've changed this to `showUID dv`.

balacij: Should `implVarAU'` be taking a `UID` instead, perhaps? It would be bad if we were doing `UID -> String -> UID`. Then again, re-using `UID`s would also be bad.

Xinlu-Y: That makes sense. I agree this constructor should take a `UID` directly rather than relying on UID text.

Issue #4804, opened 2026-03-07
`drasil-system` shallow file analysis


balacij:
# Per-File Shallow Analysis

Issues are noted with checkmark lists.

### `Drasil.System`

Contains the grand `System`' data type that contains all data necessary to build various different `SystemKind`s, including:

1. Just an SRS.
2. An SRS along with a code solution. 
3. A "notebook."
4. A website.

The `System` datatype contains two hacks (`refbyTable` and `traceTable`) that are exclusively used for SRS generation. The other major hack (`refTable`) only exists there because we duplicate `UID`s in the `Reference` table and other chunks in the `ChunkDB`).

* [ ] `System` really looks like it wants to be a typeclass and for each `SystemKind` to be a datatype unto itself that just captures what is _absolutely necessary_ for the respective generators to do their work.
* [ ] An alternative name for `System` might be `Project`. It might be fruitful to think of `System` as an abstract git repository (for us to give it a meaningful definition).

balacij:
I'm going to pause work on this until the above linked PRs are resolved. If merged, my plan is as follows:

1. Collapse `OldCodeSpec` into `CodeSpec`. There was intention to refactor this and it doesn't quite look like that's been happening. Merging them is better for the next few steps. -- #5072 
2. Mark `CodeSpec` as a kind of "system" and figure out if it truly needs to carry an entire `SmithEtAlSRS` (currently named `System`).
3. If it needs to carry an entire `SmithEtAlSRS`, figure out how to split it into $n$ phases so we have "systems" that are pure code without an SRS.
4. Move code related to the `SmithEtAlSRS -> CodeSpec` generation phase to `drasil-gen`, including the ICO solution search. Perhaps libraries, too.
5. MAYBE:
   1. Build a "raw artifacts" system. This is a "system" that contains purely a `Map RelativeFile Document` (in addition to the normal system meta-information).
   2. Re-route `IO ()` "generation" phases through the "raw artifacts" system.
6. Dissect `CodeSpec` into $m$ phases (as per the analysis we've started in the DrasilOrg/Notes repo).

balacij:
The original work is completed.

PR #4862, opened 2026-04-05
Initialize `DrasilWebsite` system

balacij:
Builds on #4861 

Contributes to #4804 

What creating this system variant teaches me is that our encoding of the website is weak. Realistically, this could have been a "RawArtifacts" system, but I kept it specialized to our website so that we know it needs to be fleshed out more. Fleshing it out can begin with enumerating a list of the examples to be highlighted on the website.


balacij:
Thanks! Merged in `main` now. Ready for re-review.

It is indeed a sideways movement.

I'm completely fine with redesigning later.

PR #4877, opened 2026-04-08
Collapse `PrintingConfiguration` into `PrintingInformation`.

balacij:
Why? There was only 1 option. We already have options in `PrintingInformation` (e.g., `Stage`). We realistically need to entirely re-build these types anyways. Downsizing before re-designing is helpful.


JacquesCarette (approved):
Normally, I would think that splitting things up into tiny pieces would actually help a re-design even more... But I can see how this change does make sense.

Issue #1697, opened 2019-07-11
Generate Input/Output Requirements from SystemInformation


samm82:
As mentioned in the meeting this morning, the requirements for inputs and outputs should be generated automatically from the inputs and outputs contained in the `SystemInformation` for an example.

smiths:
Yes, this is a good idea.  As we discussed, we essentially have the same information in two places.  Eventually it makes sense to me that the inputs and outputs aren't part of the System Information, but part of the requirements.  This doesn't really matter though, as long as we have the information in Drasil; we can worry about the best way to say it later.  Since the System Information already knows the inputs (for code generation purposes), we shouldn't have to repeat this information again in the requirements.

samm82:
A potential issue I see with this is when we want to reference the input requirement, we will probably have to use a "dummy" reference, but I'll see if/when that comes up.

samm82:
@smiths What should be the cutoff for if the list of inputs/outputs gets put in a list vs. a table? I think five quantities is still small enough to work in a list and any more should be in a table, but it's up to you.

smiths:
@samm82, if it is easier, I'm fine if the lists always go into a table.  If you decide to provide both options, then the cut-off between lists and a table should be an easily customizable number.  This would be a parameter of variation that the user could determine.  If you are looking for a default value for this parameter, yes, I think 5 is a good choice.  (You wouldn't likely want to go higher than that.)

samm82:
Always having a table would be easier (for now, at least) and would have cleaner code so I'll go that route. 😁 We can always add the extra functionality if we desire it later.

smiths:
Sounds good.  Yes, once the information is in Drasil, we can refine and improve the format and presentation, and flexibility.

samm82:
I'm also thinking we should standardize a bit - should we use "parameters", "quantities", or "values" in the I/O requirements (and in the rest of them as well)? @smiths

smiths:
@samm82, I don't understand your question.  Are you talking about the text of the requirements?  If it is the text of the requirement, I think we can say we are reading "input values".  They aren't variables; they are values.  Parameters to me implies a function interface.  I don't really like "quantities" as a standard terminology.

samm82:
There is great variation of the Requirements section when referring to values: 
- "Output-Quantities: Output the quantities from Table:ReqOutputs." (SWHS)
[...]

PR #4883, opened 2026-04-13
Remove `findAllLabelledContent` by manually aggregating them instead

### LINKED ISSUE #4712: `Drasil.Database.SearchTools`: Remove `findAllLabelledContent`
Spawned from #4300

`findAllLabelledContent` is used in exactly two spots:

1. Line 145: https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-docLang/lib/Drasil/DocumentLanguage.hs#L130-L152
  For building a `Reference` table.
2. For `mdBook` generation, all external assets must be enumerated for the compiler to allow itself to open said files, so we use `findAllLabelledContent` to build a list of all used and necessary for the SRS:
  https://github.com/JacquesCarette/Drasil/blob/a415dcc1bd8a1c800c50dffc6c316dd64ccea6d6/code/drasil-printers/lib/Language/Drasil/Markdown/Config.hs#L40-L45

There are at least two ways we can go about removing `findAllLabelledContent` by 'building' our own list of used `LabelledContent`:

1. Manually: building a list of all needed `LabelledContent`, putting them in `System`, and using said list.
2. Automatically: scanning the generated documents to build a list of used `LabelledContent`.

I tend towards (1) if only because the work involved with #4699 (i.e., #4700) has become a bit messy. I believe we need a more robust system before going towards further automation.

Reminder: Why are we looking to remove `findAllLabelledContent`? Because it is a kind of meta-knowledge. None of our renderers should be using it unless their sole purpose is to _analyze_ a `ChunkDB`.

balacij:
Closes #4712

A list of all `LabelledContent` is needed for the SRS-mdBook generator. Realistically, this is needed for all mdBook generator (hint-hint, this code is not perfect). Since we only need the mdBook generator for the SRS (and the only one we generate some L-C-s for), I added a list of all `LabelledContent` to the `SmithEtAlSRS` system type. This will need to be *manually* filled in for each of our case studies.

Hence, this is not a perfect solution. See https://github.com/JacquesCarette/Drasil/issues/4712#issuecomment-3828675046 : this is definitely not meant to be permanent. Long-term, we want this to be automatically aggregated. All external-Figure-typed `LabelledContent` that *were rendered* (and hence that an mdBook project requires) that is.


code/drasil-example/projectile/lib/Drasil/Projectile/Body.hs
```diff
@@ -130,7 +130,10 @@ si = mkSmithEtAlICO progName
   [purp] [background] [scope] [motivation]
   tMods genDefns dataDefs iMods
   inputs outputs (map cnstrw' constrained) constants
-  symbMap allRefs
+  labelledContent' symbMap allRefs
+
+labelledContent' :: [LabelledContent]
+labelledContent' = labelledContent ++ funcReqsTables
```
balacij, line 135: This piece of code is bizarre. Why didn't I add `funcReqsTables` to `labelledContent`, no? It creates a dependency cycle!

I will spare the code-heavy explanation of why this happens and focus on why we needed `funcReqsTables` to begin with. When we have examples with $n > m$ inputs/outputs, it is reasonable to move them into a table and refer to said table from the corresponding input/output requirement. We have a constructor `inReqWTab` that creates a paired `LabelledContent` and `ConceptInstance` (totally reasonable piece of code to have) that does this.

On the bright side, now that we have `SmithEtAlSRSICO` (the constructor), we can associate certain kinds of functional requirements automatically rather than having to manually build them for each example! For the situations where there are many input/output calculations and we want to move them into a table referenced in the corresponding input/output requirement (`ConceptInstance`), we should have our doc generator _know when it is appropriate_ to create a table or not. So, we should be able to erase `funcReqsTables` entirely!

balacij: Aha! I encountered this before when (coincidence) I was dealing with functional requirements: https://github.com/JacquesCarette/Drasil/pull/4465#discussion_r2532534945

It also looks like I was culprit for this dodgy code! Time for a new issue.

balacij: Or not. #1697 already exists 😄

JacquesCarette (approved):
On a surface level, this feels like a serious step backwards! But I'll proceed anyways, as this also seems like something preventing progress.

PR #4895, opened 2026-04-14
Don't use `unit_symb` for creating numeric ranges in `Sentence`s.

balacij:
`unit_symb` very clearly exists for a good reason: getting the unit symbol of a quantity. It has one strict requirement, that the input quantity _always_ has a quantity. However:

1. As we've recently been discussing, units should be more tightly integrated with types.
2. `unit_symb` is used exclusively in creating numeric ranges. We ought to have a better way of doing that (and in a way that adds the unit symbols automatically, enforcing that both sides are the same, further using the unit of the thing it is creating a range for).

### DIFF
--- a/code/drasil-example/swhs/lib/Drasil/SWHS/Assumptions.hs
+++ b/code/drasil-example/swhs/lib/Drasil/SWHS/Assumptions.hs
@@ -10,16 +10,15 @@ import qualified Language.Drasil.Sentence.Combinators as S
 
 import Data.Drasil.Concepts.Documentation (system, simulation, model,
   problem, assumpDom)
-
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):
We should rethink this. I kind of like "this has units of temperature" as more semantic that "this has centigrade as units". But that should be part of larger rethink, being uniform about this is better for now.

smiths:
Similarly, "this has units of length (distance?)" is better than "this has units of m".  I don't want us to invest a lot of effort in imperial units, but it would be nice to be able to easily switch from m to ft.  We also have to worry about the case where someone wants to display in mm instead of m.

PR #4900, opened 2026-04-15
Make `newtonLawNote` accept more than just `UnitalChunk`s as quantity input being 'found'.

balacij:
Ulterior motive: seeing where `UnitalChunk` is absolutely necessary. It's not in this case, and this note can be simplified to make it more readable than reusable.

### DIFF
--- a/code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
+++ b/code/drasil-example/swhs/lib/Drasil/SWHS/GenDefs.hs
@@ -78,7 +78,7 @@ htFluxPCMFromWaterQD = mkQuantDef htFluxP htFluxPCMFromWaterExpr
 htFluxPCMFromWaterExpr :: ModelExpr
 htFluxPCMFromWaterExpr = sy pcmHTC $* (apply1 tempW time $- apply1 tempPCM time)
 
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
We should, of course, be moving in the opposite direction, i.e. of more reusable knowledge. But any abstraction that is used only once is not (yet) a good target for that. We should be abstracting from real examples (i.e. 3+ uses) not hoped-for examples.

JacquesCarette:
Build failed - apparently there is another use?

balacij:
Looks like I was wrong, and it was used more than once (within the same file, too!). I think I had a bad cache somehow. It worked just fine on my laptop (and on another `make`), but after a `git clean -fdx`, it failed!

That being said, I will re-purpose this PR. The `newtonLawNote` combinator can be further parameterized. Should units be associated with the input quantity? Yes, but we don't require this anywhere else, so this `UnitalChunk`-as-a-requirement here does not fit everything else we do.

I'm very skeptical of `UnitalChunk`. It looks like it needs to be re-designed. I understand the point of wanting to enforce that some quantities have units, but we don't make use of that information tangibly in any way, really. And our "quantities" are really ripe for being re-built after our most recent discussions. So, I think we're on a path to deleting them (`UnitalChunk`s that is) right now.

JacquesCarette (approved):

PR #4902, opened 2026-04-16
Add .utf8 encoding to new String calls in Swift generator.

balacij:
Newer versions of swiftc will give large "ambiguous encoding" warnings when building the Swift code unless the encoding is explicitly noted.

This PR adds it. That being said, I'm not sure I took the right approach.


JacquesCarette (approved):
Not sure either, but good enough for what we need right now.

JacquesCarette (approved):

PR #4909, opened 2026-04-22
Generate Functional-style Vector Library as GOOL test in `drasil-code`

balacij:
I was very Java-brained with this code and only tested the Java code. The other ones are almost certainly broken, at the very least because `.clone` does not exist for their array types.


code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+
+thisV :: OOVariableSym r =>SVariable r
+thisV = objVarSelf localV
+
+-- Adapted from GProc's arrayElem
+arrayElem :: OOProg r => SVariable r -> SVariable r -> SVariable r
+arrayElem vr' ir' = do
+  vr <- vr'
+  ir <- ir'
+  var (variableName vr ++ "[" ++ variableName ir ++ "]") double
```
balacij: I don't see anything related to array accessing. Only list accessing. So I had to hack this in using GProc's `arrayElem` as a base.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+    forRange i (litInt 0) (objMethodCallNoParams int (valueOf v1) "dimension") (litInt 1) (bodyStatements [
+      arrayElem res i &= (listAccess (valueOf res) (valueOf i) #+
+                     listAccess (valueOf (objVar v2 localV)) (valueOf i))
+    ]),
+    returnStmt (newObj (obj "Vector") [valueOf res])
+  ]
+  where v1 = vecVar "v1"
+        v2 = vecVar "v2"
+        res = var "res" (arrayType double)
+        i = var "i" int
```
balacij: `add` is a `static` method. This is in contrast with `scale`, which non-static.

We can figure out what the right API is for this class later.

I think we should have two variants of each vector operation doing mutation or returning anew. I mixed these two here with `add` and `scale`.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+    ]),
+    returnStmt (newObj (obj "Vector") [valueOf res])
+  ]
+  where s = var "s" double
+        res = var "res" (arrayType double)
+        i = var "i" int
+
+print_ :: OOProg r => SMethod r
+print_ = docFunc "Prints the vector elements to console." [] Nothing $
+  pubMethod "print" void [] $ oneLiner $ printLn (libFuncApp "java.util.Arrays" "Arrays.toString" string [valueOf thisV])
```
balacij: Another Java-brained function that does not exist in the other languages, but the code generator is very happy to do regardless.

code/drasil-code/package.yaml
```diff
@@ -50,14 +50,6 @@ executables:
...
     main: Main
     source-dirs: test
-    other-modules:
-    - FileTests
-    - HelloWorld
-    - Helper
-    - GOOL.Observer
-    - GOOL.PatternTest
-    - VectorTest
-    - NameGenTest
```
balacij: `hpack` does the heavy lifting here. We should use it.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+
+      varDecDef (var "m" double) mainFn (objMethodCall double (valueOf v1) "magnitude" []),
+      printStr "Magnitude of v1: ", printLn (valueOf $ var "m" double),
+
+      varDec (var "vAdd" (obj "Vector")) mainFn,
+      var "vAdd" (obj "Vector") &= funcApp "Vector.add" (obj "Vector") [valueOf v1, valueOf v2],
+      printStr "v1 + v2: ", valStmt $ objMethodCallNoParams void (valueOf (var "vAdd" (obj "Vector"))) "print",
+
+      varDec (var "vUnit" (obj "Vector")) mainFn,
+      var "vUnit" (obj "Vector") &= objMethodCall (obj "Vector") (funcApp "Vector.add" (obj "Vector") [valueOf v1, objMethodCall (obj "Vector") (valueOf v2) "scale" [litDouble 2]]) "norm" [],
```
balacij: See `Vector.add`. This is a hack.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+        methods = [dimension, magnitude, norm, dot, add, scale, print_]
+
+vecVar :: (VariableSym r, OOTypeSym r) => String -> SVariable r
+vecVar v = var v (obj "Vector")
+
+localV :: VariableSym r => SVariable r
+localV = var "v" (arrayType double)
+
+thisV :: OOVariableSym r =>SVariable r
+thisV = objVarSelf localV
```
balacij: I don't like that the scope that these are in. This seems odd to me, but it does clean up the code. It would be nice if Drasil could "know" that `v` belongs to the `Vector` class.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+
+vectorClass :: OOProg r => SClass r
+vectorClass = docClass "Vectors of doubles and common vector-related operations." $
+  extraClass "Vector" Nothing [stateVar private dynamic localV]
+  [docFunc "Construct a vector from an array of doubles." ["The doubles."] Nothing $
+   constructor [param argV] [] (body [
+     block [
+       assert (valueOf (objVar argV (var "length" int)) ?> litInt 0)
+              (litString "Vector dimension must be > 0."),
+       objVarSelf localV &= objMethodCallNoParams (arrayType double) (valueOf argV) "clone"
```
balacij: `.clone()` is Java-specific. We need an array cloning function from GOOL.

balacij (commented):
It works for Java, but it ain't pretty yet.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
+{-# LANGUAGE PostfixOperators #-}
+module OOVector (ooVector) where
+
+import Drasil.GOOL hiding (arrayElem)
```
balacij: This version of `arrayElem` only deals with a specific `Integer`, not arbitrary expression inputs for the index. That's why we needed our hacky `arrayElem`.

code/stable/gooltest/python/OOVector/OOVector.py
```diff
@@ -0,0 +1,84 @@
...
+    
+    ## \brief Calculate the dot product of two vectors.
+    # \param v1 First vector.
+    # \param v2 Second vector.
+    # \return The dot product.
+    def dot(self, v1, v2):
+        assert v1.dimension() == v2.dimension(), "Vector dimensions must match for dot product."
+        res = 0.0
+        for i in range(0, v1.dimension(), 1):
+            res = res + v1.v[i] * v2.v[i]
```
balacij: Do we have a relative addition operator @B-rando1 ? I couldn't find one.

code/stable/gooltest/swift/OOVector/main.swift
```diff
@@ -0,0 +1,108 @@
...
+    func print() -> Void {
+        print(Arrays.toString(self.v))
+    }
+}
+
+var ds1: [Double] = [1.0, 2.0, 3.0]
+var ds2: [Double] = [4.0, 5.0, 6.0]
+var v1: Vector = Vector(ds1)
+var v2: Vector = Vector(ds2)
+print("v1: ", terminator: "")
```
balacij, line 90: We should consider generating a `DrasilUtils` file that masks some of noisy code like this one.

balacij:
I will work on the design of the vector class later, but I think this is a good point for review.

code/drasil-code/test/OOVector.hs
B-rando1: Is type inference what you're getting at?

B-rando1:
This makes me think of #3991, particularly the "missing features" section. In some ways it feels like the initial work on GOOL involved a "best guess" of the syntax that would be most helpful for us. Some of the less good designs have stuck around since then, because we've mainly focused on extending GOOL rather than refining it.

TL;DR - I suspect we might need to redesign some parts of the interface (e.g. `arrayElem`) if they're not working for us.

code/drasil-code/test/OOVector.hs
B-rando1: Is there a reason you want to use arrays instead of lists? Iirc GOOL's lists are implemented as arrays more often than not.

Either way, I agree that `arrayElem` is odd and should probably be fixed.

B-rando1: I did some more digging (apparently we don't have any test cases for `arrayType` in GOOL, only `listType`!), it looks like in Java, `array`s are translated to static C-style arrays and `list`s are translated to `ArrayList`s.

code/drasil-code/test/OOVector.hs
B-rando1: So the good news is that there is a way to tell GOOL that a method is static when you're creating it in a class: the `Permanence` argument of `method`. The bad news:
- Half the renderers (e.g. Python) ignore the `Permanence` argument and assume dynamic. We'll need to modify their implementation of `intMethod` (this seems to be the place, at least for Python) to add support for this.
- We don't seem to have a way to tell GOOL that a method is static when we _call_ it. We might need to create a family of `classMethodCall` functions to mirror the `objMethodCall` family of functions.

code/drasil-code/test/OOVector.hs
JacquesCarette: Why? I think you might be writing code at too low a level here.

What function do you actually need here? What are you trying to accomplish? Good GOOL code is more intentional than directly executable.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+              (litString "Vector dimension must be > 0."),
+       objVarSelf localV &= objMethodCallNoParams (arrayType double) (valueOf argV) "clone"
+     ]
+   ])]
+  methods
+  where argV = var "v" (arrayType double)
+        methods = [dimension, magnitude, norm, dot, add, scale, print_]
+
+vecVar :: (VariableSym r, OOTypeSym r) => String -> SVariable r
+vecVar v = var v (obj "Vector")
```
JacquesCarette, line 27: I'm concerned about the exact word 'Vector' here. Feels like it's either duplication or wrong.

code/drasil-code/test/OOVector.hs
JacquesCarette: At least `thisV` is not exported from this file, so its scope isn't as bad as it seems.

But both Brandon and I could be misinterpreting your meaning!

code/drasil-code/test/OOVector.hs
JacquesCarette: This is definitely a horrible hack. There should be a concept of "indexing" that should map to what you want. I do know indexing is under-supported though.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+arrayElem :: OOProg r => SVariable r -> SVariable r -> SVariable r
+arrayElem vr' ir' = do
+  vr <- vr'
+  ir <- ir'
+  var (variableName vr ++ "[" ++ variableName ir ++ "]") double
+
+dimension :: OOProg r => SMethod r
+dimension = docFunc "Returns the dimension of this vector." [] (Just "The dimension of the vector.") $
+  method "dimension" public dynamic int [] $ bodyStatements [
+    returnStmt $ valueOf $ objVar thisV (var "length" int)
```
JacquesCarette: The string "length" should not be embedded here.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+  ]
+
+magnitude :: OOProg r => SMethod r
+magnitude = docFunc "Calculate the Euclidean norm (magnitude) of this vector."
+  [] (Just "The magnitude.") $
+  pubMethod "magnitude" double [] $ oneLiner $
+    returnStmt (funcApp "Vector.dot" double [valueOf self, valueOf self] #/^)
```
JacquesCarette: You should not hard-code dot as accessor.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+
+norm :: OOProg r => SMethod r
+norm = docFunc "Calculate unit vector of this vector."
+  [] (Just "A new unit vector.") $
+  pubMethod "norm" (obj "Vector") [] $ bodyStatements [
```
JacquesCarette: lots of hard-coding of string Vector. And scale. And magnitude.

code/drasil-code/test/OOVector.hs
```diff
@@ -0,0 +1,149 @@
...
+        i = var "i" int
+
+add :: OOProg r => SMethod r
+add = docFunc "Calculate the resultant vector of two vectors."
+  ["First vector.", "Second vector."] (Just "The resultant vector.") $
+  method "add" public static (obj "Vector") [param v1, param v2] $ bodyStatements [
+    assert (objMethodCallNoParams int (valueOf v1) "dimension" ?==
+            objMethodCallNoParams int (valueOf v2) "dimension")
+           (litString "Vector dimensions must match for addition."),
+    varDecDef res local (objMethodCallNoParams (arrayType double) (valueOf (objVar v1 localV)) "clone"),
```
JacquesCarette: That hard-coding of "clone" is definitely very bad - and low-level. You need to program at a higher level, even in GOOL.

code/drasil-code/test/OOVector.hs
JacquesCarette: We should have two Vector implementations, one that's imperative, one that is functional. We should not mix them in the same class.

code/drasil-code/test/OOVector.hs
JacquesCarette: The code generator has very little sanity checking. It is not meant to be written "by hand", really.

We should make some hacks harder to achieve though. Some functions should not take strings at all -- that would be a good start.

code/stable/gooltest/python/OOVector/OOVector.py
```diff
@@ -0,0 +1,84 @@
...
+    # \return A new scaled vector.
+    def scale(self, s):
+        res = self.v.clone()
+        for i in range(0, self.dimension(), 1):
+            res[i] = s * res[i]
+        return Vector(res)
+    
+    ## \brief Prints the vector elements to console.
+    def print(self):
+        print(java.util.Arrays.Arrays.toString(self.v))
```
JacquesCarette: Indeed.

JacquesCarette (changes requested):
GOOL should be viewed as a fair bit more declarative, or at least intent-ful, than the low-level programming you are using it for here.

code/drasil-code/test/OOVector.hs
B-rando1: I came across this in stable:

https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/stable/projectile/projectile_m_p_nol_b_u_v_d/src/java/Projectile/Control.java#L22

So we do have static method calls. I'll have to circle back to see how it's generated; I'm guessing there's a hack somewhere but maybe not.

B-rando1: Getting back to this now, it is indeed a hack. In [genMainFunc](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs#L96), it calls [genCalcCall](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs#L64), which calls [genFuncCall](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs#L83), which calls [fApp](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-code/lib/Language/Drasil/Code/Imperative/GenerateGOOL.hs#L143). When the function being called is in a different module than the module being generated, this uses [extFuncAppMixedArgs](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L235) (external function call), which just so happens to produce exactly the same syntax as a static method call.

B-rando1: And one minute later, I realized that's not quite the whole story. In Java, exported module functions sort of have to be expressed as static methods, and therefore function calls to an external module have to be expressed as static method calls. Checking the Python version of this file, the calculation functions are just functions, not static methods. So the above is not a hack, it's just the way this pattern is expressed in Java.

A more complete answer the question "does GOOL support static methods and static method calls" is that it supports module-level functions and module-level function calls, which may or may not be expressed as static methods.

(it also inexplicably supports creating static methods directly in some languages, but not calling static methods directly :) )

B-rando1: We can talk about whether we should change this, of course. Static properties offer a higher level of namespacing than module-level variables/functions, which could be useful in some cases. On the other hand, that might be getting too "low-level" again, as @JacquesCarette keeps warning us about. It might be good to discuss this in a separate issue.

JacquesCarette: We want to be able to talk about such things we can explain in a language-independent way, assuming some base concepts. While we might secretly know that this would involve `static` in Java, we don't want to talk about `static` unless that is a clear "external" concept.

Note that it's okay to talk about concepts that don't 'exist' in some languages; as long as erasing that information does not cause the execution of the code to change, that's fine. It's like type-erasure in that sense.

balacij: `static` appears to be supported by all of our supported OO languages we support:

* Swift has [`class`/type methods](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/methods/#Type-Methods)
* Python has `@staticmethod` and `@classmethod`
* [C# has static methods](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-classes-and-static-class-members)
* [C++ static methods](https://www.geeksforgeeks.org/cpp/static-member-function-in-cpp/)

We should be able to support generating this code. That being said as @JacquesCarette mentioned, for languages that don't support it (which we should assume exist), we should be able to erase this information.

There are at least two issues here:

1. `genCalcCall` _intends_ to generate a _static function call_. Why? `genCalcCall` should be abstract over how we access these functions. If our `Calculations` class were something we needed to instantiate first, we would need to augment `genCalcCall` to support this behaviour. I think we need a more abstract notion of a "handle" (akin to Java's [MethodHandle](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandle.html) at the Drasil level) that we should be using. Is this particular idea far-fetched? I don't think so. Currently, our `main` functions know the order in which to call things, but we can switch our `Calculator` module to using a lazy evaluator design -- using caching + variable dependency checking to calculate intermediate variables as needed. I'll attach an example below.
2. We don't support `PermanenceSym` (as @B-rando1 mentioned earlier) completely. This is just coding/design work that needs to be done, it seems.

I think a core part of the issue here is that we _assume it to be static_ in `genCalcCall`. Why is that information relevant to this generation function here that @B-rando1 mentioned? In retrospect, it might be because of how "modular" was previously thought of in Drasil -- if we expect the program to be "modular," it will most likely be in another module (?). That being said, the reason is not so important -- that's not the point of the question.


The code example for (1) -- Lazy initialization/evaluation pattern:

```java
public class Calculator {

    private Optional<double> distance;
    private Optional<int> time;
    
    private int calculateTime(...) {
       this.time = ...;
       
       return this.time;
    }

    public double calculateDistance(int x, int y, int dx, int dy) {
        if (!this.time.isPresent()) {
            this.calculateTime()
        }
        
        this.distance = ...;
        
        return this.distance;
    }

}
```

code/drasil-code/test/OOVector.hs
balacij: Considering how we do things right now, I would like to see `arrayElem` defined for arrays within whichever GOOL-type-class we use for defining the array interface. We can clean up that code a bit more later.

So, it looks like we should convert this to an issue once GitHub Issues is working again.

code/drasil-code/test/OOVector.hs
B-rando1: Are there ever situations where we'd want to mix-and-match? I don't have an example off the top of my head, but I'm thinking of how Julia has `push` (pure) and `push!` (impure).

code/drasil-code/test/OOVector.hs
balacij: I think this function (`Arrays.toString`) in particular wants to be part of the array interface in GOOL. Once GitHub Issues is back, we can file an issue.

> It is not meant to be written "by hand", really.

By this, I assume you mean that writing GOOL can be more painful to write "by hand" in the same sense that WebAssembly is painful to write by hand, and not that it has no safety rails?

code/drasil-code/test/OOVector.hs
balacij: So, my writing above was confusing, just to be clear. The code here was just "odd" to me, and I couldn't quite put my finger on "why" at the time.

That being said, while talking to @B-rando1 right now, we have at least two issues:

1. `localV`: we don't teach Drasil that `localV` is a private variable belonging to the `Vector` class. It would be nice if the GOOL API, when referencing variables, could know automatically when to use `this.` or not, as well if it can even access such variable from its current usage site. It feels like we are doing too much work for GOOL (when GOOL should be doing it).
2. My original issue was what @JacquesCarette mentioned here. I'm not sure if our Haskell-level scope is appropriate. It's convenient to have it where it is, but it should be inaccessible as well (perhaps at the type-level?) to things that shouldn't be able to reference it (which we only enforce by being careful about exports). Starting with putting all of these functions into a "where" block of the `Vector` definition is a good start. But if we do that, how should we export `magnitude`, `scale`, etc.? Maybe we shouldn't be doing that?

code/drasil-code/test/OOVector.hs
balacij: This is part of the constructor for a `Vector`. Currently it only supports taking in an array. So I was thinking the most appropriate, naive thing to do would be to clone the inputted data so that we know it wont be mutated later when we eventually operate on it.

code/stable/gooltest/python/OOVector/OOVector.py
balacij: @B-rando1 helped with this one: https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L359-L367

We should consider renaming that typeclass, I think. We should also add documentation to it.

code/stable/gooltest/swift/OOVector/main.swift
balacij: This was meant to be in regards to the `print` line, not the 3 preceding it.

It would be nice if we had a `print' s = print(s, terminator: "")`.

balacij: We can do this to tidy up our generated logging code as well. Hopefully to also avoid how many times we open the log file, too.

code/drasil-code/test/OOVector.hs
balacij: I definitely see duplication, but I'm not sure if I'm seeing the same issue you are?

code/drasil-code/test/OOVector.hs
balacij: Agreed. We need a `arraySize`/`arrayLength` function at the array GOOL interface level.

code/drasil-code/test/OOVector.hs
balacij: In the definitions, they are unavoidable (they need to be encoded somewhere). For references, I agree, but I don't quite know what the right way to avoid it is. It would be nice if we had "reference variables" (at the Haskell level) to these functions, but I don't see anything like this in our codebase (at least not from my skim), so I think there will need to be design work?

code/drasil-code/test/OOVector.hs
JacquesCarette: As other discussions made clear, I would rather move away from exposing 'array' from GOOL's interface. Too low-level.

code/drasil-code/test/OOVector.hs
JacquesCarette: While it might make sense in "optimized code" to mix, at the higher level, I don't think it does. We don't have the reasoning machinery to enable that, so I'd rather not right now.

code/drasil-code/test/OOVector.hs
JacquesCarette: GOOL should not have an `array` interface at all...

And yes, I meant it in that first sense, not the safety aspects.

code/drasil-code/test/OOVector.hs
JacquesCarette: Be wary that different language using the word `static` does not mean that they are conceptually the same! We need to understand the actual meaning of the keyword in each language.

And yes, we have some implicit assumptions in our code (such as `genCalcCall` that we should surface.

code/drasil-code/test/OOVector.hs
JacquesCarette: I'm all for having more information being embedded in what we declare (such as scope, here that `localV`'s scope is the `Vector` class). We started implementing scoping, but never finished.  This should not be re-started until other refactors of  `GOOL` are done though.

We likely should not export the raw `magnitude` and `scale`, but them being methods in `Vector` should be known.

code/drasil-code/test/OOVector.hs
JacquesCarette: The better way to think of it is that the input is passed in *by value* (and could be declared as such). As an implementation detail, language like Java that by default pass by reference would *in the implementation* take care of not touching the input data -- where cloning could indeed be what you can do.

You're trying to implement *code* in `GOOL` while `GOOL` is meant to implement slightly higher-level means of getting things accomplished.

code/stable/gooltest/swift/OOVector/main.swift
JacquesCarette: Conceptually, what is `print'` doing?

JacquesCarette:
Apart from the discussion, we did agree to go about this differently, right?

balacij:
Yes 😄 We just need to split this up into issues.

This approach is still feasible, and we can return to it if we so choose later (after @B-rando1 augments GOOL's container APIs), but, we'll go with the manual-created variant first.

B-rando1:
After pulling in changes to the `Array` API and the fixes to classmethods, I got the generated code working for Java, C#, Python, and Swift. C++ still fails because of some incorrect levels of pointer indirection, but hopefully it's not too hard to fix those.

B-rando1:
There are a few issues with the generated C++ code.

First, in the header file, we declare `vector<double> &v;`, but we want to drop the `&`. The `Vector` instance should own its own array for vector operations. Unfortunately, this `&` is hard-coded in the implementation of `stateVar`:
https://github.com/JacquesCarette/Drasil/blob/a01e5adc95285530d194893143c2ff9ba66d08ba/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs#L2344-L2348

This code is used (correctly) elsewhere, so we can't just hard-code it for this PR. We'll need to find a way to properly encode reference types into GOOL so that the C++ renderer knows whether to add a `&`.

Next, there are a bunch of places where it has `this.method()` that should be `this->method()`. I need to look into this more, but I wouldn't be surprised if this has never been tested. I don't know of any case where `this.method()` would be valid.

In the same vein, whenever we pass `this` as a parameter, we need to dereference it, i.e. pass `*this`. Our C++ renderer doesn't really have a model for doing this, so I'll need to look at the design and figure out how to get this without too many hacks.

Finally, we have to figure out how to handle methods that take other `Vector`s. Right now we take `Vector` values (i.e. `Vector v1`), which means that the `Vector`s are copied when the method is called. The better way would be to take a `const Vector &v1` - that is, a reference that won't be modified. The `const` does matter, because otherwise we could only call the method on lvalues, which is overly restrictive. This would open a can of worms, though. Firstly, we'd need to refine GOOL's understanding of parameters. It already has `param` and `pointerParam`, but we'd need to add `constPointerParam`. Second, we'd need to add the concept of a `const` method. The reason is that a method taking a `const Vector&` can call methods on that parameter, and those methods then have to be marked as `const`, i.e. not modifying `this`. So we'd have to teach GOOL about both `const` reference types and `const` methods.

The first three points are things we have to figure out, but we can come back to the last one later, since the code will function without it and it'll require larger changes to GOOL's type system.

smiths:
Great summary @B-rando1.

B-rando1:
I pulled in changes from #5231. Fingers crossed, this should pass CI :eyes: 

Once it does, we can finally talk about the specifics of the interface, mainly making everything involve pure vector operations.

B-rando1:
I made the changes to the API that made sense to me. Can everyone take a look again when they have time?

balacij:
Thank you for fixing up the code, @B-rando1! It's looking great!

JacquesCarette:
So is this ready for a re-review?

balacij:
Yes, it is ready for review. @B-rando1's work allowed this PR to generate some more practical code. In subsequent PRs we can try to plug this vector library in (as a choice).

JacquesCarette (approved):
As a test of GOOL, I am fine with this. I would not want it to make it 'in' to GOOL or Drasil -- then I would have more comments about the design.

JacquesCarette:
Oops - merging this broke the build! GOOL's interface had changed in between, and this didn't detect that. @B-rando1 @balacij

PR #4912, opened 2026-04-24
haddock docs: only build for `projectile` and `drasil-*` deps

balacij:
Note that this means `drasil-website` is also not built!

The reasoning behind this is the noise in the haddock documentation:

1. 'normal': https://jacquescarette.github.io/Drasil/docs/index.html
2. 'fully exposed modules': https://jacquescarette.github.io/Drasil/docs/full/index.html

The case studies clutter up the vertical space. So, with this change, we only build haddock docs for one case study (`projectile`). Hence, the other case studies and `drasil-website` are removed from the documentation.


balacij:
I didn't 'ask' before writing this code because I didn't know if it was even possible, but it only took 30 or so minutes, so if this is not desired, we can close this.

smiths:
I don't really have an opinion on this change.  If de-cluttering is likely to save people time, it seems like a good change.  However, someone might be looking for the documentation for the web-site or for one of the non-projectile examples.  It isn't that common a use case, but it could come up.

balacij:
Yeah, I'm not quite perfectly happy with it either. I would prefer if we can create two subsections within the same Haddock docs, the first for our libraries and the second for the case studies. But, this isn't supported by Haddock. And neither is organizing the package list arbitrarily.

The rationale isn't necessarily just to save time for us, I think it will make the documentation more appealing to read if the index isn't cluttered, especially with files that have little to no documentation (our case studies are not very 'complete' in this regard).

JacquesCarette (approved):
This is a good change. Our haddocks should be for Drasil, and our examples should be a separate thing.

PR #4913, opened 2026-04-24
Remove unused function: `foldlscC`

balacij:
Removing dead code.


JacquesCarette (approved):
I'm all for removing dead code. I'm surprised this is dead! I'm fairly sure that if our text was more structured, this would be needed. But it's easy to rewrite when we next need it.

PR #4914, opened 2026-04-24
Enable `-Wmissing-export-lists` in `drasil-database`.

balacij:
With this GHC flag enabled, we will be warned of any module missing an export list.

For the CI, that means warnings turning into errors, meaning incomplete export lists become more noticeable.


JacquesCarette (approved):
I guess you'll incrementally add this to most of our .yaml files then?

balacij:
Yes. Rationale is partially code style and partially (a bit of a stretch here) because it forces authors remove dead code before committing new files.

Note that I will not be touching `drasil-data` however!

PR #4915, opened 2026-04-24
Remove unused boolean variables in `drasil-printers`.

balacij:
Just 3 unused booleans.


JacquesCarette (approved):
These seem like good options to have... but dead code deserves to be deleted.

Issue #4929, opened 2026-04-27
Extend the GOOL Array/List APIs


B-rando1:
This comes out of work on #4909 

In GOOL, the [List](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L258-L282) API is at least on the right track: it has most of the common actions one would expect. Array support, on the other hand, is much more spotty. Basically the only action we let users do with arrays is indexing via [arrayElem](https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L108), which is very badly designed.
- First of all, it's in the `VariableSym` typeclass, which means that `arr[i]` is considered a _variable_ in GOOL 🫨.
- More practically, it takes a Haskell `Integer` instead of a GOOL `SValue` for the index, which of course is quite limiting.

We should create an `Array` typeclass modelled after the `List` one. Because they'd share a few methods like `indexToInt`, we'd probably want a parent typeclass `Indexable` that they both inherit from.

We _might_ want to get even fancier than that. @JacquesCarette had a [comment](https://github.com/JacquesCarette/Drasil/pull/4909#discussion_r3133470917) that suggested something to that effect, though I wasn't completely sure what he was getting at.

B-rando1:
I tried changing the return type of `arrayElem` from `SVariable` to `SValue`, and realized that we can't because `arrayElem` is used to create L-Values. So we might need to leave it returning `SVariable` until we figure out #4398.

B-rando1:
On top of the methods provided by the list array, @balacij's work on #4909 requires `arrayToString` and `clone` methods.

JacquesCarette:
We really want a "directly addressable" container type (i.e. abstracting over arrays) and something more Haskell-list-like for lists.

But yeah, we'll need to fix lvalues first.

Furthermore, as we've discussed, I think #4909 is a dead end and should be closed.

B-rando1:
I'm facing some problems changing the array API.

For context, I've been working to split this:

```hs
arrayElem :: SValue r -> SVariable r -> SVariable r
```

into this:

```hs
-- | A fixed-size, indexable container, such as C-style arrays
class (IndexTranslator r) => Array r where
[...]

PR #4930, opened 2026-04-27
Made `arrayElem` take `SValue` as input instead of `Integer`

B-rando1:
Contributes to #4929 and #4909


balacij (approved):
🙏 Thank you, this will be good for the vectors PR.

JacquesCarette (approved):

PR #4933, opened 2026-04-27
Move `drasil-utils/lib/Utils/Drasil/Document.hs` to `drasil-build-artifacts/lib/Drasil/Build/Artifacts/Document.hs`.

balacij:
Rationale: This is a file of tools related to building artifacts. That
is now the goal of `drasil-build-artifacts`.


JacquesCarette (approved):
I'll allow it -- but this is still sub-optimal. ½ the stuff moved are generic pretty-printing routines that don't belong in `drasil-build-artifacts` either! 

But we can worry about that when we move to a more modern pretty-printing library.

PR #4934, opened 2026-04-28
feat: create a concrete `FileLayout` representation for laying out `Doc`s to disk

balacij:
Contributes to #4475 and #4610 

Laying out artifacts is the final thing we do in our generation pipeline.

For example, the following code lays out files for SRS Format-Zoo generation:

https://github.com/JacquesCarette/Drasil/blob/5db1a3d6d5cd0b7d2614918e09e5fda6b7bd1d42/code/drasil-gen/lib/Drasil/Generator/SRS.hs#L31-L148

Laying out each different kind of artifact comes with its own host of quirks. For example, mdBook requires a `.drasil-requirements.csv` and configuration `.toml` file, HTML requires a `.csv`, and so on. Extending this code is problematic because we need to be careful about overwriting files, writing to the correct directories, etc.

An easy way to clean up this code is to re-organize the 'helper functions' to operate on a per-directory-bases rather than by files as it currently does. So, this means we would see a `writeMdBook` function that aggregates all the `mdBook`-related artifacts and writes them all at once in the same function.

However, writing this kind of code is very algorithmic (i.e., ripe for abstraction).

The goal of this PR is to introduce a concrete `Artifact` representation which we can hopefully use in the future to render all of our software artifacts in a more maintainabl fashion.

The design of this code is intentionally very simple. The `Artifact` data type is a tree with `Document` branches and `File` leaves, with sub-`Artifact`s being mapped by name. Highlights:

1. Both constructors have names and contents.
2. Files and documents are not allowed to share names within the same parent path. This is partially a restriction imposed by macOS disallowing this (and so Drasil would not be as portable as we would like -- so this is a fairly uncommon restriction to impose on our other users).
3. Using `String` as the type for name/path components. Specifically, no [`os-string`](https://hasufell.github.io/posts/2024-05-07-ultimate-string-guide.html#osstring-posixstring-and-windowsstring). While `OsString` is the correct string type to use in 2026, using it is a hassle that just doesn't add much value to our needs because it goes too low-level. Our paths can be `String`s all the way up until we do the actual rendering, where we can then convert our `String`s to `OsString`s to write to absolute paths. This is overkill for our needs.
4. No `insert`, `delete`, `lookup`, `union`, `intersection`, `diff` operations. These are just not operations we take in Drasil. These can all be implemented in the future if necessary, but they're odd features for us to have right now.
5. `Artifact` has a type parameter for the content-types. This is partially for #4475, where we will be switching the backend. Note that file-writing with `prettyprinter` has an "efficient" style that immediately writes `Doc`s to disk instead of creating a giant in-memory `String` or `Text` representation that `pretty` requires. So, we create a `Renderable` typeclass for writing to specific file paths immediately.
    ```haskell
    -- | Render a document and write it to a file. Useful for rendering and writing
    -- immediately.
    class Renderable doc where
      renderToFile :: FilePath -> doc -> IO ()
    ````

Demonstrated in #4941.


balacij:
So, what I realize by re-writing the OP is that what `Artifact` is really about is _building an artifact layout_. So, perhaps a better name for it is `ArtifactLayout`? But then again, it isn't strictly about laying out artifacts about a file-path nicely, it does also carry the artifact encodings themselves (all carried as some sort of `Doc` type -- i.e., with no information about the files themselves). Any thoughts about the name? I still prefer `Artifact` for now. In fact, after our meeting the other day, I think renaming the package to `drasil-artifact-layout` might be good (or something in similar mindset).

There are some other, less-important open-ended threads we can discuss too:

1. Authorship. This might be a good area to inject authorship information.
2. Metadata. We currently generate `.drasil/` folders for our SRS artifacts when rendering with debugging enabled. Should we make `.drasil` a reserved name and potentially treat it as slightly different in this code?
3. Authenticity/File checksums. There might be something we can do here, but I don't think we really need to do this. Perhaps leave it as a future TODO note in a TODO.md file within the `drasil-build-artifacts` project?
4. Generator awareness. Should we experiment with using the `Artifact` data type to capture generators? I can vaguely see us being able to capture more information about what generator was used and where using an applicative instance for `Artifact` and a tree-diffing mechanism. This would be a large ask right now, however. And I'm not sure how much value it would bring to the table.

balacij:
With #4941 as a demo, this PR is ready for review.

balacij:
Ha! #4941 brings up an interesting issue: ending newlines with files. Who should be responsible for this? `x -> Doc` renderer or `Doc -> IO ()` file writer? Previously we relied on `hPutStrLn` (`Ln` is the variant of `hPutStr` that appends a newline character) to write to disk, so it was the `Doc -> IO ()` phase (_technically_, since there was a `String` IR of the `Doc`, it was the `String -> IO ()` phase, but I'm not sure if this is relevant).

For now, I will default to the same.

balacij:
If/when we merge this, once fully integrated, we should be able to make `drasil-build-artifacts` consist of exclusively these files added by this PR. Note however that this is dependant on #4942, #4941, and converting the (considerably more troublesome) code and SRS generators. Those will take a few more "clean up first" PRs before we should even bother tackling that.

balacij:
@B-rando1 Since this _may_ affect bits of code in or around `drasil-code` and `drasil-gool`, I would appreciate if you could review as well!

balacij:
`merge` might actually be a useful operation for two `Directory`s (considering we generate multiple 'artifact format zoos' -- i.e., the SRS format and code flavours), but I'm going to try to delay writing this code until absolutely necessary because (a) the code will be ugly, (b) it _can_ lead to hard errors at runtime (e.g., what do we do when merging two directories with [the same name when] a file in both is shared with different contents?). This particular operation is going to be very delicate.

balacij:
A far-fetched idea for `Artifact` is to capture the most generic possible details about them. We had a paper we discussed some time ago (I don't recall the name), but IIRC, it discussed the most common things to find in a code repo. For example, a build system (Makefile, etc.), frontmatter (README[.md]), usage instructions (we don't do this), code, scripts, CI configuration files, docs, etc.

What do we do with this information? Where do we capture? Completely unsure. In `Artifact`s (the data type created by this PR)? Maybe not? But maybe we could have a data type _above_ `Artifact` meant to capture the high-level _kind_ and _features_ of the software artifact (e.g., an executable program should definitely have source code, usage information, etc., and _that_ can be rendered into different flavours, a document artifact should definitely have an assets/ folder, a build system, etc.). Is this `System`? Close but not quite. [One abstraction _down_ (towards the final artifacts).](https://github.com/DrasilOrg/Notes/blob/ghrcCodeVariants/projects/Re-architecting%20drasil-code%20for%20new%20generation%20options%202025/notes/VARIABILITIES.md)

smiths:
@balacij, I think the list of recommended artifacts is in the paper [State of the Practice for Lattice Boltzmann solvers](https://link.springer.com/article/10.1007/s11831-023-09981-2).  In particular, the list is in Table 9 (attached).

RecommendedArtifacts

balacij:
That's the one! Thank you (for writing the paper as well)!

This reminds me of our discussion about modularity and how it is a property of software artifacts. I'm not quite sure what to do with these ideas yet -- too vague, but they're not important right now, so I'll write them down for later.

B-rando1:
> The design of this code is intentionally very simple. The Artifact data type is a trie with Document nodes and File leaves, with sub-Artifacts being mapped by name...

Should this say "Directory nodes and File leaves"?

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,63 @@
...
+import Data.Foldable qualified as F
+import Data.Map.Strict qualified as M
+import Drasil.Build.Artifacts.Render (Renderable (..))
+import System.Directory (createDirectory, doesPathExist)
+import System.FilePath (pathSeparator, (</>))
+
+-- | A software artifact may contain other software artifacts (i.e., be a
+-- directory of other artifacts), or just be a single file.
+data Artifact d
+  = Directory String (M.Map String (Artifact d))
```
B-rando1: I'm a little unclear why this needs a `Map` -- couldn't it just be a `List`/`Set` of `Artifact d`? Or what does the `String` key represent here?

balacij:
> Should this say "Directory nodes and File leaves"?

Right, I meant _branches_ and leaves because the `Artifact` is just a tree.

EDIT: Updated the OP! Thank you!

JacquesCarette:
In this context, I hate the raw word *Artifact*. It's right up there with 'thing' and 'object' and 'chunk' (!) as being devoid of useful meaning.

*Artefact produces by a software engineering process* is significantly better, if ridiculously long.

Plus, as you yourself note, that's not even what this data-structure is. I'll now look at the details, and will likely put in another comment with further thoughts.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+
+-- | A software artifact may contain other software artifacts (i.e., be a
+-- directory of other artifacts), or just be a single file.
+--
+-- Polymorphic over the internal representation of the contents of the
+-- artifacts. For rendering, 'writeArtifact' requires the file content
+-- representation satisfy 'Renderable'.
+data Artifact d
+  = -- | A directory with optionally many nested artifacts.
+    Directory String (M.Map String (Artifact d))
```
JacquesCarette: Maybe we can take this opportunity to use `Text` rather than `String`.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+import System.Directory (createDirectory, doesPathExist)
+import System.FilePath (pathSeparator, (</>))
+
+-- | A software artifact may contain other software artifacts (i.e., be a
+-- directory of other artifacts), or just be a single file.
+--
+-- Polymorphic over the internal representation of the contents of the
+-- artifacts. For rendering, 'writeArtifact' requires the file content
+-- representation satisfy 'Renderable'.
+data Artifact d
```
JacquesCarette: I would be happier with `FileLayout`. This has nothing to do with artifacts (especially as the `d` will usually be `Doc` !!).

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+  = -- | A directory with optionally many nested artifacts.
+    Directory String (M.Map String (Artifact d))
+  | -- | A file with content (of an unspecific type).
+    File String d
+  deriving (Show, Functor, Foldable, Traversable)
+
+-- | Get the name of an 'Artifact'.
```
JacquesCarette: This is not really a name, is it? It is more like 'root label', no?

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
JacquesCarette: Is this suppose to map the relative or absolute addressing?

Shouldn't the first `String` be an abstract representation of something rather than a "name"?

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+-- | Equality on artifacts is dependent solely on the _names_. Note that this
+-- implies a file and directory of the same name cannot exist within the same
+-- directory. This is a restriction partially imposed by macOS.
+instance Eq (Artifact d) where
+  (==) = (==) `on` name
+
+-- | Internal: Check if a path component (i.e., text before/between/after path
+-- separators) is a valid component. Here, validity being defined by not being
+-- any of: ., .., ~, or the system-local path separator.
+checkValid :: String -> String
```
JacquesCarette: This should produce, at worst, a `ValidString`, i.e.
```haskell
newtype ValidString = VS String
```

B-rando1 (commented):
I think this looks good to me, other than one question. Using a tree structure makes sense and lines up with our discussion in #4749

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Render.hs
```diff
@@ -0,0 +1,24 @@
+module Drasil.Build.Artifacts.Render
+  ( Renderable (..),
+  )
+where
+
+import Prettyprinter qualified as PNew
```
JacquesCarette, line 6: Good!

JacquesCarette (changes requested):

balacij:
> In this context, I hate the raw word Artifact. It's right up there with 'thing' and 'object' and 'chunk' (!) as being devoid of useful meaning.

Note that this was the original thought behind this data type. I don't really see this as something to hate. It is **strictly write-only**, no code should be analyzing an `Artifact`. It should exclusively be used to writing the final `Doc` artifacts to disk.

Only now am I thinking it might be more about "laying out arbitrary file things." In this sense, there is _some meaning_ capture (if only that tiny bit of information about the file tree) that can be used, but I'm not sure of how useful this new information is right now.

JacquesCarette:
This is a `MultiFileLayout` or `FileBundle` or some such. No 'artifact' in sight.

JacquesCarette:
Being able to (incrementally) build a `FileBundle` before writing it all out to disk is useful.

But certainly it would make sense to have an `ArtifactBundle` data-structure that would be considerably more semantic and know about 'code', 'readme', 'makefile', 'licence', etc, with lots of accompanying meta-data.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
balacij: As a tree structure, "node label" is most appropriate.

As a file layout structure, the `String`s are just path components. We do not talk about root labels because everything is written with respect to a final "basePath" (in the `writeFiles` function). So, `C:/` and `/`, the real 'root labels' should never be seen in any `Artifact`. "basename" would the more precise name of the terminal path components.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
balacij: Relative.

Can you explain more? I'm not quite following.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
balacij: Using a map is helpful in this situation because sub-artifacts (sub-paths) must be uniquely named.

B-rando1: Does that mean that each sub-artifact has its name stored both in the key and in itself?

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
JacquesCarette: You have a function just below called `name` that you apply to `Artifact` that retrieves this. So why do you think this first `String` is a name? The alternate explanation is that that function has a bad name (no pun intended).

You should comment that this `String` is meant to be a valid single directory name that is meant to be relative. Even better than a comment would be a data-structure (as `newtype`) for that.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
JacquesCarette: See above. You should have `ValidDirectory` and `ValidFileName` newtypes.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
balacij: Correct. And, agh. I should fix this. I don't know how I ended up doing that, lol.

Thank you for spotting that!

balacij:
Thank you for the feedback, I will need a bit of time to fix things up!

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Legacy/Classes.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE FunctionalDependencies #-}
-module Drasil.Build.Artifacts.Classes (
+module Drasil.Build.Artifacts.Legacy.Classes (
```
balacij, line 2: As discussed earlier, this PR should give us code that can replace the majority of the other modules within `drasil-build-artifacts`. So, this PR also moves the other modules to a `Legacy` subfolder.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
```diff
@@ -0,0 +1,74 @@
...
+  }
+  deriving (Functor, Foldable, Traversable)
+
+-- | Internal: File layout tree.
+data FileTree doc
+  = -- | A directory with optionally many nested artifacts.
+    Directory (M.Map PathSegment (FileTree doc))
+  | -- | A file with content (of an unspecific type).
+    File doc
+  deriving (Functor, Foldable, Traversable)
```
balacij: Now, rather than having `Artifact`, we have `FileLayout` and `FileTree`. The `FileLayout` has a "path segment" that it is meant to declare. This path segment is the name of the file/directory (the `FileTree`) it contains.

The `FileTree` is a binary tree that carries directories and files.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
```diff
@@ -0,0 +1,74 @@
...
+-- | Create a file 'FileLayout'.
+file :: PathSegment -> doc -> FileLayout doc
+file fp = FileLayout fp . File
+{-# INLINE file #-}
+
+-- | Create a directory 'FileLayout', optionally containing any number of nested
+-- files.
+directory :: (Foldable f) => PathSegment -> f (FileLayout doc) -> FileLayout doc
+directory fp = FileLayout fp . Directory . F.foldr' insert mempty -- FIXME: Need to deal with duplicate PathSegs due to insensitivity
+{-# INLINE directory #-} -- NOTE: Inlining here and in 'directory' allows for the held @f (FileLayout doc)@s to never actually build 'FileLayout's!
```
balacij: Credits to @TOTBWF for teaching me how to read Core! 😄

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
```diff
@@ -0,0 +1,74 @@
...
+  if exists
+    then error $ "Path already exists: " ++ show targetPath
+    else go basePath layout
+  where
+    go currentPath (FileLayout fname (File content)) =
+      renderToFile (currentPath </> fname) content
+    go currentPath (FileLayout dname (Directory children)) = do
+      let nextPath = currentPath </> dname
+      createDirectory nextPath
+      F.traverse_ (\(n, c) -> go nextPath (FileLayout n c)) (M.toList children)
```
balacij, line 90: The crescendo of refinement is meant to lead to this!

I believe that calling this function will be part of the last line in each of our `app/Main.hs` files.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FilePath.hs
```diff
@@ -1,28 +1,69 @@
...
+import System.FilePath (pathSeparator)
+import System.IO (Handle, IOMode (WriteMode), hPutStr)
+import System.OsPath (OsPath, encodeUtf)
+import System.OsPath qualified as FP ((</>))
+import Prelude hiding (writeFile)
+
+-- | Represents a valid segment of a path (e.g., a file or directory /name/, or
+-- the name of one parent folder).
+newtype PathSegment = PS {unPS :: OsPath}
+  deriving (Eq, Ord)
```
balacij: `PathSegment` is a sanitary path input.

Note that it wraps an [`OsPath`, not `Text`/`String`/etc.!](https://hasufell.github.io/posts/2024-05-07-ultimate-string-guide.html#a-note-on-filepaths)

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FilePath.hs
```diff
@@ -1,28 +1,69 @@
...
+
+-- | 'QuasiQuoter' for building 'PathSegment's.
+ps :: QuasiQuoter
+ps =
+  QuasiQuoter
+    { quoteExp = qPathSeg,
+      quotePat = unpermitted,
+      quoteType = unpermitted,
+      quoteDec = unpermitted
+    }
```
balacij, line 57: And we have a quasi-quoter for building path segments because (a) majority will be static `String`-based ones (for now), and (b) it will provide GHC-provided compilation errors for bad inputs rather than Drasil-runtime errors.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FilePath.hs
```diff
@@ -1,28 +1,69 @@
...
+-- | Internal: Check if a path segment (i.e., text before/between/after path
+-- separators) is a valid segment. Here, validity being defined by not being any
+-- of: ., .., ~, or the system-local path separator.
+qPathSeg :: String -> Q Exp
+qPathSeg s
+  | s `elem` [".", "..", "~"] = fail $ "invalid path segment: " ++ show s ++ "."
+  | pathSeparator `elem` s = fail $ "cannot create path segment with " ++ show pathSeparator ++ " in the name."
+  | otherwise = case encodeUtf s :: Either SomeException OsPath of
+      Left err -> fail $ "invalid os path: " ++ show err
+      Right osp -> [|PS osp|]
```
balacij: The criteria for a valid `PathSegment` is fairly straightforward for now:

1. It must be possible for the current system (`encodeUtf`).
2. It must not be `.`, `..`, nor `~`.
3. It may not contain a parent directory/sub-directory.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Legacy.hs
```diff
@@ -0,0 +1,15 @@
...
+  module Drasil.Build.Artifacts.Legacy.FileIO,
+  module Drasil.Build.Artifacts.Legacy.FilePath
+) where
+
+import Drasil.Build.Artifacts.Legacy.Classes
+import Drasil.Build.Artifacts.Legacy.Document
+import Drasil.Build.Artifacts.Legacy.Directory
+import Drasil.Build.Artifacts.Legacy.FileData
+import Drasil.Build.Artifacts.Legacy.FileIO
+import Drasil.Build.Artifacts.Legacy.FilePath
```
balacij, line 15: Re-exporting all legacy modules through the new `Drasil.Build.Artifacts.Legacy` module.

balacij (commented):
Ready for more feedback, @B-rando1 @JacquesCarette !

I do want to also build a testing suite for this code as well. I'm just not sure if I should do it in this PR or another.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
JacquesCarette: Strictly speaking, `FileTree` is a Rose Tree, not a binary tree! But that is pedantry.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
JacquesCarette: I would think that calling this function ought to be internalized in some function in `drasil-gen` and not exposed to applications at all. This is too low-level a concern for them to worry about.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FilePath.hs
JacquesCarette: Hurray!

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FilePath.hs
JacquesCarette: Nice!

JacquesCarette (changes requested):
Hasn't this now morphed into more of a `drasil-file-handling` package? It sure would make sense for such a package to be at the "bottom" of the dependency hierarchy, i.e. not depend on anything else.

B-rando1 (approved):
It makes sense to me the way you've laid this out. It's hard to know for sure if this is the right design without using it, but at the very least this should be a big step up over our current methodology.

balacij:
> Hasn't this now morphed into more of a drasil-file-handling package? It sure would make sense for such a package to be at the "bottom" of the dependency hierarchy, i.e. not depend on anything else.

Agreed. We can do this in a subsequent PR.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
balacij: I agree that `drasil-gen` should be the only package using this function. But I believe it is best to keep this code within `drasil-file-handling` because the internals might change in the future. This code relies heavily on said internals, and I think we want to avoid exposing `.Development`/`.Internal`/`.Core` modules unless absolutely necessary.

balacij:
There is one major caveat with using `Os(String|Path)`: we can't do case-insensitive path equality.

My first idea was to change check lowercase equality, but that's when I realized as we also can't manipulate `OsString`s at the `OsChar` level. And it looks like this is by design:

> It captures the notion of syscall specific encoding (or the lack thereof) to avoid roundtrip issues and memory fragmentation by using unpinned byte arrays. Bytes are not touched or interpreted.

https://hackage-content.haskell.org/package/os-string-2.0.10/docs/System-OsString.html

Otherwise, this is ready for review, @JacquesCarette 

We should build a testing suite in a subsequent PR as well.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts.hs
```diff
@@ -1,15 +1,6 @@
...
-  module Drasil.Build.Artifacts.Classes,
-  module Drasil.Build.Artifacts.Directory,
-  module Drasil.Build.Artifacts.Document,
-  module Drasil.Build.Artifacts.FileData,
-  module Drasil.Build.Artifacts.FileIO,
-  module Drasil.Build.Artifacts.FilePath,
-) where
+module Drasil.Build.Artifacts
+  ( module Drasil.Build.Artifacts.FileLayout,
+  )
```
djarabek, line 5: The QuasiQuoter `ps` required to construct `PathSegment`s isn't exported, so there's no way to construct `PathSegment`s.

I also don't see a way to construct a `PathSegment` from a `String`, for instance the code in #4941:
https://github.com/JacquesCarette/Drasil/blob/072ce765899df877f6a721361d798d1710645e87/code/drasil-gen/lib/Drasil/Generator/Website.hs#L24

djarabek (changes requested):

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts.hs
djarabek: I suppose I could change the type of `fileName` to `PathSegment` and then construct it as `[ps|index|]` in website's `Main.hs`, however there would still need to be an exported operation to concatenate path segments to add the `.html` extension.

balacij: Right. Thank you! I will (1) export `ps` and (2) figure out how to pass in variables/arbitrary Haskell expressions.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/FileLayout.hs
JacquesCarette: Agreed. I just did not want to see this function showing up in `app/Main.hs` files.

JacquesCarette (approved):

JacquesCarette:
Have you fixed the items that @djarabek spotted, so we are now waiting on his re-review?

balacij:
Building a complete testing suite is a bit of a hassle right now. I'd rather not block this PR so I've just implemented a very tiny unit-test-based testing suite for the quasiquoter at least.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts.hs
balacij: This is done in (1) [49242de](https://github.com/JacquesCarette/Drasil/pull/4934/commits/49242de065f1d548fafb08afe3457a4bd81bf38f) and (2) [24b411c](https://github.com/JacquesCarette/Drasil/pull/4934/commits/24b411c390fa9719dbc66ecf7e68a9d304e33312).

balacij:
> Have you fixed the items that @djarabek spotted, so we are now waiting on his re-review?

Yep!

PR #4938, opened 2026-04-29
Renamed some LValue-related concepts

B-rando1:
Contributes to #4398. It's likely that I missed a few, but these should be the main ones. Hopefully we'll catch/clarify any that I missed or misclassified.

I also added a bunch of TODOs about future work.


balacij (approved):
Wow there's a lot of references.

Regarding missing some, did you use a find and replace tool (e.g., `find`+`sed`)?

JacquesCarette:
Needs a merge.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
```diff
@@ -126,12 +126,13 @@ class OpElim r where
 class ScopeElim r where
   scopeData :: r (Scope r) -> ScopeData
 
-class RenderVariable r where
-  varFromData :: Binding -> String -> VSType r -> Doc -> SVariable r
+-- TODO: [Brandon Bosman, 04/29/2026]: This is an odd method, and it's used for more lvalues than just variables.
```
B-rando1, line 129: This is an example of an ambiguous case - `varFromData` takes a binding, which I believe is specific to variables, but functions like `mkVar` (renamed to `mkLValue`) use it for all types of lvalues.

B-rando1:
> Regarding missing some, did you use a find and replace tool (e.g., `find`+`sed`)?

This was mostly manual. `vim` helps a bit, but not as much as `find` + `sed` would've. I could've sped some of it up with that method, but it's more difficult for keywords like `variable` that appear in comments etc. all over the place. Maybe there's a syntax-directed find-and-replace tool that can handle the scope of Drasil; that would be handy.

The real reason it's possibly incomplete is ambiguous cases like the one I mentioned above. We'll need to rework some of these functions to more cleanly distinguish between variables and lvalues so that we can give them accurate names.

B-rando1:
Conflict is resolved now.

code/drasil-gool/lib/Drasil/Shared/RendererClassesCommon.hs
JacquesCarette: We should definitely separate out the various concepts.

code/drasil-code/test/GOOL/Observer.hs
```diff
@@ -22,11 +22,11 @@ observer = fileDoc (buildModule observerName [] [] [docClass observerDesc
   helperClass])
 
 -- | Makes a variable @x@.
-x :: (VariableSym r) => SVariable r
+x :: (LValueSym r) => SLValue r
 x = var "x" int
 
 -- | Acces the @x@ attribute of @self@.
-selfX :: (OOVariableSym r) => SVariable r
+selfX :: (OOLValueSym r) => SLValue r
```
JacquesCarette, line 29: Hmm, is `self` really an LValue? You can't really put it on the left of an assignment, can you?

JacquesCarette:
So I'm wondering if this leaves the `GOOL` code less confusing or equally confusing?

What we had before was the "syntactic" concept of a variable, as it occurs in programming languages. However, that was constraining, because such variables actually have different roles in different contexts. So it makes sense to move from using purely syntactic tags (like variable) to somewhat more semantic tags (like lvalue). So far so good. *But* you've moved all the variables over to lvalue!

In some sense, what we had before was 'bad' but now I fear that some of the code is 'wrong', when read semantically.

I'm not saying that I won't merge this as a potentially valuable intermediate step. But I do want to be convinced that the next step(s) will be to fix those places where lvalue is just wrong!

code/drasil-code/test/GOOL/Observer.hs
B-rando1: I guess `self` is a sort of constant, which means it has the same ambiguity that constants do.

I've come across two definitions of LValue: "locator value" (i.e. a value that locates a piece of memory) and "left value" (i.e. something that can be assigned to). Under the first definition, constant references (including `self`) could be classified (and are, in C++) as "non-mutable locator values". Under the second, they are definitely not "left values".

The practical part of this is that we need a way to construct LValues from constant references one way or another. Either definition can work, but will have different tradeoffs:
- "Locator value": In this interpretation, which GOOL has implicitly been following, constant references are LValues. This means we can use things like `arrayElem` and `objVar` directly on them without anything weird. The downside is that at the Haskell type-level, we can't prevent people from treating them like mutable LValues and trying to modify them. We could add a generation-time check for this, though.
- "Left value": In this interpretation, constant references are RValues, so we can't construct LValues from them in the same way that we can from variable references. We would have to create custom fields like `objVarSelf` (which does already exist, coincidentally), `constArrayElem`, `constObjVar`, etc. to get around this.

@TOTBWF didn't seem to like the first interpretation when we were talking about this the other day, but to me it seems much cleaner conceptually. All it would require is adding a piece of state to LValues answering the question "am I mutable?" and causing an error if GOOL code tries to assign into a non-mutable lvalue. This shouldn't difficult, since GOOL already carries other state for LValues.

B-rando1:
That is a fair critique. Just to clarify, are you mainly talking about the lack of binding forms or is it more widespread than that?

code/drasil-code/test/GOOL/Observer.hs
JacquesCarette: We should pick an interpretation (and, if need be, name) that coincides with the information we need for GOOL to get its job done properly.

This might mean that we need both `locator` and `assignable` !!!

JacquesCarette:
More widespread. I think there are many different semantic concepts appearing that are now all called `lvalue` even though some of them serve a different purpose.

Note that we might need `locator`, `assignable`, `assignablelocator`, `readonly`, `binder` and who knows what else, to replace 'variable'.

B-rando1:
This will be a good reference, but it seems like we should be going for more of a bottom-up approach. We should teach GOOL about the various concepts first, and update their names as we go.

JacquesCarette:
So should we delete this branch?

B-rando1:
Sure. I'll keep my local copy because its comments about what should be changed to a binding form should be helpful, but we don't need to keep it here.

PR #4941, opened 2026-04-30
refactor(website): use new `Artifact` representation for laying out artifacts

balacij:
Builds on #4934 

This PR uses the new artifact layout mechanism #4934 introduces for laying out artifacts for the website generator.


code/drasil-gen/lib/Drasil/Generator/Website.hs
```diff
@@ -2,32 +2,28 @@ module Drasil.Generator.Website (
...
 
-import Prelude hiding (id)
 import Control.Lens ((^.))
-import System.IO (hClose, hPutStrLn, openFile, IOMode(WriteMode))
-import Text.PrettyPrint.HughesPJ (render)
 
-import Drasil.Build.Artifacts (createDirIfMissing)
-import Language.Drasil (Stage(Equational), Document)
-import Language.Drasil.Printers (makeCSS, genHTML, Notation(Engineering), piSys)
+import Drasil.Build.Artifacts (createDirIfMissing, directory, file, writeArtifact)
```
JacquesCarette: The previous version of this file was a mess, but now it feels even more odd, as it has IO here and in `writeArtifact`. Feels like the responsibility to write a `FileBundle` should be in a single place. In other words, this refactor did not go quite far enough in figuring out who had what secrets.

JacquesCarette (changes requested):

code/drasil-gen/lib/Drasil/Generator/Website.hs
balacij: It is a bit odd, I agree, but I'm not sure if it's any more odd than before?

I'm not sure if I'm understanding your concern correctly, but one thing I was flip-floppy on was about what `exportWebsite` should output. Return the `Artifact` or if it should also write it (via `writeArtifact`)? If latter, then what code should do the `writeArtifact`?

We have a lot of insider knowledge built into `drasil-gen` that make it work for us for demonstrating everything Drasil can do (for the SRS/code format zoos we generate). Is this a secret that should be contained within `drasil-gen`?

How far do we go in this PR?

JacquesCarette: We had an extensive discussion of this PR at our meeting - is this clearer now?

balacij: Yes. I'm not entirely sure how this code will change yet, but it will require some non-trivial design.

code/drasil-gen/lib/Drasil/Generator/Website.hs
```diff
@@ -1,33 +1,31 @@
...
-  createDirIfMissing True dir
-
-  outh <- openFile (dir ++ "/" ++ fileName ++ ".html") WriteMode
-  hPutStrLn outh $ render $ genHTML fileName pd
-  hClose outh
-
-  outh2 <- openFile (dir ++ "/" ++ fileName ++ ".css") WriteMode
-  hPutStrLn outh2 $ render $ makeCSS doc
-  hClose outh2
+  writeFiles localPath website
```
balacij, line 31: Now this code is responsible for knowing which generator to use and where to put the results of each one (a pure "recipe"?).

Instead of doing the file-writing here as well, we can also return the `FileLayout Doc` instead. Then `drasil-website/app/Main.hs` could be responsible for doing the final `writeFiles localPath website`.

The website generator is a bit odd regardless because it just wasn't quite fleshed out enough.

JacquesCarette (approved):

Issue #4884, opened 2026-04-13
Need new chunk: `ProgramName`


balacij:
> I'm thinking a new `ProgramName` chunk, with fields `name` (or `label`) and some optional description. It should not implement the `term` lens. We're overloading `CI` too far here.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/4363#discussion_r2743727488_

For our `SmithEtAlSRS`, we need to be able to associate a "codename/program name" with our SRS. It is not necessarily a name intended to be injected into our Table of Abbreviations and Acronyms. Hence, we need a new chunk type for this along with a way of rendering said chunk type into `Sentence` (in a similar fashion we do symbols and terms) and code-appropriate names.

PR #4942, opened 2026-04-30
refactor(lesson-plan-generator): use new `FileLayout` representation for laying out artifacts

balacij:
Builds on #4934

Similar to #4941 

This PR uses the new artifact layout mechanism https://github.com/JacquesCarette/Drasil/pull/4934 introduces for laying out artifacts for the website generator.


code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+--
+-- Polymorphic over the internal representation of the contents of the
+-- artifacts. For rendering, 'writeArtifact' requires the file content
+-- representation satisfy 'Renderable'.
+data Artifact d
+  = -- | A directory with optionally many nested artifacts.
+    Directory String (M.Map String (Artifact d))
+  | -- | A file with content (of an unspecific type).
+    File String d
+  deriving (Show, Functor, Foldable, Traversable)
```
balacij: We might also consider creating something like:

```haskell
data CompleteArtifact d
  = CA ArtifactName Authors License (M.Map String (Artifact d))
```

This data type would be one top-level directory that carries extra information we only care for carrying _once_ (at the top-level), where the directory name is also a [bit more meaningful](#4884).

balacij: If we were to do this, we might even need to do an extra check on the held artifacts -- that they don't use a "special, scary, very conventional name" (e.g., README[.md], LICENSE[.md], CONTRIBUTING[.md], etc.).

code/drasil-gen/lib/Drasil/Generator/LessonPlan.hs
```diff
@@ -2,30 +2,26 @@ module Drasil.Generator.LessonPlan (
   exportLessonPlan
 ) where
 
-import Prelude hiding (id)
 import Control.Lens ((^.))
-import System.IO (hClose, hPutStrLn, openFile, IOMode(WriteMode))
-import Text.PrettyPrint.HughesPJ (render)
 
-import Drasil.Build.Artifacts (createDirIfMissing)
+import Drasil.Build.Artifacts (directory, file, writeArtifact)
```
JacquesCarette: Same comment here as on previous PR (for website)

JacquesCarette (changes requested):

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Core.hs
JacquesCarette: See some of my previous comments. You're slowly re-inventing the "software dossier" as the data-structure that sits above this, you just have not abstracted enough (yet).

djarabek:
Ready for review.

balacij (commented):
Unfortunately I can't approve my own PR, but the updates look good to me. Thanks, @djarabek!

JacquesCarette (approved):

PR #4943, opened 2026-04-30
refactor(-gen): remove unused, ill-supported `Plain` SRS output format

balacij:
The "Plain" SRS type is unused and a bit pointless. If/when we want more formats, there are better output types to produce.


JacquesCarette (approved):
The point was to have a "plain text" printer for at least parts of SRS, with the goal of being able to use it for code comments.

Getting rid of the top-level format is fine, as it doesn't affect that goal.

Issue #4874, opened 2026-04-08
Pre-summer Wiki Work


balacij:
We have 5 key pages to review regarding day-to-day Drasil development:

1. [ ] [Contributor's Guide](https://github.com/JacquesCarette/Drasil/wiki/Contributor's-Guide)
2. [ ] [Workflow](https://github.com/JacquesCarette/Drasil/wiki/Workflow)
3. [ ] [New Workspace Setup](https://github.com/JacquesCarette/Drasil/wiki/New-Workspace-Setup) - See https://github.com/JacquesCarette/Drasil/issues/3728
4. [ ] [Git2Know for Drasil](https://github.com/JacquesCarette/Drasil/wiki/Git2Know-for-Drasil)
5. [ ] [Debugging in Drasil](https://github.com/JacquesCarette/Drasil/wiki/Debugging-in-Drasil)

smiths:
It isn't on the Wiki, but we should also update the [onBoarding document](https://github.com/JacquesCarette/Drasil/blob/main/doc/OnBoarding/onBoarding.pdf) before the summer.

balacij:
I like the onboard document. It doesn't look like there's a lot to do for that one except for updating the names. Did you have anything specific in mind that I can help with?

Regarding the Git2Know page, I'm not quite sure what our goal is for that page. Should it introduce Git from scratch? How in-depth should it be? As it is now, I'm not sure who its target audience is. If it is meant as reference material for people already familiar with git, that's good, we just need to update its content with more advanced topics. If it's meant for those not already familiar with git, then it could use a bit more introductory discussion, introducing terminology and basic VCS+git concepts.

<details>

<summary>For example, this is more of a layout (with a bit of content I filled in) that I would expect to see on a "Git2Know" page for prospective first year summer students.</summary>

<hr>

# What is `git`?

`git` is a distributed [version control system](https://en.wikipedia.org/wiki/Version_control) commonly used for software development. `git` is used to share a set of files while keeping information about who made changes, when they were made, and why they were made (through user-provided explanations). This guide will establish some basic terminology and walk through some common steps that contributors will need to know. However, it will not teach you the inner workings of `git` nor is it a substitute for more in-depth resources. If you'd like to learn more about `git`, you may enjoy reading the [Pro Git](https://git-scm.com/book/en/v2) book.

If you haven't already installed `git`, you should make sure to install and configure it following the [New Workspace Setup](New-Workspace-Setup#git) guide.

## Terminology

[...]

PR #4946, opened 2026-05-01
Edit New Workspace Setup guide

balacij:
Contributes to #4874 

1. Fixes typos.
2. Updates advice regarding Windows: WSL preferred only for beefier systems.
3. General tidying up and clarifying minor things.


balacij:
One thing I noticed about the New Workspace Setup wiki page is that we don't mention installing the compilers and interpreters necessary for our generated _executable software artifacts_, only our documentation-related software artifacts.

wiki/New-Workspace-Setup.md
```diff
@@ -1,32 +1,34 @@
...
-2. If you use the `nix` package manager and/or the `NixOS` operating system, you may alternatively use the `shell.nix` ad-hoc development environment provided, however, you should still install [VSCode](#vscode) + [Extensions](#extensions) and [install the `lmodern` and `lmmath` fonts](https://nixos.wiki/wiki/Fonts).
+This guide will provide you with a straightforward workspace recommendation for developing and contributing to Drasil.
+
+In order, you will be installing the following _bare minimum requirements_ for working with Drasil's Haskell codebase:
+
+1. [Basic Development Tools](#basic-development-tools): `git`, `GNU Make`, etc.
+2. [Stack](#stack-via-ghcup): A Haskell development toolchain.
+3. [VSCode](#vscode) + [Extensions](#extensions): A commonly used code editor with good Haskell support provided through a [plugin](https://github.com/haskell/haskell-language-server).
+
+As Drasil generates a host of different artifacts, you will also need to install various other tools to work with them:
```
JacquesCarette: I think this list needs to be more subtle. I don't have some of these tools installed myself, and can work with Drasil just fine!

For doing certain tasks, you absolutely need these (locally). Same as with what is mentioned later, aka compilers.

So the tools to install should be qualified by task dependency. If you don't work on certain parts, then you don't need some of these tools.

JacquesCarette (changes requested):
Just one thing to fix, otherwise this is a great set of updates.

wiki/New-Workspace-Setup.md
```diff
@@ -73,11 +78,15 @@ The Drasil directory structure relies on symbolic links. Windows 10/11 has suppo
...
+If you're using a recent version of Mac (> Mavericks/10.9), please open up a terminal window (`Command + Space`, then search "Terminal"), and type in `git --version` into it, and press enter. A GUI installer will appear prompting you to install Git. Please follow the on-screen steps, leaving all options default. Once the installation is complete, you should be able to type `git --version` into your Terminal window again and receive a version code back instead of having the GUI pop up again. If you're using an older version of Mac, you may follow through the steps of installing Git using any of the [official repositories](https://git-scm.com/download/mac).
```
smiths: I didn't know about this option for installing `git`.  Neat!  It probably doesn't matter, but if the Mac user is installing `brew` for later steps, would it make sense to install `git` via `brew` as well?  The xcode solution is simpler, but the `brew` solution might be better for future updates?

I'm not requesting a change, just asking a question.  :smile:

smiths:
Thank you @balacij for the detailed instructions.  I don't know if I mentioned it, but I used your instructions to jog my memory when I set up my latest laptop and desktop this past winter.  :smile:

wiki/New-Workspace-Setup.md
balacij: That makes sense to me. After finding out that macOS ships with extremely old versions of `grep`, `sed`, etc., I am definitely on board with this.

Fixed in e6b84b3b1750378051f21bccd9e7ba4177120b9d

wiki/New-Workspace-Setup.md
balacij: How does c68047fecd3390e0242078eb8b25be243aca4d37 look?

balacij:
The New Workspace Setup guide is a bit long-winded. There are a lot of instructions that we can simplify, especially for macOS and apt-based Linux distros, where we can also possibly ship a script that does 95% of the work!

Ah, thank you, @smiths ! I'm glad to hear it! Did you find any issues with it?

smiths:
I don't remember any issues.  I would have said something if I noticed a problem.  :smile:

wiki/New-Workspace-Setup.md
```diff
@@ -1,32 +1,41 @@
...
-2. If you use the `nix` package manager and/or the `NixOS` operating system, you may alternatively use the `shell.nix` ad-hoc development environment provided, however, you should still install [VSCode](#vscode) + [Extensions](#extensions) and [install the `lmodern` and `lmmath` fonts](https://nixos.wiki/wiki/Fonts).
+This guide will provide you with a straightforward workspace recommendation for developing and contributing to Drasil.
+
+In order, you will be installing the following _bare minimum requirements_ for working with Drasil's Haskell codebase:
+
+1. [Basic Development Tools](#basic-development-tools): `git`, `GNU Make`, etc.
+2. [Stack](#stack-via-ghcup): A Haskell development toolchain.
+3. [VSCode](#vscode) + [Extensions](#extensions): A commonly used code editor with good Haskell support provided through a [plugin](https://github.com/haskell/haskell-language-server).
```
JacquesCarette: This shouldn't be in the 'bare minimum'. I work on Drasil comfortably without. "Highly recommended", sure.

wiki/New-Workspace-Setup.md
```diff
@@ -1,32 +1,41 @@
...
+If you intend to work with the artifacts that Drasil generates, you should also install:
+
+* For **document** artifacts*:
+    1. [Graphviz](#graphviz)
+    2. A [LaTeX](#latex) toolchain
+    3. [Inkscape](#inkscape)
+    4. [mdBook](#mdBook)
+    5. [Jupyter](#jupyter)
+* For **code** artifacts: [Doxygen](#doxygen)
+* For **linting** Drasil's codebase: [ShellCheck](#shellcheck)
```
JacquesCarette: Stan? hlint?

JacquesCarette (changes requested):

wiki/New-Workspace-Setup.md
JacquesCarette: It would be nice to get this wrapped up - bad to have open PRs hanging.

wiki/New-Workspace-Setup.md
balacij: Resolved in 5b62fda85

wiki/New-Workspace-Setup.md
balacij: Added some rough instructions in 6c9e14eed

JacquesCarette (approved):

PR #4949, opened 2026-05-04
CSharp: fix run target to use mono instead of native binary

### LINKED ISSUE #4927: C# Makefiles are Broken for Linux
Log:

```
$ cd stable/gooltest/csharp/HelloWorld/
$ make
mcs -out:HelloWorld HelloWorld.cs Helper.cs
HelloWorld.cs(273,20): warning CS0219: The variable `dummy' is assigned but its value is never used
HelloWorld.cs(274,18): warning CS0219: The variable `myObj' is assigned but its value is never used
HelloWorld.cs(292,23): warning CS0219: The variable `myList' is assigned but its value is never used
Compilation succeeded - 3 warning(s)
$ make run
./HelloWorld 
./HelloWorld: ./HelloWorld: cannot execute binary file
make: *** [Makefile:26: run] Error 126
```

I believe this is because `mcs` produces a .NET assembly file, not a native binary file. The solution is to run `mono HelloWorld` instead of `./HelloWorld`.

Purav-Kanda:
Fixes #4927

mcs compiles C# to a .NET assembly, not a native binary. Running ./HelloWorld 
fails on Linux with "cannot execute binary file". Changed csRunnable from 
nativeBinary to interpMM "mono" so the generated Makefile run target uses 
`mono HelloWorld` instead of `./HelloWorld`. Updated all stable C# Makefiles 
to match.


B-rando1 (approved):
The code looks good to me, and the makefiles work on my machine 👍

JacquesCarette (approved):
I will merge this PR because this is more general -- but probably we should have an option to control this. But I think a generation-time option, not a generated-Makefile level option!

PR #4953, opened 2026-05-06
Use Data.List.NonEmpty for ICO problem inputs and outputs

### LINKED ISSUE #3652: Enforcing that examples have outputs
While generating output requirements in #3259, Dr. Carette mentioned that "If you attempt to generate an output requirement with no outputs, that should be a hard error." ([source](https://github.com/JacquesCarette/Drasil/issues/3259#issuecomment-1673738596)) In a meeting, he also mentioned that this should be handled at a higher level than at the requirement-generation level; currently, this check is done in 15823ccc7768f998d7e54d5a9121610242b67cd5.

Alongside this, we've also been having discussion about what it means for a program to have an output, as well as the different kinds of output. In the same comment, Dr. Carette said "our concept of `Output` is too naïve. There are too many meanings of 'output' (ranging from the result of a function to printing on the screen). A relational specification can be refined to point to some values as being 'output'. Some theories have multiple things that can be considered 'outputs'. We should have more precise names for each of these concepts." Honestly, I'm not sure at what level this discussion should take place, but it seems related to the enforcement of programs having outputs.

The real question: assuming that each program has some sort of output (regardless of what kind of output it is), how/where should we enforce that outputs are specified for each example? Is this a valid assumption to make?

djarabek:
Closes #3652

TODO:
- fix running the template (replace undefined with a dummy variable):
```
template: Prelude.undefined
  undefined, called at lib/Drasil/Template/Body.hs:79:4 in template-0.1.24.0-2cKFhjLwZ0m3eciTvFpV5w:Drasil.Template.Body
make[1]: *** [Makefile:329: template_gen] Error 1
make[1]: Leaving directory '/home/dj/projects/Drasil/code'
make: *** [Makefile:261: test] Error 2
```
- investigate pushing the `NonEmpty` type further down to `OldCodeSpec`


JacquesCarette:
Main comment I was going to make: try to not use `fromList` at creation site, when you can. `toList` at use site is often the right thing to do, as often the use site no longer cares about the invariant.

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -42,8 +43,8 @@ glassTypeCon, nomThick :: ConstrConcept
 
 {--}
 
-inputs :: [DefinedQuantityDict]
-inputs = map dqdWr inputsWUnitsUncrtn ++ map dqdWr inputsWUncrtn ++
+inputs :: NE.NonEmpty DefinedQuantityDict
+inputs = NE.fromList $ map dqdWr inputsWUnitsUncrtn ++ map dqdWr inputsWUncrtn ++
```
JacquesCarette: A couple more `fromList` to avoid in here.

code/drasil-example/ssp/lib/Drasil/SSP/Unitals.hs
```diff
@@ -78,11 +79,11 @@ inputsWUncrtn = [slopeDist, slopeHght, waterDist, waterHght, xMaxExtSlip,
 inputsNoUncrtn :: [DefinedQuantityDict]
 inputsNoUncrtn = [constF]
 
-inputs :: [DefinedQuantityDict]
-inputs = map dqdWr inputsWUncrtn ++ map dqdWr inputsNoUncrtn
+inputs :: NE.NonEmpty DefinedQuantityDict
+inputs = NE.fromList $ map dqdWr inputsWUncrtn ++ map dqdWr inputsNoUncrtn
```
JacquesCarette: one more `fromList`

code/drasil-example/swhs/lib/Drasil/SWHS/Unitals.hs
```diff
@@ -236,11 +238,11 @@ aspectRatioMax = dqd' (dcc "aspectRatioMax"
...
 
 constrained :: [ConstrConcept]
-constrained = map cnstrw' inputConstraints ++ map cnstrw' outputs
+constrained = map cnstrw' inputConstraints ++ map cnstrw' (NE.toList outputs)
 
 -- Input Constraints
-inputs :: [DefinedQuantityDict]
-inputs = map dqdWr inputConstraints ++ map dqdWr [absTol, relTol]
+inputs :: NE.NonEmpty DefinedQuantityDict
+inputs = NE.fromList $ map dqdWr inputConstraints ++ map dqdWr [absTol, relTol]
```
JacquesCarette: another `fromList`

JacquesCarette (changes requested):

code/drasil-example/template/lib/Drasil/Template/Body.hs
```diff
@@ -74,7 +76,7 @@ si = mkSmithEtAlICO
   progName [authorName]
   [] [] [] []
   ([] :: [TheoryModel]) ([] :: [GenDefn]) ([] :: [DataDefinition]) ([] :: [InstanceModel])
-  ([] :: [DefinedQuantityDict]) ([] :: [DefinedQuantityDict]) ([] :: [ConstrConcept]) ([] :: [ConstQDef]) []
+  (undefined :: NE.NonEmpty DefinedQuantityDict) (undefined :: NE.NonEmpty DefinedQuantityDict) ([] :: [ConstrConcept]) ([] :: [ConstQDef]) []
```
JacquesCarette: looking at the test failure, I am not surprised that these two `undefined` give you a problem.

I think the simplest fix will be to make the Template implement the *identity function* (which has an input and an output). Any other hack will cause more pain.

JacquesCarette:
Can we try to get this resolved? What's the bottleneck?

djarabek:
I'm working on the changes for this PR (and #4958), I'll have everything pushed by the end of today.

djarabek:
Ready for review and merging!

JacquesCarette (approved):

PR #4958, opened 2026-05-07
Update 'template' example to include inputs and outputs

djarabek:
Contributes to #3652

#4953 Switches inputs and outputs to be `NonEmpty` lists, so we need some inputs and outputs in the template. Ideally these inputs and outputs form a small, but logical, example. Jason suggested a program that computes $t_1 = t_0 + \Delta t$.

In general, I think the template overall needs a lot of work to be considered a useful template. It does not follow the same structure as the other examples (it only provides `Main.hs` and `Body.hs`).

### DIFF
--- a/code/drasil-example/template/lib/Drasil/Template/Body.hs
+++ b/code/drasil-example/template/lib/Drasil/Template/Body.hs
@@ -7,6 +7,8 @@ module Drasil.Template.Body (mkSRS, si) where
 
 import Drasil.System (mkSmithEtAlICO)
 import Language.Drasil
+import Language.Drasil.Display
+import Language.Drasil.ShortHands (lT)
 import Drasil.SRSDocument
 import Drasil.DocLang (tunitNone)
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
Sure, I had suggested something even simpler, but this will do too!

balacij:
I think we should flesh out the template a tiny bit more. Adding a `DataDefinition`, displaying it and generating _just_ Python code would be great in my opinion.

JacquesCarette:
Do we need to flesh it out more in *this* PR? I don't like seeing all these Draft PRs around.

balacij:
> Do we need to flesh it out more in this PR? I don't like seeing all these Draft PRs around.

No, if we're fine with this code as it is, then this is enough to unblock #4953

djarabek:
I have more of the changes ready, however I can hold them for a separate PR if we would like to.

balacij:
A new PR is fine. @JacquesCarette can merge, this is enough to unblock #4953.

PR #4961, opened 2026-05-09
Fix #3415: Remove extra spaces after references in Jupyter output

### LINKED ISSUE #3415: Extra Spaces Added after references in generated Jupyter Notebook
When adding references in the Organization of Document section for #3318, the Jupyter Notebook generated adds extra spaces between the punctuations and the references. 

image

image

Is it ok to have the spaces, or should it be fixed?

The discussion where it was first identified can be found [here](https://github.com/JacquesCarette/Drasil/pull/3407#discussion_r1190332161)
@smiths @samm82

Purav-Kanda:
The Jupyter notebook backend was importing reflinkInfo from HTML.Helpers, which uses <+> (space-separator) before the info argument. When citations had no extra info (Cite2 EmptyS, the common case), this produced output like '<a>...</a> ,' with a trailing space before punctuation.

Mirrors HTML/Print.hs's special case for Cite2 EmptyS, routing those references through plain reflink instead. As a side effect, citation links now render as cleaner Markdown ([name](#ref)) instead of HTML <a> tags inside the notebook source.

Verified across all examples: projectile, glassbr, pdcontroller, ssp, dblpend, gamephysics, sglpend, swhs, swhsnopcm, template. No changes to HTML or LaTeX output.


JacquesCarette (approved):
Very nice

PR #4962, opened 2026-05-11
Update file-io to 0.1.6

djarabek:
Related to #4934

A bug [1] in 0.1.5 causes issues writing files with Unicode characters

[1] https://github.com/haskell/file-io/issues/45

### DIFF
--- a/code/drasil-build-artifacts/package.yaml
+++ b/code/drasil-build-artifacts/package.yaml
@@ -24,7 +24,7 @@ dependencies:
 - pretty
 - prettyprinter
 - template-haskell
-- file-io
+- file-io >= 0.1.6
 - os-string
 
 ghc-options:
--- a/code/drasil-build-artifacts/stack.yaml
+++ b/code/drasil-build-artifacts/stack.yaml
@@ -2,3 +2,7 @@ resolver: lts-22.44
 
 packages:
 - .
+
+extra-deps:
+# lts-22.44 includes 0.1.5, but we need 0.1.6+ due to https://github.com/haskell/file-io/issues/45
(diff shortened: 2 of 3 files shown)


balacij:
Nice!

Is it possible for us to move this to `drasil-build-artifacts/stack.yaml` and get the same results?

While we're at this, I think we should move the `setLocaleEncoding utf8` code to the `writeFiles` function in `drasil-build-artifacts`. We should then refine the comments in `drasil-build-artifacts` to clarify it will exclusively deal with UTF-8.

djarabek:
> Is it possible for us to move this to `drasil-build-artifacts/stack.yaml` and get the same results?

(as discussed offline) it needs to be in both, I have done this in https://github.com/JacquesCarette/Drasil/pull/4962/commits/3fe3142ecd6420a463f7c392250b418590061423.

I also added a version constraint in `package.yaml` so that stack (and hopefully in the future, cabal) will fail to build with the outdated/broken version.

> While we're at this, I think we should move the `setLocaleEncoding utf8` code to the `writeFiles` function in `drasil-build-artifacts`. We should then refine the comments in `drasil-build-artifacts` to clarify it will exclusively deal with UTF-8.

I don't think this can be done until after everything uses `writeFiles`, as otherwise some files may be written before `setLocaleEncoding` is called.

balacij:
> I don't think this can be done until after everything uses writeFiles, as otherwise some files may be written before setLocaleEncoding is called.

Sounds good to me!

balacij (approved):

JacquesCarette (approved):

PR #4972, opened 2026-05-12
refactor(srs-gen): Use new `FileLayout` for laying out artifacts

djarabek:
Depends on #4962 (otherwise, we get broken output)

Similar to #4941, #4942

The only changes to generated output is that generated GraphViz files now have a newline at the end of the file - which is a good thing.


code/drasil-gen/lib/Drasil/Generator/SRS/TraceabilityGraphs.hs
```diff
@@ -1,32 +1,31 @@
...
-  mkOutput gi "allvsall" edgesAllvsAll [assumpNF, ddNF, tmNF, gdNF, imNF, reqNF, gsNF, chgNF]
+-- | Creates a list of `FileLayout`s for the generated graphs.
+outputDot :: GraphInfo -> [FileLayout Doc]
+outputDot gi =
+  [ mkOutput gi "avsa" edgesAvsA [assumpNF],
+    mkOutput gi "avsall" edgesAvsAll [assumpNF, ddNF, tmNF, gdNF, imNF, reqNF, chgNF],
+    mkOutput gi "refvsref" edgesRefvsRef [ddNF, tmNF, gdNF, imNF],
+    mkOutput gi "allvsr" edgesAllvsR [assumpNF, ddNF, tmNF, gdNF, imNF, reqNF, gsNF],
+    mkOutput gi "allvsall" edgesAllvsAll [assumpNF, ddNF, tmNF, gdNF, imNF, reqNF, gsNF, chgNF]
+  ]
```
balacij: Should return a single directory instead of a list of files.

code/drasil-gen/lib/Drasil/Generator/SRS/TraceabilityGraphs.hs
```diff
@@ -1,32 +1,31 @@
...
 
 -- | General output function for making a traceability graph. Takes in the graph information, title, edge generator functions, and node family functions.
-mkOutput :: GraphInfo -> String -> (GraphInfo -> [(UID, [UID])]) -> [GraphInfo -> NodeFamily] -> IO ()
+mkOutput :: GraphInfo -> String -> (GraphInfo -> [(UID, [UID])]) -> [GraphInfo -> NodeFamily] -> FileLayout Doc
 mkOutput gi ttl getDirections getLabels =
-  writeFile (ttl ++ ".dot") (render $ mkDot ttl (getDirections gi) (map ($ gi) getLabels))
+  file [ps|{ttl}.dot|] (mkDot ttl (getDirections gi) (map ($ gi) getLabels))
```
balacij, line 30: Not for this PR at all, but whenever we see hardcoded `.X` such as here, we should remind ourselves that there really ought to be an intermediate phase that knows what encoding to output.

This work is not crucial right now, but it is good work to do eventually.

code/drasil-gen/lib/Drasil/Generator/SRS.hs
```diff
@@ -34,90 +35,63 @@ exportSmithEtAlSrs syst srsDecl srsFileName = do
...
-  mapM_ (prntDoc srs printfo srsFileName) [HTML, TeX, Jupyter, MDBook]
-  genDot syst' -- FIXME: This *MUST* use syst', NOT syst (or else it misses things!)!
+  let srsLayout    = directory [ps|SRS|] $
+                       map (prntDoc srs printfo srsFileName)
+                       [HTML, TeX, Jupyter, MDBook]
+      traceyLayout = directory [ps|TraceyGraph|]
+                       (outputDot (mkGraphInfo syst')) -- FIXME: This *MUST* use syst', NOT syst (or else it misses things!)!
+  -- FIXME: Ultimately, there should be a single writeFiles call.
+  writeFiles localPath srsLayout
+  writeFiles localPath traceyLayout
```
balacij, line 50: As we discussed in person, this will need to be merged in a subsequent PR that cleans up `dumpEverything` (it was a good abstraction at the time for checking for the env. variables, but bad now, it needs to be done elsewhere) and the Makefile so that we can output a single directory.

Also, we might need to refactor these `export` functions, splitting them in two:

* lay
* write

This way the composed artifact generators (the ones for the SRS+Code generator) can reuse `lay` at least. But there might be alternative solutions. This is just a naive first guess we discussed.

code/drasil-gen/lib/Drasil/Generator/SRS.hs
```diff
@@ -34,90 +35,63 @@ exportSmithEtAlSrs syst srsDecl srsFileName = do
...
-                  prntCSV  pinfo
+    directory (outPath fmt) $ (doc : mMakefile) ++ extraFiles fmt
+  where
+    doc = prntDoc' fn fmt d pinfo
+    mMakefile = maybeToList $ prntMake $ DocSpec fmt fn
+
+    outPath HTML    = [ps|HTML|]
+    outPath TeX     = [ps|PDF|]
+    outPath Jupyter = [ps|Jupyter|]
+    outPath MDBook  = [ps|mdBook|]
```
balacij: We should create a `toPathSeg :: Format -> String` function and use that here (first capturing the output in a variable).

code/drasil-gen/lib/Drasil/Generator/SRS.hs
```diff
@@ -34,90 +35,63 @@ exportSmithEtAlSrs syst srsDecl srsFileName = do
...
 
--- | Common error for when an unsupported SRS format is attempted.
+    extraFiles HTML   = [ file [ps|{fn}.css|] $ makeCSS d ]
+    extraFiles MDBook = [ file [ps|book.toml|] $ makeBook d pinfo,
+                          file [ps|.drasil-requirements.csv|] $ makeRequirements pinfo
+                        ]
+    extraFiles _      = []
```
balacij: We discussed this as well, but what this looks like (to me) is evidence that this function is doing too much at once. It should be exporting a generic "document project" that has holes for a few generic operations we might expect from a Makefile/Nix file/scripts/etc.: build (which can be a no-op!), view (which can spin up an HTTP server using Python with a hotreloading script injected or open HTML in your we browsers, or open your local PDF viewer using `xdg-open`, `open`, or...), deliverable (which packages a project into a single zip file for distribution), and clean (which cleans up any intermediate files needed to _build_ the document, e.g., TeX or a virtualenv for the Python server).

These holes are _options_ that would vary in permissible configurations about the encoding of the artifact. For example, you wouldn't spin up a Python server to view your PDF, but you might for HTML.

I'm not saying that we implement these options, but I am saying we should split this into two generator phases.

code/drasil-gen/lib/Drasil/Generator/SRS.hs
```diff
@@ -34,90 +35,63 @@ exportSmithEtAlSrs syst srsDecl srsFileName = do
...
-  hClose outh2
-
--- | Helper for generating the .toml config file for mdBook.
-prntBook :: Document -> PrintingInformation -> IO()
-prntBook doc sm = do
-  outh <- openFile fp WriteMode
-  hPutStrLn outh $ render (makeBook doc sm)
-  hClose outh
-  where
-    fp = "SRS/mdBook/book.toml"
```
balacij, line 112: This is exactly the kind of `diff` I was hoping to see!

code/drasil-gen/lib/Drasil/Generator/SRS.hs
```diff
@@ -34,90 +35,63 @@ exportSmithEtAlSrs syst srsDecl srsFileName = do
...
+-- | Internal: Creates a `FileLayout` for the Makefile of an SRS,
```
balacij: Thank you for following the `Internal:` prefix comment style. I think this is a good pattern to follow and we should maybe recommend it in our wiki.

code/stable/dblpend/TraceyGraph/allvsall.dot
```diff
@@ -85,4 +85,4 @@ digraph "allvsall" {
         rank="same";
         "instance:motionMass";
     }
-}
\ No newline at end of file
+}
```
balacij, line 88: Indeed, these are great! And a result of unifying the file writing code.

balacij (changes requested):
(Posting our discussion in-person here as well.)

djarabek:
Oh, CI hasn't run on this PR since it doesn't target main... I based it off `updateFileIo` (as those changes are required to get correct output), and I set the target branch as `updateFileIo` so that the unrelated changes wouldn't show in this PR (GitHub will automatically update the target to `main` when the other PR is merged). I guess the current actions setup doesn't work with this though.

code/drasil-gen/lib/Drasil/Generator/SRS.hs
djarabek: I pushed https://github.com/JacquesCarette/Drasil/commit/119104ab13dfcad5a40b527a970a344f890859d4 which cleans up `dumpEverything` a bit.

code/drasil-gen/lib/Drasil/Generator/ChunkDump.hs
```diff
@@ -7,42 +9,28 @@ import Control.Lens ((^.))
...
 
 -- FIXME: This is more of a general utility than it is drasil-database specific
-dumpTo :: ToJSON a => a -> TargetFile -> IO ()
-dumpTo d targetFile = do
-  trg <- openFile targetFile WriteMode
-  LB.hPutStrLn trg $ encodePretty d
-  hClose trg
+dumpTo :: ToJSON a => a -> PathSegment -> FileLayout Doc
+dumpTo d targetPath =
+  file targetPath (text $ LB.unpack $ encodePretty d)
```
balacij, line 36: Instead of `unpack`ing the bytestring and re-writing it using `pretty`, we should create an instance for `Renderable` in `drasil-file-handling`.

balacij: Agh, we can't do that! We need them all to write to the same `Doc` language.

balacij: ~~Ok, for now this is fine, but we should replace `aeson-pretty` usage with our own `Encoding -> Pretty.Doc` renderer.~~

For now, this is fine, but it would be good in a subsequent PR to:

1. Remove `aeson-pretty` dependency.
2. Replace `encodePretty` with: `renderValue . toJSON` where `renderValue :: Value -> Doc` is our own JSON value renderer.

Since nothing debug-related is committed to `stable/`, we could also do this in this PR fairly easily, but I don't want that to slow down _this PR_.

code/stable/sglpend/SRS/mdBook/.drasil-requirements.csv
```diff
@@ -1,8 +1,8 @@
...
-../../../../datafiles/sglpend/sglpend.jpg,src/assets/sglpend.jpg
 ../../../../datafiles/sglpend/SystemContextFigure.png,src/assets/SystemContextFigure.png
+../../../../datafiles/sglpend/sglpend.jpg,src/assets/sglpend.jpg
+../../../../traceygraphs/sglpend/allvsall.svg,src/assets/allvsall.svg
+../../../../traceygraphs/sglpend/allvsr.svg,src/assets/allvsr.svg
 ../../../../traceygraphs/sglpend/avsa.svg,src/assets/avsa.svg
 ../../../../traceygraphs/sglpend/avsall.svg,src/assets/avsall.svg
 ../../../../traceygraphs/sglpend/refvsref.svg,src/assets/refvsref.svg
-../../../../traceygraphs/sglpend/allvsr.svg,src/assets/allvsr.svg
-../../../../traceygraphs/sglpend/allvsall.svg,src/assets/allvsall.svg
```
balacij, line 8: The changes to the `.drasil-requirements.csv` files are due to needing to rewrite how these are gathered. TLDR: Instead of using using `LabelledContent` gathered through the `PrintingInformation`, searching the generated files (the `[LayoutObj]` lists, specifically).

(See the `assetMat` function.)

balacij:
Because of this refactor, we can remove `_lbldCntnt :: [LabelledContent]` from `PrintingInformation` and `SmithEtAlSRS`!!

And we've removed `DocSpec` because of it, too!

Excellent!

djarabek (commented):
Can't approve my own PR, but your changes look great Jason.

balacij (approved):
This PR is especially good because it brings a lot of issues to light and (in my opinion) shows us what we should be doing next:

1. The SRS code generator is used as part of the composed code generator for the SRS+Code[Zoo] generator. We will need to split up `exportSmithEtAlSrs` into one function that renders an SRS into a `[FileLayout Doc]` and one that does that in addition to writing it to disk. This way we can reuse the former function in part of a composed-variant of the latter that pairs it with rendered code as well.
2. Our SRS generator is also really an SRS Zoo generator. We should rename this.
3. `drasil-printers` is very heavily tied to the SRS generator. It's only because of this PR that I was able to write #4983, #4990, #4989, and #4975 (!). And I'm already working on [more stuff](https://github.com/JacquesCarette/Drasil/compare/artifactReprSRS...artifactReprSRS').
4. It highlights _where_ we want to expose more options (the `Formats` data type) for output-type-specific rendering options.
5. It shows us we were previously doing unsafe file-writing of the SRS. Previously the code was too okay with producing code that would overwrite files (bad, in part because it doesn't get rid of old/deleted files) or more generally doesnt require a fresh directory to write artifacts into. The `Makefile` orchestrates building the `build/exampleName/` directories. That should be done by the binary!

Overall, the PR is in good shape and looks ready to merge in my opinion. It does exactly what it says.

One notable issue I'm not overlooking but don't want to address here: It relies on a hack in `exportSmithEtAlSrs` to write to the same directory, but the way forward is not to fix that in this PR, but to fix (5) in a new PR, which will in turn remove this hack.

balacij (approved):
Re-approving because I just patched a bug that needed fixing: this breaks `make x_tex` -- [383eea3](https://github.com/JacquesCarette/Drasil/pull/4972/commits/383eea3413ca1f7bb8e5d305b961e35106e459d0)

JacquesCarette (approved):

PR #4973, opened 2026-05-12
Remove `Type` Type Family

B-rando1:
Contributes to #4360. The removal of type families seems to be necessary to allow us to remove the global `GOOLState`, so this is a good start. Thanks to @balacij for picking out `TypeSym` as a potential starting point!


JacquesCarette (approved):
We're going to have to do some serious cleanup after all this refactoring. I'm seeing a few not-so-nice things slip in (like new uses of HughesPJ). But I don't want to hold this up just for that.

Issue #4974, opened 2026-05-12
Re-splitting `drasil-printers` to decouple `drasil-lang`, `drasil-code`, and `drasil-printers`


balacij:
## Background

See #4509 

Currently, `drasil-lang` contains modules that belong in `drasil-code` (the ones prefixed with `Drasil.Code`). For example, [`CodeExpr`](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-lang/lib/Drasil/Code/CodeExpr.hs). They are "stowaways" in `drasil-lang` to avoid a circular dependency between `drasil-code`, `drasil-printers`, and `drasil-code`:

* `drasil-printers` relies on `CodeExpr` and the other modules in `Drasil.Code` (namely code variables) for describing how `CodeExpr` can be rendered into `Doc`s for human-readable, non-code purposes. ASCII variants, that is.
* `drasil-code` relies on `drasil-printers` for said rendering function and more (e.g., rendering functions for `ModelExpr`, `Sentence`, `Symbol`, etc.).

If we were to move `CodeExpr` from `drasil-lang` to `drasil-printers`, a cycle appears because `drasil-printers` contains the rendering logic for `CodeExpr` (i.e., depends on it).

## Objective

To refactor the dependency graph such that we can move the `drasil-code`-related stowaways in `drasil-lang` into `drasil-code`.

## Analysis

`drasil-printers` is very high up in our dependency graph. Why? Because it imports from a bunch of other packages:

- `drasil-build-artifacts`
- `drasil-database`
- `drasil-lang`
- `drasil-metadata`
- `drasil-theory`
- `drasil-utils`

This brings us to a major issue: `drasil-printers` does too much. It contains two things:

1. A scientific document typesetting language (similar to `Pandoc` but a bit more specialized for our needs).
2. Interpreters that typeset various expression languages (`Sentence` and `*Expr`, i.e., natural language and math language).

The part where `drasil-printers` "does too much" is really in regards to (2). This is an extensibility throttle for `drasil-printers` and something that ensures that it will always be high up in the dependency graph. We need to break this up by moving the renderers (or, typewriters/docwriters) elsewhere, such as somewhere closer to the definitions of the things being typeset. That is, `Sentence`-renderer towards `Sentence`, etc.

Now, can we make this move today and would it solve our dependency graph? Partially and not yet.

<details>

<summary>

[...]

PR #4975, opened 2026-05-13
refactor: remove `DType` type

balacij:
Contributes to #4974 and theory-related work (no issue)

This PR makes the `drasil-printer`'s `Definition` table take less information about what kind of thing is being defined. Previously it permitted allowed Theory, Data(Definition), General(Definition), and Instance(Model). This can be extended far past these 4, but even then, why? This information was only used to discriminate between `Definition`s for which CSS style class should be applied in the HTML renderer. The other renderers ignored this information. Furthermore, the CSS for the 4 types was the same!

This PR deletes `DType` and avoids reliance on it for stylizing definition tables with what CSS we normally apply.

As a result of this, we decrease coupling between `drasil-printers` and `drasil-lang`.


samm82 (approved):
These changes and their rationale make sense to me, and everything compiles on my end! Not sure if you want someone closer to this discussion to also approve it before merging it 😅

smiths (approved):
This change makes sense to me.  The new versions of stable look the same as the old ones (as far as I can tell).

JacquesCarette (approved):

PR #4980, opened 2026-05-15
refactor(code-gen) Use `FileLayout` for laying out artifacts

### LINKED ISSUE #4610: There should be no output `IO ()` outside of `drasil-gen`

djarabek:
Closes #4610


code/drasil-gen/lib/Drasil/Generator/Code.hs
```diff
@@ -30,42 +32,36 @@ import Language.Drasil.GOOL (unPP, unJP, unCSP, unCPPP, unSP, unJLP)
...
+    layout <- exportCode syst chcs
+    return $ directory [ps|{dir}|] [layout]
 
 -- | Calls the code generator.
-genCode :: Choices -> CodeSpec -> IO ()
+genCode :: Choices -> CodeSpec -> IO (FileLayout Doc)
 genCode chs spec = do
-  workingDir <- getCurrentDirectory
   time <- getCurrentTime
   sampData <- maybe (return []) (\sd -> readWithDataDesc sd $ sampleInputDD
```
djarabek, line 38: The somewhat awkward `IO (FileLayout Doc)` is needed because of `getCurrentTime` and reading the sample data. These two things should probably be done somewhere else to allow this function to be pure.

code/drasil-gen/lib/Drasil/Generator/Code.hs
```diff
@@ -30,42 +32,36 @@ import Language.Drasil.GOOL (unPP, unJP, unCSP, unCPPP, unSP, unJLP)
...
 -- | Internal: Generate a zoo of ICO-style executable softifact.
-exportCodeZoo :: SmithEtAlSRS -> [Choices] -> IO ()
-exportCodeZoo syst = mapM_ $ \chcs -> do
-  let dir = map toLower $ codedDirName (syst ^. programName) chcs
-  workingDir <- getCurrentDirectory
-  createDirIfMissing False dir
-  setCurrentDirectory dir
-  exportCode syst chcs
-  setCurrentDirectory workingDir
+exportCodeZoo :: SmithEtAlSRS -> [Choices] -> IO [FileLayout Doc]
```
djarabek: We need to return a `[FileLayout Doc]` here, because we need to create a number of directories while not being responsible for creating the parent of those directories.

code/drasil-example/projectile/lib/Drasil/Projectile/Choices.hs
```diff
@@ -32,7 +32,7 @@ choiceCombos = [
     optFeats = makeOptFeats
       (makeDocConfig [CommentFunc, CommentClass, CommentMod] Quiet Hide)
       (makeLogConfig [LogVar, LogFunc] "log.txt")
-      [SampleInput "../../../datafiles/projectile/sampleInput.txt", ReadME],
+      [SampleInput "../../datafiles/projectile/sampleInput.txt", ReadME],
```
djarabek: This in fact revealed a bug, the use of `setCurrentDirectory` in `exportCodeZoo` results in an extra `../` being needed, as the program switches it's working directory for each `Choices` in the zoo.

code/drasil-gen/lib/Drasil/Generator/Code.hs
```diff
@@ -30,42 +32,36 @@ import Language.Drasil.GOOL (unPP, unJP, unCSP, unCPPP, unSP, unJLP)
...
   let code = codeSpec syst chcs
   genCode chcs code
 
 -- | Internal: Generate a zoo of ICO-style executable softifact.
-exportCodeZoo :: SmithEtAlSRS -> [Choices] -> IO ()
-exportCodeZoo syst = mapM_ $ \chcs -> do
-  let dir = map toLower $ codedDirName (syst ^. programName) chcs
-  workingDir <- getCurrentDirectory
-  createDirIfMissing False dir
-  setCurrentDirectory dir
```
djarabek: The last `setCurrentDirectory` is dead.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -124,22 +124,52 @@ generator l dt sd chs cs = let
...
+      initFile = [file [ps|__init__.py|] mempty | l == Python]
+      packageFiles = toFileLayout (map hasPathAndDocToFileAndContents (progMods prog) ++ progDossier)
+        ++ designLogFile ++ initFile
+  in
+    directory
+      [ps|{dirName}|]
+      packageFiles
+
+-- HACK?
+toFileLayout :: [FileAndContents] -> [FileLayout Doc]
```
djarabek: This is a big hack, though it may be difficult to avoid, given how much the `[FileAndContents]` (and similar) representation is used. An observation I made however is that every langauge except Java (currently) lays out all the files in a single directory (all files are `src/*`). Java does this as well, except the `.java` files are in a nested folder (to make them a `package`).

code/drasil-example/projectile/lib/Drasil/Projectile/Choices.hs
balacij: Great catch! This might be worth pushing to another PR.

code/drasil-code/lib/Language/Drasil/Code/PackageData.hs
```diff
@@ -3,18 +3,18 @@ module Language.Drasil.Code.PackageData (PackageData(packageProg, packageAux),
...
 
 -- | The underlying data type for packages in all renderers.
-data PackageData = PackD {packageProg :: ProgData, packageAux :: [FileAndContents]}
+data PackageData = PackD {packageProg :: ProgData, packageAux :: [FileLayout Doc]}
 
-pattern PackageData :: ProgData -> [FileAndContents] -> PackageData
+pattern PackageData :: ProgData -> [FileLayout Doc] -> PackageData
 pattern PackageData prog aux <- PackD prog aux
   where
-    PackageData prog aux = PackD prog (filter (not . isEmpty . fileDoc) aux)
```
djarabek, line 16: This was a gross hack to support platforms that didn't support doxygen. `SoftwareDossierSym` has been updated to use `Maybe`, so we no longer need to filter out empty files.

code/drasil-code/test/Main.hs
```diff
@@ -62,13 +62,15 @@ main = do
 
 -- | Gathers all information needed to generate code, sorts it, and calls the renderers.
 genCode :: [PackageData] -> IO()
-genCode files =
-  createCodeFiles $ files >>= \(PackageData prog aux) ->
+genCode files = do
```
djarabek, line 68: This test needs further cleanup, I made the minimal changes to get it to pass.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -124,22 +122,52 @@ generator l dt sd chs cs = let
...
+  in
+    directory
+      [ps|{dirName}|]
+      packageFiles
+
+-- HACK?
+toFileLayout :: [FileData] -> [FileLayout Doc]
+toFileLayout fc =
+  let
+    toEntries [] m = m
```
JacquesCarette: flip the order of the parameters (most rapidly varying parameters go further right)
this is a `foldl`, no?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -124,22 +122,52 @@ generator l dt sd chs cs = let
...
+insertFile :: (String, a) -> M.Map String (Entry a) -> M.Map String (Entry a)
+insertFile (p, d) m =
+  if '/' `elem` p
+    then
+      let (fname, rest) = break (== '/') p
+          folderM = case M.findWithDefault (Folder M.empty) fname m of
+                      File _   -> dupError fname
+                      Folder f -> f
+      in M.insert fname (Folder $ insertFile (drop 1 rest, d) folderM) m
+    else M.insertWith
```
JacquesCarette: put this on a single line, there's no need for 3 lines here.

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
```diff
@@ -124,22 +122,52 @@ generator l dt sd chs cs = let
...
+
+    root = toEntries fc M.empty
+
+    entryToLayout (n, File d) = file [ps|{n}|] d
+    entryToLayout (n, Folder m) = directory [ps|{n}|] $ map entryToLayout $ M.assocs m
+  in
+    map entryToLayout (M.assocs root)
+
+data Entry a = File a | Folder (M.Map String (Entry a))
+  deriving (Show)
```
balacij, line 152: This recreates `FileLayout`. Can we avoid creating this? Does GOOL/`drasil-code` need to change more to accommodate the code we want to write?

code/drasil-gen/lib/Drasil/Generator/Code.hs
balacij: Returning a `[FileLayout Doc]` is okay. I de-couple file generation from file-writing in #4994 as well, making similar change.

code/drasil-gen/lib/Drasil/Generator/Code.hs
balacij: Where is the timestamp used?

code/drasil-gool/lib/Drasil/Shared/AST.hs
```diff
@@ -64,10 +63,6 @@ data FileData = FileD {filePath :: FilePath, fileMod :: ModData}
 fileD :: FilePath -> ModData -> FileData
 fileD = FileD
 
-instance HasPathAndDoc FileData Doc where
-  getPath = filePath
-  getDoc = modDoc . fileMod
-
```
balacij, line 69: :)

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/Legacy/FileData.hs
```diff
@@ -1,22 +0,0 @@
...
--- | Constructor for auxiliary files.
-fileAndContents :: FilePath -> Doc -> FileAndContents
-fileAndContents = FileAndContents
-
-instance HasPathAndDoc FileAndContents Doc where
-  getPath = filePath
-  getDoc = fileDoc
-
-hasPathAndDocToFileAndContents :: (HasPathAndDoc a Doc) => a -> FileAndContents
-hasPathAndDocToFileAndContents file = fileAndContents (getPath file) (getDoc file)
```
balacij, line 22: Extremely happy to see this!!

code/drasil-gen/lib/Drasil/Generator/Code.hs
djarabek: A date is optionally included in the header of generate code files: https://github.com/JacquesCarette/Drasil/blob/808465017da07c15f038b1408d2f831d144bd778/code/drasil-code/lib/Language/Drasil/Choices.hs#L222-L223

We don't currently test this, as it would make our stable artifacts unstable (perhaps something to consider for the golden testing library? some sort of way of 'masking' areas we expect to be different?)

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
djarabek: GOOL gives us a `ProgData` which has `progMods`, which is a list of `FileData` (which each have a `String` `filePath` and a `FileMod` `modData` (which contains a `Doc`)). We would need a tree returned instead of a flat list to avoid this.

JacquesCarette (approved):
Outside the items that @balacij pointed out, I'm fine with this.

djarabek:
@JacquesCarette ready for review

code/drasil-code/test/Main.hs
djarabek: #5013 Will address this

code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs
balacij: We discussed this in person. We definitely want to remove this code long-term, but doing it now will be a huge hassle.

Once this PR is merged, @djarabek, can you please move this discussion to an issue with a longer-form description?

balacij (approved):
Great work!

JacquesCarette (approved):

PR #4983, opened 2026-05-15
refactor(-printers): avoid using `HDiv` hack in Jupyter Notebook generation of equation blocks

balacij:
`LayoutObj`'s `HDiv` constructor is a hack that allows HTML documents to nest sections in a giant `div` block. This kind of constructor feels it should be avoided entirely (not sure how to do this yet).

This PR removes one unnecessary use of `HDiv` to print out single-cell equation blocks in the Jupyter Notebook generator.

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
+++ b/code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
@@ -79,12 +79,11 @@ printLO CodeBlock {}                    = empty
 
 -- printLO' is used for generating general notebook (lesson plans)
 printLO' :: LayoutObj -> Doc
-printLO' (HDiv ["equation"] layoutObs _)  = markdownCell $ vcat (map printLO' layoutObs)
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Note that having a "document division" constructor, a la HTML's DIV, is not in itself a bad idea.

code/drasil-printers/lib/Language/Drasil/JSON/Print.hs
```diff
@@ -79,12 +79,11 @@ printLO CodeBlock {}                    = empty
 
 -- printLO' is used for generating general notebook (lesson plans)
 printLO' :: LayoutObj -> Doc
-printLO' (HDiv ["equation"] layoutObs _)  = markdownCell $ vcat (map printLO' layoutObs)
```
JacquesCarette, line 82: The only reason your change "works" is because the label part is actually ignored! In theory, this is a change of semantics, at least on the 'import' side of things.

JacquesCarette (approved):

JacquesCarette:
Looks like the merge of this broke the build!?

PR #4984, opened 2026-05-19
Renamed all occurrences of `static`/`dynamic` to `classLevel`/`instanceLevel`

### LINKED ISSUE #4977: Switch "Static Method" Terminology to "Class Method"
In #4964, @smiths recommended that we switch our terminology from talking about "Static" methods to "Class" vs "Object" methods. This is less confusing for two reasons: first, it much more intuitively indicates whether the method is tied to the class or instances of the class; and second, it prevents name collisions with other things that use static - e.g. [Binding](https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-gool/lib/Drasil/Shared/AST.hs#L49).

B-rando1:
Closes #4977.

Summary of changes:
- Updated the main `PermanenceSym` typeclass:
  - Renamed `PermanenceSym` to `AttachmentSym`. This was one of names I had suggested in the issue, since the mental picture it evokes is similar to "Binding", without associations to other PL concepts.
  - Renamed the methods of `AttachmentSym` from `static`/`dynamic` to `classLevel`/`instanceLevel`. I added "Level" to the end because "class" and "instance" are still to generic on their own. `classAttachment`/`instanceAttachment` could also work, but "Level" seems good enough to me.
  - Renamed `LanguagePolymorphic`'s common implementations for `classLevel`/`instanceLevel` methods to `static`/`dynamic`.
- Updated the ADTs that back `AttachmentSym`:
  - Renamed `Binding` (the ADT backing `AttachmentSym` for most renderers) to `AttachmentTag`. I couldn't use `Attachment` because it was already taken by the `AttachmentSym` typeclass, and I couldn't use `AttachmentData` because that's a separate record type used by the C++ renderer (see next point). `AttachmentTag` seems like a reasonable name.
  - Renamed `BindData` (the ADT backing `AttachmentSym` for the C++ renderer) to `AttachmentData`.
- Reworked the terminology of variables in `OOVariableSym`:
  - Renamed `objVar` and `classVar` to `objVarAccess` and `classVarAccess`. This frees up their original names (see next point), and more accurately reflects what they do. `objVarAccess`, given an object `o` and an instance-level variable (name only) `v`, returns `o.v`. Likewise, `classVarAccess`, given a class `C` and a class-level variable (name only) `v`, returns `C.v`. Using "access" matches the terminology of the List API, i.e. with `listAccess`.
  - Broke up `staticVar'` into `classVar` and `classConst`. The amount of time I spent figuring out what `staticVar` even does is a testament to the confusing terminology we're using; hopefully this will improve things. Basically, `staticVar'` does the same thing as `var`, except (1) it marks the variable as class-level in its internal representation (`var` marks it as instance-level, which is slightly incorrect in some cases but fine for now), and (2) the first argument is a boolean for whether it's a static constant - the Python renderer converts constant names to uppercase. I decided to make these decisions more explicit by splitting up and renaming this method to `staticVar` and `staticConst`.

This ended up being a larger set of changes than I anticipated, but I think the naming is much more consistent and accurate now.


JacquesCarette (approved):

JacquesCarette:
I'm going to merge, even though we're not 100% sure the CI is fully working. It will be a test...

Issue #4989, opened 2026-05-22
Rearchitecting `drasil-printers`


balacij:
## Issues

`drasil-printers` has some ongoing issues:

1. It does not support many output options.
2. It does not use a modern pretty-printing library.
3. It does too many things at once (in its rendering phase).
4. It knows about the SRS and other parts of the code generator too much.
5. It's missing a testing suite.

### It does not support many output options.

See the following code snippets:

1. https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-printers/lib/Language/Drasil/TeX/Print.hs#L44-L46
2. https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs#L35-L37
3. https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-printers/lib/Language/Drasil/JSON/Print.hs#L30-L54
4. https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-printers/lib/Language/Drasil/HTML/Print.hs#L50-L84

These are the main entry-points of the renderers that `drasil-printers` exposes. The LaTeX generator is the only one with an option, showing a table of contents or not. This is arguably not an option that `drasil-printers` should be exposing; document formation should have completed by the time we want to do typesetting.

However, we do have some nuggets of thought for options:

https://github.com/JacquesCarette/Drasil/blob/317e28de0ccecb54868ccc6a4a4cd6fec5d37873/code/drasil-printers/lib/Language/Drasil/Config.hs#L1-L65

The above file is used to provide options for the LaTeX generator. However, it is not really exposed by `drasil-printers`. Tweaking it would require tweaking `drasil-printers`.

There are many more options we can and should be able to support, such as:

* The "naive" ones: nesting depth, whitespace usage, unicode availability, line length, etc.
* The output-format-specific ones: which node types to use, flattening divs, jupyternoteboook-cell-usage policy, TeX output format, which markdown flavour to use, when to use HTML elements instead of markdown counterparts, label/identifier naming policy, etc.
* Stylization.

### It uses an outdated pretty-printing library.

[...]

PR #4990, opened 2026-05-22
init(`drasil-data-formats`): scaffold new Drasil subproject with `CSV` renderer

balacij:
Contributes to #4989 and #4475 

`drasil-data-formats` is a package for building and rendering data in various formats, such as:

* CSV
* TOML
* YAML
* JSON
* etc.

CSV is the only one initialized for now.

Note: `drasil-data-formats` is NOT intended for anything parsing related. It should only be used to produce as many output data formats as possible, supporting as many output options as possible (e.g., whitespace policy, minification, quotation policy, number-writing policy, etc.).

The goal with this is to begin gradually re-building `drasil-printers` while supporting more kinds of generation options and using the modern pretty-printing library. This PR specifically is intended to replace `makeCSV`:

https://github.com/JacquesCarette/Drasil/blob/70a4917667b99792f2a48883a0c4cea5b87e34e7/code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs#L102-L119

We are deliberately using our own library here because:

1. We want to demonstrate as many possible output options as possible. We do not have good control of external libraries.
2. We want to have a fairly uniform API for doing everything render-related. Again, we do not have good control over external libraries. That being said, we can create our own wrappers atop others.
3. Using our own libraries will be completely optional.


balacij:
The remaining work task for this PR is to build a golden testing suite.

balacij:
Also, note: I did my best to be [RFC4180](https://datatracker.ietf.org/doc/html/rfc4180)-conformant. We should add this to the comments. This resource was great for building the code.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
```diff
@@ -0,0 +1,110 @@
...
+    CSVRenderOptions
+  )
+where
+
+import Data.List (find, intersperse)
+import Data.Text (Text)
+import qualified Data.Text as T (any, pack, replace, splitOn)
+import Prettyprinter (Doc, Pretty (..), comma, dquotes, hardline, hcat, vcat)
+
+-- | A CSV file representation.
```
JacquesCarette: add comments that this is meant to be "optional header row plus list of rows".

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
```diff
@@ -0,0 +1,110 @@
...
+data CSV = CSV (Maybe [Text]) [[Text]]
+  deriving (Show, Eq)
+
+-- | Get the header row of a 'CSV'.
+header :: CSV -> Maybe [Text]
```
JacquesCarette: why not use 
```haskell
data CSV = CSV {header :: Maybe [Text], rows :: [[Text]]}
```
so that you get these accessors for free?

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
```diff
@@ -0,0 +1,110 @@
...
+header (CSV mhr _) = mhr
+
+-- | Get all rows of a 'CSV'.
+rows :: CSV -> [[Text]]
+rows (CSV _ rs) = rs
+
+-- | Create a 'CSV'. Expects all rows to have the same length as the header (if
+-- it exists) or the first row (if the rows are non-empty).
+mkCSV :: Maybe [Text] -> [[Text]] -> Either String CSV
```
JacquesCarette: Actually, why not have
1. an `Int` parameter to `mkCSV` that asks the user for the intended length (since the header could be wrong, even when present)
2. keep that `Int` in the representation (even if redundant) ? [that way `rowCount` becomes a projection instead of a computation]

Also, don't inline the coherence verifier directly in the constructor, make that a separate function.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
```diff
@@ -0,0 +1,110 @@
...
+      ]
+  where
+    (expLen, expLenSrc) = maybe (length fr, "first row") (\r -> (length r, "header")) mhr
+mkCSV mhr [] = pure $ CSV mhr []
+
+-- | Calculate the number of rows in a 'CSV'.
+rowCount :: CSV -> Int
+rowCount (CSV _ rs) = length rs
+
+-- | Options for rendering a 'CSV'.
```
JacquesCarette: I would split all this rendering-related stuff into a separate file.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
```diff
@@ -0,0 +1,110 @@
...
+escapeCellPolicy :: DoubleQuotationPolicy -> Text -> Doc ann
+escapeCellPolicy Minimal t
+  | T.any needsQuote t = quoteAndEscape t
+  | otherwise = escHls t
+escapeCellPolicy Everywhere t = quoteAndEscape t
+
+-- | Internal: Check if a character appearing in a cell indicates that the cell
+-- /must/ be quoted.
+needsQuote :: Char -> Bool
+needsQuote c = c == '"' || c == ',' || c == '\n' || c == '\r'
```
JacquesCarette:
```haskell
needsQuote = `elem` "\",\n\r"
```
or something like that.

JacquesCarette (changes requested):

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
balacij: Those accessors are usable for overwriting the data carried within the records, which is dangerous because it means it can break the invariants we specifically intended to uphold through the smart constructors.

Unfortunately, it just looks like a GHC/Haskell language restriction. No way to declare that an accessor cannot be used for writing, only reading.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
balacij: Addressed in ffeafe36b299119f8d4cad9bcd828c40d81225d7

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
balacij: Addressed in 463e9a4561b9a4d7c85520eb538f1c80ba318dae

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
balacij: Addressed in b88782f20501236777da45fe58d3f7bde4b1f779

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
balacij: Addressed:

1 and 2: ef8193eab9732e30a680ecd2c9bee6db3ac539c8

And: df6307e796dee0291a499ac2d38802fa3087d203

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV.hs
JacquesCarette: fair

JacquesCarette (approved):

balacij:
Converting back to draft because there is (a) a merge conflict and (b) @djarabek has a testing suite partially completed that we might as well include in this PR.

balacij:
Confirmed that `stack test` works:

```console
drasil-data-formats> test (suite: drasil-data-formats-test)

Drasil.Data.Formats
  Drasil.Data.Formats.CSV
    mkCSV
      Unit Tests
        explicit mismatched header size: OK
        explicit mismatched row size:    OK
        implicit mismatched row size:    OK
        no header mismatched row size:   OK
    CSV
      Unit Tests
        rows simpleCSV:                  OK
        header simpleCSV:                OK
        header simpleCSVnoH:             OK
        rowCount simpleCSV:              OK
        columnCount simpleCSV:           OK
    renderCSV
      Golden Tests
        simpleCSV minimal:               OK
        simpleCSVnoH minimal:            OK
        simpleCSV everywhere:            OK
        simpleCSVnoH everywhere:         OK
        complexCSV minimal:              OK
        complexCSV everywhere:           OK

All 15 tests passed (0.02s)
```

This is ready for review, @JacquesCarette :)

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+    -- ** Constructors
+    mkCSV,
+  )
+where
+
+import Data.List (find)
+import Data.Text (Text)
+
+-- | The number of columns a CSV has.
+type ColumnCount = Int
```
JacquesCarette: Let's take advantage of new code to also do "the right thing" for counts: make them `Natural` instead of `Int`!

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Core.hs
```diff
@@ -0,0 +1,79 @@
...
+
+-- | The number of rows a CSV has (excluding its reader).
+type RowCount = Int
+
+-- | A CSV file representation containing an optional header and a list of rows.
+--
+-- Caches the column and row counts for future potential reference.
+data CSV = CSV (Maybe [Text]) [[Text]] ColumnCount RowCount
```
JacquesCarette: Please make this a record.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Render.hs
```diff
@@ -0,0 +1,71 @@
...
+-- /must/ be quoted.
+needsQuote :: Char -> Bool
+needsQuote c = c `elem` ['"', ',', '\n', '\r']
+
+-- | Internal: Replace all double-quotes with double-double-quotes in a cell and
+-- wrap the whole cell in double-quotes.
+quoteAndEscape :: Text -> Doc ann
+quoteAndEscape = dquotes . escapeHLs . T.replace dqs ddqs
+
+-- | Internal: Double quotes as 'Text'.
```
JacquesCarette: So we're going to have these "Double quotes as 'Text'" all over. Design question is it better to have that (as you have it here, not exported) or to have a single module where all these are collected together?

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Render.hs
```diff
@@ -0,0 +1,71 @@
...
+
+-- | Internal: Double double-qoutes as 'Text'.
+ddqs :: Text
+ddqs = dqs <> dqs
+
+-- | Internal: `prettyprinter` needs us to manually deal with hard linebreaks.
+escapeHLs :: Text -> Doc ann
+escapeHLs = hcat . intersperse hardline . map pretty . T.splitOn lf . T.replace crlf lf
+  where
+    lf = T.pack "\n"
```
JacquesCarette: Can't we use `OverloadedStrings` to not have to worry about writing `T.pack` ?

code/drasil-data-formats/package.yaml
```diff
@@ -0,0 +1,41 @@
+spec-version: 0.36.0
+name:         drasil-data-formats
+version:      0.0.1
+author:       "Jacques Carette, Spencer Smith"
```
JacquesCarette: add yourself here!

JacquesCarette (changes requested):

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Render.hs
balacij: eda466d2428c0fbf2f39562a858a21104cb1cd45

I would say that went well.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Core.hs
balacij: 752c9bdfd4392598818e0d2a3f3371f049c7bf73 & 63dc8c430cb34ab1a81b6509052e438800a62d4b

Good idea.

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Core.hs
balacij: Fixed in 66920b23722c1eb23e33e8b4cdbb2142c6b50e97

code/drasil-data-formats/package.yaml
balacij: Fixed in d10830cef3

code/drasil-data-formats/lib/Drasil/Data/Formats/CSV/Render.hs
balacij: With `OverloadedStrings` as you suggested, I figured we don't really need this to be a top-level definition now.

I guess we can ignore this issue for now until it comes up again (which I'm sure it will in the context of rendering)?

JacquesCarette (approved):

PR #4993, opened 2026-05-23
refactor: unify `Main` program entries of all case studies

balacij:
Builds on #4972

> This PR is broken off from [`artifactReprSRS'`](https://github.com/JacquesCarette/Drasil/tree/artifactReprSRS').

The goal here is to unify the IO parts of the case studies (hopefully only the main program entries) into a series of case study types. The next step will be to de-couple artifact generation/`FileLayout` production from IO.

### DIFF
--- a/code/drasil-example/dblpend/app/Main.hs
+++ b/code/drasil-example/dblpend/app/Main.hs
@@ -1,13 +1,9 @@
 module Main (main) where
 
-import GHC.IO.Encoding
-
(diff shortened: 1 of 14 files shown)


djarabek (dismissed):

JacquesCarette (approved):

JacquesCarette:
Oh, something seems to have gone wrong: there was no evidence of CI having run on this! So this could break the build.

PR #4996, opened 2026-05-24
feat(`drasil-build-artifacts`): create a miniature golden testing suite

### LINKED ISSUE #4965: Investigate using `tasty-golden` for golden testing case studies
[`tasty-golden`](https://github.com/UnkindPartition/tasty-golden) looks like a straightforward alternative to our current Makefile-based golden testing suite for our examples.

We should investigate if switching to it would simplify our Makefile setup. I suspect it would at least make stabilization a more uniform experience between the website and the primary case studies.

In #4934, I created a [tiny `tasty-hunit`-based](https://github.com/JacquesCarette/Drasil/blob/9dceb7e8e9a0da9409737aa09a1c83b3af0230f3/code/drasil-build-artifacts/test/Spec/Drasil/Build/Artifacts/FilePath.hs) testing suite for `drasil-build-artifacts`' (`drasil-file-handling` -- #4960) quasi-quoter. If/when investigating, we should start small -- and with this library.

balacij:
Closes #4965 

Contributes to #4989 

Based on [`tasty-golden`](https://hackage-content.haskell.org/package/tasty-golden-2.3.6).

Unfortunately, `tasty-golden` is just shy of being suitable for our purposes. It's just not quite well-suited to our whole-directory diff'ing style. It prefers we work on a single-test-case-per-test-file-basis, which means a lot more effort.

This PR implements a miniature version of `tasty-golden` for use in our internal libraries. I think we can also make use of it for main case studies, but it would mean we break up our `stable/` folder, or do some `..` hacking to keep the same directory structure.

Results of `stack test drasil-build-artifacts`:

```
Registering library for drasil-build-artifacts-0.1.1.0..
drasil-build-artifacts> test (suite: drasil-build-artifacts-test)

Drasil.Build.Artifacts
  Drasil.Build.Artifacts.FilePath
    ps
      Unit Tests
        ps quasiquoter - simple:                  OK
        ps quasiquoter - interpolation:           OK
        ps quasiquoter - escaped curly brace:     OK
        ps quasiquoter - mixed:                   OK
        ps quasiquoter - multiple interpolations: OK
  Drasil.Build.Artifacts.FileLayout
    writeFiles
      Golden Tests
        Nested files:                             OK
        Single file:                              OK

All 7 tests passed (0.01s)
```

And this is what it would look like if we have a build that breaks `golden`:

```
drasil-build-artifacts> test (suite: drasil-build-artifacts-test)

Drasil.Build.Artifacts
  Drasil.Build.Artifacts.FilePath
    ps
      Unit Tests
        ps quasiquoter - simple:                  OK
        ps quasiquoter - interpolation:           OK
        ps quasiquoter - escaped curly brace:     OK
        ps quasiquoter - mixed:                   OK
        ps quasiquoter - multiple interpolations: OK
  Drasil.Build.Artifacts.FileLayout
    writeFiles
      Golden Tests
        Nested files:                             OK
        Single file:                              FAIL
          Outputs differ:
          --- ./test/golden/hello-world.txt	2026-05-24 12:37:45.422018447 -0400
          +++ ./test/build/hello-world.txt	2026-05-24 12:37:48.819871413 -0400
          @@ -1,2 +1 @@
          -Hello, asdfWorld!
          -this is a string that only appears on the golden side
          +Hello, World!

          Use -p '/Single file/' to rerun this test only.

1 out of 7 tests failed (0.02s)
```

If accepted, before merging, I would like to test this a little bit more with #4990 .


code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -0,0 +1,112 @@
...
+goldenTestingGroup buildPath goldenPath groupName mkTests =
+  withResource setup (\_ -> return ()) (\_ -> testGroup groupName (map (\gt -> unGTC gt buildPath goldenPath) mkTests))
+  where
+    setup = do
+      createDirectoryIfMissing True $ localPath </> buildPath
+      createDirectoryIfMissing True $ localPath </> goldenPath
+
+-- | Internal: Tasty golden test representation.
+--
+-- 1. Adds an @--accept@ CLI option.
```
balacij, line 70: The only issue I really have with this: the complete command would be `stack test --test-arguments="--accept"`.

A `Makefile` _might_ be in order, but I would prefer we didn't have one per `drasil-` subproject...

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -0,0 +1,112 @@
...
+        removePathForcibly goldenTarget
+        action goldenPath
+        pure $
+          testPassed $
+            "Accepted fresh artifacts as golden for: " ++ goldenTargetStr ++ "."
+      else do
+        removePathForcibly buildTarget
+        action buildPath
+        (exitCode, stdout, stderr) <-
+          readProcessWithExitCode "diff" ["-ru", "--color=always", "--strip-trailing-cr", goldenTargetStr, buildTargetStr] ""
```
balacij: Note: This output is written to stdout instead of being logged to a `.log` file as we currently do. I think this is fine, but I think it's worth mentioning as a prospective future feature at least.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -0,0 +1,125 @@
...
+        action goldenPath
+        pure $
+          testPassed $
+            "Accepted fresh artifacts as golden for: " ++ goldenTargetStr ++ "."
+      else do
+        removePathForcibly buildTarget
+        action buildPath
+        (exitCode, stdout, stderr) <-
+          readProcessWithExitCode
+            "diff"
```
djarabek, line 100: I had concerns about if this would work on windows, but assuming an appropriate unix environment (cygwin/msys2) it should work. Should definitely have someone test, there might be some weirdness with line endings for instance.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -0,0 +1,125 @@
+-- | A golden testing suite based on @tasty-golden@, specialized to 'FileLayout'
```
djarabek, line 1: Should this file be in the `lib` folder? This introduces testing code (and dependencies) to all the packages that use `drasil-build-artifacts`.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -0,0 +1,125 @@
...
+          testPassed $
+            "Accepted fresh artifacts as golden for: " ++ goldenTargetStr ++ "."
+      else do
+        removePathForcibly buildTarget
+        action buildPath
+        (exitCode, stdout, stderr) <-
+          readProcessWithExitCode
+            "diff"
+            [ "-ru",
+              "--color=always",
```
djarabek, line 102: Will this cause issues in environments that don't support color?

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
balacij: We definitely need to test it on Windows, I agree.

If we have an active Windows user developing Drasil and their `make` is working fine, the `diff` used by this testing suite would be the same `diff` used by the Makefile, since the Makefile executes this code and the environment is shared. So, this should also work fine, including the line ending differences.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
balacij: Yes. This is a library feature that I would want our other libraries to make use of.

We could move this to a separate package, if we even want this code, that is.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
balacij: Changing to `auto` is probably a good idea at least, but I don't think there's anything we can do in those situations where terminals dont support color.

balacij: [96ec884](https://github.com/JacquesCarette/Drasil/pull/4996/commits/96ec884334a4230fd9f2de0114dc49cf4e207f9d)

balacij: Discussed in person: `auto` is not quite enough either. But, this is primarily helpful for active development (i.e., locally), so leaving colourization always on is probably a good idea. We can always add a environment flag for it later to disable it if needed, just don't see the necessity right now.

djarabek (approved):

JacquesCarette (commented):
This seems like a good idea to me.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
balacij: Addressed in 21183c093443ee755cdf4a90bd5551b758876b8a in a reasonable (imo) way: reminding the user how to accept failed tests! 😄

djarabek (approved):
Tested in my windows VM and it works. I found (and fixed) a different issue though: https://github.com/JacquesCarette/Drasil/pull/5011

JacquesCarette (approved):

balacij:
Thank you for testing it, @djarabek !!

PR #4998, opened 2026-05-25
Removed More Type Families

B-rando1:
Contributes to #4360.

There are a couple more hacks in the same vein as in #4973. These should disappear at a later date - either by replacing `CodeInfoX` with a better-designed type of pass, or by using functional dependencies to bring back `()` as the repr for these types in `CodeInfoX`.


JacquesCarette (approved):
Very nice simplification. Unsurprisingly, merge conflicts to resolve.

B-rando1:
Conflicts are resolved now.

JacquesCarette (approved):

Issue #4935, opened 2026-04-28
Move GOOL Arrays/Lists toward Container-Style Interfaces


B-rando1:
The idea here is that we want GOOL to present an interface for "indexable structures" akin to [C++ Containers](https://en.cppreference.com/cpp/container), rather than giving them access to "raw" arrays and arraylists. These container APIs could have multiple implementations per-language: e.g. for Java, "resizable containers" could be implemented as [ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html)s, [LinkedList](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedList.html)s, [Vector](https://docs.oracle.com/javase/8/docs/api/java/util/Vector.html)s, etc. We can give the user the high-level choice of which implementation to use, or let them not choose and choose a default for them.

Currently our thinking is that there should be two interfaces: one for "indexable structures" and one for "indexable structures that can grow". The other option would be to have them all under one umbrella, but either throw an error or switch to a dynamic implementation if the user chooses a static implementation. But it might be good to keep them as separate concepts, since situations will likely arise where the user wants to use static arrays where possible and dynamic arrays where not. Assuming we keep them separate ideas, the latter could also be viewed as a subtype of the former. Implementing it that way might be difficult in GOOL, though. I'd need to think about it some more.

There's clearly a bit of architectural work to be done here, but I think a good place to start is to choose a more abstract name than "array" and "list" and rename the current GOOL typeclasses accordingly. @JacquesCarette likes the word "Container", which seems fine to me. I guess we could call the one for "arrays" `Container` and the one for "dynamic arrays" `ResizableContainer`?

B-rando1:
One other note, on the "static array" side this will be somewhat blocked by #4929. We can't rename arrays until we have an actual typeclass for them!

JacquesCarette:
Likely more than two interfaces -- lists are not really indexable (or at least, shouldn't be!)

I'd likely prefer `FixedSizedContainer` and `ResizableContainer` for things that are indexable.

B-rando1:
Related: #1988

PR #4999, opened 2026-05-25
Created Array Typeclass

B-rando1:
Contributes to #4929, serving as a foundation for future work.

I'm not entirely sure how I feel about the addition of the `IndexTranslator` typeclass, but it makes sense to me that both `List` and `Array` should inherit from these functions under our current architecture. The only reason it's useful is to let us use the same implementations for our Julia (1-indexed) lists/arrays and our other (0-indexed) languages. We could achieve the same level of flexibility by moving `IndexTranslator` to `RendererClassesCommon` (where we put most of our "internal" GOOL API) or by having our common array implementations take an index-translator function. Or we could just have separate implementations for Julia; iirc it's only a handful of functions that are shared.


JacquesCarette (approved):
Let's go with this, we can always revisit if it turns out to be a bad idea.

I think it needs a merge though.

B-rando1:
Conflict is resolved now.

I agree that this seems to be a good intermediate step that will allow us to choose "arrays" as opposed to "lists/dynamic arrays" until we can give that choice at a higher level.

If we can enable GOOL to have multiple passes, I think it could be an interesting project to consolidate GOOL's `List` and `Array` typeclasses into a single `IndexableContainer` and have the user supply an ordered list of preferred concrete datatypes. We could then have a pass that checks which operations are being performed and finds the data structure highest on the list that can supply them. E.g. if the user ranks "C-style array" higher than "dynamic array" but they do a resizing operation on it, then GOOL will notice that and use "dynamic array" (and tell the user that it made that decision). This is one direction that we could take with #4935.

It'll be a bit before we can try that, though, so as an intermediate step I think this makes sense. Particularly for vectors, since whatever high-level interface we design for them, static arrays are probably the most sensible "default option" for the underlying data structure for most languages.

JacquesCarette (approved):

JacquesCarette:
Good to remember the "long range" plans (multiple passes) as something we're really after, and that these many refactorings are really steps towards such bigger goals.

PR #5001, opened 2026-05-25
Misc. GOOL Array Improvements

B-rando1:
Contributes to #4929.

Summary:
- Made `varDec` throw an error for arrays in C++. For that renderer specifically, it would generate broken code. Also added comments to `varDec` and `varDecDef` in `InterfaceCommon` explaining to use `arrayDec` and `arrayDecDef` for arrays.
- Fixed `litArray` in Java, so now it works both at declaration-time and after declaration.
- Fixed `arrayElem` in Julia. Previously it used the name `arr[i]` for the variable's "name" (used by the variable map), which meant that the Julia renderer couldn't find the scope of the "variable" if it was assigned to. Now it uses `arr` for the "name". It's a bit of a hack, but doesn't break anything.
- Fixed `arrayDec` in Python, Julia, and Swift. Previously `arrayDec` would create an empty list, which meant that assigning to a specific index of the "array" would cause an error. Now they create a list of `0`s of the specified length.
- Added tests for `arrayDecDef` and `arrayDec`.


code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/AbstractProc.hs
```diff
@@ -84,7 +84,7 @@ arrayElem :: (ProcRenderSym r) => SValue r -> SVariable r -> SVariable r
 arrayElem i' v' = do
   i <- IC.intToIndex i'
   v <- v'
-  let vName = variableName v ++ "[" ++ render (RCC.value i) ++ "]"
+  let vName = variableName v -- Slight hack; we used to add `++ "[" ++ render (RCC.value i) ++ "]"`
```
B-rando1, line 87: I'm not actually sure if I should call this a hack. `variableName` is only used for variable map lookups, e.g. to find if a variable name has been taken when generating a fresh name and to find the scope of a variable to know if the `global` keyword is required in Julia. As far as variable map lookups are concerned, we should just be using `variableName v`, not the rest of it. So maybe this is an improvement, and we should update the other implementations to match?

JacquesCarette:
Conflicting now, needs a merge.

B-rando1:
Conflict resolved (pending CI)!

JacquesCarette (approved):

PR #5002, opened 2026-05-25
Fix main build

balacij:


JacquesCarette (approved):
These diffs look good to me, but linter is not happy?

JacquesCarette (approved):

JacquesCarette:
Still not quite fixed. This PR indeed fixed the problem identified, but now:
> glassbr: Path already exists: "./SRS"
> CallStack (from HasCallStack):
>  error, called at lib/Drasil/Build/Artifacts/FileLayout.hs:82:10 in drasil-build-artifacts-0.1.1.0-Fv1Z0wnk9I7YkBR05DKw3:Drasil.Build.Artifacts.FileLayout

PR #5003, opened 2026-05-26
fix: make sure `make deploy` works locally

balacij:
Hopefully this will fix the CI. A bit of trial and error work here.


JacquesCarette (approved):

JacquesCarette:
Sigh - build failed, but likely because of network troubles. I re-started it.

Issue #4997, opened 2026-05-25
What to do about C++ arrays?


B-rando1:
With my work on #4929, I keep coming across issues with C++ arrays.

There are two related problems with arrays in C++: (1) how to know the length of the array when we need to allocate it, and (2) how to keep track of the length of the array after that point. Neither of these are issues in our other languages: (1) because they're garbage-collected and can freely heap-allocate an array at creation-time (our C++ renderer stack-allocates everything, which among other things means that array sizes need to be known ahead of time); and (2) because their arrays are wrapped in an object containing the length of the array.

From looking around and talking to @balacij, I've come across ~4 ways of handling arrays in C++:
- Stack-allocated arrays. This is what our renderer currently uses. There are two ways of creating them:
  ```c++
  int arr[4]; // Note that the size must be given at declaration-time
  int arr2[] = {1, 2, 3};
  ```
  Currently we have a separate `arrayDec` that takes an extra argument compared to `varDec`, used to tell C++ the size of the array. This is sufficient for most things (as long as we have a constant or a variable we can refer to for the size), but it doesn't work if we want a stack-allocated array in a class: for classes the size of the array needs to be known at compile-time. We could use templates to overcome this, but GOOL doesn't support templates.
  The other issue with C++ arrays is that in general, there doesn't seem to be a consistent way to get the length of them. In the scope they're created, we can use `sizeof(arr) / sizeof(arr[0])` to get their size, but my understanding is that if an array is passed to a function, it's converted to a bare pointer and that won't work anymore. There are two things I can think of that could work with this: either rely on the GOOL writer to supply the array length wherever it's required, or (2) somehow generate code for storing the array length in an extra variable. Neither of these are ideal, though.
- Pointers to heap-allocated arrays. These are more flexible than stack-allocated arrays because we don't need to know the length until we need to allocate them; but they have the same issue where we need to keep track of their length separately. And the larger issue is that our C++ renderer is not set up for heap-allocation.
- Stack-allocated [std::array](https://en.cppreference.com/cpp/container/array). These seem like a really cool datatype - under the hood I believe they compile to C-style arrays, but they use templates to give us static knowledge of their length. They do require us to give the length as a template-variable, but that in principle would be manageable. The only issue is that we don't support templates, so they're off the table.
[...]

PR #5004, opened 2026-05-26
Made `print` work correctly for GOOL `Array`s

B-rando1:
GOOL has a set of polymorphic print functions that generate different code based on the type of value they're given. They treat `List`s and `Set`s very specially, but they naively print arrays like primitives, which for most languages results in printing a memory address. This PR fixes the implementations for those languages, generating (hopefully) idiomatic print statements for arrays in each one.


code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
```diff
@@ -2786,6 +2786,23 @@ cppPrint newLn pf vl = do
...
+          elemVar <- genVarName [] "elem"
+          let innerType = listInnerType (onStateValue valueType v)
+              printMaybeNewLn = if newLn then printLn else print
+          multi [printStr "[",
+            varDecDef (var firstVar bool) local litTrue,
+            forEach (var elemVar innerType) v (body [block[
+              ifNoElse [((valueOf (var firstVar bool) ?!), oneLiner $ printStr ", ")],
+              print $ valueOf (var elemVar innerType),
+              var firstVar bool &= litFalse]]),
+              printMaybeNewLn $ litString "]"]
```
B-rando1, line 2803: Admittedly, this is probably not the most idiomatic way of doing this. And it likely only works in the scope the variable was declared in, where the C++ compiler can statically know the length of the array and generate the necessary iterator code...

Solving #4997 would help us generate better code here lol

JacquesCarette (approved):

PR #5012, opened 2026-05-27
Fix build errror in golden testing

djarabek:


JacquesCarette (approved):

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs
```diff
@@ -123,5 +123,5 @@ goldenTest tName layout = GTC $
           buildRoot
           goldenRoot
           (toPath (name layout))
-          (`writeFiles` layout)
+          (\p -> writeFiles NeverOverwrite p layout)
```
balacij, line 126: This should just be `OverwriteAllowed` but this is good.

balacij: Nope, disregard. I forgot that I orchestrated cleaning in the Haskell code!

balacij (approved):

PR #5025, opened 2026-05-28
Added a Logging Renderer to GOOL

### LINKED ISSUE #4985: Referring to GOOL syntax within GOOL programs
See My [comment](https://github.com/JacquesCarette/Drasil/issues/4929#issuecomment-4445211720) from #4929 for the initial motivation. The crux of the issue there is that we currently have code that can generate (just by enabling variable logging for `swhsnopcm`, for example) this:

```java
outfile.print("var 'dT_W[0]' assigned ");
outfile.print(dT_W[0]);
outfile.println(" in module ODE");
```

In the first line, we print as a string `dT_W[0]` - a piece of GOOL syntax. This is a weird case because it's a different level of use of the syntax. The way we currently get away with it is by exposing from GOOL a method `variableName` that converts a GOOL `SVariable` to a `String`. But this seems like a somewhat sketchy and unnecessarily limited solution to the problem of "referring to GOOL syntax within GOOL programs".

That we can only do this with `SVariables` is limiting and causing problems in #4929. I tried changing `arrayElem` to return an `SValue` instead of an `SVariable`. This is because we want to have multiple ways of instantiating in code what we currently call `Array`s in GOOL, some of which will involve explicit getters and setters rather than LValues. But because we want to be able to log array assignments, I wasn't able to make that work.

One solution for this would be to refine and generalize the current approach, working with `SValue`s instead of `SVariable`s.
- We could do this either externally (e.g. expose a `valueToString` method in GOOL)
- Or we could do it internally (e.g. `litString` could take `List (Either String SValue)` instead of `String` and convert everything to `String` on the fly. A bit clunky perhaps, but it would do the trick.

As another option, we could explicitly identify the patterns where we need to quote GOOL syntax, and create more tailored implementations for each. The benefit of this is that the print statements could be fine-tuned by GOOL based on downstream choices.  E.g. we could have a method like this:

```hs
-- | Given an array `a`, index `i`, and value `v`, log a message saying that `a[i]` was updated to hold `v`.
arrayAssignedLog :: SValue -> SValue -> SValue -> [MSStatement]
```

Then if the user chooses to represent GOOL `Array`s as the current, C-style arrays, `arrayAssignedLog` would produce a set of statements much like what we have now. But if the user chose `ArrayList` as the representation, then it could be changed to something like this:
```java
outfile.println("dT_W[0] updated: called `dT_W.set(0, " + dT_W.get(0) + ")` in module ODE");
```

B-rando1:
Closes #4985.

I created a new renderer, called `LoggingFor`, that converts a select subset of GOOL/GProc syntax to `Doc`. Its instances are mostly polymorphic over the language renderer, but for a few cases it dispatches based on the specific target language to provide the correct syntax.

A couple of notes:
- In `drasil-code`, the function `maybeLog` that previously took an `SVariable r` now takes an `SVariable LoggerCode` _and_ an `SVariable r`. It uses the former to print the name of the variable and the latter to print its value. It's a bit clunky, but I'm not sure if there's a nice way to get the same variable as an instance of both renderers.
- I Chose to give instances for `OOProg` and `ProcProg`, with mostly `undefined` implementations for their typeclasses, because `mkVar` and other means of creating a variable that `drasil-code` uses have `OOProg` or `ProcProg` as their type constraint. We could narrow those to be more specific so that we don't need all those empty instances, but I think this is ok too.


JacquesCarette:
> because mkVar and other means of creating a variable that drasil-code uses have OOProg or ProcProg as their type constraint. We could narrow those to be more specific so that we don't need all those empty instances,

We should definitely have our constructors ask for just the features they need, and not ask for too much. So this is a refinement that should be done (separately, and for as many of the constructors as possible). This would make it easier to use sub-languages of `OOProg` and `ProcProg`, which is desirable.

code/stable/glassbr/src/cpp/InputParameters.cpp
```diff
@@ -42,79 +42,79 @@ void InputParameters::get_input(string filename) {
     infile >> this->a;
     infile.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
     outfile.open("log.txt", std::fstream::app);
-    outfile << "var 'this->a' assigned ";
+    outfile << "var 'self.a' assigned ";
```
JacquesCarette: I forget my C++: is this a correct change? I know this is just in text, but since we're trying to get correct code to print, we should indeed do that!

code/stable/glassbr/src/csharp/InputParameters.cs
```diff
@@ -62,70 +62,70 @@ private void get_input(string filename) {
         infile.ReadLine();
         this.a = Double.Parse(infile.ReadLine());
         outfile = new StreamWriter("log.txt", true);
-        outfile.Write("var 'this.a' assigned ");
+        outfile.Write("var 'self.a' assigned ");
```
JacquesCarette: Ditto, is this correct C#?

code/stable/glassbr/src/java/GlassBR/InputParameters.java
```diff
@@ -68,70 +68,70 @@ private void get_input(String filename) throws FileNotFoundException, IOExceptio
         infile.nextLine();
         this.a = Double.parseDouble(infile.nextLine());
         outfile = new PrintWriter(new FileWriter(new File("log.txt"), true));
-        outfile.print("var 'this.a' assigned ");
+        outfile.print("var 'self.a' assigned ");
```
JacquesCarette: This definitely feels like bad Java!

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs
```diff
@@ -0,0 +1,455 @@
...
+        vrDoc = unLC vr'
+    return $ return $ clsDoc <> text "." <> vrDoc
+  extClassVarAccess = classVarAccess
+  instanceVarAccess ob vr = do
+    ob' <- ob
+    vr' <- vr
+    return $ return $ unLC ob' <> text "." <> unLC vr'
+  instanceVarSelf vr = do
+    vr' <- vr
+    return $ return $ text "self." <> unLC vr'
```
JacquesCarette: aha! This is the source of issues. Okay, what you are going to need is for `OOVariableSym` to dispatch to the right language. What this means is that `LoggerCode` should have 2 variables, with the other one indicating the language.

JacquesCarette (changes requested):
This is generally going in a good direction. It will need a more refined approach though.

code/stable/glassbr/src/cpp/InputParameters.cpp
B-rando1: This is incorrect - because I just wrote one renderer to get feedback on the design, there are a few places like this where its syntax differs from some of the languages.

code/stable/glassbr/src/csharp/InputParameters.cs
B-rando1: This is also incorrect - see my above comment.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs
B-rando1: Ah, that's probably an even better idea than having multiple renderers. It would definitely lead to less code duplication.

B-rando1: I did some looking around at [Instances](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/instances.html), trying to reduce duplicate code as much as possible. I have this:

```hs
-- ...
{-# LANGUAGE FlexibleInstances -}
-- ...

instance {-# OVERLAPPABLE #-} OOVariableSym (LoggerCode lang) where
  classVar = var
  classConst = constant
  self = return $ return $ text "self"
  classVarAccess cls vr = do
    cls' <- cls
    vr' <- vr
    let clsDoc = (typeDoc . unLC) cls'
        vrDoc = unLC vr'
    return $ return $ clsDoc <> text "." <> vrDoc
  extClassVarAccess = classVarAccess
  instanceVarAccess ob vr = do
    ob' <- ob
    vr' <- vr
    return $ return $ unLC ob' <> text "." <> unLC vr'
  instanceVarSelf vr = do
    vr' <- vr
    slf <- self
    return $ return $ unLC slf <> unLC vr'

instance {-# OVERLAPPING #-} OOVariableSym (LoggerCode JavaCode) where
  self = return $ return $ text "this"
```

But I still get a couple of warnings and errors. In `instanceVarSelf` when it calls `self`, it complains that there are overlapping instances for `self` (even though it has the `OVERLAPPABLE` flag); and in the `JavaCode` instance it warns that the instance is incomplete. @JacquesCarette or @TOTBWF do you know if there are (reasonable) ways to get around these errors? Or am I better off just abstracting these into functions and calling them separately for each renderer?

JacquesCarette: Having a 'generic' version and then some specialized version is always going to cause you grief.

What I would do:
1. split `OOVariableSym` into more classes (in particular, to isolate `self` into its own class)
2. don't use `OVERLAPPABLE` at all in the results. Either implement generically or for each printer.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs
```diff
@@ -0,0 +1,237 @@
...
+import Drasil.GOOL.LanguageRenderer.SwiftRenderer (SwiftCode)
+import Drasil.Shared.LanguageRenderer (dot)
+
+import Text.PrettyPrint.HughesPJ (Doc, text, comma, space, brackets, braces,
+  punctuate, hcat)
+import qualified Text.PrettyPrint.HughesPJ as P (char, integer, float, double)
+import Data.List (intercalate)
+import Data.Kind (Type)
+
+newtype LoggerCode (lang :: Type -> Type) a = LC {unLC :: a} deriving Functor
```
B-rando1: It might be good to rename this to something more informative. I'm currently drawing a blank; does anyone have ideas?

balacij: I could be completely off, but, isn't this for Pseudocode creation partially specialized for each language? Is there any problem with `Pseudocode`?

JacquesCarette: It shouldn't be `Pseudocode`, it should be code, that's the whole point of this design...

Why not just `Logging`? Or `LoggingFor` since it is parametrized by `lang` ?

JacquesCarette (approved):

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs
balacij: https://github.com/JacquesCarette/Drasil/blob/8c2ab23de277adf2ed460cc07a47e5de910d16a9/code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs#L116-L145

^ Is this meant to be used for all GOOL-supported languages? I'm going off of this.

JacquesCarette: Yes - and clearly this is wrong. It should dispatch to the underlying language to get its job done.  Probably should look like
```haskell
instance TypeSym lang => TypeSym (LoggerCode lang) where
  bool = bool
  int = int
```
and so on.

B-rando1: Ah yes, that should be an elegant solution. The above `TypeSym` implementation was just something I quickly wrote to see if this works, and I forgot to get back to it.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/Logger.hs
```diff
@@ -0,0 +1,168 @@
...
+import Drasil.GOOL.LanguageRenderer.CppRenderer (CppCode, CppSrcCode, CppHdrCode)
+import Drasil.GOOL.LanguageRenderer.PythonRenderer (PythonCode)
+import Drasil.GOOL.LanguageRenderer.SwiftRenderer (SwiftCode)
+import Drasil.Shared.LanguageRenderer (dot)
+
+import Text.PrettyPrint.HughesPJ (Doc, text, empty, comma, space, brackets,
+  braces, punctuate, hcat)
+import qualified Text.PrettyPrint.HughesPJ as P (char, integer, float, double)
+
+newtype (SharedProg lang) => LoggerCode lang a = LC {unLC :: a} deriving Functor
```
B-rando1: When Reed looked at this `(SharedProg lang)` constraint on Friday, he seemed very surprised that it's possible to put those on a `newtype`, and said that this isn't doing what I want it to be doing. I think he was mistaken though, because re-adding it allowed me to leave out a bunch of constraints on the typeclass instances.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LoggingFor.hs
```diff
@@ -25,25 +25,23 @@ import Drasil.Shared.LanguageRenderer (dot)
 import Text.PrettyPrint.HughesPJ (Doc, text, empty, comma, space, brackets,
   braces, punctuate, hcat)
 import qualified Text.PrettyPrint.HughesPJ as P (char, integer, float, double)
-import Data.List (intercalate)
-import Data.Kind (Type)
 
-newtype LoggerCode (lang :: Type -> Type) a = LC {unLC :: a} deriving Functor
+newtype (SharedProg lang) => LoggingFor lang a = LC {unLC :: a} deriving Functor
```
JacquesCarette: This `SharedProg` constraint means that `LoggingFor` will only be available for renderers of "whole programs". This is an over-constraint that will become a pain down the road.

You should instead use the more verbose version that puts 'tight' constraints exactly where `lang` is actually used, and nowhere else.

JacquesCarette (changes requested):

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LoggingFor.hs
B-rando1: I changed the constraint back to `Type -> Type`, and it compiled without issues. I'm a little confused now, I definitely had to add a bunch of extra constraints yesterday before I changed the constraint to `(SharedProg lang)`. But I guess less constraints needed is a good thing :smile:

JacquesCarette (approved):

JacquesCarette:
Can you merge/rebase please?

B-rando1:
Done!

JacquesCarette (approved):

Issue #4970, opened 2026-05-12
Enable `-Wmissing-export-lists` GHC option in `drasil-example`


balacij:
See #4932 for an example.

Note: You will need to make sure that you only export the things that are actually imported by other packages.

camishibata:
Enable `-Wmissing-export-lists` GHC option in `drasil-example/`:    
- [x] `dblpend`
- [x] `gamephysics`
- [ ] `glassbr`
- [ ] `hghc`
- [ ] `pdcontroller`
- [ ] `projectille`
- [ ] `sglpend`
- [ ] `ssp`
- [ ] `swhs`
- [ ] `swhsnopcm`

PR #5027, opened 2026-05-28
Enable `-Wmissing-export-lists` GHC option in `drasil-example/glassbr` and `drasil-example/hghc`

camishibata:
Contributes to #4970


code/drasil-example/hghc/lib/Drasil/HGHC/Choices.hs
```diff
@@ -1,4 +1,4 @@
-module Drasil.HGHC.Choices where
+module Drasil.HGHC.Choices () where
```
balacij: It doesn't look like there's anything of value worth saving in this module. I think we can (should) just remove it.

JacquesCarette: Agreed!

JacquesCarette (changes requested):

balacij (approved):
Nice!

JacquesCarette (approved):

PR #5029, opened 2026-05-29
Switched C++ renderer to use `std::vector` for `Array`s

### LINKED ISSUE #4997: What to do about C++ arrays?

B-rando1:
Closes #4997

Notes:
- `litList` (literal list) previously threw an error for some reason, even though `{1, 2, 3}` is a valid literal `std::vector`. I made both `litList` and `litArray` use this syntax.
- `argsList` needs a custom type, because its type is `const char**`. When I first set `arrayType = listType`, a bunch of code added `std::vector` imports because they used `argv`. Fortunately we can give it a custom type, and it doesn't seem like a hack.
- Because of syntactic differences, `arrayDecDef` could no longer rely on `arrayDec` for the first part of its implementation. I created `arrayDecBase` for both of them to call. We probably shouldn't be dealing directly with `Doc` for these functions, but that'll need to be a future refactor.
- I switched `print` back to purely using `G.print` from `LanguagePolymorphic`. Because `arrayType` is now an alias for `listType`, `G.print` automatically generates the correct code when it sees `List` as the `CodeType`.

### DIFF
--- a/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
+++ b/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
@@ -22,8 +22,8 @@ import Drasil.Shared.InterfaceCommon (SharedProg, Label, MSBody, VSType,
   extFuncApp, IndexTranslator(..), Array(..), List(..), Set(..), InternalList(..),
   ThunkSym(..), VectorType(..), VectorDecl(..), VectorThunk(..),
   VectorExpression(..), ThunkAssign(..), StatementSym(..), AssignStatement(..),
-  (&=), DeclStatement(..), IOStatement(..), StringStatement(..), FunctionSym(..),
-  FuncAppStatement(..), BinderSym(..), CommentStatement(..), ifNoElse,
+  DeclStatement(..), IOStatement(..), StringStatement(..), FunctionSym(..),
+  FuncAppStatement(..), BinderSym(..), CommentStatement(..),
   ControlStatement(..), ScopeSym(..), ParameterSym(..), MethodSym(..),
   convScope, BinderElim (..))
 import Drasil.GOOL.InterfaceGOOL (CSStateVar, OOProg, ProgramSym(..),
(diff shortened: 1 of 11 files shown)


JacquesCarette (approved):
Very nice.

PR #5033, opened 2026-05-29
init(`drasil-assets`): a package for encoding arbitrary file assets

balacij:
Builds on #5032 

For now, no 'chunk's around, but I believe that's what we want long-term for analysis purposes; e.g., to ask how many external assets does this project rely on?

Where can/should we use this?

Some examples I think this might be useful for:

* https://github.com/JacquesCarette/Drasil/blob/cbcf885d07f6dd787eed724b6515e29409375188/code/drasil-printers/lib/Language/Drasil/HTML/CSS.hs#L10-L140
* As an alternative to our mdBook projects dealing with external asset files:
   https://github.com/JacquesCarette/Drasil/blob/cbcf885d07f6dd787eed724b6515e29409375188/code/scripts/copy_example_assets.sh#L1-L27
* "Sample inputs" (not that they should be treated as arbitrary assets): https://github.com/JacquesCarette/Drasil/blob/cbcf885d07f6dd787eed724b6515e29409375188/code/drasil-example/projectile/lib/Drasil/Projectile/Choices.hs#L38-L46
* Anything and everything in our [`datafiles` folder](https://github.com/JacquesCarette/Drasil/tree/cbcf885d07f6dd787eed724b6515e29409375188/code/datafiles), except for external libraries (because those should be automatically downloaded by package managers or scripts we generate).

However, in general, we should be using this package sparingly. It is a last resort, only to be used when generating something is too difficult or not possible.


code/drasil-assets/lib/Drasil/Assets.hs
```diff
@@ -0,0 +1,64 @@
...
+import Drasil.Build.Artifacts (FileLayout, PathSegment, exactFile)
+import Language.Haskell.TH (Code (examineCode), Q, liftCode, runIO)
+import Language.Haskell.TH.Syntax (Lift, addDependentFile, makeRelativeToProject)
+import System.Directory.OsPath (doesPathExist)
+import System.File.OsPath (readFile')
+import System.OsPath (OsPath, decodeUtf, encodeUtf)
+import Prelude hiding (readFile)
+
+-- | An arbitrary file asset; a 'B.ByteString' with a description.
+data Asset = Asset String B.ByteString
```
JacquesCarette: Shouldn't new packages take the plunge and use `Text` rather than `String`?

code/drasil-assets/package.yaml
```diff
@@ -0,0 +1,48 @@
+spec-version: 0.36.0
+name:         drasil-assets
+version:      0.1.1.0
+author:       "Dan Szymczak, Steven Palmer, Jacques Carette, Spencer Smith"
```
JacquesCarette: For new packages like this, you should put yourself (and only yourself) as author!

code/drasil-assets/lib/Drasil/Assets.hs
```diff
@@ -0,0 +1,64 @@
...
+-- | Get the raw 'ByteString' content of an 'Asset'.
+content :: Asset -> B.ByteString
+content (Asset _ c) = c
+
+-- | Convert an 'Asset' to a 'FileLayout' node given a 'PathSegment' (file name)
+-- it is to be written to.
+toFile :: Asset -> PathSegment -> FileLayout
+toFile a = flip exactFile (content a)
+
+-- | Read a file from disk, providing a description of the asset.
```
JacquesCarette: Feels like this function should not be in the same file as where you define `Asset`. It is an entirely different beast.

JacquesCarette (changes requested):

code/drasil-assets/package.yaml
balacij: Fixed in 9a79e0b00dc5bd95060edfc5314248d936a062e5 .

code/drasil-assets/lib/Drasil/Assets.hs
balacij: Agreed. Thank you for the reminder!

Fixed in 7bebaca.

code/drasil-assets/lib/Drasil/Assets.hs
balacij: Agreed. Fixed in f7b257c.

balacij:
Thank you for the review! It's ready for another one :)

JacquesCarette (approved):

PR #5039, opened 2026-05-30
refactor: eliminate `WritePolicy` in `writeFile` to avoid repeated `case`s

balacij:
Arguably better because it removes repeated `case` statements, but also worse because it increases line count! I'm also not sure if it makes the code more readable.

(I really just hated the repeated `case` statements I introduced in #5031, but I'm okay with not going through with this PR.)

### DIFF
--- a/code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/WriteFiles.hs
+++ b/code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/WriteFiles.hs
@@ -27,19 +27,6 @@ data WritePolicy
   | -- | Or without one?
     ExactBytes
 
--- | Internal: 'WritePolicy' eliminator.
-policy ::
-  -- | Value returned if the 'WritePolicy' is 'AppendNewline'.
-  f ->
-  -- | Value returned if the 'WritePolicy' is 'ExactBytes'.
-  f ->
-  -- | The 'WritePolicy'.
-  WritePolicy ->
-  f
-policy f _ AppendNewline = f
-policy _ f ExactBytes = f
-{-# INLINE policy #-}
-
 -- | Write arbitrary things to a file (respecting a 'WritePolicy').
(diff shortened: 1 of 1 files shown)


code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/WriteFiles.hs
```diff
@@ -27,6 +27,18 @@ data WritePolicy
   | -- | Or without one?
     ExactBytes
 
+-- | Internal: 'WritePolicy' eliminator.
+policy ::
+  WritePolicy ->
```
JacquesCarette: The usual thing to do with eliminators is to put this argument last, not first. Lets you do composition more nicely.

JacquesCarette (changes requested):
I do like this.

code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/WriteFiles.hs
balacij: I was writing this comment after having fixed this:

Much, much better. The code continues to grow in lines, but that's only because I added comments for each argument.

(I don't know why I'm fixating on line count, but I seem to have.)

But then I realized `policy` could be inlined into `writeFile`! 246120b827571e852ed2140fb24b39ab1cfd4118

Even better now! Line count is down + better documentation + no repeated `case` statements.

balacij:
Changed the title and goal of the PR now, but I think it's better.

Ready for review, @JacquesCarette . In particular, please let me know if you would prefer we kept `policy` as well instead of getting rid of it in [246120b](https://github.com/JacquesCarette/Drasil/pull/5039/commits/246120b827571e852ed2140fb24b39ab1cfd4118).

JacquesCarette (approved):
No need - this looks good as is.

PR #5041, opened 2026-06-01
Remove acronyms list from pdcontroller example

### LINKED ISSUE #3538: remove the definitions of `acronyms` from all the examples in drasil-examples
Note: this will most likely not be straightforward. Some of the things in `acronyms` are likely redundant, others will belong elsewhere while others still will be related to infrastructure and should be inserted in the various databases via other means. I definitely expect a number of side-issues to get spawned off of this work.

Purav-Kanda:
Closes #3538

Grepped the full drasil-example directory and confirmed pdcontroller was the only example with an explicit `acronyms :: [CI]` definition. The four CIs (pdControllerCI, proportionalCI, piCI, pidCI) are referenced directly in `ideaDicts` instead, removing the intermediate list. No change to generated output — same elements, same order.


code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
```diff
@@ -110,8 +110,8 @@ ideaDicts =
   -- Actual IdeaDicts
   concepts ++
   -- CIs
-  nw progName : map nw acronyms ++ map nw mathcon'
-
+  nw progName : map nw [pdControllerCI, proportionalCI, piCI, pidCI] ++ map nw mathcon'
```
JacquesCarette: Is it still needed to hand-insert all of them? What happens if you take some of them out of this list?

JacquesCarette (changes requested):

code/drasil-example/pdcontroller/lib/Drasil/PDController/Body.hs
Purav-Kanda: I'll try removing them from the list and running the example to compare the generated output against stable. I'll report back what breaks, if anything.

Purav-Kanda: I tried removing them and the example crashes with Term: pdControllerCI not found in TermMap. So they still need to be listed explicitly. Is there a way the infrastructure could handle this automatically?

balacij: Hi @Purav-Kanda Thank you for contributing!

It looks like you might have slightly misunderstood the goal of the issue. My [original comment](https://github.com/JacquesCarette/Drasil/issues/3538#issuecomment-1942867350) discussing it a bit more is a bit dated now because we now have `CommonKnowledge` to help: https://github.com/JacquesCarette/Drasil/blob/1e815e60c154ef1130ddc38ceebcd9cfc824923c/code/drasil-gen/lib/Drasil/Generator/CommonKnowledge.hs#L22-L103

Note that I'm definitely not saying to move PDControllers `CI`s to`drasil-gen`. Those look like they should stay where they are, but perhaps that they should be reduced to normal `IdeaDict`s.

Also, this issue applies to all examples in `drasil-example`. It might help to survey what elements are in these `acronyms` lists first. You can discuss your findings here if you'd like some help :)

Purav-Kanda: surveyed map nw usage across all example Body.hs files — here's what i found, let me know if i'm on the right track:
possibly should move to basisIdeaDicts:

mathcon' (de, leftSide, ode, pde, rightSide) — shows up in 8/9 examples and doesn't seem to be in the basis yet
physicCon' (oneD, twoD, threeD) — 6/9 examples, same situation. also noticed gamephysics re-exports threeD/twoD from drasil-data through its own Concepts.hs, so that might be cleanable too

seems example-specific, would stay local:

progName — every example but unique per example
centreMass (gamephysics), phsChgMtrl (swhs/swhsnopcm), iGlass/lGlass (glassbr)
pdcontroller's 4 CIs — i'm going to check if their abbreviations actually show up in the generated SRS; if they don't i'm thinking they could be downgraded to plain IdeaDict, but not sure

if this looks right, my plan would be to add mathcon' and physicCon' to basisIdeaDicts and remove them from the individual examples — but wanted to check before touching anything

balacij: Adding `mathcon'` and `physicCon'` to `basisIdeaDicts` sounds sensible to me! I would start with that and then unpack more of your findings w.r.t. the other ones in the main issue.

Purav-Kanda: Based on the survey from this PR, opened #5061 to add mathcon' and physicCon' to basisIdeaDicts since they appeared in 8/9 and 6/9 examples respectively. Will continue investigating the remaining findings (pdcontroller CIs, etc.) in issue #3538.

Purav-Kanda: Based on balacij's note, the real fix is downgrading these 4 CIs to plain IdeaDict if their abbreviations aren't used in the generated output. I'll investigate that in #3538. For now they need to stay since removing them crashes the example. Happy to address this in a follow-up PR.

balacij:
Since you mentioned you would be investigating this elsewhere, I think we can close this PR then, @Purav-Kanda ?

To make a long-story short with the investigation, you would need to check which one of those `CI`s actually need to be `CI`s and which can just be `IdeaDict`s without harming output.

Purav-Kanda:
Closing — will continue in #3538.

Issue #5037, opened 2026-05-29
Should we golden test our case studies with #4996's code?


balacij:
In #4996, we introduced a miniature golden testing suite. Early usage in #5033 and #4990 shows that it works well.

Our case studies are also golden tested but that testing is done with the `Makefile` and a bash script.

Question:

1. Should we convert our case study Makefile-based golden testing to one based on #4996?
2. If so, should we also break up stable?

There are some reasons for (1) at least, such as:

1. We simplify our Makefile.
2. We make dealing with the the golden testing suites uniform. We currently have `stable/`, `stable-website/`, and the new `drasil-**/test/golden/` subfolders, each of which need a different command.
3. We force ourselves to flesh out our golden testing suite further.
4. We're encouraging ourselves to do more testing. (Maybe this is a stretch?)
5. It wouldn't take too long to do.

There are also some reasons to not, such as:

1. It's not strictly a necessary change right now.
2. The Makefile _works_ for the most part right now.
3. We only just recently implemented #4996, so we might see this as too early.
4. As #4996 stands, we would not generate `.log` files for our diffs, but just spit them out to console. This can be implemented in #4996, but I'm not sure if these `.log` files are valuable enough? For checking diffs, I prefer using `difftastic` and an interactive log viewer (that amounts to `DFT_SKIP_UNCHANGED=true DFT_COLOR=always difft "$1" "$2" | less -R`). I find this more helpful, but unless others use it as well (or want to do so), I'm probably wrong here.

Speculating a bit here, I believe that this would be a good thing to do. I think we should pursue it and break up `stable/`. I can see us moving more of our Makefile-orchestrated testing towards testing orchestrated by Drasil itself through `stack test` (e.g., #5028), including orchestrating compiling and running code and building documents in addition to the basic golden testing. This could be a viable route to beginning to deal with #3557 .

If we do (1), then (2) would also make sense to do, if only to keep golden artifacts closer to home.

JacquesCarette:
`Makefile` are an awkward technology. Maybe we could consider moving some of it to `shake`?

I'm all for upgrading our testing tech. The `.log` are because it's easy to do that. Having multiple places where tests go is not great. At least it can be explained: they test very different parts of Drasil that happen at different 'stages'. But that's probably a sufficient reason.

B-rando1:
Sorry, just getting back to this now 😅 

[...]

PR #5047, opened 2026-06-02
Simplify website generation

djarabek:
Removes the unneeded `HTML` sub directory and renames the output directory from `Website` to `website` for consistency.

Marked as draft due to some `Makefile` issues.


code/Makefile
```diff
@@ -153,8 +153,8 @@ DOCS_FOLDER = $(DOCS_FOLDER_NAME)/
 TRACEY_GRAPH_FOLDER_NAME = TraceyGraph
 TRACEY_GRAPH_FOLDER = $(TRACEY_GRAPH_FOLDER_NAME)/
 TRACEY_GRAPH_NAMES = allvsall allvsr avsa avsall refvsref
-WEBSITE_FOLDER_NAME = Website
-WEBSITE_FOLDER = $(WEBSITE_FOLDER_NAME)/HTML/
+WEBSITE_FOLDER_NAME = website
```
djarabek, line 151: Changing this results in a duplicate `website_stabilize` target, as the one generated by `filter` and the one declared explicitly now both have the same name (previously we had `Website_stabilize` and `website_stabilize` 🙃)

The explicit one is just a wrapper with some echos, but removing it then results in `make help` not listing it.
https://github.com/JacquesCarette/Drasil/blob/df0fda3a4ecffdeac82718e56b5c8a564add4730/code/Makefile#L280-L285

I'm not sure what the best approach here is. I don't see any reason we use the addsuffix/filter approach for the website at all (there's only one website), can we just get rid of it?

code/Makefile
```diff
@@ -394,10 +394,9 @@ $(GOOLTEST)$(STABILIZE_E_SUFFIX): $(GOOLTEST)$(GEN_E_SUFFIX)
 
 # The website stabilization. First builds the website, and then copies the
 # contents from the website folder into stable-website.
-# Target will have the form Website_stabilize.
+# Target will have the form website_stabilize.
 $(filter %$(STABILIZE_E_SUFFIX), $(STABILIZE_WEBSITE)): %$(STABILIZE_E_SUFFIX): %$(BC_E_SUFFIX) website
 	rm -rf "./$(STABLE_WEBSITE_FOLDER)"
-	@mkdir -p "$(STABLE_WEBSITE_FOLDER)"
```
djarabek, line 400: If the `stable-website` folder exists, the below copy will copy the `website` folder inside `stable-website`. If the `stable-website` folder does not exist, the below copy will create the `stable-website` folder and copy the contents of `website` into it. So we don't want to make the directory in advance.

code/Makefile
balacij: 1. Thank you for fixing the name. That was bothering me recently.
2. Can you explain the `make help` issue a bit more?
3. You can remove the `website_stabilize` creation. I think we should move towards replacing how stabilization is done regardless -- #5037, but that's still pending discussion.

djarabek: Part of `make help` is automatically generated https://github.com/JacquesCarette/Drasil/blob/df0fda3a4ecffdeac82718e56b5c8a564add4730/code/Makefile#L581-L592

If we remove the explicit `website_stabilize`, it disappears from the generated portion of `make help`.

djarabek: I pushed https://github.com/JacquesCarette/Drasil/pull/5047/changes/c940c7d7768d37449b5056c77993aa8b8f4b390c which removed the addsuffix/filter for the website, solving this problem.

code/Makefile
```diff
@@ -362,23 +361,11 @@ $(filter %$(TEST_E_SUFFIX), $(TEST_EXAMPLES)): %$(TEST_E_SUFFIX): $(CLEAN_GF_PRE
...
-
 # Individual example build cleans. Removes a specific example from the build folder.
 # Targets will have the form exampleName_build_clean.
 $(filter %$(BC_E_SUFFIX), $(BC_EXAMPLES)): %$(BC_E_SUFFIX):
 	rm -rf "./$(BUILD_FOLDER)$(EDIR)"
 
-# The website build cleans.
-# Target will have the form Website_build_clean
-$(filter %$(BC_E_SUFFIX), $(BC_WEBSITE)): %$(BC_E_SUFFIX):
-	rm -rf "./$(WEBSITE_FOLDER)"
```
balacij, line 380: Is this target now generated with the rest of the build clean targets? Or are we just deleting this in favour of moving `stable-website` to `drasil-website/test/golden`?

djarabek: Sorry, I meant to add a comment explaining this when I pushed. The `clean_website` target does the same thing, so I removed this in favour of that.

```
dj@framework:~/projects/Drasil/code$ make clean_website
rm -rf "./website/"
```

balacij: Ah, nice!

balacij (approved):
Thank you! This is a nice clean up.

JacquesCarette (approved):

Issue #4306, opened 2025-07-29
GamePhysics' Output Data Constraints table is empty


balacij:
https://jacquescarette.github.io/Drasil/examples/gamephysics/SRS/HTML/GamePhysics_SRS.html#Table:OutDataConstraints

<img width="1156" height="344" alt="Image" src="https://github.com/user-attachments/assets/db8555fd-bdc7-448a-93ed-4489df607570" />

This table should either not be there, or show the "reasonable values" as described in GamePhysics' Unitals.hs:

```haskell
outputConstraints :: [UncertQ]
outputConstraints = map (`uq` defaultUncrt) 
  [posOutCons, veloOutCons, orientOutCons, angVeloOutCons]

...

posOutCons        = constrained' QP.position           [] (exactDbl 0)
veloOutCons       = constrained' QP.velocity           [] (exactDbl 0)
orientOutCons     = constrained' QM.orientation        [] (exactDbl 0)
angVeloOutCons    = constrained' QP.angularVelocity    [] (exactDbl 0)
```

camishibata:
Do we want to display only the variables that have constraints, or, if there are any constraints, do we want to display the entire table, including the variables without them?

balacij:
I think that's best a question asked to @smiths but my thoughts are that if no variables have any constraints, there shouldn't be a table generated at all, and that if any variables have constraints, only the variables with constraints should be shown in the SRS data constraints tables. That being said, I see in the GamePhysics manually-created SRS that it contains 3 variables with "None" values for their constraints.

smiths:
@camishibata we should display variables that do not have constraints.  I think it is better to think of the table as a list of constraints that involve variables, rather than a list of variables with their constraints.  A constraint can involve more than one variable.  If there is no constraint, then there is nothing to list.  :smile:

PR #5050, opened 2026-06-02
Add value to GamePhysics' Output Data Constraints table

camishibata:
Contributes to #4306


code/stable/gamephysics/SRS/mdBook/src/SecCorSolProps.md
```diff
@@ -4,11 +4,11 @@ The [Output Data Constraints Table](./SecCorSolProps.md#Table:OutDataConstraints
...
-|\\(\boldsymbol{p}\text{(}t\text{)}\\)|
-|\\(\boldsymbol{v}\text{(}t\text{)}\\)|
-|\\(ϕ\\)                              |
-|\\(ω\\)                              |
+|Var                                  |Physical Constraints     |
+|:------------------------------------|:------------------------|
+|\\(\boldsymbol{p}\text{(}t\text{)}\\)|--                       |
+|\\(\boldsymbol{v}\text{(}t\text{)}\\)|--                       |
+|\\(ϕ\\)                              |\\(0\leq{}ϕ\leq{}2\\,π\\)|
+|\\(ω\\)                              |--                       |
```
balacij, line 12: I know that you mentioned you found this constraint from somewhere. Can you please link it here? Do the other variables have constraints?

Also, I realize now that we need to split the original issue in two: one is to add the correct constraints (as you do here, and is the original goal of the issue), but a separate issue is that the 1 column table should not have been generated to begin with. We should look into `drasil-docLang` for the code that generates this table and figure out how to stop it from being generated if there are no constraints.

JacquesCarette (approved):
So should I merge as is or wait?

balacij:
Merging as-is is fine with me. The code looks good to me. Let's just not close the original issue.

Just switched OP from Closes to Contributes.

code/stable/gamephysics/SRS/mdBook/src/SecCorSolProps.md
camishibata: I consulted section 5.2.6 of this document: [gamephys](https://github.com/smiths/caseStudies/blob/master/CaseStudies/gamephys/docs/SRS/GamePhysicsSRS.pdf). The other variables have no constraints in it.

balacij: Thank you!

balacij (approved):

PR #5051, opened 2026-06-02
Add `HasRationale` for optional data constraint rationale

Xinlu-Y:
Adds an optional `rationale` field to `ConstrConcept` so a constrained quantity can carry a short sentence explaining its typical value. If any quantity has one, the input Data Constraints table gets an extra "Rationale" column.

Also adds two constructors: `constrainedWithRationale` and `uqDirect`.

It's backward compatible — the new field defaults to `Nothing`, so existing examples generate the same output. A follow-up PR adds a Binary Star System example that uses this.


code/drasil-lang/lib/Language/Drasil/Chunk/Constrained.hs
```diff
@@ -32,9 +32,10 @@ import Language.Drasil.Stages (Stage)
 --
 -- Ex. Measuring the length of a pendulum arm could be a concept that has some reasonable value
 -- (between 1 cm and 2 m) and the constraint that the length cannot be a negative value.
-data ConstrConcept = ConstrConcept { _defq    :: DefinedQuantityDict
-                                   , _constr' :: [ConstraintE]
-                                   , _reasV'  :: Maybe Expr
+data ConstrConcept = ConstrConcept { _defq       :: DefinedQuantityDict
+                                   , _constr'    :: [ConstraintE]
+                                   , _reasV'     :: Maybe Expr
+                                   , _rationale' :: Maybe Sentence
```
balacij, line 38: Should rationale for the constraints be associated per-constraint or per-variable?

Aside: I wonder how we deal with variables that are constrained by each other?

JacquesCarette: We deal with those kinds of constraints via 'relations'. `ConstrConcept` is specifically for single variables. So the rationale here should be likewise.

code/drasil-lang/lib/Language/Drasil/Classes.hs
```diff
@@ -86,6 +87,11 @@ class HasReasVal c where
   -- | Provides a 'Lens' to the possible reasonable value.
   reasVal     :: Lens' c (Maybe Expr)
 
+-- | A chunk that may have a rationale explaining why a value or constraint was chosen.
```
JacquesCarette, line 90: Because this is 'may have' rather than has, the class should be `MayHaveRationale`. The lens name should not change.

JacquesCarette (changes requested):
Good addition, just a small naming issue to change.

balacij:
It looks like this PR was partially merged into `main` as a result of #5052 but it still needs to be fixed up (the conflicts, that is) and merged as well to address @JacquesCarette's comment.

Xinlu-Y:
> It looks like this PR was partially merged into `main` as a result of #5052 but it still needs to be fixed up (the conflicts, that is) and merged as well to address @JacquesCarette's comment.

fixed that

JacquesCarette (approved):

PR #5052, opened 2026-06-02
Add Binary Star System (BSS) example

Xinlu-Y:
Adds a new Binary Star System case study. It uses the optional rationale feature from #5051, so it is stacked on that PR.


code/drasil-printers/lib/Language/Drasil/Plain/Print.hs
```diff
@@ -54,7 +54,7 @@ pExprDoc _ (Ident s) = text $ toPlainName s
 pExprDoc _ (Label s) = text $ toPlainName s
 pExprDoc _ (Spec s) = specialDoc s
 pExprDoc f (Sub e) = text "_" <> pExprDoc f e
-pExprDoc f (Sup e) = text "^" <> pExprDoc f e
+pExprDoc f (Sup e) = text "_" <> pExprDoc f e
```
balacij, line 57: I believe this is a regression?

balacij (commented):
This project looks really neat! I will need to sit down and read it thoroughly at some point. I just quickly noticed the one issue.

JacquesCarette:
As this is quite a large PR, it will also take me some time to review it. Hopefully I can get to it tomorrow, Friday at the latest.

code/drasil-printers/lib/Language/Drasil/Plain/Print.hs
Xinlu-Y: I thought it was a regression lol but after thinking twice, I intended to do that. This printer only feeds code generation, not the SRS docs. Superscripts as ^ make invalid identifiers (v_x1^0 isn't valid Python), so they have to be _. That's what the ODE codegen fix was for.

code/drasil-example/bss/lib/Drasil/BinaryStar/Assumptions.hs
```diff
@@ -0,0 +1,78 @@
...
+twoBody = cic "twoBody"
+  (S "The system consists of exactly two stars with masses" +:+
+   ch mass_1 `S.and_` ch mass_2 +:+. S "" +:+
+   S "Third-body gravitational perturbations are ignored" !.)
+  "twoBody" assumpDom
+
+isolated = cic "isolated"
+  (S "Non-gravitational forces (e.g., drag, thrust, radiation pressure)" +:+
+   S "are neglected; the only interaction modeled" `S.is`
+   S "mutual Newtonian gravitation" +:+
```
JacquesCarette: There are a lot of words/phrases in these sentences (like 'gravitation', 'Newtonian gravitation', 'Newton's law of universal gravitation', etc) which should be chunks.

code/drasil-example/bss/lib/Drasil/BinaryStar/Body.hs
```diff
@@ -0,0 +1,258 @@
...
+import qualified Language.Drasil.Sentence.Combinators as S
+import Data.Drasil.Concepts.Theory (inModel)
+import Data.Drasil.Concepts.Math (ode)
+import Data.Drasil.Concepts.Documentation (assumption, endUser, input_,
+  interface, output_, physical, software, sysCont, softwareConstraint,
+  softwareSys, system, user, environment, product_, datum)
+import Data.Drasil.Concepts.Education (undergraduate, calculus)
+import Data.Drasil.Concepts.Physics (gravity, twoD)
+import Data.Drasil.Quantities.Physics (energy, gravitationalConst)
+import Data.Drasil.Quantities.PhysicalProperties ()
```
JacquesCarette: If you need nothing from these, don't import them!

code/drasil-example/bss/lib/Drasil/BinaryStar/Body.hs
```diff
@@ -0,0 +1,258 @@
...
+--------------------------------------------
+
+probDescIntro :: Sentence
+probDescIntro = foldlSent_
+  [S "simulate the motion of a binary",
+   S "star system under mutual gravitational interaction"]
+
+physSystParts :: [Sentence]
+physSystParts = map (!.)
+  [S "The first star with mass" +:+ ch mass_1,
```
JacquesCarette: I see in a later file that "first star" and "second star" are known phrases, so should be used here. Same for 'mass'.

code/drasil-example/bss/lib/Drasil/BinaryStar/DataDefs.hs
```diff
@@ -0,0 +1,10 @@
+module Drasil.BinaryStar.DataDefs (dataDefs) where
+
+import Theory.Drasil (DataDefinition)
+
+-- | The hand-written SRS defines DD1 as the gravitational constant G,
+-- which in Drasil is handled as a constant (gravitationalConstValue)
+-- in Unitals.hs rather than a DataDefinition.
+-- The separation distance formula is in TM6 (relPosTM).
+dataDefs :: [DataDefinition]
```
JacquesCarette: since you have no data definitions, you should just delete this file

code/drasil-example/bss/lib/Drasil/BinaryStar/Body.hs
```diff
@@ -0,0 +1,258 @@
...
+  [S "The total mechanical energy of the system must remain",
+   S "constant over time (up to numerical tolerance),",
+   S "since the system is isolated", sParen (refS isolated),
+   S "and masses are constant", sParen (refS constantMass)]
+
+corrSolnEqn :: Contents
+corrSolnEqn = unlbldExpr energyExpr
+
+-- E = const
+energyExpr :: ModelExpr
```
JacquesCarette: Why isn't this in `Drasil.BinaryStar.Expressions` ?

code/drasil-example/bss/lib/Drasil/BinaryStar/Requirements.hs
```diff
@@ -0,0 +1,90 @@
...
+---------------------------------------------------------
+
+nonFuncReqs :: [ConceptInstance]
+nonFuncReqs = [correct, portable, maintainable, usable]
+
+correct :: ConceptInstance
+correct = mkCorrectNFR "correct" "Correctness"
+
+portable :: ConceptInstance
+portable = mkPortableNFR "portable" ["Linux", "Mac OSX", "Windows"] "Portability"
```
JacquesCarette, line 76: I'm surprised those 3 OS names have to be hardcoded here?

code/drasil-example/bss/lib/Drasil/BinaryStar/Requirements.hs
```diff
@@ -0,0 +1,90 @@
...
+
+maintainable :: ConceptInstance
+maintainable = cic "maintainable"
```
JacquesCarette: This feels like a generic statement that should be abstracted out of here, so it can be used for many examples.

JacquesCarette (changes requested):
This is really good, and very close to great. I imagine a few rounds of reviews will be needed to get it there.

code/drasil-example/bss/lib/Drasil/BinaryStar/Requirements.hs
Xinlu-Y: This is the current pattern across all examples. They all hardcode the same OS list. Should I open an issue to track abstracting this into drasil-srs?

JacquesCarette: Yes please - and then I'll not worry about this part.

JacquesCarette (approved):
While there is still a bunch of things I would like to see done (mostly abstracting out more terms from the hard-coded `String`s), this is solid enough to be added as a new example now.

PR #5053, opened 2026-06-03
Added `arrayClone`

B-rando1:
Contributes to #4929.


balacij (approved):

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -284,6 +284,10 @@ class (IndexTranslator r) => Array r where
   -- TODO [Brandon Bosman, 05/19/2026]: Change return type to SValue
   -- | Given array `a` and index `i`, creates `a[i]`
   arrayElem :: SValue r -> SVariable r -> SVariable r
+  -- TODO [Brandon Bosman, 05/21/2026]: Consider switching this to a polymorphic `clone`,
+  -- more like how `print` currently works
+  -- | Given a source array, create a (shallow) clone of it
+  arrayClone :: SValue r -> SValue r
```
JacquesCarette: should be called 'arrayCopy'. Clone is too specific (vocabulary-wise). We should name things closer to what they mean rather than their implementation details.

JacquesCarette (changes requested):

JacquesCarette:
And now it needs a merge...

B-rando1:
Merged now.

JacquesCarette (approved):

PR #5055, opened 2026-06-03
Switched `instanceVarAccess` to take `SValue` instead of `SVariable`

B-rando1:
This was annoying me for a while, and would've forced me to make `arrayLength` also take a variable for the array, because some languages encode `length` as an instance variable.

It'll also allow us to chain method calls together, e.g.

```python
[1..10].sorted().filter(isEven).map(\x -> x^2)
```
Obviously that code isn't quite valid, but the idea of it works. Each intermediate method returns a new array, which lets you chain method calls together. Letting GOOL take a value instead of a variable for the instance will allow us to do the same.


JacquesCarette (approved):
Nice!

PR #5062, opened 2026-06-04
Add skeleton MatlabRenderer modules (cloned from Julia)

Xinlu-Y:
This lays the groundwork for a matlab backend in GProc. 
Nothing generates real matlab code yet. It's just the skeleton modules cloned from the Julia renderer so I have a place to start building.


code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -0,0 +1,48 @@
...
+  optimizeDox = error doxError
+
+  makefile _ it cms = G.makefile Nothing (G.noRunIfLib it mlRunnable)
+                            (G.docIfEnabled cms (DocConfig [] []))
+
+  unReprDoc = unMLP
+
+-- | Default runnable information for MATLAB files
+mlRunnable :: Maybe Runnable
+mlRunnable = interpMM "octave"
```
B-rando1, line 45: From a quick search, it looks like Octave is an open-source alternative to MATLAB with mostly compatible syntax. @smiths have you heard of Octave, and will it work for our purposes?

B-rando1 (commented):
Barring my question about Octave, this looks good. There are pros and cons to starting from a copy of the Julia renderer vs starting from scratch. If you start from a copy of the Julia renderer, you'll need to be careful to go over every function implementation and check that it makes sense for MATLAB. I know I made the mistake a few times with Julia of following another renderer too closely and either generating non-idiomatic Julia or taking a roundabout route to generate idiomatic Julia. Just something to keep in mind; you shouldn't blindly follow another renderer's implementation without knowing why it does something a certain way and whether it makes sense to do the same for MATLAB.

code/drasil-code/lib/Language/Drasil/Code/Imperative/GOOL/LanguageRenderer/MatlabRenderer.hs
smiths: Yes, Octave is an open source alternative to Matlab.  Scilab is another open source alternative.  When I last tried these options, I found I liked the graphics on Octave more, but Scilab seemed to have more users.  The basics of Octave and Scilab were identical (maybe almost identical?) to Matlab.  The differences are in the extra packages that Matlab provides.  We won't be using those extras, so the code we generate should almost be able to work in all three options.  I think this is a case of low-hanging fruit.  We might as well support scilab, octave and Matlab.  The advantage of this is that not everyone has access to Matlab.  We have a site license at McMaster, but not all users of Drasil will be at Mac.  :smile:

I did a bit of a google search.  Scilab is a little more different from Matlab than I thought.  Octave is designed to a clone of Matlab, but Scilab is more independent.  From a google search, some differences fro Scilab are:

- System Constants: Scilab requires a percent sign prefix for built-in constants. Where MATLAB/Octave use pi, 
- Scilab expects %pi.Imaginary Numbers: MATLAB/Octave accept i or j seamlessly. 
- Scilab requires %i.String Enclosures: MATLAB historically uses single quotes ('string') for standard character arrays. 
- Scilab strongly prefers double quotes ("string")."

We should definitely support Octave.  We could wait on Scilab and add it in the future as a little project for someone.

JacquesCarette:
Should this be marked as `Draft` as it is not ready for merging?

JacquesCarette:
I agree with @B-rando1 . It is probably a better idea to remove all implementations (replaced them all with 'undefined') and then think them through one by one.

Xinlu-Y:
My plan is that merge this skeleton first, then add the real implementation in small follow-ups, one piece at a time (program/file/module, functions, variables, statements, …). 
Does that work for you? @JacquesCarette

JacquesCarette (approved):
Let's try it the piecemeal way and see!

JacquesCarette:
Oh no, merging this broke the build! There was an interaction with #5115 from @B-rando1 that was not otherwise caught. Please fix asap @Xinlu-Y .

Xinlu-Y:
Thanks! @balacij

PR #5066, opened 2026-06-04
refactor(printers-mdBook): Use drasil-data-formats CSV

djarabek:
Contributes to #4989

### DIFF
--- a/code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
+++ b/code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
@@ -1,16 +1,21 @@
-{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE OverloadedStrings, QuasiQuotes #-}
 
 -- | Defines main Markdown printer functions.
 module Language.Drasil.Markdown.Print (genMDBook, pSpec) where
 
 import Prelude hiding (print, (<>))
 import qualified Prelude as P ((<>))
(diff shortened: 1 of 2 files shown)


code/drasil-printers/lib/Language/Drasil/Markdown/Print.hs
```diff
@@ -71,17 +76,21 @@ mkTitle rm t = text "\"" <> pSpec rm t <> text "\""
 
 -- | Prints the .csv file mapping the original filepaths of assets to the
 -- location mdBook uses.
-makeRequirements :: Project -> Doc
-makeRequirements p = makeCSV $ ["Original", "Copy"] : assetMat p
+makeRequirements :: Project -> PNew.Doc ann
+makeRequirements p =
+  let
+    mCSV = mkCSV (Just 2) (Just ["Original", "Copy"]) (assetMat p)
+    csv = fromRight (error "expected csv") mCSV
```
JacquesCarette: This code loses the carefully crafted error that `mkCSV` would have put in `Left`. You should use it! And using it will mean using `either` instead of `fromRight`, where the latter is almost always a "code smell".

JacquesCarette (changes requested):

balacij (approved):

djarabek:
@JacquesCarette Ready for rereview

JacquesCarette (approved):

Issue #5042, opened 2026-06-01
Switch GOOL from file based output `FileLayout`-like tree based output


djarabek:
In https://github.com/JacquesCarette/Drasil/pull/4980, we migrated the code generation to `FileLayout` (allowing for `drasil-code` to be `IO`-free). For the most part, this wasn't difficult and greatly improved the code, however there was one sticking point: the `ProgData` returned by GOOL (which contains the generated source code), gives us a `[FileData]`, which is a flat list of file paths and `Doc`s, not a hierachial representation compatible with `FileLayout`. In #4980, we used the following workaround to convert the flat list into a `FileLayout`:

https://github.com/JacquesCarette/Drasil/blob/dbfdf3aeba32d774e4d0f197e3288f27cc5fd94a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs#L140-L141

GOOL should return the generated source code as a tree that can be directly converted to `FileLayout`.

It seems like GOOL currently only has limited support for generating modules/packages, only in Java does it generate source code with `package` declarations (since this is required to be able to import across generated files). The code that enables this is specific to Java:
https://github.com/JacquesCarette/Drasil/blob/dbfdf3aeba32d774e4d0f197e3288f27cc5fd94a/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JavaRenderer.hs#L137-L141

Resolving this issue may make sense as part of a larger refactor that enables more support for packages/modules in GOOL.

balacij:
@B-rando1 Do you have any thoughts on how this and how it can be done? Is `FileLayout` sensible for GOOL to generate the way we want to move it towards?

B-rando1:
I think there might be hope with moving `FileLayout` into GOOL. It's hard to see where the sticking points will be without trying it, though.

[...]

PR #5071, opened 2026-06-04
 Migrate codegenTest to use golden testing framework

### LINKED ISSUE #5013: Switch gool/code codegen testing suite to #4996-based one
#4996 introduced a miniature golden testing library that we should make use of to orchestrate the golden test suite for `drasil-build-artifacts`.

We should make use of this library for the golden testing suite for `drasil-code` and `drasil-gool`. In fact, it looks like some of the existing components of the testing suite in `drasil-code` really belong to `drasil-gool`?

If #4980 isn't merged before we start working on this issue, we should base this work off of #4980.

djarabek:
Closes #5013


code/drasil-code/test/Main.hs
```diff
@@ -31,49 +26,114 @@ import GOOL.PatternTest (patternTest)
...
+    "gooltest"
+    [ goldenTest "java" $ directory [ps|java|] $ genCode (classes unJC unJP),
+      goldenTest "python" $ directory [ps|python|] $ genCode (classes unPC unPP),
+      goldenTest "csharp" $ directory [ps|csharp|] $ genCode (classes unCSC unCSP),
+      goldenTest "cpp" $ directory [ps|cpp|] $ genCode (classes unCPPC unCPPP),
+      goldenTest "swift" $ directory [ps|swift|] $ genCode (classes unSC unSP),
+      goldenTest "julia" $ directory [ps|julia|] $ genCode (jlClasses unJLC unJLP)
+    ]
+    
+-- codeGenTestGroup :: TestTree
```
djarabek: Me and Brandon talked and wanted a more structured approach to running the tests, however I ran into type errors getting it to work.

code/drasil-code/test/Main.hs
```diff
@@ -100,12 +170,29 @@ jlClasses unRepr unRepr' = zipWith
   (map (Proc.unCI . (`evalState` initialState)) [helloWorldProc,
     fileTestsProc, vectorTestProc, nameGenTestProc])
 
-------------------
--- IO Functions --
-------------------
+-- | Internal: Converts a list of `FileData` to a `FileLayout`.
+toFileLayout :: [FileData] -> [FileLayout]
```
djarabek: This is a temporary measure, #5042 may result in this being moved/removed

code/drasil-code/test/Main.hs
balacij: We should avoid trying to run the generated code _right now_. That should be done in a separate PR (and there's a real design discussion to be had about how this should be done because it extends to our other generated software artifacts, too). The work of switching to the new golden testing code is complex enough.

djarabek: Sorry, I meant running the tests as in doing the golden testing.

balacij: Golden! My bad :)

code/drasil-code/test/Main.hs
```diff
@@ -31,49 +26,114 @@ import GOOL.PatternTest (patternTest)
...
+--           goldenTest "NameGenTest" $ directory [ps|NameGenTest|] $ genCode [classes'' lang nameGenTestOO]
+--         ]
+--     ]
+--   where
+--     classes'' :: (OOProg r) => String -> OO.GSProgram r -> PackageData
+--     classes'' "java" = classes' unJC unJP
+--     classes'' "python" = classes' unPC unPP
+--     classes'' "csharp" = classes' unCSC unCSP
+--     classes'' "cpp" = classes' unCPPC unCPPP
+--     classes'' "swift" = classes' unSC unSP
```
balacij: This looks like a better organization of the testing suites (i.e., `testName/lang/..`). Was there anything wrong with this approach?

djarabek: Yes - me and Brandon are trying to fix the type errors right now. This specific code should have been deleted before I pushed, we decided on the reverse structure `{test_case}/{language}`.

balacij: What's the rationale for the change? This way makes more sense to me because we can group tests by related typeclasses (1 group = 1 `test_case`) and expected inputs/outputs/interactions. This way we can quickly check that the typeclasses are used to generate the "same" code across each language.

That being said:

1. @B-rando1 works on GOOL, not me. If he prefers the opposite (e.g., workflow reasons), I'm not going to argue haha
2. (I just realized this.) Not making this change in _this PR_ is completely fine with me because the current `stable/gooltest/` generates `lang/testName`. That is, we shouldn't be making two changes at once.
3. It might be most appropriate to have both suggestions!
4. I'm also not entirely sure if the existing tests are meant to test grouped pools of typeclasses. I could be entirely wrong here!

B-rando1: We were able to get the type gymnastics to work; @djarabek is re-packaging them now. I was thinking grouping `test/lang` (as opposed to the current `lang/test`) might be better, because the majority of GOOL development is on implementing/modifying a single feature for all languages, rather than adding a new language and testing all features. So either in this PR or a future one I think it would be an improvement to switch it to `/test/lang`.

balacij: It looks like we are talking past each other but agreeing? (I hope! 😄)

* I originally read the code as though it were generating `testName/lang` (hence my first comment). I was wrong about this.
* Reading "yes" in response to my question further affirmed my incorrect reading of what the code did.

As I mentioned in both comments, I too think `testName/lang` is much more reasonable!

JacquesCarette: Separately: we should not dispatch on a `String`. Why aren't we using a proper type here?

code/drasil-code/test/Main.hs
balacij: Is there any way we can at least de-duplicate this code from the code in:

https://github.com/JacquesCarette/Drasil/blob/7686ce4155dd2e74d07c1bdca6a8021aed9fd7da/code/drasil-code/lib/Language/Drasil/Code/Imperative/Generator.hs#L141-L167

?

I'm fine with exporting the code from `Language.Drasil.Code.Imperative.Generator` for the purposes of the testing suite. It should just be moved to a separate module.

code/drasil-code/test/Main.hs
djarabek: This was just some code I was playing around with, it wasn't meant to get pushed and isn't in the final version.

djarabek (commented):
~~Just realized CI will also need to be updated to detect changed files in the new location.~~

Edit: `make codegenTest` is run unconditionally, so no need to make changes.

code/drasil-code/test/Main.hs
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5071/changes/03d6d9b899e195f3f7e3da800f852891a3b79a1d. I didn't move it, if you have an idea for a better location I can move it.

balacij (approved):
Great work! This simplifies things nicely.

code/drasil-code/test/Main.hs
balacij: Eh, it's fine where it is.

JacquesCarette (approved):

PR #5080, opened 2026-06-05
refactor(lesson-plan): move all code to new package; `drasil-lesson-plan`

balacij:
Another side-step.

`drasil-docLang` is too large, carrying too many unrelated things. In this PR, we reduce the package size by moving all code related to the lesson plan to a new package. I do not attempt to clean up the code at all. That should be done in a subsequent PR.


code/drasil-docLang/package.yaml
```diff
@@ -40,8 +40,10 @@ library:
   - Drasil.DocLang
   - Drasil.DocLang.SRS
   - Drasil.Document.Contents
-  - Drasil.DocumentLanguage.Notebook
   - Drasil.DocumentLanguage.Units
   - Drasil.DocumentLanguage.TraceabilityGraph
   - Drasil.Sentence.Combinators
   - Drasil.SRSDocument
+  - Drasil.GetChunks
```
JacquesCarette, line 47: These 3 exports are very suspicious. Please open an issue to figure out what to do with them.

JacquesCarette (approved):
(Still approving, even though there's an outstanding task to be done.)

code/drasil-docLang/package.yaml
balacij: These will be removed with #5085 :)

PR #5083, opened 2026-06-05
 init(drasil-testing-kit): a package for common testing code

### LINKED ISSUE #5034: Introduce testing library package
Me and Jason discussed and decided it would be a good idea to introduce a package for our testing-only code (currently just [GoldenTesting](https://github.com/JacquesCarette/Drasil/blob/cbcf885d07f6dd787eed724b6515e29409375188/code/drasil-build-artifacts/lib/Drasil/Build/Artifacts/GoldenTesting.hs)).

We also need to add a test main function that wraps Tasty's `defaultMain` with a call to `setLocaleEncoding utf8`, to ensure tests will work correctly on Windows.

I'm not sure what to call this package, maybe just `drasil-testing` (or perhaps `drasil-test-lib`)?

djarabek:
Closes #5034


code/drasil-testing-kit/lib/Drasil/TestingKit/FileHandling.hs
```diff
@@ -0,0 +1,8 @@
+-- Re-exports drasil-file-handling to avoid dependency issues
+-- in the drasil-file-handling tests.
+module Drasil.TestingKit.FileHandling
```
djarabek: Based on [Jason's suggestion in #5034](https://github.com/JacquesCarette/Drasil/issues/5034#issuecomment-4614046572). I would be open to other solutions though.

code/drasil-testing-kit/lib/Drasil/TestingKit/TestMain.hs
```diff
@@ -0,0 +1,9 @@
+module Drasil.TestingKit.TestMain (testMain) where
+
+import GHC.IO.Encoding (setLocaleEncoding, utf8)
+import Test.Tasty (defaultMain, TestTree)
+
+testMain :: TestTree -> IO ()
+testMain testTree = do
+  setLocaleEncoding utf8
+  defaultMain testTree
```
balacij, line 10: Just needs some comments :)

code/drasil-testing-kit/package.yaml
```diff
@@ -0,0 +1,36 @@
...
+- -Wall
+- -Wredundant-constraints
+- -Wmissing-export-lists
+
+library:
+  source-dirs: lib
+  exposed-modules:
+  - Drasil.TestingKit.FileHandling
+  - Drasil.TestingKit.GoldenTesting
+  - Drasil.TestingKit.TestMain
```
balacij: I think it's fine to have `GoldenTesting` as a separately exported module. @JacquesCarette might have a different opinion (?). However, in general, we prefer single-module exports. So, in this case, we would create `Drasil.TestingKit` that re-exports everything from (at least) `FileHandling` and `TestMain`.

balacij (commented):
Only minor comments, looks good!

code/drasil-testing-kit/package.yaml
djarabek: My logic on `FileHandling` is it's only meant to be used in the `drasil-file-handling` tests, so it should be exposed separately. Unless we want to make this the main way to access `drasil-file-handling` in a test?

I agree that `TestMain` could be exported from `Drasil.TestingKit`.

balacij: > Unless we want to make this the main way to access drasil-file-handling in a test?

It's not so much that we want to be the main or only way to access `drasil-file-handling` in a test, it's only that when importing a "testing kit," it doesn't hurt if it also exports some file-related features. Now that I write this I refine my thought a bit more (it should be coupled with the _golden testing_ code, not in the general testing kit), I see two options at least:

* `Drasil.TestingKit`: Re-export `TestMain`.
* `Drasil.TestingKit.Golden`: Re-export `.GoldenTesting` and `Drasil.FileHandling`. Though, you can also rename `GoldenTesting` to just `Golden` and have it re-export some stuff from `Drasil.FileHandling`.

OR

* `Drasil.TestingKit`: Re-export everything.

I think I prefer the former at face-value.

However, seeing that `TestMain` only contains one thing: 

```haskell
testMain :: TestTree -> IO ()
testMain testTree = do
  setLocaleEncoding utf8
  defaultMain testTree
```

I wonder if it would be appropriate for us to have a parameter in `writeFiles'` of `drasil-file-handling` that _always_ sets the locale encoding, defaulting `writeFiles` to UTF8. In which case, we would not need `TestMain` at all. And as a result, we would only need to export one module from `drasil-testing-kit`:

* `Drasil.TestingKit.Golden`.

djarabek: > * `Drasil.TestingKit`: Re-export `TestMain`.
> 
> * `Drasil.TestingKit.Golden`: Re-export `.GoldenTesting` and `Drasil.FileHandling`. Though, you can also rename `GoldenTesting` to just `Golden` and have it re-export some stuff from `Drasil.FileHandling`.


This makes sense to me.

In case we ever read files during a test, I think it would be prudent to set the encoding at the start of the test. Setting the encoding when we write files could result in different behavior depending on the order tests run.

JacquesCarette:
Can you please fix the conflict in the Makefile?

code/drasil-testing-kit/package.yaml
balacij: Spoke over Discord. Going with the former option here makes sense because we also _read_ from files!

code/drasil-testing-kit/lib/Drasil/TestingKit/TestMain.hs
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5083/changes/d4d90f9f31178f95f8c4558c72161bc4fb976f16

code/drasil-testing-kit/package.yaml
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5083/changes/d4d90f9f31178f95f8c4558c72161bc4fb976f16

code/drasil-testing-kit/lib/Drasil/TestingKit/FileHandling.hs
balacij: I think this turned out pretty well!

balacij (approved):
Looks like we forgot to merge this one?

I just merged in `main` to make sure the CI wont break with any of the latest changes to our packages

JacquesCarette (approved):
Had not forgotten - I had never reviewed it, because it had conflicts, and I was waiting for those to be resolved.

PR #5089, opened 2026-06-06
refactor: move `drasil-docLang/../Document/Contents` and `../Sentence/Combinators` to `drasil-docLang`s `Language.Drasil.Document` namespace

balacij:
`drasil-docLang` primarily contains things related to SRS production. This PR moves more generic 'scientific document'-related things back into `drasil-lang` as a step towards:

1. Renaming `drasil-docLang` to `drasil-srs`.
2. Moving `drasil-docLang`-related things from `drasil-lang` into a new `drasil-docLang` package.

Disclaimer: I did not prioritize making each import explicit. These will change again very soon.


JacquesCarette (approved):
The funny thing, I'm pretty sure those all used to be in `drasil-lang` and I'm the one who moved them, because they were only used in docLang!

But I completely understand why this is a good (temporary) move.

PR #5098, opened 2026-06-08
refactor: avoid `filterEmpty` by using `Maybe`s

balacij:
This is not necessarily the cleanest code, but that's not the goal here. That can be done in a subsequent PR when we re-write this "Markdown renderer."


JacquesCarette (approved):
The 'right' fix would likely involve `catMaybes`.

balacij:
Yeah, I'm not quite a fan of this file entirely. `filterEmpty` was just very frustrating to see.

I did use `catMaybes` in `makeMd`, but the rest of the code needs a fair amount of clean up work.

It would be a lot nicer if this code weren't responsible for dealing with the Markdown generation as well! In fact, it would have been nice if this code were only generating `LayoutObj`s and we could then _choose_ what README output type we wanted: `.rst`, `.txt` (or no extension, assumed `.txt`), `.md`.

PR #5099, opened 2026-06-08
refactor: move `CreateMd` to `drasil-code`

### LINKED ISSUE #5069: Move `CreateMd.hs` from `drasil-printers` to `drasil-code`
https://github.com/JacquesCarette/Drasil/blob/cfe75edc5a348cd202ab90eefb2785b734c85556/code/drasil-printers/lib/Language/Drasil/Markdown/CreateMd.hs#L1-L155

The top comment is a great indicator of why it should be moved! This code knows far too much about the code generator, when it ought to know nothing of the sorts of any usage of `drasil-printers`.

An "aside" issue with the code above is that it doesn't really use a Markdown code AST. Rebuilding this code should be a part of #4989.

balacij:
Builds on #5098 
Closes #5069 

`upcase` is a questionable tool. We should look into how its used and why it's important.


code/drasil-code/lib/Language/Drasil/Code/Imperative/README/Render.hs
```diff
@@ -40,6 +39,11 @@ maybeSub :: String -> Maybe String -> Doc
 maybeSub role = maybe empty (\content-> doubleSep <> text ("> " ++ role ++ ":")
   <+> upcase content)
 
+-- | Capitalize first letter of string.
```
JacquesCarette, line 42: I'm 99% sure that this exists somewhere already. Plus it should be factored as something that upper-cases a `String` and then applies `text`.

code/drasil-printers/lib/Language/Drasil/Printing/Helpers.hs
```diff
@@ -102,15 +99,6 @@ sqbrac  x = "[" ++ x ++ "]"
 -- | Wraps in angular brackets ("<>").
 angbrac x = "<" ++ x ++ ">"
 
--- | Format strings and convert to Doc.
```
JacquesCarette, line 105: Maybe the `String -> String` version of these should be in `Drasil.Utils` (for now), and then the `text` can be done at call site?

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/README/Render.hs
balacij: Operational design aside, I just don't see why _this_ renderer should be doing any string manipulation!

(Excluding dealing with some quirks of the output language, of course.)

balacij: `Utils.Drasil.English` has `capitalize`:

https://github.com/JacquesCarette/Drasil/blob/98d5024c00e7bb3af84deca055f3c2ce0cc9933c/code/drasil-utils/lib/Utils/Drasil/English.hs#L6-L9

but using it results in the following (bad) diff:

```console
diff --strip-trailing-cr -r -X ../.gitignore '--exclude=*.prof' '--exclude=.drasil' stable/dblpend/src/cpp/README.md build/dblpend/src/cpp/README.md
14c14
< > Scope: The analysis of a two-dimensional (2D) pendulum motion problem with various initial conditions.
---
> > Scope: The analysis of a two-dimensional (2d) pendulum motion problem with various initial conditions.
```

The whole string is given to `upcase` (!), and so `capitalize` removes all capitalization in the rest of the sentence. So, it's not usable for our purposes.

balacij: In the name of avoiding scope creep and merging what should be a fairly simple PR in, I think it would be best for us to move this to an issue with some greater discussion about `Sentence`. Specifically discussing how they should be formed and rendered.

JacquesCarette (approved):

code/drasil-code/lib/Language/Drasil/Code/Imperative/README/Render.hs
balacij: Moved to https://github.com/JacquesCarette/Drasil/issues/5253

PR #5101, opened 2026-06-09
refactor(data-formats-CSV): flip parameters to `renderCSV`

djarabek:
Depends on #5066

This order makes more sense for the soon to be added `renderJSON` (as it is recursive), and the APIs should be consistent.


balacij (dismissed):
Good idea!

JacquesCarette (approved):

PR #5112, opened 2026-06-10
chore(`drasil-lesson-plan`): remove dead code and unnecessary dependencies

balacij:
The `multiplate` stuff is nice, but unused. I'm also fairly certain it's overkill for the lesson plan. For the SRS, it does seem to work out nicely, however.


JacquesCarette (approved):
We need to really understand what problem 'multiplate' solves, document that, and remember it (and probably use it more!). I agree that it is nice and can be overkill. So fine to delete it from here.

Issue #5111, opened 2026-06-10
GOOL: Handling Pointers in C++


B-rando1:
As we found in #4909, the C++ renderer has a very weak notion of pointers that is causing problems for us. This shows up in two ways:
- When passing `this` to a function expecting a `Vector`, we need to dereference it (pass `*this` instead of `this`) so that it's a `Vector` and not a pointer to a `Vector`. GOOL doesn't know how to do this, currently.
- For some method calls, the library calls `objMethodCall` on `self`, which produces code like `this.method()`. The current "intended solution" in GOOL is to call `selfFuncApp` instead, which produces the correct code `this->method()`.

The current solution to the second point feels like a hack. Why should `selfFuncApp` have a fundamentally different implementation than `objMethodCall`? It would make more sense to me if `selfFuncApp` was a helper function that called `objMethodCall` as part of its implementation.

To make this work, we could add `Pointer CodeType` as a new option for `CodeType`, and make `self` have type `Pointer (Object ...)` in C++. Then `objMethodCall` would need to check whether `obj` has type `Object ...` or `Pointer (Object ...)` and use `.` or `->` accordingly.

Handling the first point will require more work. It would be really nice if we could do something similar: have `valueOf` check the type of its variable and add a `*` if it's a pointer and leave it out otherwise; but that would conflict with the updated `objMethodCall`, which would want the pointer itself. I think instead we'd have to add an explicit `dereference` operator to GOOL's external interface. It's not optimal, but I can't think of a reliable way to automatically add `*` as needed. The only other option I can think of is, if we kept track of functions' input types, we could have call sites compare the type of their arguments to the expected type of their parameters, and implicitly add dereferences to coerce the arguments to the correct type. But that sounds a little far-fetched.

The above is what I think I want to do, because it seems fairly elegant and I don't think it'll result in technical debt if we do it right. If that looks like too much work, there is a hacky alternative that will suffice for our current problem. That solution would be to render the GOOL keyword `self` as `(*this)` in C++. We'd need to change `selfFuncApp` and `selfVarAccess` to use `.` instead of `->`, but I believe that after those changes all our generated code would still work, and the vectors code would also work. The downside is that the generated code would be more ugly (from `this->method()` to `(*this).method()`), and it would add the implicit assumption that we only ever care about `this` as a value rather than a pointer.

JacquesCarette:
[...]

PR #5113, opened 2026-06-10
Simplified and renamed selfFuncAppMixedArgs

B-rando1:
#5111 explains my understanding of why `selfFuncAppMixedArgs` (now called `selfMethodCallMixedArgs`) exists in the first place. Except for C++ it should be a single function that calls `objMethodCallMixedArgs'`, but C++ needs it to use `->` instead of `.` between `this` and `method()`. I'd still like to push that decision into `objMethodCallMixedArgs'`, but for now this simplifies and clarifies the method.


JacquesCarette (approved):
Let's see how this all evolves. This does feel like a 'locally better' way to handle things, given your analysis.

Issue #4782, opened 2026-03-02
Insert more types of chunks into the `ChunkDB`


balacij:
Spawns from https://github.com/JacquesCarette/Drasil/issues/3351#issuecomment-3987108969

Previously, chunks were built in a nesting fashion where 1 `UID` could lead to many different chunk types. Since #2873 at least, we have a hard requirement that each `UID` must uniquely refer to a single chunk type to a single, unique, reusable piece of information.

Our `ChunkDB` design involved exporting the "internal" chunks from the "larger" chunks (the ones carrying others, inheriting their `UID`) and placing them in the `ChunkDB`. Now, with the new design, we should move away from this. For example, `CI` is a 

The goal of this issue is to move away as much as possible from inserting only the "inner" chunks.

For example:

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L127-L129

We currently limit ourselves to 12 chunk types being inserted.

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L116-L125

For our `ConceptChunk`s inserted, we extract them from other chunks when we should be inserting the entire chunks as they are.

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs#L110-L114

We don't insert `CI`s, only their inner `IdeaDict`s.

https://github.com/JacquesCarette/Drasil/blob/8ed6fa31f179c6010afe495cb70a6614427a88d4/code/drasil-example/gamephysics/lib/Drasil/GamePhysics/Unitals.hs#L27-L28

We only insert `DefinedQuantityDict`s when we have other variable-defining-like chunk types.

JacquesCarette:
You have an incomplete sentence "For example, `CI` is a".

PR #5123, opened 2026-06-10
fix(bss): insert `ConceptChunk`s as they are

balacij:
Contributes to https://github.com/JacquesCarette/Drasil/pull/4829#issuecomment-4092462316
Contributes to #4782 
Contributes to #2873 

Builds on #5120

Along the way, I add `gravity` to the list of `ConceptChunk`s, which in turn adds it to the list of terms and defs. Is this okay? Not sure, but we can fix it!


JacquesCarette (approved):
Code is fine. And while it is not wrong to have 'gravity' in the Terminology section, it feels like overkill. So I definitely want @smiths to look at it. And @Xinlu-Y too.

Xinlu-Y:
The refactor looks good to me. 
The only side effect is it now lands in T&D (overlapping Gravitational interaction), because `TermsAndDefs` just uses the whole `conceptChunks` list. I'm fine to leave for now, but ideally we could keep `gravity` in the DB without listing it in T&D?

smiths:
I agree with @Xinlu-Y.  I don't like having the concept of "gravity" and "gravitational interaction".   I'm okay with defining one of them, since gravity is so fundamental to this problem, but I don't like defining the concept twice.

code/drasil-example/bss/lib/Drasil/BinaryStar/Concepts.hs
```diff
@@ -28,10 +28,12 @@ starTwo = compoundNC second_ star
...
-defs :: [ConceptChunk]
-defs = [binaryStarSys, starBody, gravInteraction, newtonLUG,
-        initialConditions, trajectory, centerOfMass, inertialRefFrame,
-        simTimeSpan]
+conceptChunks :: [ConceptChunk]
+conceptChunks = gravity : ccsFortermsAndDefsTbl
+
+ccsFortermsAndDefsTbl :: [ConceptChunk]
+ccsFortermsAndDefsTbl = [binaryStarSys, starBody, gravInteraction, newtonLUG,
+  initialConditions, trajectory, centerOfMass, inertialRefFrame, simTimeSpan]
```
balacij, line 36: Removing the addition of `gravity` to the Terms and Defs list is as straightforward as breaking up this list of `ConceptChunk`s in two.

The problem is that we _only_ go about it in this way, while this should be sorted automatically for us, with no real reason to have two lists (or any list for that matter). Background definitions brought about by background theories should be kept in the background. Of course, some can be highlighted for reminder purposes, but only when explicitly asked to do (e.g., it can be an opt-in option).

balacij:
Ready for review again :)

JacquesCarette (approved):

PR #5128, opened 2026-06-11
refactor: avoid `nw` usage in symbol convention declarations

balacij:
Builds on #5127 

Removes more `nw` usage in an operational way.

Does not contribute to using `UID`s/`UIDRef`s instead of embedding whole chunks, but it might be a good one to give as an example to fix. From a quick search with github issues, I don't see anything discussing this. I'll have to look for that later.


JacquesCarette (approved):
Nice to see the `classy lenses`-based polymorphism coming in to make things more direct, i.e. removing needless casts.

PR #5129, opened 2026-06-11
new(data-formats): Introduce JSON representation and renderer

djarabek:
Contributes to #4989

[RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259) served as my primary reference for the implementation.


code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Core.hs
```diff
@@ -0,0 +1,25 @@
...
+where
+
+import Data.Text (Text)
+import qualified Data.Text as T (pack)
+import Data.Scientific (Scientific)
+import Data.String (IsString(..))
+
+-- | A JSON representation.
+data JSON =
+    JObject [(Text, JSON)]
```
djarabek, line 17: There was a moment of panic when I thought that empty and duplicate keys weren't allowed, but the JSON spec allows these (with the caveat that implementations may handle them in different ways). I think it's fine to leave things like this for the sake of simplicity.

balacij: Good catch and thank you for noting that! It would be good to add this as a note to `JObject`'s documentation for future readers.

balacij (approved):

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Core.hs
djarabek: Done in 6cb2b76bc2d4a7103bdf4ece8dcd623239e4b2bb

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Core.hs
```diff
@@ -0,0 +1,27 @@
...
+    JSON(..),
+  )
+where
+
+import Data.Text (Text)
+import qualified Data.Text as T (pack)
+import Data.Scientific (Scientific)
+import Data.String (IsString(..))
+
+-- | A JSON representation.
```
JacquesCarette: Put the link to RFC 8259 also here please

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON.hs
```diff
@@ -0,0 +1,8 @@
+module Drasil.Data.Formats.JSON
+  ( module Drasil.Data.Formats.JSON.Core,
```
JacquesCarette, line 2: why export everything? We should try to export as little as possible

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Render.hs
```diff
@@ -0,0 +1,87 @@
...
+newtype JSONRenderOptions = JSONRO { style :: JSONStyle }
+
+-- | JSON Style: How the overall JSON document is formatted.
+data JSONStyle =
+    -- | With as few characters as possible and no newlines.
+    Minified
+    -- | In a "pretty" human readable way, with given indent size.
+  | Pretty Natural
+
+-- | Internal: Returns True if the given 'JSONStyle' is 'Pretty'.
```
JacquesCarette: Is this used? It is a bit of an anti-pattern

JacquesCarette (changes requested):

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON.hs
djarabek: I copied what was done in the CSV module. Core and Render already only export what should be exposed.

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Core.hs
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5129/commits/9cdf8e4ac0eb0a0a8e6b4d6febb85fec35819c20

code/drasil-data-formats/lib/Drasil/Data/Formats/JSON/Render.hs
djarabek: Removed this in https://github.com/JacquesCarette/Drasil/pull/5129/commits/93afd6ffa3398a67ec763ce7c2025159ccf947bc

JacquesCarette (approved):

PR #5132, opened 2026-06-11
drasil-utils: rename Lists→Data.List.Extras, Strings→Data.String.Extr…

Purav-Kanda:
Renames `Utils.Drasil.Lists` → `Data.List.Extras` and `Utils.Drasil.Strings` → `Data.String.Extras` within `drasil-utils`, as discussed in #1211. Updates the `Utils.Drasil` re-export module and `package.yaml` accordingly. `Utils.Drasil.English` is left unchanged pending the future `drasil-natural-language` package.

### DIFF
--- a/code/drasil-utils/lib/Utils/Drasil.hs
+++ b/code/drasil-utils/lib/Utils/Drasil.hs
@@ -1,10 +1,10 @@
 -- | Re-exports all utilities.
 module Utils.Drasil (
   module Utils.Drasil.English,
-  module Utils.Drasil.Lists,
-  module Utils.Drasil.Strings,
+  module Data.List.Extras,
+  module Data.String.Extras,
 ) where
 
 import Utils.Drasil.English
-import Utils.Drasil.Lists
-import Utils.Drasil.Strings
+import Data.List.Extras
+import Data.String.Extras
--- a/code/drasil-utils/lib/Utils/Drasil.hs
+++ b/code/drasil-utils/lib/Utils/Drasil.hs
@@ -1,10 +1,5 @@
 -- | Re-exports all utilities.
 module Utils.Drasil (
   module Utils.Drasil.English,
-  module Data.List.Extras,
-  module Data.String.Extras,
 ) where
-
 import Utils.Drasil.English
-import Data.List.Extras
-import Data.String.Extras
--- a/code/drasil-utils/package.yaml
+++ b/code/drasil-utils/package.yaml
@@ -27,3 +27,7 @@ library:
   source-dirs: lib
   exposed-modules:
   - Utils.Drasil
+  - Data.List.Extras
+  - Data.String.Extras
+  other-modules:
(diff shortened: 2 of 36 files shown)


code/drasil-utils/lib/Utils/Drasil.hs
```diff
@@ -1,10 +1,10 @@
 -- | Re-exports all utilities.
 module Utils.Drasil (
   module Utils.Drasil.English,
-  module Utils.Drasil.Lists,
-  module Utils.Drasil.Strings,
+  module Data.List.Extras,
+  module Data.String.Extras,
```
balacij: I think we should also remove these `Data.X.Extras` from the export list. Code can directly import them instead since they're now exposed.

code/drasil-utils/package.yaml
```diff
@@ -27,3 +27,7 @@ library:
   source-dirs: lib
   exposed-modules:
   - Utils.Drasil
+  - Data.List.Extras
+  - Data.String.Extras
+  other-modules:
+  - Utils.Drasil.English
```
balacij: I don't think you need to add this, do you?

balacij (changes requested):

code/drasil-utils/package.yaml
Purav-Kanda: Good catches — will remove the other-modules entry and strip Data.List.Extras/Data.String.Extras from the re-export, updating importers to use the new modules directly.

balacij: Looks like you meant to comment this on my other comment?

code/drasil-utils/lib/Utils/Drasil.hs
Purav-Kanda: Done — removed Data.List.Extras and Data.String.Extras from the re-export list and updated all importers to use them directly. Fixed in the second commit.

code/drasil-utils/package.yaml
Purav-Kanda: My bad, I made a mistake commenting on the wrong thread — done, removed the other-modules entry. Fixed in the second commit.

balacij: Thank you @Purav-Kanda ! Let me know when you've pushed.

Purav-Kanda: done just pushed.

balacij: The other-modules entry is still there.

Purav-Kanda: Sorry about that  removed now and pushed.

balacij: Nice. This is much better. Thank you, @Purav-Kanda .

balacij (approved):
Calling @JacquesCarette for his review now too.

JacquesCarette (approved):

PR #5137, opened 2026-06-11
refactor: simplify `cuc'` constructor

balacij:
Not a fan of this change, but this at least reduces some redundancy.


JacquesCarette (approved):
Every simplification helps, and they tend to compound to enable more simplifications.

PR #5138, opened 2026-06-11
Generating Functional Requirements for outputs 

camishibata:
Contributes to https://github.com/JacquesCarette/Drasil/issues/3259


code/drasil-example/glassbr/lib/Drasil/GlassBR/Requirements.hs
```diff
@@ -91,12 +93,9 @@ checkGlassSafetyDesc = foldlSent_ [S "If", eS $ sy isSafePb $&& sy isSafeLR,
...
 
-outputValuesDesc :: Sentence
-outputValuesDesc = foldlSent [titleize output_, D.toSent $ pluralNP (the value), S "from the table for", namedRef outputValuesTable (S "Required Outputs")]
-
-outputValuesTable :: LabelledContent
-outputValuesTable = mkValsSourceTable (mkQRTuple iMods ++ mkQRTuple r6DDs) "ReqOutputs"
-                              (S "Required" +:+ titleize' output_)
+outputValuesTable :: Maybe LabelledContent
+outputValues :: ConceptInstance
+(outputValues, outputValuesTable) = outReq Nothing (NE.fromList $ mkQRTuple iMods ++ mkQRTuple r6DDs)
```
balacij, line 98: This is definitely better than before!

Can we further hide this `outReq` code behind the `mkSmithEtAlICO` smart constructor:

https://github.com/JacquesCarette/Drasil/blob/4ca2449e7df991dfde9f82505a7b09088df7add4/code/drasil-system/lib/Drasil/System/SmithEtAlSRS.hs#L73-L87

balacij: @camishibata I just pushed e57258eda6f0612b76c19f796982defbee022a26 as demo code you can use to get this going. With this, you should be able to remove any use of `outReq` from the examples, opting to move its usage to this file.

Note: You _might_ need to move this file from `drasil-system` to `drasil-srs`. I did this for [`LessonPlan`](https://github.com/JacquesCarette/Drasil/pull/5116) already. This work is eventually necessary, so its good that it's coming up as a natural issue.

Note 2: It's not perfect, it seems to crash for one of our examples, but it should serve as a good base and will show us how to fix other code.

JacquesCarette:
What is needed to 'unstick' this?

JacquesCarette:
As this sits as Draft, it is accumulating conflicts. What's the sticking point @balacij ?

balacij:
Right. Apologies for the delay.

This is hard blocked on being able to infer where each `QDefinition` used in the final calculation scheme originates (in particular, we want traceability back to IMs and DDs). Why is this hard blocked on that? Because not all `QDefinition`s used in the final calculation scheme are traceable back to IMs nor DDs.

The greater problem for generating the input and output FRs is that there's a mismatch between the SRS (captured in `SmithEtAlSRS`) and the "code" (`CodeSpec`) that is difficult to reconcile. It's not a crazy complex problem to fix, it just needs a bit of analysis.

PR #5139, opened 2026-06-11
jot down thoughts on the idealized process

JacquesCarette:
My reading has congealed somewhere, so I jotted that down. I expect to have to refine this multiples times. And, of course, the Meta-Requirements section is likely going to be 10 times larger than the current document! (And so I will likely put it elsewhere.)

But in the meantime, I would very much like to hear from @balacij and @smiths on this. But anyone on the Drasil team is also quite welcome to comment.

### DIFF
--- a/wiki/Idealized-Process.md
+++ b/wiki/Idealized-Process.md
@@ -1,5 +1,63 @@
 # Idealized Process
 
+See at the end for older notes on this topic.
+
+## Thoughts on the Idealized Process
+
+So I (Jacques) have been reading (on DDD and 'meaning' in software and other
+similar things), and this has helped me see what the steps could be. I will
+give an enumerated list of steps but, just like Parnas' Rational Process,
+these are not meant to be taken literally. In practice, there will always be
+iteration.
+
+I will first list all the steps, then give some commentary on them. So the
+first list is meant to be succinct and, because of that, may well be
+ambiguous. Convention: steps in _italics_ are optional. A step with a
+**bolded** verb means that this is (currently) an "on paper" step.
+
+1. **Gather** all relevant theories needed to describe the problem domain.
+2. Choose (or implement), from Drasil's database all theories gathered in step (1)
+3. _Generate a theory Manual_.
+4. Weave a context(s) out of the above theories for problem description.
+5. Describe the problem that needs to be solved, using the woven contexts.
+6. _Generate a Problem Description Document_
+7. **Gather** together the relevant theories to describe the tools (domain
+specific and software/computing) involved
+in the solution, as well as the theories for describing the properties
+necessary for a good solution.
+8. Choose (or implement), from Drasil's database, all theories gathered in step (7).
+9. _Generate a Solution Landscape Document_
+10. Describe the functional requirements for a solution, and the properties of a good solution
+11. Create an SRS, that weaves together the problem description of step (5) and the information in step (10)
+12. _Generate an SRS_
+13. Refine the SRS so that it is implementable.
+14. _Generate a Design Document_
+15. _Prove that the designed solution meets its requirement_
+16. _Generate the proof document_
+17. Refine the implementation.
+18. _Generate a software dossier_
+
+## Commentary
+
+- Pretty much all 'Generate' steps will have _options_ that control how it
+looks as well as various details specific to it.
+- Many items are implicit/tacit above. Things like 'problem frames' are not mentioned at all. These will arise when checking the adequacy of the contexts.
+- The subtlety between natural language generation and code generation is also omnipresent yet implicit.
+- Basically: every single step above encapsulates many pages worth of description
+- 16 of the 18 steps will generate one or more DSL that is well-suited to that
+step.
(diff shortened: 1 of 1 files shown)


balacij (commented):
Writing a review as a single comment instead of inline so we don't clutter inline reading of the document.

1. What is (3)?
2. How does (4), (5), and (6) differ?
3. How much does work involved with (7) reuse work from (4-6)?
4. (8) is confusing: I was fully expecting this to be work done in (4-6)?
5. (9): Solution _Landscape_ Document is interesting. I like it. How does it differ from (7)?
6. Is it strictly necessary to include `_Generate a[n] X_` as steps? Or can they just be folded into the previous step? Or do you mean to imply that there exists choices that can occur at those steps? (Ah, ok, I read your commentary now, and I see what you mean.)
7. How does (13) work? What exactly are we refining here? Making the SRS encoding more concrete? Cloning the theories and gradually concretizing them?
8. How is the proof document related to the SRS? Some terms we use are just terms, e.g., portability, that we happen to build generators for that produce portable code. Does (15) require automated verification, manual, or a hybrid?
9. What is (17) about? Would the implementation not need to be fully finished before (16)?
10. (18): Shouldn't it be before (16) as well?
11. Commentary (v): Very true.

"Solution _Landscape_" is really nice. Is that common terminology in DDD?

JacquesCarette:
Answering each above, using the same numbering

1. The 'theory manual' is documentation of just the background theories in play. It is pure reference material. The relevant parts of it currently end up in the SRS (too). This could be more detailed, and arranged very differently.
2. (4) create a pure context in which to talk about the problem, (5) uses that context to describe the problem, and (6) chooses how to render the problem description.
3. We have to try it to know for sure. But (7) for one of our examples would involve "numerical solution of ODEs" which would not have been needed for (4-6). Similarly, (7) will involve many "software theories" that would not be in (4-6)
4. See point above: problem space vs solution space.
5. The document is a "chosen view" of the theories gathered. The theories might contain lots of irrelevant stuff, so we need somewhere to narrow the view.
6. Yes: choices abound in the generation step.
7. I have no idea how (13) works! This is for us to discover. But this is where 'choices' happen (this or that algorithm, etc)
8. The SRS is a specification document that only hints at parts of the solution. The proof document (which would be unavailable as an option during your PhD, BTW) would need to know about the actual solution to be feasible.
9. (17) is about many of the options you're currently working (when you're working on the Human Readable Code" paper) where lots of code-level choices are made
10. (18) and (16) can be independent.

So: what parts should I be clarifying in the steps / commentary / forthcoming meta-requirements?

smiths:
I've started thinking about the idealized process.  My approach was to write down my own version of the [idealized process from the perspective of someone doing scientific computing](https://github.com/JacquesCarette/Drasil/blob/idealICO/wiki/idealICO-SCperspective.md).  I wanted to do this before thinking too much about the version by @JacquesCarette, so that I didn't just do the same thing.  The next step is to compare the two views of the ideal process.

My initial thoughts on the @JacquesCarette process are as follows:

- what is meant by theory?  Is it intended to be something like Bill's little theories?  A theory is more than just an equation, right?
- are the gathered theories the only ones needed, or are they are superset of the needed theories? That is, is this idealized process supporting exploration, or are we faking it so that this idealized version is the final version after all the exploration.
- what is a context?
- I wonder if we should split the steps with "and" into separate steps for a proper separation of concerns?
- is there theory refinement in the process?  I'm assuming that there is.  Where does it come into the process?

I used examples from projectile motion in my version of the process.  Examples would be good in the Jacques version too.  :smile:

JacquesCarette:
Your write-up is great @smiths ! You are correct that we should be comparing that process with the one that I suggest would be done via future-Drasil. Indeed, many of the parts that you carefully document would have arisen from making mine more explicit in its sub-steps. And also, I'm quite sure that  a careful comparison will identify small issues that need to be resolved.

As to your questions (same order of bullets):
- a 'theory' is a set of declarations of types, constants, functions, relations, axioms and definitions. A Drasil 'theory' will likely contain non-mathematical types and constants (i.e. our vocabulary), which will have attached information that is usually not present in 'mathematical' theories. So like little theories, just richer. Equations show up as either relations or definitions.
- superset. Refinement happens later, which narrows the theories down to what is actually used.
- a 'context' is a set of definition of variables, valid within a theory.
- I already split some steps that had 'and' (previous to pushing). The ones remaining might be a *vocabulary* problem where I am forced to use 'and' because I had not word(s) that properly described the 'unique' task in that steps. Unless I erred, at which time we should split.
- refinement happens inside steps 2 and 6. But more refinement will happen in "real life" when the process is run and one is forced to go back to previous steps because of unforeseen needs. But also, if I had split each step into sub-steps, that should have been visible.

JacquesCarette:
@smiths @balacij What should I change/add to this PR before it is merge-able?

smiths:
@JacquesCarette, I think we can merge what is currently in this PR.  We all know that more work will be necessary, but this is a good starting point.

smiths (approved):

balacij (approved):
I agree with @smiths.

I don't think I have anything I can contribute right this instant, and I don't any reason to block this PR.

PR #5141, opened 2026-06-12
feat(`ConceptChunk`+`IdeaDict`): create new smart constructors that avoid/discourage chunk casting and nesting, deprecate old ones

balacij:
Contributes to #4541
Contributes to #4829

~~DISCLAIMER: This PR _will_ fail in the CI because I `DEPRECATED` the old smart constructors. We should file PRs targetting this one that remove uses of the old ones. Until all are removed and the CI is green, this PR should not be merged.~~

Why?

Let's have a look at the old ones:

```haskell
-- | Smart constructor for creating a concept chunks with an abbreviation
-- Takes a UID (String), a term (NounPhrase), a definition (String), and an abbreviation (Maybe String).
dccA :: String -> NP -> String -> Maybe String -> ConceptChunk
dccA i ter des a = ConDict (mkIdea i ter a) (S des) []
```

Uses a `String` for the description. We should be strongly discouraging this, even if it makes the code shorter, it does it in a way that encourages bad practice.

```haskell
dccAWDS :: String -> NP -> Sentence -> Maybe String -> ConceptChunk
dccAWDS i t d a = ConDict (mkIdea i t a) d []
```

Better than `dccA`, but has a confusing name, no sane description and was only a compatibility layer for when `ConceptDomain` was created and not propagated yet. We will keep this one, renaming it.

```haskell
dcc :: String -> NP -> String -> ConceptChunk
-- | Smart constructor for creating concept chunks given a 'UID',
-- 'NounPhrase' ('NP') and definition (as a 'String').
dcc i ter des = dccA i ter des Nothing
-- ^ Concept domain tagging is not yet implemented in this constructor.
```

See `dccA`'s `String` discussion.

```haskell
-- | Similar to 'dcc', except the definition takes a 'Sentence'.
dccWDS :: String -> NP -> Sentence -> ConceptChunk
dccWDS i t d = dccAWDS i t d Nothing
```

Like `dccAWDS`, more reasonable, but very confusing name.

```haskell
-- | Constructor for projecting an idea into a 'ConceptChunk'. Takes the definition of the
-- 'ConceptChunk' as a 'Sentence. Does not allow concept domain tagging.
cc' :: Idea c => c -> Sentence -> ConceptChunk
cc' n d = ConDict (nw n) d []
```

It takes something that has an `Idea` instances (and possibly already a definition), and attaches a new definition to it. This is bad for two reasons:

1. It down-casts (/projects) another chunk type.
2. It creates data that is potentially out-of-sync with the `UID` and expected type the `UID` should reference.

```haskell
-- | Similar to 'cc'', but allows explicit domain tagging.
ccs :: Concept d => IdeaDict -> Sentence -> [d] -> ConceptChunk --Explicit tagging
ccs n d l = ConDict n d $ map (^. uid) l
```

Providing a raw `IdeaDict` encourages bad practice. The `ConceptChunk` being able to nest an `IdeaDict` internally and have a unique `UID` is an operational detail (kind of hacky to de-duplicate code) that we should be hiding. Good, but bad. Also doesn't have the abbreviation support.

```haskell
-- | For projecting out to the 'ConceptChunk' data-type.
cw :: Concept c => c -> ConceptChunk
cw c = ConDict (nw c) (c ^. defn) (cdom c)
```

Down-casts another chunk type. We should never be needing to do this.

This PR creates 3 new constructors based on the old ones: `cncpt`, `cncpt'`, and `cncpt''`. `cncpt` is strongly preferred over the other two, but having the other two is good if only to avoid having to write `Nothing` everywhere to avoid abbreviations and `[]` to avoid having to write empty 'concept domains'.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
@@ -3,7 +3,8 @@
 module Language.Drasil.Chunk.Concept (
   -- * Concept Chunks
   -- ** From an idea ('IdeaDict')
-  ConceptChunk, cncpt, cncpt', cncpt'', dcc, dccA, dccAWDS, dccWDS, cc', cw,
+  ConceptChunk, cncpt, cncpt', cncpt'', cncpt''',
+  dcc, dccA, dccAWDS, dccWDS, cc', cw,
   -- ** From a 'ConceptChunk'
   ConceptInstance, cic
   ) where
@@ -15,12 +16,15 @@ import Drasil.Database (HasUID(uid), nsUid, UID)
 import Language.Drasil.Classes (Idea, ConceptDomain(cdom), Concept)
 import Language.Drasil.Chunk.Concept.Core (ConceptChunk(ConDict), ConceptInstance(ConInst))
 import Language.Drasil.Sentence (Sentence(S))
-import Language.Drasil.Chunk.NamedIdea(mkIdea, nw, nc, idea)
+import Language.Drasil.Chunk.NamedIdea(mkIdea, nw, nc, idea, idea')
 import Language.Drasil.NaturalLanguage.English.NounPhrase (NP, pn)
 import Language.Drasil.ShortName (shortname')
 import qualified Language.Drasil.Classes as D (defn)
 
---FIXME: Temporary ConceptDomain tag hacking to not break everything.
+-- FIXME: There should only be two smart constructors ultimately for
+-- `ConceptChunk`s. One with an abbreviation, the other without. In other words,
+-- only `cncpt` and `cncpt'` should exist. The other ones should not. The
+-- problem here is that dealing with the other ones requires domain analysis.
 
 -- | Construct a 'ConceptChunk'.
 cncpt :: Concept dom =>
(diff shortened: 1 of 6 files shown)


code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
```diff
@@ -11,49 +12,88 @@ import Control.Lens ((^.))
...
+import qualified Language.Drasil.Classes as D (defn)
 
 --FIXME: Temporary ConceptDomain tag hacking to not break everything.
 
--- | Smart constructor for creating a concept chunks with an abbreviation
--- Takes a UID (String), a term (NounPhrase), a definition (String), and an abbreviation (Maybe String).
+-- | Construct a 'ConceptChunk'.
+cncpt :: Concept dom =>
+  -- | The 'UID'.
+  String ->
```
balacij: This will cause a huge diff in the future, but the UID should really be of type `UID`.

JacquesCarette (commented):
Seems like a good move. But I should wait until the CI is green?

balacij:
On second thought, merging this PR all at once might be an extremely large task, so it might be best for us to:

1. Stop upgrading `DEPRECATION` warnings to hard errors (but only these warnings).
2. Apply `DEPRECATION`s generously.
3. Do this work in very small pieces against `main` instead of targetting this PR.

I just pushed (1), (2) was contributed to by this PR, and (3) will be done in subsequent PRs.

code/drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs
```diff
@@ -64,6 +49,26 @@ instance NamedIdea IdeaDict where term = np
 -- | Finds the abbreviation of the 'IdeaDict'.
 instance Idea      IdeaDict where getA = mabbr
 
+idea :: UID -> NP -> String -> IdeaDict
+idea u t accAbbr = IdeaDict u t (Just accAbbr)
+
+idea' :: UID -> NP -> IdeaDict
+idea' u t = IdeaDict u t Nothing
```
balacij, line 56: I've also added two new constructors for `IdeaDict` and marked the rest as `DEPRECATED`. I won't add any more or else this PR will become far too large. I won't make any attempt at replacing any uses here, however. This was 'bad' enough.

balacij: (Added comments to them in [771513f](https://github.com/JacquesCarette/Drasil/pull/5141/commits/771513fbcdcea55cd82c48810ca91365e07d7d8b))

JacquesCarette (approved):

PR #5144, opened 2026-06-12
refactor: avoid `nc` constructor in favour `idea'`

balacij:
Builds on #5143 

Deals with more GHC deprecation warnings introduced in #5141.

Deletes `nc` along the way.


JacquesCarette (approved):
This change doesn't feel like a clear win on its own (as most of the code got longer...). But I'll accept it as part of a larger refactor that is worth it.

PR #5145, opened 2026-06-12
refactor: avoid `mkIdea` constructor in favour `idea[']`

balacij:
Builds on #5143 

Deals with more GHC deprecation warnings introduced in #5141.

Deletes `mkIdea` along the way.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
@@ -16,7 +16,7 @@ import Drasil.Database (HasUID(uid), nsUid, UID, mkUid)
 import Language.Drasil.Classes (Idea, ConceptDomain(cdom), Concept)
 import Language.Drasil.Chunk.Concept.Core (ConceptChunk(ConDict), ConceptInstance(ConInst))
 import Language.Drasil.Sentence (Sentence(S))
-import Language.Drasil.Chunk.NamedIdea(mkIdea, nw, idea, idea')
+import Language.Drasil.Chunk.NamedIdea(nw, idea, idea')
 import Language.Drasil.NaturalLanguage.English.NounPhrase (NP, pn)
 import Language.Drasil.ShortName (shortname')
 import qualified Language.Drasil.Classes as D (defn)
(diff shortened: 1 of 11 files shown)


code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
```diff
@@ -81,10 +81,18 @@ cncpt''' u trm defn = ConDict (idea' u trm) defn []
...
+    go (Just accAbbr) = idea u ter accAbbr
+    go Nothing        = idea' u ter
 
 dccAWDS :: String -> NP -> Sentence -> Maybe String -> ConceptChunk
-dccAWDS i t d a = ConDict (mkIdea i t a) d []
+dccAWDS i ter def a = ConDict (go a) def []
+  where
+    u = mkUid i
+    go (Just accAbbr) = idea u ter accAbbr
+    go Nothing        = idea' u ter
```
balacij: Not pretty, but these are also marked as deprecated.

code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
```diff
@@ -81,10 +81,18 @@ cncpt''' u trm defn = ConDict (idea' u trm) defn []
 -- a UID (String), a term (NounPhrase), a definition (String), and an
 -- abbreviation (Maybe String).
 dccA :: String -> NP -> String -> Maybe String -> ConceptChunk
-dccA i ter des a = ConDict (mkIdea i ter a) (S des) []
+dccA i ter def a = ConDict (go a) (S def) []
+  where
+    u = mkUid i
+    go (Just accAbbr) = idea u ter accAbbr
```
JacquesCarette:
```haskell
go = maybe (idea' u ter) (idea u ter) 
```

code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
JacquesCarette: See above.

JacquesCarette (changes requested):

code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
balacij: Thanks! fc6919c70f483d06169a788caeee7e6a12440aaa

balacij:
Ready for review again!

JacquesCarette (approved):
I'm not entirely sure that a (local, noisy) call to `mkUID` surrounding a string is really an improvement... but I'll see where this goes.

Issue #3472, opened 2023-06-08
Investigate the use of the `abrv` method from class `CommonIdea`


JacquesCarette:
Not so interested in who defines it but rather, where is it used? We need to know what code relies on abbreviations existing for certain (like `getA`, which is fine). This will also include uses of `getAcc`, `getAccStr` and `prependAbrv` since those effectively to the same thing.

The results of this investigation will tell us how to move forward on implementation.

JacquesCarette:
Now that #3473 has been merged, this Issue is now blocking wrt moving forward on this project.

harmanpreet-sagar:
`getAcc` is used in examples in order to obtain the acronym of a `CommonIdea` and returns it in the form of a `Sentence`.

The examples that use this are:
`DblPendulum`, `GamePhysics`, `GlassBR`, `Projectile`, `SSP`, and `SWHS`.

The following code snippet showcases the exact locations of where this function is utilized and defined.

```console
drasil-docLang/lib/Drasil/Sections/Introduction.hs:  S "follows the", phrase template, S "for an", getAcc Doc.srs, S "for",
drasil-example/dblpendulum/lib/Drasil/DblPendulum/Assumptions.hs:twoDMotionDesc = atStartNP (the pendMotion) `S.is` phrase twoD +:+. sParen (getAcc twoD)
drasil-example/dblpendulum/lib/Drasil/DblPendulum/Body.hs:  sParen (getAcc twoD), phrase pendMotion, phrase problem,
drasil-example/gamephysics/lib/Drasil/GamePhysics/Assumptions.hs:assumpODDesc = allObject (getAcc twoD)
drasil-example/gamephysics/lib/Drasil/GamePhysics/Assumptions.hs:{-assumptions_list = enumSimple 1 (getAcc assumption) $ map (foldlSent)
drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs:purp = foldlSent_ [S "simulate", getAcc twoD, phrase CP.rigidBody,
drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs:scope = foldlSent_ [phraseNP (the physicalSim) `S.of_` getAcc twoD,
drasil-example/gamephysics/lib/Drasil/GamePhysics/Body.hs:  plural physLib, S "Similar", getAcc twoD, plural physLib, S "are"]
[...]

PR #5148, opened 2026-06-13
refactor: avoid and delete `cc'`, delete `nw`, simplifying and deleting some chunks along the way

balacij:
Builds on #5147 

Refactors glassbr to avoid all remaining `cc'` usage and deletes `cc'`. With `nw` now being only used in `cw`, this PR also inlines `nw` into `cw` so as to avoid any potential future use of `nw`.

I will annotate this PR with some commentary because I did a few things arguably out of scope that will help future refactors, and caught some problematic things.

GlassBR will need some work in the future to de-duplicate its chunks.


code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
```diff
@@ -103,15 +103,14 @@ dcc i ter des = dccA i ter des Nothing
...
-cc' :: Idea c => c -> Sentence -> ConceptChunk
-cc' n d = ConDict (nw n) d []
+{-# DEPRECATED cw
+  "Chunk down-casting is strongly discouraged. If you want to construct a `ConceptChunk`, use one of its normal constructors." #-}
 
 -- | For projecting out to the 'ConceptChunk' data-type.
 cw :: Concept c => c -> ConceptChunk
-cw c = ConDict (nw c) (c ^. D.defn) (cdom c)
+cw c = ConDict (go $ getA c) (c ^. D.defn) (cdom c)
+  where go (Just accAbbr) = idea  (c ^. uid) (c ^. term) accAbbr
```
JacquesCarette:
```haskell
go = maybe (idea' (c ^. uid) (c ^. term)) (idea (c ^. uid) (c ^. term))
```

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
```diff
@@ -119,7 +115,15 @@ nomThick = cuc' "nomThick" (nounPhraseSP "nominal thickness")
...
-  [sfwrElem $ mkSet String $ map (str . abrv . snd) glassType]
+  -- FIXME: This is a hack for two reasons:
+  -- 1. `fromMaybe`.
+  -- 2. It should have never been using the abbreviation because that's not a
+  --    known _value_. This type-checks at the Haskell level, but not the Drasil level.
+  -- 3. This value is kind of like having `pi` to 4 decimal places. We still call
+  --    it pi, but it corresponds to a specific value. Here is similar where we
+  --    want a Drasil-level constant variable to reference an output-level literal string.
+  -- 4. What we really want here are "enumeration types" (sum types) for `glassType`
+  [sfwrElem $ mkSet String $ map (str . fromMaybe (error "bad fromMaybe; glassTypeCon") . getA . snd) glassType]
```
JacquesCarette: This is what `abrv` was *for*, not needing `fromMaybe` ! So by switching to `getA`, you're the one bringing in the need for it. We should think more deeply about this.

JacquesCarette (changes requested):

code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs
balacij: Thanks! Fixed in bd6b984ce1d3fec9614a030ee46d1c1c50056b38

code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs
balacij: Is f61e0be027 a bit more convincing?

I don't see `abrv` as any better than `fromMaybe` in this case. But, we're missing out on capturing some real knowledge, opting for hacking in an operational fix `abrv`. We're not capturing relationships between pieces of information. `abrv` is just duct tape to make the code look pretty. But, are any of the glass types really `CI`s? Ignoring capitalization, `FT` would normally mean "feet" (unit of length) to most readers. The information that `FT` means "fully tempered" is not captured by this code either way.

What I am doing here is making this hack a little bit more visible. And with f61e0be027's large comment, a lot more visible.

Aside: In general, I'm very skeptical of the necessity of `abrv`. I just don't understand why it is ever absolutely necessary for something to have a short-form, excluding as a critique of phrase frequency in a document, where, whenever possible, we should be using abbreviations and creating them as necessary to avoid heavily repeated, common phrases. Why is this something that someone writing an abstract SRS document or its components ever need to encode? It seems like a concrete decision that should be made when rendering an abstract document into its final form to add an abbreviation-usage-preference-policy, not something that a person encoding ~~relational~~ knowledge should be concerned about.

JacquesCarette: Not strongly so. The old code did not need `fromMaybe`, the new one does. I don't see why this is an improvement (unless `abrv` itself uses a `fromMaybe`? Then you are right, this is just moving the hack around).

I agree that *requiring* a short-form or abbreviation is dubious. I think the issue is that we have not captured the real meaning of what's going on in those cases. Short-form and abbreviation are too 'surface' and display-oriented. We need to understand what it all really means.

balacij: `abrv` does not use `fromMaybe`. It uses a duplicate copy (pointer) of the abbreviation `String`.

> Short-form and abbreviation are too 'surface' and display-oriented. We need to understand what it all really means.

Agreed.

I see what you're getting at.

Ok, I can revert this change. I'll open an issue for this as well.

EDIT: Instead of creating an issue, this is really a problem that is part of #3472.

JacquesCarette (approved):

Issue #5149, opened 2026-06-13
Replace all chunk constructors with ones that discourage duplication, casting, and raw `String` usage


balacij:
There are a few issues at play here:

1. We want all chunks to have a unique `UID`.
2. We want to insert all of our chunks in our chunk database, which is impossible if we have chunks with the same `UID` but different types/data.
3. We previously had a chunk and chunk database design that encouraged nesting, projecting out "smaller" chunks (subsets of data) from "larger" chunks (ones with more data), and "growing" chunks (i.e., nesting smaller ones in bigger ones). We moved away from this in #2873 but since not all chunks have been inserted into the `ChunkDB` yet, we haven't caught all the nesting issues yet. We want to insert all chunks in the chunk database (#4782), in particular without chunk casting. See #5141 or `nw` and/or `cw` from the codebase pre-[5141](https://github.com/JacquesCarette/Drasil/pull/5141).

In #5141, I created two new constructors for `IdeaDict`:

https://github.com/JacquesCarette/Drasil/blob/5eb3d16e695bb38985e899cb40cc37fed9a161d3/code/drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs#L52-L68

And marked the rest as DEPRECATED:

https://github.com/JacquesCarette/Drasil/blob/5eb3d16e695bb38985e899cb40cc37fed9a161d3/code/drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs#L70-L94

Now there are a few things different about the new constructors compared to the old ones:

1. There is no 'casting' constructor, e.g., `nw`. This is bad under the new rules of the `ChunkDB`. We should discourage this. In #5148, we remove `nw` altogether, but it is not the only case of this happening, for example.
2. We don't bother adding the `String`-based `UID` option. Users should be forced to provide explicit `UID`s, not `String`s. A part of the reason why here is that the `UID` is a type that indicates how the `String` would be used. The other part is that the `String` bit is an operational detail of creating a `UID`. If we were to change how `UID`s were constructed, we would need to change all constructors to also permit dealing with `Text`-based `UID`s, for instance.
3. If the chunk being constructed nests another one, such as in the case of `ConceptChunk`s new ones:
   https://github.com/JacquesCarette/Drasil/blob/5eb3d16e695bb38985e899cb40cc37fed9a161d3/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs#L29-L75
[...]

PR #5153, opened 2026-06-15
refactor(`DifferentialModel`): replace `UnitalChunk` requirement with just `DefinedQuantityDict`

balacij:
The independent variables just don't _strictly need_ units.

### DIFF
--- a/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/IMods.hs
+++ b/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/IMods.hs
@@ -49,7 +49,7 @@ eBalanceOnWtr = im (newDEModel' eBalanceOnWtrRC)
 eBalanceOnWtrRC :: DifferentialModel
 eBalanceOnWtrRC =
   makeASystemDE
-    time
+    (dqdWr time)
     tempW
     coeffs
     unknowns
--- a/code/drasil-theory/lib/Theory/Drasil/DifferentialModel.hs
+++ b/code/drasil-theory/lib/Theory/Drasil/DifferentialModel.hs
@@ -18,9 +18,9 @@ import Drasil.Database (HasUID(uid), HasChunkRefs(..))
 
 import Language.Drasil
   (ConceptChunk, dccWDS, Express(..), ConceptDomain(..), Definition(..), Idea(..), NamedIdea(..)
-  , ModelExpr, NP, Sentence, Expr, UnitalChunk, ModelExprC(nthderiv, equiv)
+  , ModelExpr, NP, Sentence, Expr, ModelExprC(nthderiv, equiv), DefinedQuantityDict
   , ExprC(..), columnVec, ConstrConcept, LiteralC(exactDbl, int), RequiresChecking (requiredChecks)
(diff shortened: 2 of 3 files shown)


code/drasil-theory/lib/Theory/Drasil/DifferentialModel.hs
```diff
@@ -67,7 +67,7 @@ type LHS = [Term]
 -- | Describe the structural content of a system of linear ODEs with six necessary fields
 data DifferentialModel = SystemOfLinearODEs {
   -- | independent variable, often time
-  _indepVar :: UnitalChunk,
+  _indepVar :: DefinedQuantityDict,
```
balacij, line 70: Because we use `makeLenses` on `_indepVar` and use `indepVar`, I opted to not convert this to a GADT and have `_indepVar :: Quantity q => q` because that would mean that `makeLenses` would no longer generate `indepVar`. We can do this in a subsequent PR.

For now, this PR serves strictly for #5154 to highlight that `UnitalChunk`s are not strictly required anywhere in Drasil.

code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/IMods.hs
```diff
@@ -49,7 +49,7 @@ eBalanceOnWtr = im (newDEModel' eBalanceOnWtrRC)
 eBalanceOnWtrRC :: DifferentialModel
 eBalanceOnWtrRC =
   makeASystemDE
-    time
+    (dqdWr time)
```
balacij, line 52: 1. #5154 mentions deleting `UnitalChunk`, which would mean converting `time` to a `DefinedQuantityDict` anyways.
2. #5149 will deal with `dqdWr` if #5154 doesn't happen first.

JacquesCarette: I really don't like this. This explicitly forgets information that is valuable.

A much better route to attack would be to resurrect our unit-checking code instead. *Then* we could remove the units from places where they are not needed.

JacquesCarette (changes requested):

code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/IMods.hs
balacij: I assume your thoughts on this change with https://github.com/JacquesCarette/Drasil/issues/5154#issuecomment-4711746942 ?

`time` would eventually become a plain `DefinedQuantityDict`, in which case, `dqdWr` would be removed. And in either case, there is no real loss of information other than the (unused and unnecessary) static Haskell-level type information that the variable provided has a unit.

balacij: Resurrecting unit-checking code sounds like a good idea regardless.

JacquesCarette: Yes, because we're not losing the units information.

JacquesCarette (approved):

PR #5157, opened 2026-06-15
Always build stack tests

djarabek:
This prevents stack from rebuilding the whole package when it comes time to run the tests.

See: https://github.com/JacquesCarette/Drasil/issues/5037#issuecomment-4693843789

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -168,6 +168,7 @@ GHCTHREADS ?= 2
 
 override GHCFLAGS += -Wall -j$(GHCTHREADS)
 override stackArgs += --ghc-options="$(GHCFLAGS)"
+override stackBuildArgs += $(stackArgs) --test --no-run-tests
 
 # When running 'stack exec', the following stack arguments should be passed.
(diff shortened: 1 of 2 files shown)


code/Makefile
```diff
@@ -168,6 +168,7 @@ GHCTHREADS ?= 2
 
 override GHCFLAGS += -Wall -j$(GHCTHREADS)
 override stackArgs += --ghc-options="$(GHCFLAGS)"
+override stackBuildArgs += $(stackArgs) --test --no-run-tests
```
balacij, line 171: Out of curiosity, why not merge this into `stackArgs`? Would it have any negative consequences?

djarabek: When we run tests, we do not want `--no-run-tests`, or the tests will not run.

balacij: Perfect. Thank you!

balacij (approved):

JacquesCarette (approved):

PR #5164, opened 2026-06-15
Fix stack tests which were breaking build

djarabek:


JacquesCarette (approved):
will wait for CI to go green.

balacij (approved):

PR #5170, opened 2026-06-16
Remove unneeded hedgehog dependency

djarabek:
We don't currently use it for any tests.


balacij (approved):
Yeah, I never ended up using them. They should be used eventually, however. We can remove them for now.

JacquesCarette (approved):

Issue #5167, opened 2026-06-16
Should the generated MATLAB code also run in real MATLAB, not just Octave?


Xinlu-Y:
The MATLAB renderer can generate `Projectile.m`. The code runs in Octave and gives the right answer.   

  Two things in the generated code only work in Octave, not in real MATLAB:

1. Reading the input file name from the command line
  - Now (Octave): `filename = argv(){1};`
  - `argv()` does not exist in MATLAB.
  
2. How the file is laid out
  - Now (Octave): the file starts with `1;`, then the functions, then the main code.
  - MATLAB wants the opposite: main code first, functions at the end, and no `1;`.

  So right now it runs in Octave but not in plain MATLAB.

In [this comment on #5062](https://github.com/JacquesCarette/Drasil/pull/5062#discussion_r3356883599), Dr. @smiths said the goal is code that works in both MATLAB and Octave, not just Octave.
  
A `.m` file is a script if it starts with normal commands, or a function file if it starts with `function`. Our output is a script right now (I copied the shape from the Julia renderer).

Scripts are exactly what cause the two problems above: a script has to use `argv()` to get its input, and it has the function-layout issue.   We don't really need MATLAB scripts anyway these generated programs just "take an input, compute, and write output," which is exactly what a function does.

 My current idea is:

  ```matlab
  function Projectile(filename)
      ...        % main code
  end
  function [v_launch, theta, p_target] = get_input(filename)
      ...
  end
  
  run with octave --eval "Projectile('input.txt')" or
  matlab -batch "Projectile('input.txt')".
```
  The small change this needs: the MATLAB main becomes a function that takes the file name as a parameter, instead of reading it from the command line so we don't need `argv()`, and local functions just work (the first function is the entry, the rest are local) in both MATLAB and Octave.

  Does this sound OK, or is there a better approach?

smiths:
Since the program is taking an argument (the filename), I think it makes more sense as a function than as a script.  Scripts in Matlab don't take arguments.  A common use case would be to run the function from within Matlab, but I like that you have the option of running it from the command line.

I think there are cases in Matlab where a script would make sense to generate, but not when there are run-time arguments.  (If we set all our variabilities at generation time, we won't need any arguments and a script makes sense.)

PR #5171, opened 2026-06-16
Add MATLAB rendering support for Projectile

Xinlu-Y:
It implements the MATLAB rendering pieces needed for Projectile, including printing, file I/O, command line arguments, conditionals, assignments, and variable declarations.


code/stable/projectile/projectile_u_p_nol_u_wi_v_d/src/matlab/Projectile.m
```diff
@@ -0,0 +1,107 @@
+% Projectile.m
+% A program to predict whether a launched projectile hits its target.
+% - Authors: Samuel J. Crawford, Brooks MacLachlan, and W. Spencer Smith
+% - Note: Generated by Drasil v0.1-alpha
+
+1;
```
samm82: It's been a while since I've used MATLAB (and don't think I have access to it through McMaster anymore), but is this `1;` necessary? (Genuine question 😅)

Xinlu-Y: Yes, it does look odd. Right now, the generated MATLAB code is using the Octave.

In Octave, if the file starts with a function, Octave treats the file as a function file. But our current file is meant to be a script with some local functions. So I put `1;`. 

It is not part of the real program logic. I am planning to change the output to a real function file, as discussed in #5167. After that, we should be able to remove it.

samm82: Awesome! Just double checking; glad this has been discussed! 😁

balacij: An alternative might be to clear the active workspace (e.g., `clear;`) instead of just outputting `1`. That way it's at least a somewhat 'good' feature?

It might also be good to add a comment that explains what the line is for.

Xinlu-Y: Yesss That is what I didn't think about.
but I think if we switch to a real function file, then we would not need `1;` or `clear;`.

Xinlu-Y:
heads-up: based on the meeting discussion, I'll change the program entry from a script to a function. I'll update this PR with that change, and open a separate PR for the designLog issue.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -447,9 +481,13 @@ instance CommentStatement MatlabCode where
 instance ControlStatement MatlabCode where
   break = undefined
   continue = undefined
-  returnStmt = undefined
+  -- MATLAB has no `return <expr>`: a function returns by assigning its named
+  -- output, so a return becomes `result = <value>;`.
+  returnStmt v' = do
```
B-rando1, line 464: We should avoid working directly with `Doc` as much as possible. It should be possible to get the same result with something like this:

```hs
returnStmt v' = do
  v <- zoom lensMStoVS v'
  tp <- getType v
  (var mlRet tp) &= v' 
```

(or something like that; the exact types can be finicky)

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -447,9 +481,13 @@ instance CommentStatement MatlabCode where
 instance ControlStatement MatlabCode where
   break = undefined
   continue = undefined
-  returnStmt = undefined
+  -- MATLAB has no `return <expr>`: a function returns by assigning its named
+  -- output, so a return becomes `result = <value>;`.
```
B-rando1, line 463: Would this catch cases of returning early? e.g.

```python
def max(a, b):
  if a > b:
    return a
  return b
```

would translate to (modulo my incorrect MATLAB syntax):

```matlab
function result = max(a, b)
  if (a > b)
    result = a
  end
  result = b
end
```

Is there any way to end function evaluation after assigning to `result`?

JacquesCarette:
I will wait until the dust settles a bit more before reviewing this.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
Xinlu-Y: Could we do `result = <value>; return;`?

B-rando1: Oh yeah, that should work. I didn't know if MATLAB let you do that, but since it does then that seems good to me.

JacquesCarette (approved):

PR #5174, opened 2026-06-16
Switch `String` to `UID` in `commonIdeaWithDict`

djarabek:
Contributes to #5149

This will likely conflict with #5148, and I suggest merging that PR first as it makes more complicated changes.


JacquesCarette:
There is now a conflict to resolve -- but at least #5148 is merged.

JacquesCarette (approved):

JacquesCarette (approved):

PR #5176, opened 2026-06-17
Remove `unitWrapper`

djarabek:
Contributes to #5149 

We do so by replacing `(IsUnit u) => u` with `UnitDefn`, which removes the need for `unitWrapper`. `unitWrapper` was effectively an identity function, since it constructs a `UnitDefn` from something that is an instance of `IsUnit`, however the only instance of `IsUnit` is `UnitDefn`.

Additionally, me and Jason discussed and determined that it doesn't make sense for units to be polymorphic. We should have everything we need to capture any unit in `UnitDefn`. I will investigate removing `IsUnit` entirely in a followup PR.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil.hs
+++ b/code/drasil-lang/lib/Language/Drasil.hs
@@ -99,7 +99,7 @@ module Language.Drasil (
   , scale, shift
   , derUC, derUC', derUC''
   , fund, fund', compUnitDefn, derCUC, derCUC', derCUC''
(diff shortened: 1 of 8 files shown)


JacquesCarette (commented):
I have a tiny doubt on "doesn't make sense for units to be polymorphic". But not enough to argue against this.

JacquesCarette (approved):
Oops, that was meant to be Approve, not Comment.

PR #5184, opened 2026-06-19
Added Reference Type to GOOL

B-rando1:
Contributes to #5111.
Builds on #5180.

This adds a Reference type to GOOL/GProc. Full list of changes:
- I changed C++ `self` to be of type `Reference (Object ...)`
- I changed `objMethodCall` in C++ to use `->` instead of `.` if it's passed a reference.
  - At first, I tried doing more typechecking, throwing an error if `objMethodCall` is given anything other than `Object ...` or `Reference (Object ...)`. The difficulty there is that we also call methods on `List`, `InFile`, `OutFile`, etc. I guess we could still do this if we're more fine-grained, and throw an error if we call it on `Integer`, `Float`, `Char`, etc. Should I update this?
  - I also made `selfMethodCall` a function instead of a method, and it just calls `objMethodCall` directly.
- I changed `instanceVarAccess` in C++ to use `->` instead of  `.` if it's given a reference.
  - I also made the `instanceVarSelf` implementation in C++ mirror the other languages in just calling `instanceVarAccess`.
  - I wanted to make `instanceVarSelf` a function (similar to `selfMethodCall` above), but I faced an issue with `LoggingFor`. In `LoggingFor`, it represents `Value`s directly as `Doc` (by design), which means that it throws out the `Value`'s type. This means that when logging for the C++ renderer, it can't do the trick that the C++ renderer now does where it checks if `selfMethodCall` is called on a `Reference` and using `->` or `.` accordingly.  This is something that we need to fix, both so that we can make `selfMethodCall` a function instead of a method and so that logging statements are correct whether we use `selfMethodCall` or `objMethodCall` on `self`. I don't completely know what the best way around this is; any ideas?

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/GenerateGOOL.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/GenerateGOOL.hs
@@ -26,8 +26,9 @@ import Drasil.Metadata (watermark)
 
 import Drasil.GOOL (VSType, SVariable, SValue, MSStatement, SMethod,
   CSStateVar, SClass, NamedArgs, SharedProg, OOProg, ValueSym(..), Argument(..),
-  ValueExpression(..), OOValueExpression(..), FuncAppStatement(..),
-  OOFuncAppStatement(..), ClassSym(..), CodeType(..), getCodeType)
+  ValueExpression(..), OOValueExpression(..), SelfSym(..), VariableValue(..),
+  FuncAppStatement(..), OOFuncAppStatement(..), ClassSym(..), CodeType(..),
+  objMethodCallMixedArgs, getCodeType)
 import qualified Drasil.GOOL as OO (SFile, FileSym(..), ModuleSym(..))
 
 -- | Defines a GOOL module. If the user chose 'CommentMod', the module will have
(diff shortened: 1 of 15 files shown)


B-rando1:
I'm thinking we'll have to add more machinery to `LoggingFor` one way or another - which is exactly what we didn't want to do :pensive: I can think of two ways we could do this:
- We could follow the main renderers, rendering `Value`s to `ValData`. But this would force us to expose a `renderValue` function to `drasil-code` that converts `SValue (LoggingFor lang)` to `Doc`.
- It _might_ work to have separate instances of all the methods implemented by `LoggingFor` that render `Value`s to `CodeType`. We could then have `instanceVarAccess` call those methods to get the type of the object and handle the Reference/non-reference cases accordingly.

B-rando1:
I was talking with Reed about this, and his suggestion was to flip the Logging architecture around. Instead of having `LoggingFor` and `LangCode` be separate entities, with `drasil-code` taking the output of `LoggingFor` and inserting it into a print statement in `LangCode`, `LoggingFor` should wrap `LangCode`, overloading `assign` to both do the original assignment and produce logging statements, using `LangCode` to get the name of the variable, etc.

It's a different architecture and might take a bit of work to get it working, so I'm not sure if I'm _completely_ sold. But it does seem a little more principled that what we have for logging. Any thoughts @JacquesCarette @balacij?

JacquesCarette (approved):

JacquesCarette:
I think Reed's idea is good.

JacquesCarette:
(Going to close/re-open to adjust the merge base fully).

JacquesCarette (approved):

PR #5185, opened 2026-06-19
Remove unused tol constant from projectile example

### LINKED ISSUE #5183: Remove unused constant in `projectile`: `epsilon`
In #4420, we removed the single use for $\epsilon$. Now it appears in the SRS with no reason.

We should remove that.

It would have been nice if it were automatically removed to begin with, but alas.

Purav-Kanda:
Closes #5183


samm82 (approved):
Other than the linting issue, this change (and its rationale that I'm now caught up on 😅) looks good to me!

smiths (approved):
Since we no longer have requirements related to hitting the target, the changes to stable look good to me.

Purav-Kanda:
Thanks for the reviews! I pushed a fix for the linting issue — could one of you re-approve when you get a chance?

JacquesCarette (approved):

Issue #5059, opened 2026-06-03
Add `NativeVector` Typeclass to GOOL/GProc


B-rando1:
Note that this issue only makes sense if we follow #5057 and remove the current vector typeclasses. We might decide to start with what we have there instead of starting from scratch here.

We're hopefully nearing the end of our quest to get vectors working in Drasil. A few days and a few PRs from now we should be at the point where we can try to resurrect #4909. Either by merging that or by writing classes per-language by hand, we'll have `drasil-code` tell GOOL to use vector classes when dealing with GOOL, since none of the supported languages have native support for vector operations.

Julia and MATLAB are different for two reasons: they don't support classes, and they do natively support vectors. Thus, the approach needs to be different. As discussed in #5005, we need to create a typeclass (name suggestions: `NativeVector`, VectorAsArray`, or simply `Vector`) that allows GOOL/GProc users to specify vector operations directly.

A couple of implementation notes:
- It should go in `InterfaceCommon`, since there's nothing inherently "procedural" about native vector support.
- It shouldn't be a part of `SharedProg`, since we don't want all our renderers to implement it.
  - We might want to make it part of `ProcProg`, because otherwise we'll have to add a `(NativeVector r) => ...` constraint to everything in `drasil-code` that uses it. Though we'd have to do this anyway if we added a procedural language that doesn't have native vectors... I can see it either way.
- It should have methods for everything in #4909

@Xinlu-Y would you be able to look at this? Thanks!

Xinlu-Y:
Thanks Brandon (and Jason)! Happy to take #5059
Sounds like the plan is to go with a fresh lightweight `NativeVector`, not reusing the old Thunk classes. I'll keep it out of `SharedProg` and add it to `ProcProg`, then implement the ops natively in the Matlab/Julia renderers. Should I wait on #5057 first, or start now?

JacquesCarette:
I'm fine with `NativeVector`. I would like `Vector`, but that might be premature.

Now, one of the things that has bitten us before is that there are two interfaces to vectors (vectors-as-expressions and stateful-vectors) that are different. In Drasil itself, as in math, I think being expression-based is more natural? For low-level code, stateful can be faster (but for high-level code, it's more complicated than that.)

B-rando1:
@Xinlu-Y with #5186 merged, I think you're good to start work on this whenever you have time. Let me know if you encounter any difficulties!

Xinlu-Y:
[...]

PR #5186, opened 2026-06-19
Removed `Vector` typeclasses

### LINKED ISSUE #5057: Remove Current `Vector` Typeclasses?
Just to refresh people's memory, I'm referring to these typeclasses:

https://github.com/JacquesCarette/Drasil/blob/82a37a9f849a09ca3fb36142cd09bd7b0ea4f0cd/code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs#L343-L371

They allow us to create inline vector operations, as shown in the `VectorTest` files:

https://github.com/JacquesCarette/Drasil/blob/82a37a9f849a09ca3fb36142cd09bd7b0ea4f0cd/code/stable/gooltest/java/VectorTest/VectorTest/VectorTest.java#L9-L19

It looks like these classes were added three years ago in #3504. It was a good effort, but there are some issues. The biggest one is that they inline all the vector operations, which produces rather ugly code. They also have types that are even more difficult to work with than the rest of GOOL.

It's a bit too bad; I don't completely understand why these typeclasses were created and then ignored. In theory we _could_ have used them to get vectors in Drasil, though the end result wouldn't have been as nice as with the approach we're taking now. Either way, the approach we're taking to get vectors in Drasil doesn't use these typeclasses, and there doesn't seem to be much value in keeping them around if they're not being used.

B-rando1:
Closes #5057 

I couldn't quite tell from the the issue if there was more discussion to be had, so we can keep talking if necessary. I'm still pretty sure that this is the way to go. Removing these was trivial because they're not used anywhere; and the typeclasses have such a "unique" design that starting from scratch with #5059 will be easier than adapting these ones.


JacquesCarette (approved):
I agree, let's not burden ourselves with a bad API. Best to start from scratch.

PR #5187, opened 2026-06-19
New logging architecture

B-rando1:
Contributes to #5111, allowing us to navigate around difficulties found in #5184. I'm keeping this a draft for now because it'll take a bit of work to plug this into `drasil-code`, but there's already some design decisions to talk about.

To summarize, this PR, if merged, will push our logging from `drasil-code` into `drasil-gool`. We do this by changing `LoggingFor` from a separate renderer that `drasil-code` uses to render variables to `Doc` to give to the real renderer, to a wrapper of the real renderer that modifies anything that needs to be logged. @TOTBWF said this is similar to a monad transformer. I think it might also be similar to the Decorator Pattern.

What I have now only logs variable assignment; I think we'll need a second one that logs function calls. I haven't completely thought how that'll work.

Then in `drasil-code`, we can get rid of the explicit logging, and instead choose either a "base" renderer for no logging, or a wrapper or a wrapper of a wrapper depending on logging options selected.

This feels like a big change and I'm still not completely sure how it'll work out, but I'm hoping that I can get it working and that it'll be a good change.


code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LoggingFor.hs
```diff
@@ -1,118 +1,65 @@
...
+-- TODO [Brandon Bosman, 06/19/2026]: This should be passed down from drasil-code
+logName :: (Literal r) => SValue r
+logName = litString "log.txt"
+
+instance (SharedProg lang) => AssignStatement (LoggingFor lang) where
+  (&-=) = (&-=)
+  (&+=) = (&+=)
+  (&++) = (&++)
+  (&--) = (&--)
+  assign x e = do
```
B-rando1: This is the main code that actually does anything. It wraps `assign` with basically what [loggedVar](https://github.com/JacquesCarette/Drasil/blob/153efe0243b3c5eb1cb6ec086458f5dcd40c8795/code/drasil-code/lib/Language/Drasil/Code/Imperative/Logging.hs#L29) currently does.

JacquesCarette (commented):
Right now, this is looking good. Let's see if plugging it in comes up with obstacles.

B-rando1:
Summarizing our discussion of `listSet`"

`drasil-code` has this particular bit of code for logging `listSet`:
```hs
convStmt (FAsgIndex v i e) = do
  e' <- convExpr e
  vlog <- mkVar v
  v' <- mkVar v
  t <- codeType v
  let asgFunc (C.List _) = valStmt $ listSet (valueOf v') (litInt i) e'
      asgFunc (C.Array _) = assign (arrayElem (litInt i) v') e'
      asgFunc _ = error "FAsgIndex used with non-indexed value"
      vilog = arrayElem (litInt i) vlog
      vi = arrayElem (litInt i) v'
  l <- maybeLog vilog vi
  return $ multi $ asgFunc t : l
  ```

It boils down to using `arrayElem` to get `arr[i]` to hand to the logger, since `listSet` doesn't produce an LValue to log. In our meeting we decided that using `arr[i]` to represent the target of `listSet` is still fine, but we shouldn't use `listSet` to do it.

Also, we need to make `listSet` be an `MSStatement` instead of an `SValue` so that we can add logging to it.

File: code/stable/glassbr/src/java/GlassBR/ReadTable.java [hunk 2]
```diff
@@ -67,19 +67,19 @@ public static void read_table(String filename, ArrayList<Double> z_vector, Array
...
-        outfile = new PrintWriter(new FileWriter(new File("log.txt"), true));
-        outfile.print("var 'x_matrix' assigned ");
-        outfile.print(x_matrix);
-        outfile.println(" in module ReadTable");
-        outfile.close();
-        outfile = new PrintWriter(new FileWriter(new File("log.txt"), true));
-        outfile.print("var 'y_matrix' assigned ");
-        outfile.print(y_matrix);
-        outfile.println(" in module ReadTable");
-        outfile.close();
```
B-rando1, line 82: This was the one pattern that I couldn't make a 1-to-1 translation from `drasil-code` to `drasil-gool`. I believe this code is generated by:
https://github.com/JacquesCarette/Drasil/blob/0a482386d9a5292a26faa61d7cc481e418d00845/code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs#L964-L977

Basically, it's working at a higher-level than GOOL knows about, so GOOL has no choice but to put the logging statements at the level of its highest-level, i.e. `stringListLists`, which is inside the loop and acting on temporary variables.

Our choices are to either accept this slightly worse logging, or encapsulate this pattern within GOOL so that we can log it properly.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LoggingFor.hs
```diff
@@ -1,177 +1,474 @@
...
+
+instance LiftLogging Float Float where
+  liftLogging = id
+  lowerLogging = id
+
+instance (LiftLogging u1 l1, LiftLogging u2 l2) => LiftLogging (u1, u2) (l1, l2) where
+  liftLogging = bimap liftLogging liftLogging
+  lowerLogging = bimap lowerLogging lowerLogging
+
+-- TODO [Brandon Bosman, 06/19/2026]: This should be passed down from drasil-code
```
B-rando1, line 76: This is fine for now, since `drasil-code` already hard-codes the variable name.

code/drasil-gool/lib/Drasil/Shared/LanguageRenderer/LoggingFor.hs
```diff
@@ -1,177 +1,474 @@
...
+varLogFile :: (VariableSym r) => SVariable r
+varLogFile = var "outfile" outfile
+
+valLogFile :: (VariableValue r) => SValue r
+valLogFile = valueOf varLogFile
+
+-- TODO [Brandon Bosman, 06/19/2026]: This should be passed down from drasil-code
```
B-rando1, line 83: We should probably fix this, since part of `Choices` for logging is the name of the logging file. @TOTBWF said that `ReaderT` is what we want to be using, though it might be difficult to work with because of the other state we're already dealing with.

B-rando1:
There were some concepts that I expected would be included with variable logging, but aren't. Namely:
- `stringSplit` (splitting a string on a character into a list of strings)
- `stringListVals` (given a list of variables and a list of strings, assign the $i^\text{th}$ element of the list of strings to the $i^\text{th}$ variable)
  - What a fascinating concept to be captured inside GOOL, btw... I feel like it could use some refinement
- `getFileInputLine` (reading the next line from a file handle into a variable)
- `getFileInputAll` (reading an entire file handle into a variable)
- `listDecDef` (declaring a variable of type "List" and assigning to it right away)
- `arrayDecDef`
- `setDecDef`

Enabling these would make the logs more verbose, but it might be good to think about at least.

JacquesCarette:
Can you open an issue with those "I expected but" please? Indeed, we should be able to log more.

code/stable/glassbr/src/java/GlassBR/ReadTable.java
JacquesCarette: Let's accept this slightly worse logging (for now) to gain the much nicer architecture. We can revisit later.

JacquesCarette (approved):

Issue #3454, opened 2023-05-31
Traceability Matrix assumes there will be more than one CI


ManvendraJani:
When generating the traceability matrix section the matrix definitions assume that there will be more than one of the `ConceptInstances`. This also assumes all examples use the same CIs. It should be built using what is actually there.

eg. if there is only one goal statement it should say `goal statement` instead of `statements`.

refer to #3447 for a more detailed explanation

Purav-Kanda:
Fixed in #5188 (merged). traceMatOtherReq now uses singular noun phrases (via a new plOrSg helper) when only one assumption/model/etc. exists, instead of always defaulting to plural. Feel free to close this whenever convenient.

PR #5188, opened 2026-06-21
Fix traceMatOtherReq to use singular noun phrases when only one insta…

Purav-Kanda:
…nce exists (#3454)


smiths (commented):

JacquesCarette (approved):
Nice.

PR #5190, opened 2026-06-22
Use new `DefinedQuantityDict` constructors in most places

djarabek:
Anywhere where it was 'trivial' (all the required arguments were already available inline) has now been converted to the new constructors.

Contributes to #5149


JacquesCarette (approved):
Nice de-nesting of constructors.

PR #5199, opened 2026-06-23
Switch PhysicalProperties quantities to new constructors

djarabek:
Contributes to #5149

This is a blueprint for what much of the other required work in `drasil-data` will look like. I'm open to feedback, as I'd like to get this right (especially on the formatting) before moving on, as the other modules are much larger.


JacquesCarette:
Very glad you paused right here!

Basically because this is going too far. We do need the concepts of (for example) 'density' as well as be able to declare a quantity for that concept. Here is where we need to start doing something different, namely to start the migration to 'referencing' concepts. I'll come up to discuss.

JacquesCarette:
Closing this one, as we've established it was not going in the 'right' direction.

PR #4879, opened 2026-04-09
Re-write traceability graph generator using `Doc`s.

balacij:
See GlassBR's traceability graphs (TGs):

Screenshot 2026-04-09 at 12 18 30 AM

They don't really make a lot of sense.

On looking at the traceability graph `.dot` generator, I realized a few things:

1. We do not have an internal `.dot` representation.
2. We are generating our TG-related `.dot` files with immediately file writing rather than using an intermediate encoding.
3. Our generated `.dot` files produce invalid syntax.

The goal of this PR is fix at least (2) and (3). This PR re-writes the TG-generating code to use `Doc` and fixes the output syntax. For reference, locally, the traceability graphs for GlassBR are as follows:

Screenshot 2026-04-09 at 12 21 37 AM

<hr>

Note: The goal of this PR is _not_ to perfect the outputted graphs. We can deal with that another day.

PR #5201, opened 2026-06-23
feat(`System`): create `ToFiles` typeclass for defining `System` software generators

balacij:
The meaning of "rendering" a `System` is to realize its final software artifacts. That is, to pull the final project code repository.

In this sense, I propose we rename `[Ss]ystem` to `[Pp]roject`. Refined explanation to be written (when I finish up the code).


JacquesCarette:
We had been mostly using "render" to mean "create a human-readable version of". I still like that meaning.

balacij:
Is this in conflict with that? I don't really see that being so.

We discussed 'rendering' a few meetings ago; that it was the name of the action we take after a 'system' has been fully concretized (up to a supported generator) that takes the system and renders it into a final set of software artifacts (files). So, 'human-readable' is implied by the 'software artifacts (files)' bit.

balacij:
I'm not sure of a good name yet, but I'm not as much of a fan of "render" anymore with 599d3cc1e2e0cc0dd10e5f834d58a4dbc7fa451d. Perhaps splitting `Render` into "generate" and "transform" is best, where `Generate` targets generating `[FileLayout]`s and `Transform` targets generating anything. Or perhaps `Transpile` and `Render`.

Ok, so, let's take a step back and discuss _what issue I was actually trying to tackle_ first (next comment).

balacij:
Background: we have a lot of generators/renderers/transpilers/transformers/etc. that transform/transpile things into other things and produce final software artifacts.

Issues: 

1. Options are _often_ placed as Haskell-level function arguments, but the groupings are not clearly explained and at times, misplaced.
2. We often compose and/or sequence the generators/transpilers/etc. without explicitly realizing it/mentioning it. 
3. We don't expose the various options of the "later-used generators/transpilers/etc." we should be able to (fairly trivially) expose.
4. We call all of our generators/transpilers/transformer/renderers/etc. different things. Using standard terminology is helpful, especially so we can clearly know what we are doing and where we do them.

The goals amount to:

1. Making it clear when we define a generator/transpiler/renderer/etc. and use them.
2. Clearly grouping and forcing good explanation (through documentation) why certain groups of generator/transpiler/renderer/etc. options exist and are grouped together.
3. Making the code generally more self-explanatory, understandable. Improving our explanations of what our own code does will make it easier to reuse said code and, of course, to explain how Drasil works to outsiders. In general, improving maintainability of our generators/transpilers/renderers/etc..
4. (Secondary) Exposing more generation options everywhere we can, or at least making it easier to do so.

Next post: Explaining the approach I tried and using `drasil-lesson-plan` as the example to find an initial design.

smiths:
Sounds great @balacij.  There is a clear connection between your discussion of generators/renderers/transpilers/transformers/etc and the paper on generating human readable code.  :smile:  This work will definitely make the paper easier to write.

JacquesCarette:
Yes, this "stepping back" is great. Deriving names through explanations (and via concrete examples) is definitely something I strongly support!

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
```diff
@@ -1,30 +1,69 @@
...
+
+      -- 3. Transpile SDL to TDL (Typesetting Document Language).
+      printSetting = piSys (plan ^. systemdb) (plan ^. lsnPlanRefs) Equational Engineering
+      pd = makeDocument printSetting nb
+
+      -- 4. Generate `Prettyprinter.Doc` body using TDL.
+      doc = genJupyterLessonPlan pd
+
+      -- 5. Emit final files (with `Prettyprinter.Doc` body).
+      files = [file [ps|{lsnFileName}.ipynb|] doc]
```
balacij: 1. Passing around a `ChunkDB` looks like a weak hint to us really wanting at least a `Reader`-based monadic environment. However, a regular `State` is more likely as we need the `ChunkDB` produced after the SRS generator has completed to generate code: https://github.com/JacquesCarette/Drasil/blob/98704433ce56657d8169c4c3e3f5876244cd9bb8/code/drasil-gen/lib/Drasil/Generator/CaseStudyVariants.hs#L63-L71
   I think this is now a 'strong' hint. It's something we've been suspecting for a while.
2. Optimally, I would like to see code that looks more like:

```haskell
    generate ...
>>> transpile ...
>>> generate ...
>>> emit ...
```

balacij: 3. I believe this is a "recipe". Perhaps that's what we should keep calling it for now.
4. I made a counting typo because I removed a step that I thought existed. I fixed this in a subsequent force-push.
5. The `Options` is really a roll-up of 3 different recipe options.

code/drasil-example/projectile-lesson/lib/Drasil/Projectile/Lesson/Body.hs
```diff
@@ -55,7 +55,7 @@ symbols = [horiz_velo, Qs.iSpeed, Qs.ixSpeed, Qs.iySpeed, Qs.speed, Qs.constAcce
   Qs.iPos, Qs.height]
 
 projectileMotionLesson :: CI
-projectileMotionLesson = commonIdea (mkUid "projMotLsn") (pn "Projectile Motion Lesson") "Projectile Motion" []
+projectileMotionLesson = commonIdea (mkUid "projMotLsn") (pn "Projectile Motion Lesson") "projectile-lesson" []
```
balacij, line 58: 6. Digging our heels deeper into this `CI`/`abrv` usage issue so we gain more information about how to deal with it properly. #4884 suggests using a chunk. I think we should just create a `RepoName` newtype that wraps a `Symbol` for now. Using valid `git` repository names as a benchmark here, that `Symbol` could be anything. However (!), we commonly restrict ourselves to names that wouldn't produce different URL slugs when shipped off to a git server. That is, we may consider restricting ourselves to names with the same [percent encoded version](https://en.wikipedia.org/wiki/Percent-encoding). This is strictly a convention, however. Alternatively, we can have it be an explicit `OsPath`.

balacij: I said `Symbol` because I thought it would let us easily create repo names such as $x^2$, but we don't have a unicode text renderer for `Symbol`. So, a plain ole `String` (or `Text`) is probably fine, too. If we only use it for knowing which file/folder to write to, then `OsPath` is probably best.

balacij:
Thank you both! Indeed, this is spawning in from work on `drasil-code`, `drasil-gen`, and `drasil-system`. `drasil-lesson-plan` is just the 'smallest' project, so its akin to `Projectile` in that its one of the better case studies to examine Drasil through.

I really want the code to be immediately clear in what's happening with the 'larger' actions (https://github.com/JacquesCarette/Drasil/pull/5201#discussion_r3476202611).

> 1. Making it clear when we define a generator/transpiler/renderer/etc. and use them.
> 2. Clearly grouping and forcing good explanation (through documentation) why certain groups of generator/transpiler/renderer/etc. options exist and are grouped together.
> 3. Making the code generally more self-explanatory, understandable. Improving our explanations of what our own code does will make it easier to reuse said code and, of course, to explain how Drasil works to outsiders. In general, improving maintainability of our generators/transpilers/renderers/etc..
> 4. (Secondary) Exposing more generation options everywhere we can, or at least making it easier to do so.

So, what did I do?

First, I grouped the options for the lesson plan generator (code comments omitted):

```haskell
data Options = Options {
  lsnDesc :: LsnDesc,
  titleComb :: CI -> CI -> Sentence,
  lsnFileName :: String
}
```

This way if we have a function we call a 'generator', we can more easily tell which options belongs to it and we can force ourselves to tell a coherent story about why these options are grouped together to perform one stage-jump-down in the generator pipeline. Then we can propagate the individual options/knobs as needed to used generators. When we have nested generators, we can nest them too. For example:

```haskell
data Options = Options {
  -- Basic generation options
  lsnDesc :: LsnDesc,
  titleComb :: CI -> CI -> Sentence,
  
  semanticOptions :: SemanticDocumentOptions,
  jupyterOptions :: JupyterNotebookOptions,

  -- File layout options (single-file generator, no options because we already decide on exclusively generating Jupyter notebooks)
  lsnFileName :: String
}
```

And of course, these options would just belong to two generators that are used in the generation pipeline. Doing this would also simplify our lives in exposing more generation options.

Next, I created a _single_ typeclass, `Render`, that contains a single function, `render`:

```haskell
-- | A "system" can be rendered into a set of (human-readable) software
-- artifacts. An instance of `Render` is approximately what it means to define a
-- off-the-shelf software generator.
class HasSystemMeta sys => Render sys opts | opts -> sys where
  render ::
    -- | The system.
    sys ->
    -- | The rendering options.
    opts ->
    -- | The final, rendered software artifacts.
    [FileLayout]
```

As I mentioned earlier, I believe this is an abstract "recipe" type declaration/a fully furnished software generator. A 'system' can be fully concretized and rendered into a final set of software artifacts about a specific set of options. Perhaps a better name for this typeclass would be `GenerationRecipe`, `Recipe`, or `Generator`, but that's not important _right now_ (I do want to fix the names).

Then, I reworked the software artifact generator for `drasil-lesson-plan` to instantiate this typeclass:

```haskell
instance Render LessonPlan Options where
  -- | Renders a 'LessonPlan' using a 'LsnDesc' (a description of the document
  -- contents and organization) and a title combinator merging "notebook" with the
  -- name of the 'LessonPlan'.
  render plan Options{..} = files
    where
      -- Steps:

      -- 1. Generate SDL (Semantic Document language) using `LessonPlan`.
      nm = notebook `titleComb` (plan ^. sysName)
      as = foldlList Comma List $ map (S . fullName) $ plan ^. authors
      nb = Notebook nm as $ mkSections (plan ^. systemdb) lsnDesc

      -- 2. Transpile SDL to TDL (Typesetting Document Language).
      printSetting = piSys (plan ^. systemdb) (plan ^. lsnPlanRefs) Equational Engineering
      pd = makeDocument printSetting nb

      -- 3. Generate `Prettyprinter.Doc` body using TDL.
      doc = genJupyterLessonPlan pd

      -- 4. Emit final files (with `Prettyprinter.Doc` body).
      files = [file [ps|{lsnFileName}.ipynb|] doc]
```

I tried really hard to show as many concrete stages/forms `LessonPlan` takes until it becomes a final software artifact.

I think this approach is relatively uncontroversial. It's a fairly minor refactor, but it does clean things up nicely I would say. It would be nice if we could see the `Generate`, `Transpile`, and `Emit` comments become real typeclasses and functions we use to further clean up said code, but I'm not quite sure what they would look like:

```haskell
class Transformer inp opts out | opts -> inp, opts -> out where
  transform :: opts -> inp -> out

class Generator inp opts out | opts -> inp, opts -> out where
  generate :: opts -> inp -> out

class Transpiler inp opts out | opts -> inp, opts -> out where
  transpile :: opts -> inp -> out
```

They all look the same to me right now. I'm not sure if there's any additional type information we can add to capture properties of the inputs/outputs.

And now, as I was writing this, @JacquesCarette passed by and we had a brief chat about this. Indeed, we do not yet have enough information. We can refine this code further, but I think we should do that in a subsequent PR + more analysis. We should also try to re-work the code generator for `drasil-srs` and `drasil-code` through this.

Two questions:

1. Should we rename `System` to `Project`? (I need to write more analysis before we discuss this.)
2. Should we rename `Render` to `Recipe` for now while we figure out terminology (transform/generator/transpile/etc.)?

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
```diff
@@ -1,30 +1,72 @@
...
+  lsnDesc :: LsnDesc,
+  -- FIXME: `titleComb` seems a bit odd. We only use `S.forT`. Two things:
+  --
+  -- 1. We generate `Notebook for Projectile Motion Lesson` as the title for the
+  -- projectile motion lesson plan. This should probably just be "Projectile
+  -- Motion Lesson".
+  -- 2. The title should probably just directly come from the system name.
+  --
+  -- i.e., we should remove this.
+  titleComb :: CI -> CI -> Sentence,
```
balacij: I could be completely wrong about this comment. It's really "future work" but I'm noting it here.

balacij:
Regarding the rename to `Project`, I will leave that as out of scope for this PR.

For now, I think this PR is ready for review.

code/drasil-example/projectile-lesson/lib/Drasil/Projectile/Lesson/Body.hs
JacquesCarette: I think we still want a chunk, because we want to be able to track provenance and output traceability information -- and we don't want multiple mechanisms for doing that.

Just because we currently under-use the name of a lesson doesn't mean we won't start using it as widely as we do for other artifacts.

code/drasil-gen/lib/Drasil/Generator/CaseStudyVariants.hs
```diff
@@ -80,12 +79,11 @@ caseStudyMainSRSWCodeZoo syst srsDecl srsFileName choices = do
   let layout = directory [ps|{exampleName}|] $ docLayouts ++ zooLayouts
   writeFiles OverwriteAllowed localPath layout
 
-caseStudyMainLsnPlan :: String -> LessonPlan -> LsnDesc -> String -> IO ()
-caseStudyMainLsnPlan exampleName plan nbDecl lsnFileName = do
+caseStudyMainLsnPlan :: LessonPlan -> LsnDesc -> String -> IO ()
+caseStudyMainLsnPlan plan nbDecl lsnFileName = do
```
JacquesCarette: I really don't like that rename from `exampleName` to `lsnFileName`. What we need here is a proper unit of information (based on the lesson name, that would let code downstream figure out a file name).

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
```diff
@@ -1,30 +1,72 @@
...
--- contents and organization) and a title combinator merging "notebook" with the
--- name of the 'LessonPlan'.
-render :: LessonPlan -> LsnDesc -> (CI -> CI -> Sentence) -> Document
-render plan dd comb = Notebook nm as $ mkSections (plan ^. systemdb) dd
-  where
-    nm = notebook `comb` (plan ^. sysName)
-    as = foldlList Comma List $ map (S . fullName) $ plan ^. authors
+-- | Options for the single-file-generating Jupyter notebook renderer.
+--
+-- FIXME: Rename to something more LessonPlan-focused?
```
JacquesCarette: Yes please.

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
```diff
@@ -1,30 +1,72 @@
...
+data Options = Options {
+  -- | Describe the organization of the final lesson plan.
+  lsnDesc :: LsnDesc,
+  -- FIXME: `titleComb` seems a bit odd. We only use `S.forT`. Two things:
```
JacquesCarette: I think two things are going on here:
1. the `titleComb` was maybe a little over-engineering (as we have a single example)
2. you are drawing too many conclusions from a single example.

I think it makes sense (but still is over-engineering) to be able to 'generate' a title from a collection of information.

code/drasil-system/lib/Drasil/System/Render.hs
```diff
@@ -0,0 +1,67 @@
...
+
+import Drasil.FileHandling (FileLayout, directory, ps, writeFiles, OverwritePolicy(..))
+import Language.Drasil (CommonIdea(abrv))
+
+import Drasil.System.Core (HasSystemMeta(..))
+
+-- | A "system" can be rendered into a set of (human-readable) software
+-- artifacts. An instance of `Render` is approximately what it means to define a
+-- off-the-shelf software generator.
+class HasSystemMeta sys => Render sys opts | opts -> sys where
```
JacquesCarette: why should it be that the options tell us what system we have? I am puzzled by the order of this fundep.

Suggestions for names:
1. `ToFiles`
2. `Concretize`

code/drasil-system/lib/Drasil/System/Render.hs
```diff
@@ -0,0 +1,67 @@
...
+    -- FIXME: Both `abrv` usage and `sysName` usage here is dubious. We need to
+    -- replace this field with something better, such as project name and
+    -- project shortname (repo name).
+    --
+    -- In some sense, I want to rename `System` to `Project`. Hence,
+    -- `SystemMeta` becomes `ProjectMeta`. This is nice because `SystemRepo`
+    -- then also becomes `ProjectRepo` (which is more commonly understood repo
+    -- and vague in a positive way).
+
+-- | 'renderSystemRepo', but it also writes the directory to disk (relative to a
```
JacquesCarette: This should not be in this package. I don't think packages outside of `drasil-gen` should do IO. Plus, I don't even think we want this function! We should see this composition of steps at the call site.

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
```diff
@@ -1,30 +1,72 @@
...
+  --
+  -- i.e., we should remove this.
+  titleComb :: CI -> CI -> Sentence,
+  -- | The name of the output file (no extension, @.ipynb@ is added later).
+  lsnFileName :: String
+
+  -- FIXME: Output formats? This can be rendered as a literate notebook as well.
+}
+
+instance Render LessonPlan Options where
```
JacquesCarette: This feels like premature generalization. Other than when doing weird Haskell trickery, having a single instance of a class is a "design smell".

JacquesCarette: Also: given that you have nicely extracted 4 steps, those should be 4 functions, which may, in time, become 4 classes if that makes sense.

JacquesCarette (changes requested):

code/drasil-gen/lib/Drasil/Generator/CaseStudyVariants.hs
balacij: Both existed before this. I only deleted the former.

`exampleName` was used to create the project `stable/` subfolder, and `lsnFileName` was the name of the specific `.ipynb` file.

code/drasil-system/lib/Drasil/System/Render.hs
balacij: I was trying to be helpful to the type-checker :smile: If the output type were another type variable, adding `opts -> out` would also be helpful.

If we create a set of options, it's more likely than not that we create it with a specific system type in mind and wouldn't use the same top-level options set for another system type. For example, latex rendering options set would not be used as the primary rendering options for `CodeSpec`, but if we were generating literate code, we might use it as part of the rendering options (nested).

This is a strong one-way association only as well. (We might want to instantiate many renderers (about different options sets) for the same system.)

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
balacij: > This feels like premature generalization. Other than when doing weird Haskell trickery, having a single instance of a class is a "design smell".

The goal is to also instantiate this for `CodeSpec`, `SmithEtAlSRS`, and `DrasilWebsite`.

> Also: given that you have nicely extracted 4 steps, those should be 4 functions, which may, in time, become 4 classes if that makes sense.

I'm not sure where we'll want to go with this yet, but towards the end of https://github.com/JacquesCarette/Drasil/pull/5201#issuecomment-4804170665 I hinted at creating more typeclasses as well, but I'm not so sure they would be 'meaningful enough' yet. I would want them to capture unique properties/use different type constraints/information somehow before I propose anything else.

code/drasil-system/lib/Drasil/System/Render.hs
balacij: (I think you meant to attach this to `renderSystemRepo'`? `renderSystemRepo` generates a `FileLayout`.)

Some of our previous `IO ()` functions spread around the codebase were doing 'bad' things (e.g., #4879), but this one is fairly tame. It merges two things that I imagine we'll be doing in each of our `main` functions: `writeFile . render`?

Admittedly, I want to move as many things _out_ of `drasil-gen` as possible. It knows too much about the secrets of our packages (`drasil-code` and `drasil-srs` to be specific).

(But this also isn't a hill I care to die on 😄)

code/drasil-system/lib/Drasil/System/Render.hs
balacij: `ToFiles` and `Concretizee` sounds better. I was trying to reserve `Concretize` for when the output type were also moved to a type variable.

Maybe we should take your suggestion from yesterday instead?

```haskell
class Transform in opts out | opts -> in, opts -> out where
  transform :: opts -> in -> out
```

code/drasil-lesson-plan/lib/Drasil/LessonPlan/Renderer.hs
balacij: > you are drawing too many conclusions from a single example.

😄 Agreed. I'm trying to extrapolate as much as possible, which is perhaps the wrong goal.

balacij: > I think it makes sense (but still is over-engineering) to be able to 'generate' a title from a collection of information.

Fair enough, I will clean up the documentation then.

code/drasil-system/lib/Drasil/System/Render.hs
balacij: `Recipe` also works well.

JacquesCarette: `Transform` = `Recipe` = program with options. So you would basically be lifting lots of coding from specific functions to going through instances. And would likely end up in Haskell hell whenever you try to be the least bit polymorphic with your options.

My feeling is that this is asking for trouble.

code/drasil-system/lib/Drasil/System/Render.hs
JacquesCarette: I want to centralize most `IO`. Right now, that's `drasil-gen`. I'm fine with `drasil-gen` being re-invented.

I want whatever package is responsible for doing "final orchestration" to be the one to call `writeFile`, with other packages being responsible for creating the 'data bundles' needed by that.

code/drasil-system/lib/Drasil/System/DrasilWebsite.hs
```diff
@@ -1,23 +1,36 @@
...
-data DrasilWebsite = DW {
-  _sm :: SystemMeta,
-  _indexDoc :: Document,
-  _webRefs :: M.Map UID Reference
-}
+data DrasilWebsite = DW
+  { _sm :: SystemMeta,
+    _indexDoc :: Document,
+    _webRefs :: M.Map UID Reference
+  }
```
balacij, line 33: In a subsequent PR, we can (should) move this to `drasil-website`.

code/drasil-system/lib/Drasil/System/DrasilWebsite.hs
```diff
@@ -28,3 +41,29 @@ mkDrasilWebsite :: SystemMeta -> Document -> [Reference] -> DrasilWebsite
 mkDrasilWebsite m doc rs = DW m doc refs
   where
     refs = M.fromList $ map (\r -> (r ^. uid, r)) rs
+
+-- | HTML generation options for the 'DrasilWebsite'.
+newtype DrasilWebsiteGenOptions = DWGO
+  { -- | What CSS should be loaded?
+    css :: Doc
+  }
```
balacij, line 49: Yay! Our first generation option for the website!

code/drasil-gen/lib/Drasil/Generator/CaseStudyVariants.hs
```diff
@@ -79,19 +70,3 @@ caseStudyMainSRSWCodeZoo syst srsDecl srsFileName choices = do
...
-      layout = directory [ps|{exampleName}|] [genJupyterLessonPlan plan nbDecl lsnFileName]
-  writeFiles OverwriteAllowed localPath layout
-
--- | The Drasil website binary is expected to build a `Website/HTML/` folder
--- containing the actual website artifacts (`index.html` and `index.css`).
-caseStudyMainDrasilWebsite :: DrasilWebsite -> IO ()
-caseStudyMainDrasilWebsite dw = do
-  setSystemLocale
-  writeFiles OverwriteAllowed localPath $
-    directory [ps|website|] $ genWebsite dw
```
balacij, line 97: I'm going to restrict this PR to only these two generators because I want to keep the line count down -- the SRS and code generators are a bit more complicated.

What do you think @JacquesCarette ? Should I just finish this off in this PR?

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
```diff
@@ -0,0 +1,76 @@
...
+data WriteOptions = WO
+  { -- | Where should we write to?
+    basePath :: OsPath,
+    -- | Are we allowed to overwrite files or not?
+    overwritePolicy :: OverwritePolicy,
+    -- | What is the name of the subfolder to be created?
+    localDirName :: forall sys. HasSystemMeta sys => sys -> PathSegment,
+    -- | What is the expected text encoding scheme?
+    textEncoding :: TextEncoding
+  }
```
balacij: Very generic options meant to be usable for any and all 'system's that support `ToFiles`.

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
```diff
@@ -0,0 +1,76 @@
...
+--
+-- 1. 'basePath': Always generate to local path.
+-- 2. 'overwritePolicy': Always allow file-overwriting.
+-- 3. 'dirName': The example's abbreviation.
+-- 4. 'textEncoding': UTF-8.
+drasilMakefileReqOpts :: WriteOptions
+drasilMakefileReqOpts = WO localPath OverwriteAllowed dirName utf8
+  where
+    dirName sys = let ab = abrv $ sys ^. systemMeta . sysName
+                   in [ps|{ab}|]
```
balacij, line 50: Drasil-system-level assumptions become crystal clear.

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
```diff
@@ -0,0 +1,76 @@
...
+
+-- FIXME: Both `abrv` usage and `sysName` usage above is dubious. We need to
+-- replace this field with something better, such as project name and project
+-- shortname (repo name).
```
balacij, line 54: Going back to earlier discussion about `ProgramName`, I will refrain from changing how things are currently done to keep this PR focused.

code/drasil-system/lib/Drasil/System/Core.hs
```diff
@@ -32,6 +32,8 @@ data SystemMeta = SystemMeta
 
 makeClassy ''SystemMeta
 
+-- FIXME: Rename `System` to `Project`?
```
balacij: With respect to `ToFiles`, a system is an abstraction over a concrete set of software artifacts. I think it's fair to call it an `AbstractProject` or something of the sorts.

Note that we don't even have `System` nor `SystemInformation` anymore. So the only thing we would be renaming is `SystemMeta` to `ProjectMeta[data]` (which contains repo-level metadata; very common in all software projects).

code/drasil-system/lib/Drasil/System/DrasilWebsite.hs
balacij: Since we lack a non-`Doc`-based system for stylizing our HTML, we just use `Doc`. Not a real priority at all, but, this is arguably a good usecase for `drasil-assets`.

code/drasil-system/lib/Drasil/System/Transformations.hs
```diff
@@ -0,0 +1,28 @@
...
+    -- | The generation options.
+    opts ->
+    -- | The final, rendered software artifacts.
+    [FileLayout]
+
+-- FIXME: Should we try to name our generators? Should we associate them with
+-- chunks so we can refer to them in code as well? Perhaps this will give us
+-- usage statistics, if we have a fine-grained enough `Render`-reliant scheme
+-- and were polymorphic about the output type. Going monadic here can help give
+-- us usage statistics on what things went on in the generation pipeline.
```
balacij, line 28: @JacquesCarette 

(I want to bring your attention to this comment after our latest discussion about chunks in `drasil-code`, this comment feels a bit more relevant.)

balacij: You can disregard the bit about being polymorphic about the output type in `toFiles`.

code/drasil-example/projectile-lesson/app/Main.hs
```diff
@@ -1,8 +1,12 @@
...
 
+import Drasil.LessonPlan (JupyterGenOptions(..))
 import Drasil.Projectile.Lesson.Body (si, nbDecl)
+import qualified Language.Drasil.Sentence.Combinators as S
 
 main :: IO ()
-main = caseStudyMainLsnPlan si nbDecl "Projectile_Lesson"
+main = concretizeAndWrite si opts drasilMakefileReqOpts
+  where
+    opts = JupyterGenOptions nbDecl S.forT "Projectile_Lesson"
```
balacij, line 12: Until we have reason to make smart constructors for `JupyterGenOptions`, I will leave the record fully exposed. I don't see a real reason to future proofing it until the entire project is fleshed out more.

balacij (commented):
Ready for another review, @JacquesCarette !

JacquesCarette:
And yet more conflicts have arisen.

JacquesCarette (commented):
This set of deltas is good. I don't want to hit 'Approve' quite yet, as it feels like github is not showing everything that I really want to see/review.

code/drasil-gen/lib/Drasil/Generator/CaseStudyVariants.hs
JacquesCarette: Fine to focus on just those two for now.

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
```diff
@@ -0,0 +1,76 @@
...
+setSystemLocale :: IO ()
+setSystemLocale = setLocaleEncoding utf8
+
+{-# DEPRECATED setSystemLocale
+  "Use `concretizeAndWrite` instead of directly setting system locale before file-writing." #-}
+
+-- | Configuration options for writing a repository of software artifacts.
+data WriteOptions = WO
+  { -- | Where should we write to?
+    basePath :: OsPath,
```
JacquesCarette: I feel like most of the 'outside' should never read/write `basePath`.

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
JacquesCarette: I agree that something like `ProgramName` is the (current best) solution, but to not worry about it now.

code/drasil-system/lib/Drasil/System/Core.hs
JacquesCarette: Let's figure out our full pipeline first, with all its concepts before we do any more big renames.

code/drasil-system/lib/Drasil/System/Transformations.hs
JacquesCarette: Being able to gather information while generating is a good idea. But we don't need the stages to be monadic to do that, we just need an 'orchestrator' that handles all the dispatch. *It* can be monadic, while called purely functional pieces.

JacquesCarette (changes requested):
Almost there!

code/drasil-system/lib/Drasil/System/Core.hs
balacij: Sounds good. Removed in cab9771524a3f741296b579ca6eb159411cc8c18.

code/drasil-gen/lib/Drasil/Generator/WriteSystem.hs
balacij: I can see what you mean.

Fixed in a460a83a03e225c3764481332e3933529b28dc28

balacij:
Ok, ready again for review, @JacquesCarette!

JacquesCarette (approved):

PR #5209, opened 2026-06-24
Fix the definition of a Newton

djarabek:
Doesn't seem to affect stable, since I assume we only ever print "N".


JacquesCarette (approved):
Nice catch. And really sad that this fix changes nothing in stable!

PR #5213, opened 2026-06-25
Don't allow UIDs containing colons

djarabek:
This is part of a rabbit hole that separating quantity and concept UIDs sent me down.

When `show`ing a UID, colons are used to separate the namespaces, as such if they are allowed in a non-namespaced UID, we can have the following subtle (and undesirable) behaviors: `mkUid "a:b" == nsUid "a" (mkUid "b")` is False, but `show (mkUid "a:b") == show (nsUid "a" (mkUid "b"))` is True. With these changes, evaluating the LHS results in an error in both cases.

### DIFF
--- a/code/drasil-theory/lib/Theory/Drasil/MultiDefn.hs
+++ b/code/drasil-theory/lib/Theory/Drasil/MultiDefn.hs
@@ -90,8 +90,7 @@ instance RequiresChecking (MultiDefn Expr) Expr Space where
 instance Express e => Express (MultiDefn e) where
   express q = equiv $ sy q : NE.toList (NE.map (express . (^. expr)) (q ^. rvs))
 
--- | Smart constructor for MultiDefns, does nothing special at the moment. First
--- argument is the 'String' to become a 'UID'.
+-- | Smart constructor for MultiDefns, does nothing special at the moment.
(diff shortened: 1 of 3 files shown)


code/drasil-theory/lib/Theory/Drasil/MultiDefn.hs
```diff
@@ -92,19 +92,19 @@ instance Express e => Express (MultiDefn e) where
...
-      "MultiDefn `" ++ u ++ "` created with non-unique list of expressions"
+      "MultiDefn `" ++ show u ++ "` created with non-unique list of expressions"
   where
     dupsRemovedLen = length $ NE.nub des
 
 -- Should showUID be used here?
 
 -- | Smart constructor for 'MultiDefn's defining 'UID's using that of the 'DefinedQuantityDict'.
 mkMultiDefnForQuant :: DefinedQuantityDict -> Sentence -> NE.NonEmpty (DefiningExpr e) -> MultiDefn e
-mkMultiDefnForQuant q = mkMultiDefn (showUID q) q
```
djarabek, line 107: This appears to be the only spot where there's an attempt to "round trip" a UID via String. This meant that using any namespaced UID here would cause issues.

JacquesCarette: good catch

JacquesCarette (approved):

PR #5215, opened 2026-06-25
Added `maybeDeref` operation to GOOL

### LINKED ISSUE #5111: GOOL: Handling Pointers in C++

B-rando1:
Closes #5111


JacquesCarette (approved):
Looks good - but conflict needs fixed.

B-rando1:
Conflicts resolved!

B-rando1:
I accidentally pushed changes that were meant for a different branch :facepalm: hence the force-push to undo.

JacquesCarette (approved):

PR #5219, opened 2026-06-26
refactor: avoid unnecessary `CodeVarChunk` roundtrip casting through `CodeChunk`

balacij:
Builds on #5217 

`quantvar` casts a `DefinedQuantityDict` to a `CodeVarChunk`. `codeIdea` casts things (including `CodeVarChunk`) to a `CodeChunk`. Previously this code:

1. Casted a DQD to a CVC.
2. Casted the CVC to a CC.
3. Casted the CC back into a CVC.

This PR removes the last two casts.


JacquesCarette (approved):
Nice!!!

PR #5226, opened 2026-06-28
Add lessonName field to LessonPlan, flatten lesson output folder (#5165)

### LINKED ISSUE #5165: Add `lessonName` to `LscDesc`
Much like the `progName` in `SmithEtAlSRS`, we need something we can use to derive an output folder name from.

> In a subsequent PR, we should create something like it (`lessonName` ?) for lessons. That should then be part of `LsnDesc`.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/5161#discussion_r3415853241_

We can also simplify the output by removing the redundant inner `Lesson` folder.

> We might as well remove the extra `Lesson` folder while we're at it. This will simplify our lives a bit later when we render the lesson plan to HTML and show it on the website.

_Originally posted by @balacij in https://github.com/JacquesCarette/Drasil/pull/5161#discussion_r3416205278_

Purav-Kanda:
Closes #5165

## Summary
Adds a `lessonName :: String` field to `LessonPlan` and uses it to derive the output folder for generated lesson artifacts, removing the need to pass the folder name as a separate argument and eliminating the redundant nested `Lesson/` subfolder.

## Changes
- **`Drasil.LessonPlan.Core`**: added `_lessonName :: String` field to the `LessonPlan` record, exposed via `makeLenses` as `lessonName`. `mkLessonPlan` now takes the name as its first argument.
- **`Drasil.Generator.CaseStudyVariants`**: `caseStudyMainLsnPlan` no longer takes a separate `exampleName` argument — it reads `plan ^. lessonName` instead. Output layout is flattened to `directory [exampleName] [genJupyterLessonPlan ...]`, removing the extra `Lesson` subdirectory.
- **`projectile-lesson` example** (`Body.hs`, `Main.hs`): updated to pass `"projectile-lesson"` into `mkLessonPlan` and updated the call site to match the new `caseStudyMainLsnPlan` signature.

## Why this design
`LsnDesc` is just a type alias (`[LsnChapter]`), not a record, so the field couldn't go there. Putting `lessonName` on `LessonPlan` mirrors how `SmithEtAlSRS` already carries `programName`. The literal folder name `"projectile-lesson"` is preserved (not auto-derived) since it's hardcoded in `code/Makefile` (`PROJECTILE_LESSON_DIR`) for the build/diff pipeline.

## Testing
- `stack build drasil-lesson-plan drasil-gen projectile-lesson` builds cleanly.
- Ran the `projectile-lesson` executable and confirmed `Projectile_Lesson.ipynb` is generated directly in `build/projectile-lesson/`, with no nested `Lesson/` subfolder.


JacquesCarette (approved):
Looks good to me - but since @balacij is actively changing projectile lesson, he ought to have a look before it gets merged.

Purav-Kanda:
Thanks for the review, @JacquesCarette! @balacij — happy to wait for your look before this gets merged. Let me know if there's anything in projectile-lesson you'd like adjusted to fit the direction you're taking it.

balacij (approved):
Nice!

This does solve the ticket. I'm not sure if a repo name is core to a lesson plan encoding, but we're moving things around in #5201.

Thank you for moving the lesson `.ipynb` out of the `Lesson/` folder as well!

PR #5227, opened 2026-06-28
Native vector

Xinlu-Y:
The changes covers the rendering layer only, teaching GOOL how to *write* MATLAB vector code. I need to start wire it into `drasil-code` so real models reach it. My next step is to split out the procedural vector functions in `drasil-code` so the procedural path can route them to the `NativeVector` operations (`VAdd → vecAdd`, `Dot → vecDot`, `Scale → vecScale`, `Norm → vecMag`, ...), while the OO path keeps using the generated vector class. It covers the full #4909 vector API on the native side.

Related: #5059


code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -630,6 +638,28 @@ mlArg n' = do
   s <- string
   mkVal s (text "varargin" <> braces (RC.value n))
 
+-- | Indexes into a vector: @v(i + 1)@. MATLAB is 1-indexed while GOOL is
+--   0-indexed, so we shift the index by one.
```
B-rando1: The comment about indexing seems to be from an earlier commit before you started using `intToIndex`.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -340,7 +340,7 @@ instance Array MatlabCode where
   arrayCopy = undefined
 
 instance List MatlabCode where
-  listSize = undefined
+  listSize v = funcApp "length" int [v]   -- length(v)
```
B-rando1: This is fine, but it might be better to match the Julia renderer by calling `CS.listSize`. It expands to exactly this, but I think it's slightly cleaner.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -403,6 +403,15 @@ instance Set JuliaCode where
   setRemove s e = funcApp "delete!" void [s, e]
   setUnion a b = funcApp "union!" void [a, b]
 
+-- TODO: implement native vector operations for Julia (currently MATLAB-only).
```
B-rando1, line 409: Is the plan to add Julia in a subsequent PR?

(that's fine with me, just double-checking your plan)

code/drasil-code/test/golden/VectorTestProc/matlab/VectorTest.m
```diff
@@ -0,0 +1,16 @@
+% VectorTest.m
+% Tests native vector operations.
+% - Authors: Drasil
+% - Note: Generated by Drasil v0.1-alpha
+
+function result = vectorOps(a, b)
```
B-rando1, line 6: This looks very idiomatic!

One question, have  you tried composing multiple operations together in one line? Like `a * b + c`? It might be good to have tests like that.

code/drasil-code/test/VectorTest.hs
```diff
@@ -0,0 +1,39 @@
...
+  [GProc.docMod "Tests native vector operations." watermark ["Drasil"] "" $
+    GProc.fileDoc (GProc.buildModule "VectorTest" [] [vectorOps])]
+
+-- | Takes two vectors and stores each vector operation's result, returning
+-- their dot product.
+vectorOps :: (ProcProg r) => SMethod r
+vectorOps =
+  function "vectorOps" public double [param (var "a" double), param (var "b" double)]
+  (bodyStatements
+    [ var "made" (listType double) &= litList double [litDouble 1.0, litDouble 2.0, litDouble 3.0]  -- [1.0, 2.0, 3.0]
```
B-rando1: The GOOL programmer shouldn't need to know that vectors are rendered as `List`s. We should have a separate `vecType` and `litVec`, which can point to `listType` and `litList` in most cases. This also frees us to use other types for vectors if needed.

(a quick note on `vectType`, `litVec` - they should probably go in `NativeVector`, not `TypeSym` and `Literal`, so that only languages with native vectors need to implement them)

B-rando1 (changes requested):
A few notes, but overall I really like the direction this is going!

JacquesCarette (commented):
I don't have comments beyond those of @B-rando1 .

code/drasil-code/test/VectorTest.hs
```diff
@@ -0,0 +1,42 @@
...
+  [GProc.docMod "Tests native vector operations." watermark ["Drasil"] "" $
+    GProc.fileDoc (GProc.buildModule "VectorTest" [] [vectorOps])]
+
+-- | Takes two vectors and stores each vector operation's result, returning
+-- their dot product.
+vectorOps :: (ProcProg r tp vis smt) => SMethod r
+vectorOps =
+  function "vectorOps" public double [param (var "a" double), param (var "b" double)]
+  (bodyStatements
+    [ var "made" (vecType double) &= litVec double [litDouble 1.0, litDouble 2.0, litDouble 3.0]  -- [1.0, 2.0, 3.0]
```
B-rando1: These should use `varDecDef` so that GOOL knows to add these variables to its variable map. The Julia renderer relies on information inserted there.

code/drasil-gool/lib/Drasil/GProc/InterfaceProc.hs
```diff
@@ -9,10 +9,10 @@ module Drasil.GProc.InterfaceProc (
   ) where
 
 import Drasil.Shared.InterfaceCommon (Label, SMethod, SharedProg,
-  MethodSym)
+  MethodSym, NativeVector)
 import Drasil.Shared.State (GS, FS)
 
-class (SharedProg r tp vis smt, ProgramSym r tp vis smt
+class (SharedProg r tp vis smt, ProgramSym r tp vis smt, NativeVector r tp
```
B-rando1: I'm not sure if `NativeVector` should be part of `ProcProg`. There are procedural languages that don't have native vector support (e.g. C) and probably OO languages that do, so I don't think it makes sense to tie vector implementation to language paradigm.

B-rando1 (changes requested):
Getting close, just a couple more things.

code/drasil-gool/lib/Drasil/GProc/InterfaceProc.hs
Xinlu-Y: Agreed it shouldn't be tied to paradigm. I removed `NativeVector` from `ProcProg`

B-rando1 (approved):
Looks good now 👍

JacquesCarette (approved):

PR #5228, opened 2026-06-29
DblPend: Fix UID reuse in GenDefs

djarabek:
This once again comes from a rabbit hole found while working on #5149. There will be a number of similar PRs fixing issues I've discovered in the examples while changing how chunks/UIDs work.

The key issue here is the reuse of the UID `force`, which results in incorrect definitions being used. Changing this UID results in an error, as the chunk is never inserted into the database. So, I moved the chunks to `Unitals.hs` (while also cleaning up the definitions to be in line with the rest of the quantities) and added them to the list of chunks that are inserted. This also required using unique symbols for the x and y components, which seems to be in line with the rest of the definitions.

However, the output still appears to be suspicious in a lot of ways, particularly the rendered formulas (the use of vectors in a formula that defines a vector component doesn't seem right). @smiths could you help determine what the expected output is here, and if this is in fact a step in the right direction?


code/drasil-example/dblpend/lib/Drasil/DblPend/GenDefs.hs
```diff
@@ -222,11 +222,8 @@ xForceGD_1 = gdNoRefs (equationalRealmU "xForce1" xForceMD_1)
         (getUnit force) (Just xForceDeriv_1) "xForce1" []
 
 xForceMD_1 :: MultiDefn ModelExpr
-xForceMD_1 = mkMultiDefnForQuant quan EmptyS defns
-    where quan  = quantAU (mkUid "force") (horizontalForce `onThe` firstObject)
```
JacquesCarette, line 226: Certainly the removal of these inline quantity definitions (with duplicate UID to boot) is the right thing to do.

code/stable/dblpend/SRS/Jupyter/DblPend_SRS.ipynb
```diff
@@ -1032,14 +1036,14 @@
...
-        "$$\\symbf{F}=m \\symbf{a}\\text{(}t\\text{)}=-{\\symbf{T}_{1}}\\,\\sin\\left({θ_{1}}\\right)+{\\symbf{T}_{2}}\\,\\sin\\left({θ_{2}}\\right)$$\n",
+        "$${F_{\\text{x}1}}=m \\symbf{a}\\text{(}t\\text{)}=-{\\symbf{T}_{1}}\\,\\sin\\left({θ_{1}}\\right)+{\\symbf{T}_{2}}\\,\\sin\\left({θ_{2}}\\right)$$\n",
         "</td>\n",
         "</tr>\n",
         "<tr>\n",
         "<th>Description</th>\n",
         "<td>\n",
         "\n",
-        "<li>$F$ is the force ($N$)</li>\n",
+        "<li>$F_x1$ is the horizontal force on the first object ($N$)</li>\n",
```
JacquesCarette, line 1046: I'm pretty sure this should be $F_{x1}$ as otherwise this is going to display in funny ways.

code/stable/dblpend/SRS/Jupyter/DblPend_SRS.ipynb
```diff
@@ -1066,7 +1070,7 @@
         "\n",
         "#### Detailed derivation of force on the first object:\n",
         "<a id=\"GD:xForce1Deriv\"></a>\n",
-        "$$\\symbf{F}=m \\symbf{a}\\text{(}t\\text{)}=-{\\symbf{T}_{1}}\\,\\sin\\left({θ_{1}}\\right)+{\\symbf{T}_{2}}\\,\\sin\\left({θ_{2}}\\right)$$\n",
+        "$${F_{\\text{x}1}}=m \\symbf{a}\\text{(}t\\text{)}=-{\\symbf{T}_{1}}\\,\\sin\\left({θ_{1}}\\right)+{\\symbf{T}_{2}}\\,\\sin\\left({θ_{2}}\\right)$$\n",
```
JacquesCarette, line 1073: `F_{\\text{x}1}` is quite weird printing!

JacquesCarette (approved):
All my comments are outside the scope of this PR, in that all the weirdnesses were already there and not new here. 

I see these changes as a strict improvement over what was there before. There is still work to be done, but I don't think it should be done in this PR. I would be happy to merge as-is.

djarabek:
I opened #5232 for the incorrect definitions. Seeing as you agree that this is an improvement, you can merge this.

JacquesCarette:
As we discussed in the meeting: this moves the definitions from "complete nonsense" to "somewhat nonsense" (and issue file to get to not-nonsense). The 'right' solution involves vectors...

PR #5231, opened 2026-06-29
Fixed C++ Class Members Having Reference Types by Default

### LINKED ISSUE #5216: GOOL: C++ Class Variables Need To Not Be References by Default
With #5215, #4909 is one `&` away from generating functioning C++ code.

The issue is that the C++ renderer automatically assumes that all of a class's variables of type `List` or `Set` should be references. E.g. the header file for `Populate` in `dblpend` (look at the bottom where it declares `theta`):
https://github.com/JacquesCarette/Drasil/blob/98704433ce56657d8169c4c3e3f5876244cd9bb8/code/stable/dblpend/src/cpp/Populate.hpp#L16-L31

The reason is that the only times up til now that we've generated classes with `List` members was for the Boost ODE library. This library requires us to give it an object with a reference to an array for it to store the result in, because it doesn't have the option to return the result separately. This gives us a problem: for Boost we need to have a class with either a pointer or a reference to a list in them, and for OOVector we need to have a class with a list (not a reference or pointer; that will quickly break logic) in it.

The obvious solution is to utilize the new `Reference` type that GOOL has. When defining a class in C++, for each of its variables we check the type, and insert `*` or `&` if it's a `Reference` type or nothing if it's not. The question is, `*` or `&`?

The case for `&` is that we can implement it faster. The translation of the previous paragraph, and propagating it into the Boost part of `drasil-code`, are the only changes we'd need to make. The reason I don't want to do this is that GOOL `Reference`s map to C++ pointers, not C++ references. We could make this work as a hack for the present, but it would come back to bite us later.

Using `*` is obviously better in the "conceptual" sense, since we already have `Reference` mapping to C++ pointers. The trouble with this is that the syntax for working with pointers is slightly different than with references. Some of this is automatically handled by GOOL, but one case where it isn't is the case of converting a value to a pointer using `&`. To do this, we'll have to teach GOOL a new concept, `toReference`, that converts a "value" to a "reference to a value".

To summarize, I think we should make the C++ renderer make variables values by default and pointers if they have type `Reference`. To make the ODE library work with this, we'll need to add a `toReference` operation to GOOL that will translate to `&` in C++. This will give us the flexibility we need to work with both Boost and OOVector.

B-rando1:
Closes #5216 

All the generated C++ code compiles (`make gool`), and the `output.txt` generated by the C++ Double Pendulum example is identical to what it previously was.

Feel free to critique my `drasil-code` and `drasil-lang` code @balacij @djarabek. After this, I would love to stay away from those packages for a while and stay in `drasil-gool` where it's safe and things make sense :sweat_smile:

### DIFF
--- a/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
+++ b/code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
@@ -573,7 +573,7 @@ odeintCall info = externalLibCall [
       stepSize $ odeOpts info] ++ [
     customObjArgFill [privStateVar $ referenceToCodeVarChunk $ solListVar info]
       (customClassFill [constructorInfoFill [unnamedParamFill $ solListVar info]
-         [(referenceToCodeVarChunk $ solListVar info, sy $ solListVar info)] [],
+         [(referenceToCodeVarChunk $ solListVar info, UnaryOp MakeRef (sy $ solListVar info))] [],
          methodInfoFill [] [appendCurrSolFill $ referenceToCodeVarChunk $ solListVar info]])]]
   where chooseMethod RK45 = (0, map (callStepFill . libCallFill . map
           basicArgFill) [[], [absTol $ odeOpts info, relTol $ odeOpts info]])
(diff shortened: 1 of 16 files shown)


JacquesCarette (approved):
When Drasil works, it's great. So few changes, and certainly nothing to do with the "design log", and yet, here we are.

code/drasil-code/lib/Data/Drasil/ExternalLibraries/ODELibraries.hs
```diff
@@ -698,6 +698,16 @@ solListVar info = quantvar $ implVarAU' (dv +++ "sol")
   (Vect (dv ^. typ)) (symbol dv Implementation) (getUnit dv)
   where dv = depVar info
 
+-- | Given a codeVarChunk, changes its Space to be a Reference to the original Space
+referenceToCodeVarChunk :: CodeVarChunk -> CodeVarChunk
```
djarabek: I really don't like this, what could work instead is making `solListVar` take an `isRef` boolean.

B-rando1: Fixed now. I agree, it is much simpler.

JacquesCarette (approved):

PR #5234, opened 2026-06-30
new(data-formats): Introduce HTML AST and renderer

camishibata:
Contributes to #4989

PR #5242, opened 2026-07-04
Update authors lists and active team members

balacij:
Builds on (and targets) #5234


JacquesCarette (commented):
Should onBoarding.pdf be in the repo?

balacij:
I have no issue with small PDFs in the repo, I commit them in my own repos all the time. IIRC, @smiths preferred this as well (at least in this case!) because it simplifies work on our end when sending the PDF to students.

If your concern is repo size and preferring to limit how many PDFs are committed, we can restrict ourselves to only committing this PDF going forward as well. We would still have the old PDFs in our git commit history. No way of truly getting rid of that without rewriting the git commit history entirely (which I would not be opposed to, especially because there are tools that can help us do this).

(If we choose to rewrite our commit history, we should also get rid of the past committed `.exe`s.)

JacquesCarette:
Some PDFs are ok, as long as they are not (re)generated very often.

JacquesCarette (approved):

smiths:
With respect to pdfs under version control, what I've started doing in my template capstone repo and in my other programming projects is to use a [GitHub action](https://github.com/smiths/turbulent-flow/blob/main/.github/workflows/latex-pages.yml) to build the pdfs and deploy them to a [GitHub page](https://smiths.github.io/turbulent-flow/).  The nice thing is that the pdfs don't take up room in the commits, but the most recent version is always available.

We could put the onBoarding document on our [Drasil Github page](https://jacquescarette.github.io/Drasil/).  I don't think it is worth the effort right now, but it is an idea to keep in mind, especially if we find there are other pdfs that we are keeping under version control that we would rather not, but we still want to be able to point to the generated pdf.

If we ever generate GitHub repos for our case studies (instead of, or in addition to, keeping them in the main Drasil repo), we should consider using a template repo that has a generic GitHub action for building the pdfs and deploying them to a Github page.

PR #5247, opened 2026-07-04
refactor: move common chunk atom `SingleLine` renderers to `drasil-printers` and use them more

balacij:
Builds on #5246

PR #5248, opened 2026-07-04
refactor: remove exported `.Import` module in `drasil-printers`

balacij:
Builds on #5247

After this PR, the ones it depends on and #5201, another shallow file analysis of `drasil-printers` and its uses is in order. This will be helpful for both the `drasil-code` work (so we can finally move the `Drasil.Code` packages from `drasil-lang` to `drasil-code`) and #4989.


JacquesCarette (dismissed):
Looks good, but I won't merge while the build is broken.

JacquesCarette (approved):

Issue #5233, opened 2026-06-30
Removing Type Families from GOOL


B-rando1:
We've been talking about and working on this for a while, but other than #4360 (apparently Discussions don't display links to them?) we don't have a lot on here. I'm creating this so that we have something we can point to, and have further discussions.

Completed work:
- #4973 
- #4998 
- #5102 
- #5225 

I've removed all the type families that always resolved to the same type (and `Type`, which was `()` in the `CodeInfoX` renderers). The others will be more difficult, as they'll need to be replaced with functional dependencies.

I had one idea that might make switching from type families to fundeps easier. I believe that all of the type family concrete types are subtypes of each other, up to isomorphism. E.g. `Type` was `TypeData` for all the renderers and `()` for `CodeInfoX`. Or `StateVar` is `Doc` in most renderers but `StateVarData` (which contains a `Doc` and some other info) in C++. What if we temporarily specialized all of the types to the "most informative" specialization, which would allow us to remove the type family? Then we could add fundeps after that to remove the extra information from renderers that don't need it. Would that be easier?

B-rando1:
Reed said my idea was a bad idea 😄 

I think I'll start by finishing work I started on Friday to make `TypeSym` have a fundep (so that `CodeInfoX` can use `()` again). Then I'll start the larger refactor of converting our type families to fundeps.

B-rando1:
As discussed in our meeting, I'm looking into `CodeInfoX` to see if we can rework it, because most of our type families resolve to the same type everywhere except `CodeIntoX`.

Starting with an "Observational" analysis, I see the following changes when I remove `CodeInfoOO` from our pipeline:
- In Java, functions stop denoting the exceptions that they throw. Iirc `CodeInfoOO` does some work of building up a function call map, and Java must use this to figure out which exceptions a function might throw.
  - Java also automatically adds import statements for the exceptions that are referenced.
  - Swift also loses `throws` declarations on it function declarations, as well as `try` statements at call sites of those functions.
- C++ and Python lose (some) import statements for modules in the same package.

The import statements part is surprising to me. Not all of them go away, so clearly we have a mechanism for adding imports while rendering a module, but it's only used some of the time. I'll need to look into this some more.

The exception tracking was mainly what I was worried about. I don't believe we have a mechanism for updating a function declaration while generating its body, the way we do for adding imports while generating a module. It might be feasible to add one, though.

B-rando1:
[...]

PR #5256, opened 2026-07-06
Parameter fundep

B-rando1:
Contributes to #5233 

The type signatures are getting very long with all the typeclass parameters, so I started using a new formatting scheme that Reed suggested for them. The diff might be harder to parse because of that, especially here on GitHub since it doesn't ignore whitespace. Let me know if there's anything I can do to make it easier to read.


JacquesCarette:
There's a conflict that needs resolved.

JacquesCarette (approved):
Actually, `diff` was contextual enough that it was still clear what had been changed. Also, I'm not so concerned with long signatures -- things will continue to gyrate for a while as we refactor, and then we'll be able to see new patterns, from where we'll be able to shrink things back down.

B-rando1:
Conflicts resolved!

B-rando1:
`Parameter` is instantiated as `ParamData` in all our renderers, so I'll leave this as draft while I look into `CodeInfoX`.

B-rando1:
Closing in favour of #5274

PR #5257, opened 2026-07-07
fix(html-gen): malformed `a.href` attributes for local links

balacij:
All of our `href` attributes for local links did not wrap the values in parentheses.

PR #5245, opened 2026-07-04
docs: old shallow file analysis of `drasil-utils`

balacij:
Builds on #5237 

Posting the old shallow file analysis of `drasil-utils`. The code has since improved.

PR #5258, opened 2026-07-07
hacks: remove `repUnd` hack that previously fixed bad labels

balacij:
Builds on #5257 

Contributes to #5245 

Story time:

1. `repUnd` was a hack introduced in 3ceab2950 to deal with 'bad reference names' (used as both document-internal labels and displayed labels) that could not be readily used in TeX text.
2. It was removed in 3e7976396.
3. It was then reintroduced in 95a19f87e for the same reason it was initially introduced; the labels were not valid TeX.
4. In 857b58711, a more robust fix for dealing with generating 'bad' TeX text appeared: escaping characters. Thank you @samm82!

So, as of 857b58711, I believe we can remove `repUnd`.


code/stable/swhs/SRS/PDF/SWHS_SRS.tex
```diff
@@ -794,9 +794,9 @@ \subsubsection{Data Definitions}
 \noindent
 \begin{minipage}{\textwidth}
 \begin{tabular}{>{\raggedright}p{0.13\textwidth}>{\raggedright\arraybackslash}p{0.82\textwidth}}
-\toprule \textbf{Refname} & \textbf{DD:waterVolume.pcm}
+\toprule \textbf{Refname} & \textbf{DD:waterVolume\_pcm}
```
JacquesCarette, line 797: this won't render properly -- I believe you'll get just the p as a subscript, but it should be all of `pcm`.

JacquesCarette (changes requested):
Unfortunately, this is going to have not-so-good changes to the actually rendered LaTeX.

JacquesCarette:
Plus this somehow got the build to fail!?!

code/stable/swhs/SRS/PDF/SWHS_SRS.tex
balacij: Image

Renders as above, which aligns with how all of the previously-written `DD:waterVolume\_pcm` text was written elsewhere.

Our code related to TeX rendering could use some attention. The `lspec` solution works, but we should have more types to denote `\label{..}`s vs plain text. I had to propagate `lspec` usage a bit more. `\label{..}` bodies can carry unescaped underscores, but plain text needs them to be escaped.

balacij: Ah well, this image relates to the DD for SWHSNoPCM.

One thing with the current way the labels are: we show both `DD:waterVolume_nopcm` and `DD:waterVolume.nopcm` shown in the SRS. This at least lets us just use the former (which, from commit history, is what was intended).

balacij:
> Plus this somehow got the build to fail!?!

It looks like our `make tex` target was broken for Linux/Windows... It was looking for `.blg` and `.log` files named `projectile_SRS.X`, but only `Projectile_SRS.X` existed in the TeX artifact directories (applies to all case studies). For macOS, where the file system is case insensitive, this hasn't given me issue since I switched to mac!

balacij:
Ok, now this PR does 3 things:

1. Fixes the `make tex` target (and the CI by extension).
2. Propagates `lspec` usage (i.e., the function for rendering labels correctly in TeX files).
3. Removes `repUnd`.

code/stable/swhs/SRS/PDF/SWHS_SRS.tex
JacquesCarette: A little ugly, but acceptable.

JacquesCarette (approved):

PR #5260, opened 2026-07-07
Add parentheses around sum/product/integral body in printer

### LINKED ISSUE #1080: Missing parentheses for sum quantifier in HTML
Drasil does not seem to be auto-generating parentheses where they are required in a sum quantifier in HTML. An example in SSP (Equation for IM1) is shown below, where there should be parentheses to show that the "_R<sub>n</sub>_" and "_T<sub>n</sub>_" are not included in the summation:

image

Looking at #273, this was working in the past but must have broken at some point since then.

Purav-Kanda:
Fixes #1080.

### Problem
`eopAdds` and `eopMuls` in `Printing/Import/ModelExpr.hs` and
`Printing/Import/Expr.hs` rendered the operator body with no
precedence check. In TeX/MathJax, `\sum R_i \prod \Psi_v + R_n`
parses as `(Σ R_i)(Π Ψ_v) + R_n` rather than `Σ(R_i Π Ψ_v) + R_n`.

### Fix
Replace bare `P.Row [modelExpr e sm]` / `P.Row [expr e sm]` with
`modelExpr' sm (precA Mul + 1) e` / `expr' sm (precA Mul + 1) e`.
Threshold `precA Mul + 1 = 191` parenthesizes Add (180), Subt (180),
and Mul (190) bodies while leaving Pow (200) and atoms (500) bare.

### Verification
SSP IM:fctSfty now renders correctly:

Output:
image

### Known limitation
Nested sums render as `Σ_i (Σ_j f(i,j))` — parens around the inner
sum are unnecessary but mathematically correct.


code/drasil-printers/lib/Language/Drasil/Printing/Import/Expr.hs
```diff
@@ -79,23 +79,23 @@ call sm f ps = P.Row [
 eopAdds :: PrintingInformation -> DomainDesc t Expr Expr -> Expr -> P.Expr
 eopAdds sm (BoundedDD v Continuous l h) e =
   P.Row [P.MO P.Inte, P.Sub (expr l sm), P.Sup (expr h sm),
-         P.Row [expr e sm], P.Spc P.Thin, P.Ident "d", symbol v]
+         P.Row [expr' sm (precA Mul + 1) e], P.Spc P.Thin, P.Ident "d", symbol v]
```
JacquesCarette: Why wrt `Mul` and not `Add` since integration is a genralized kind of summation? And why `+ 1` ?

Purav-Kanda: Good catch. I was thinking "parenthesize anything non-atomic" but that's too aggressive — ∑ aᵢ·bᵢ is already standard notation and doesn't need extra parens. The parens are only necessary when the body has addition/subtraction, since ∑ a+b would be read as (∑a)+b. So precA Add + 1 is the right cutoff — the +1 is because expr' uses strict <, so we need to go one above precA Add to actually catch Add-precedence bodies. Will update both files and re-stabilize.

JacquesCarette (approved):

JacquesCarette:
Can you double-check that this still fixes the original problem? I looked at the changes in SSP, and it was not at all clear to me that this changed the right expression!

Purav-Kanda:
Here's the current rendering of IM:fctSfty — without explicit parens around the Mul body, the sum visually appears to only cover Rᵢ, making it ambiguous with (∑Rᵢ)(∏Ψᵥ) + Rₙ. This suggests the threshold may need to include Mul-precedence bodies after all. Would precA Mul + 1 be the right cutoff here, or is there a different approach you'd prefer?
image

smiths:
I agree that the brackets make the equation easier to read.  They also match how the equation was shown in the [original source](https://hub.hku.hk/bitstream/10722/42061/1/102322.pdf) (see Equation 6).

Purav-Kanda:
Thanks @smiths. Based on the screenshots showing ambiguous rendering and the match with the original source, I'll revert to precA Mul + 1. @JacquesCarette — the visual output with precA Add + 1 leaves Mul-body sums like ∑ Rᵢ·∏Ψᵥ + Rₙ ambiguous in practice, even if mathematically conventional. The explicit brackets seem necessary here. Awaiting your thoughts before making the change.

Purav-Kanda:
Just checking in — happy to proceed with reverting to precA Mul + 1 if that looks right to you, or merge as-is with Add + 1 if you'd prefer. Let me know how you'd like to move forward.

balacij:
Hi @Purav-Kanda Do you think you could give a few different examples of how this would change things? What were expressions like before and after (using pictures of LaTeX or providing the actual LaTeX)?

Adding (+ 1) is also a bit unconventional. What do you think about using something along the lines of `precBigOp BigAdd` and `precBigOp BigMul`? This way the precedence numbers are a bit easier to understand/readily available.

Purav-Kanda:
Here are some examples showing the difference:

∑ (aᵢ + bᵢ) — both thresholds fix this: renders as \sum \left(a_i + b_i\right) ✓
∑ aᵢ · bᵢ — with Add + 1 (181) no parens added since Mul body (prec 190) is above threshold; with Mul + 1 (191) it renders as \sum \left(a_i \cdot b_i\right) ✓

SSP's IM:fctSfty has a Mul-body sum that stays ambiguous under Add + 1, which is why I think Mul + 1 is needed — it also matches the source paper @smiths linked.
On precBigOp — that makes sense and is cleaner. Would it be something like precBigOp Add = precA Add + 1 for eopAdds and precBigOp Mul = precA Mul + 1 for eopMuls? Just want to confirm before I add it.

JacquesCarette:
I have not forgotten about this - I just need a decent chunk of time to make sure all the details line up properly.

PR #5264, opened 2026-07-07
Removed `CodeInfoProc`

B-rando1:
Well that was easy :smile: 

I had the thought in the back of my mind, that I may have created `CodeInfoProc` because that's how GOOL did it and I didn't feel like figuring out how to _not_ do it the same way. Turns out that it's super easy to remove, and as I guessed, it doesn't affect our procedural renderers at all.


B-rando1:
Actually I may have spoken too soon. Removing `CodeInfoOO` causes changes in stable for not just Java, but for all our renderers. Most likely we just don't have test cases for Julia of the features that it helps with.

JacquesCarette:
Need to figure out if the changes to stable are good/bad.

B-rando1:
After taking a longer look at `CodeInfoProc`, I'm more confident that it's only used for the class map and function exception map.  Julia and MATLAB don't currently use either of these pieces of information, so I think we're good to remove `CodeInfoProc`.

Now, there is an argument for keeping it: once we add structs, we'll need something like the clas map; and if we ever add a procedural language with explicit `throws` declarations, then we'll need a function exception map. But we don't currently have plans for adding a language with explicit `throws` declarations, and as far as I know structs are still a few months away at least (@Xinlu-Y correct me if I'm wrong). It probably makes sense to figure out what `CodeInfoOO` looks like post-refactor and copy that to `GProc` as needed.

JacquesCarette:
Since the features will indeed be easier to add post-refactor, it is better to remove `CodeInfoProc` (as it is a source of inflexibility that we don't need).

JacquesCarette (approved):

PR #5265, opened 2026-07-07
Remove `toPlainName`

balacij:
The usecase for this function can be intuited at each of its usage sites. All are about creating (locally valid) identifiers. However, as discussed in the comments of `toPlainName`, it is not exhaustive, and as discussed in <https://github.com/JacquesCarette/Drasil/pull/5245#discussion_r3525155397>, it is confusing.

We know that `toPlainName` is helpful for `drasil-code`. However, each language has its own permissible-identifier scheme that we should use/permit.

Currently, `toPlainName` is not currently (effectively) used. If/when these possible issues arise, we should rebuild `toPlainName` correctly.


JacquesCarette (approved):
So I guess the upshot is that now it is easier to mistakenly name a function something that won't compile... a bit of a regression, but not so bad.

Issue #3181, opened 2022-12-03
Github Actions failing on ubuntu-latest


JacquesCarette:
See [this build failure](https://github.com/JacquesCarette/Drasil/actions/runs/3608923700/jobs/6081806688) for an example. Seems to be related to lualatex and fonts? Might be new, even though this bug seemed to appear in the past (on the lualatex side, not ours).

Sam T.-H. on twitter heartily recommends using [ACT](https://github.com/nektos/act) for locally debugging such things.

smiths:
ACT does look like a good option to run GitHub Actions locally.  The README for ACT doesn't mention another advantage that occurs to me, running the Actions locally doesn't use up the free minutes.  When we last spoke about the free minutes, there wasn't much worry that we were going to exceed them, but it is still nice to have a local option, in case there is a concern in the future.

balacij:
I'm very surprised these errors are popping up now, only after GH updated their image to 22.04. I'm using NixOS with unstable packages (most packages are newer than Ubuntu's), and I have never seen any noticeable errors like these.

I don't think the lualatex fonts thing is an "error," but a warning message. I have it appear on my machine as well. However, the bit about "$TERM" is odd. I think you fixed that now though.

ACT looks like a great tool for debugging the CI!

balacij:
Relevant: https://github.com/JacquesCarette/Drasil/pull/5271#discussion_r3545456430

There's not much we can do to debug the errors back in 2022 and it looks like we're at the mercy of GitHub and its provided package sets here. We haven't had issue with tex other than build times, but that's being worked on in #5271 . I think we can close this issue because there's not much we can really do here other than wait for the next issue to pop up and try to deal with it.

PR #5271, opened 2026-07-08
Rewrite PR CI Workflow

djarabek:
The workflow is now split up to run several steps in parallel, as well as being faster my minimizing the amount of dependencies that need to be installed for each step.

In future PRs, I will improve the other workflows (mainly the Deploy workflow) in a similar manner to this one.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -13,69 +13,43 @@ defaults:
 jobs:
   build:
     name: "Build & Test"
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
+    outputs:
+      tex: ${{ steps.changes.outputs.tex}}
+      md: ${{ steps.changes.outputs.md }}
+      gencode: ${{ steps.changes.outputs.gencode }}
     steps:
       - uses: actions/checkout@v7
 
-      - name: "Search changes for applicable tests"
+      - name: "Detect changes"
         uses: dorny/paths-filter@v4
         id: changes
         with:
           filters: |
+            # This section is referenced in all the below sections
+            # Paths added here are core/build files that should trigger all optional steps
+            build: &build
+              - '.github/workflows/Build.yaml'
+              - 'code/Makefile'
+              - 'code/scripts/**'
             tex:
+              - *build
               - 'code/stable/**/*.tex'
             md:
+              - *build
               - code/stable/**/*.md
               - code/stable/**/book.toml
-            artifacts:
+            gencode:
+              - *build
               - 'code/stable/**/src/**'
-
-      - name: "Update apt package list"
-        run: |
-          sudo apt-get update
-
-      - name: "Install core system requirements"
-        run: |
-          sudo apt-get install -y --fix-missing libgmp-dev python3 graphviz g++ default-jdk mono-devel
-
-      - name: "Install TeX requirements"
-        if: ${{ steps.changes.outputs.tex == 'true' }}
-        run: |
-          sudo apt-get install -y --fix-missing fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-luatex inkscape
-
-      - name: "Install mdBook"
-        if: ${{ steps.changes.outputs.md == 'true' }}
-        run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          cargo install mdbook
-
-      - name: "Update PATH"
-        run: |
-          echo "$HOME/.local/bin" >> $GITHUB_PATH
-          echo "$HOME/.swift/usr/bin" >> $GITHUB_PATH
-          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
-
-      - name: "Install Stack"
-        uses: haskell-actions/setup@v2
-        with:
-          enable-stack: true
-          stack-no-global: true
-          stack-version: 'latest'
+              - 'code/drasil-code/test/golden/**'
 
       - name: "Cache dependencies"
         uses: actions/cache@v6
         with:
           path: |
             ~/.stack
-            ~/.cabal/packages
-            ~/.cabal/store
-            code/.stack-work/
-            code/website/.stack-work/
-            ~/.local/bin/graphmod
-          key: ${{ runner.os }}-store-${{ hashFiles('code/stack.yaml') }}
-
-      - name: "Clean previous run"
-        run: make clean
+          key: ${{ runner.os }}-stack-${{ hashFiles('code/stack.yaml') }}
 
       - name: "Install dependencies"
         run: make stackArgs="--no-terminal" deps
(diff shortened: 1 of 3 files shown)


code/Makefile
```diff
@@ -325,17 +328,21 @@ $(filter %$(BUILD_P_SUFFIX), $(BUILD_PACKAGES)): %$(BUILD_P_SUFFIX): check_stack
 # Specific example targets will have the name exampleName_gen.
 # Alternatively, you can run an example by running `stack exec exampleName`.
 $(filter %$(GEN_E_SUFFIX), $(GEN_EXAMPLES)): %$(GEN_E_SUFFIX):
+ifneq ($(SKIP_GEN), 1)
```
djarabek, line 331: I'm not super thrilled about this, but it's the best I could come up with.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+      - name: "Compile GOOL Examples"
+        run: make codegenTest SKIP_GEN=1
+      
+      - name: "Compile generated software artifacts"
+        run: make gool SKIP_GEN=1
+
+  latex:
+    name: LaTeX
+    needs: build
+    runs-on: ubuntu-26.04 # Ubuntu 26 has newer TeX packages that are significantly faster
```
djarabek, line 144: TexLive 2025 is significantly faster. Ubuntu 26 runners are still in beta, but hopefully no issues. There are other options (such as using TexLive in a docker image) that I'm still exploring.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+          sudo apt-get install -y graphviz
 
-      - name: "Test Built Website against Stable Version"
-        run: make test_website NOISY=yes
+      - name: "Install mdBook"
+        env:
+          MDBOOK_VERSION: 0.5.4
+        working-directory: ${{ runner.temp }}
+        run: |
+          curl -L -O https://github.com/rust-lang/mdBook/releases/download/v$MDBOOK_VERSION/mdbook-v$MDBOOK_VERSION-x86_64-unknown-linux-gnu.tar.gz
```
djarabek, line 191: I looked into using https://github.com/peaceiris/actions-mdbook, but it currently causes deprecation warnings, indicating it isn't very actively maintained.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
         uses: dorny/paths-filter@v4
         id: changes
         with:
           filters: |
+            # This section is referenced in all the below sections
+            # Paths added here are core/build files that should trigger all optional steps
+            build: &build
+              - '.github/workflows/Build.yaml'
+              - 'code/Makefile'
+              - 'code/scripts/**'
```
djarabek, line 35: There might be some other things I forgot that would be good to include here.

code/Makefile
balacij: Until the Makefile is properly audited and restructured to accommodate skipping `_gen` targets, this is perfectly reasonable.

code/Makefile
```diff
@@ -347,7 +354,7 @@ $(filter %$(INSTALL_E_SUFFIX), $(INSTALL_EXAMPLES)): %$(INSTALL_E_SUFFIX):
 # Same as above targets; run individual examples but create traceability graphs
 # for each example run. Needs Graphviz to run. Graphs generate in svg, pdf, and png formats.
 # Specific example targets will have the name exampleName_trace_graph.
-$(filter %$(TRACE_GRAPH_SUFFIX), $(TRACE_GRAPH_EXAMPLES)): %$(TRACE_GRAPH_SUFFIX): %$(GEN_E_SUFFIX) graphmod
+$(filter %$(TRACE_GRAPH_SUFFIX), $(TRACE_GRAPH_EXAMPLES)): %$(TRACE_GRAPH_SUFFIX): %$(GEN_E_SUFFIX)
```
balacij, line 357: `graphmod` isn't required?

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+        run: make test test_website NOISY=yes
 
-      - name: "Compile GOOL examples"
-        run: make codegenTest
+      - name: "Upload logs on failure"
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: diff-logs
+          path: code/logs
```
balacij, line 74: This will be very nice to have!

code/Makefile
djarabek: Nope, the graphmod target just installs the `graphmod` command, which we clearly don't use here.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
-        run: make website
+      - name: "Install requirements"
+        run: |
+          sudo apt-get install -y graphviz
 
-      - name: "Test Built Website against Stable Version"
-        run: make test_website NOISY=yes
+      - name: "Install mdBook"
+        env:
+          MDBOOK_VERSION: 0.5.4
```
balacij, line 188: We should probably document this version number somewhere. In fact, we should be documenting compiler/interpreter version numbers tested against for every language we generate.

Maybe for now we just note this as a comment near the mdBook generator code, and later (in a subsequent PR) we add this as well to the wiki?

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
-            ~/.cabal/packages
-            ~/.cabal/store
-            code/.stack-work/
-            code/website/.stack-work/
-            ~/.local/bin/graphmod
-          key: ${{ runner.os }}-store-${{ hashFiles('code/stack.yaml') }}
-
-      - name: "Clean previous run"
-        run: make clean
+          key: ${{ runner.os }}-stack-${{ hashFiles('code/stack.yaml') }}
```
djarabek, line 55: I did realize one problem, because I changed this it's now out of sync with the main CI (Deploy.yaml), so PRs won't get caches until after their first commit. We might want to wait until the changes to Deploy.yaml are ready to merge this.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+        uses: actions/download-artifact@v8
+        with:
+          name: generated-gooltest
+          path: code/drasil-code/test/build
+
+      - name: "Compile GOOL Examples"
+        run: make codegenTest SKIP_GEN=1
+      
+      - name: "Compile generated software artifacts"
+        run: make gool SKIP_GEN=1
```
balacij: Can you set `SKIP_GEN` as an environment variable for this job?

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+    name: LaTeX
+    needs: build
+    runs-on: ubuntu-26.04 # Ubuntu 26 has newer TeX packages that are significantly faster
+    if: ${{ needs.build.outputs.tex == 'true' }}
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: "Update apt package list"
+        run: |
+          sudo apt-get update
```
balacij, line 153: If the package lists are now fresher than when the CI was originally made, maybe we don't need to bother doing an `apt-get update`? Maybe we should experiment with that too... (in a subsequent PR).

djarabek: I believe an update is always necessary if you're doing an apt install as the package lists in the image could be very (weeks/months) outdated depending when the image was built.

code/Makefile
```diff
@@ -271,6 +271,9 @@ test_website: $(CLEAN_GF_PREFIX)$(LOG_FOLDER_NAME) website ##@Website Test fresh
 	@echo ---------------------------------------
 	@LOG_FOLDER="$(LOG_FOLDER)" LOG_SUFFIX="$(LOG_SUFFIX)" NOISY=$(NOISY) "$(SHELL)" "$(SCRIPT_FOLDER)log_check.sh"
 
+stack_test: code ##@General Run stack tests
+	stack test $(stackArgs) --no-interleaved-output
```
balacij, line 275: Nice. This will be good locally as well. Especially if we go through with #5037.

.github/workflows/Build.yaml
balacij: This list is great. We can add later if needed.

code/Makefile
balacij: Awesome!

.github/workflows/Build.yaml
smiths: How hard would it be to create our own `requirements.txt` file, like that used by Python projects to install their dependencies?  I really like the idea of explicitly recording our version numbers and making that visible, but I don't like the idea of recording the information in two different places.  That feels like a maintenance problem.  If we put the information in a `requirements.txt` (another name would be better, since requirements is already used in our project), then we could pull that information into the `Makefile` and into our documentation.

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
 jobs:
   build:
     name: "Build & Test"
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
balacij: Since we used ubuntu-26.04 in one, we should use it for all other jobs as well. Each ubuntu package release has a different package set, and that _might_ bite us one day.

.github/workflows/Build.yaml
djarabek: We currently rely on the actions image/Ubuntu packages for most of the dependencies, so getting specific versions of them is hard. I do agree that for the cases where we do have control over version numbers, that information should be centralized.

We should definitely document the major versions/language standard we target though (Java 8, C++11, etc).

.github/workflows/Build.yaml
djarabek: I was avoiding this for now because ubuntu 26 runners are in beta.

smiths: In a much simpler [yaml file](https://github.com/CynthiaLiu0805/BridgeChlorideExposurePredictor/blob/main/.github/workflows/latex-pages.yml) than this one, I used `runs-on: ubuntu-latest`.  Is that a valid option for us?  Probably not, but it would solve the problem of having to change the version number in the future.  Of course, it would cause a problem when everything breaks at some point in the future.  :smile:

.github/workflows/Build.yaml
balacij: If we were to switch completely to nix- or devenv-based tooling, this shouldn't be too crazy at all.

.github/workflows/Build.yaml
djarabek: We could use this, `ubuntu-latest` points to the latest stable version (which in this case is `ubuntu-24.04`). So to get the speed improvements in the latex job we'd still need to specify `ubuntu-26.04` there (at least for the next few months).

.github/workflows/Build.yaml
```diff
@@ -13,99 +13,183 @@ defaults:
...
+          sudo apt-get install -y graphviz fonts-lmodern texlive-bibtex-extra texlive-latex-extra texlive-science texlive-luatex inkscape
+
+      - name: "Download generated artifacts"
+        uses: actions/download-artifact@v8
+        with:
+          name: generated
+          path: code/build
+
+      - name: "Compile generated TeX artifacts"
+        run: make -j4 tex SKIP_GEN=1 SUMMARIZE_TEX=yes
```
djarabek: I was never able to pinpoint the exact reason for the sporadic LaTeX failures, but my best guess was some kind of memory/resource constraint. Limiting to 4 parallel jobs (as the runner has 4 CPUs) seems to have fixed it.

.github/workflows/Build.yaml
balacij: Relevant: #3181

In the past, `ubuntu-latest` was giving us a headache with sporadic network errors. I think it would be good to try it out again for everything other than the tex-related job.

djarabek: Sounds good. In my testing I didn't run into any actions related issues (and ran the jobs dozens of times total), so that issue is likely resolved.

balacij:
One thing I'm realizing now is that our Build & Test workflow doesn't render the Jupyter notebooks into HTML. It looks like #4816 accidentally skipped that part when the CI was split. Would you be able to add that back here?

.github/workflows/Build.yaml
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5271/commits/6da7bfd66a8b033170b3c6e4473d52946fa5a817

.github/workflows/Build.yaml
djarabek: Done in https://github.com/JacquesCarette/Drasil/pull/5271/commits/049a65c5da1178ec1a8213288364400964e29314

djarabek:
> One thing I'm realizing now is that our Build & Test workflow doesn't render the Jupyter notebooks into HTML. It looks like #4816 accidentally skipped that part when the CI was split. Would you be able to add that back here?

Done in https://github.com/JacquesCarette/Drasil/pull/5271/commits/3b74e1c40e056e004861e966a656d14a334286b0

.github/workflows/Build.yaml
djarabek: Updated in https://github.com/JacquesCarette/Drasil/pull/5271/commits/58e006a309c8e0759c324d149d5c4c85a95f9032

balacij (approved):
Great work!

JacquesCarette (approved):
Seems good to me, but my approval is much shallower here than on Haskell code.

PR #5272, opened 2026-07-08
Replace manual `instance hasChunkRefs X` with `declareHasChunkRefs ''X`

### LINKED ISSUE #5159: Replace manually-written `instance HasChunkRefs X` with `declareHasChunkRefs ''X`
`declareHasChunkRefs ''X` uses a strategy that will force correct gathering of all `UID` references. Currently, we are not. Proof:

https://github.com/JacquesCarette/Drasil/blob/4ca2449e7df991dfde9f82505a7b09088df7add4/code/drasil-theory/lib/Theory/Drasil/GenDefn.hs#L21-L40

`GenDefn` carries a derivation, which contains `Sentence`s. Those are not considered in the current instance of `HasChunkRefs` for `GenDefn`.

Note: `declareHasChunkRefs` uses `TemplateHaskell`, which means a bit of a compilation-time tradeoff, but if that becomes a huge issue, we can create some hand-crafted instances if absolutely necessary.

For reference, we use `declareHasChunkRefs` for only `Citation`s and `IdeaDict`s:

```console
drasil-lang/lib/Language/Drasil/Chunk/Citation.hs
19:import Drasil.Database (HasUID(..), UID, showUID, mkUid, declareHasChunkRefs,
47:declareHasChunkRefs ''Citation

drasil-lang/lib/Language/Drasil/Chunk/NamedIdea.hs
14:import Drasil.Database (mkUid, UID, HasUID(..), declareHasChunkRefs,
40:declareHasChunkRefs ''IdeaDict
```

camishibata:
Closes #5159

Replacing some of the manually-written `instance HasChunkRefs X` with `declareHasChunkRefs ''X`
Some types required keeping their manual instances due to the limitation of `declareHasChunkRefs ''X`, those are:
- `QDefinition`, `ConstraintSet`, `ModelKinds`, `MultiDefn`, `DefiningExpr`, `UIDRef`, `ConstrConcept`, `CodeDefinition` and `DifferentialModel` contain parameterized Kinds (* -> *)
- `Sentence` and `RelationConcept` contain Non-Vanilla Constructors
- `DefinedQuantityDict` and `PluralRule` / `NP` contain embedded Functions
- `chunk` instances:  `Drasil.Database.TH` depends on `Drasil.Database.Chunk`
- `Section` and `SecCons` are mutually dependent

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Chunk/NamedArgument.hs
+++ b/code/drasil-code/lib/Language/Drasil/Chunk/NamedArgument.hs
@@ -8,7 +8,7 @@ module Language.Drasil.Chunk.NamedArgument (
 
 import Control.Lens ((^.), makeLenses, view)
 
-import Drasil.Database (HasUID(..), HasChunkRefs(..))
+import Drasil.Database (HasUID(..), declareHasChunkRefs, Generically(..))
 import Language.Drasil (HasSpace(..), HasSymbol(..),
   Idea(..), MayHaveUnit(..), NamedIdea(..), Quantity,
   DefinedQuantityDict, Concept, dqdWr, Definition (defn), ConceptDomain (cdom))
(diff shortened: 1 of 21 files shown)


balacij (approved):
It looks like this catches a lot of dependencies previously unaccounted for!

For the list of things you mentioned, it just looks like we will need to go manual for now. Thank you for explaining each one, they all make sense.

Does this PR then close #5159 or does it only contribute to it?

JacquesCarette (approved):

PR #5273, opened 2026-07-08
Remove type fundep

B-rando1:
Contributes to #5233, reverting #5238.

Types are represented as `TypeData` in all our renderers except `CodeInfo`, so we can simplify `TypeSym` to use `TypeData` directly.


JacquesCarette (approved):
Too bad about the "hand revert". Nice changes to `CodeInfo`.

PR #5282, opened 2026-07-09
Rewrite Deploy CI

djarabek:
The workflow is now split up to run some steps in parallel, as well as being faster my minimizing the amount of dependencies that need to be installed for each step.

The workflow isn't split up quite as much as the PR CI one, as we need to combine the artifacts at the end to be able to publish the website. It's still about 2 times faster than the current one, though.

A sample run can be seen [in my fork](https://github.com/djarabek/Drasil/actions/runs/29025918436), and the published website is available at https://djarabek.github.io/Drasil/. I've clicked through the website and all the links work.


balacij (approved):
Narrowing the control scope of the needed permissions is nice.

Great work!

JacquesCarette (approved):

PR #5283, opened 2026-07-09
Add workaround for inkscape crash

djarabek:
Not really tested, but hopefully fixes the issue. I've only had the crash come up once since limiting to 4 LaTeX processes at a time (`-j4`).

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -145,8 +145,6 @@ jobs:
     if: ${{ needs.build.outputs.tex == 'true' }}
     env:
       SKIP_GEN: 1
-      # Workaround rare inkscape crash: https://gitlab.com/inkscape/inkscape/-/work_items/4716#note_1898150983
-      SELF_CALL: 1
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/Build.yaml
```diff
@@ -145,6 +145,8 @@ jobs:
     if: ${{ needs.build.outputs.tex == 'true' }}
     env:
       SKIP_GEN: 1
+      # Workaround rare inkscape crash: https://gitlab.com/inkscape/inkscape/-/work_items/4716#note_1898150983
+      SELF_CALL: 1
```
balacij: Shouldn't we add this to the Makefile instead of the CI?

djarabek: Yes, that does make a lot more sense. Moved to `tex_build.sh` in https://github.com/JacquesCarette/Drasil/pull/5283/commits/fbae51b672d0ba01b009b592ab2a9f45547ae48e

balacij (approved):

JacquesCarette (approved):

PR #5284, opened 2026-07-09
Specialized use of `SharedProg`

B-rando1:
This PR specializes uses of `SharedProg` in `drasil-gool` and `drasil-code`. For a large percentage of cases, I was able to remove one or both of `vis` (visibility) or `smt`  (statement). This and future PRs targetting `OOProg`, `ProcProc`, `CommonRenderSym`, `OORenderSym`, and `ProcRenderSym` should help reduce the workload of #5233.

I introduced an umbrella typeclass `SharedStatement` that ties together anything that commonly appears in a statement. Otherwise there were a dozen or so functions in `drasil-code` that required 8+ separate constraints.


JacquesCarette (approved):
That made some really nice improvements (both syntactically and semantically) in quite a feel places.

PR #5237, opened 2026-07-02
docs: shallow file analysis of `drasil-code`

balacij:
Posting the shallow file analysis of `drasil-code`.

After #5201, I will break up some of the tangible pieces of work found in this analysis into issues for us to discuss in more detail.

PR #5287, opened 2026-07-09
refactor(`SmithEtAlSRS`): move to `drasil-srs`

### LINKED ISSUE #5268: Move `SmithEtAlSRS` from `drasil-system` to `drasil-srs`
Spawned from https://github.com/JacquesCarette/Drasil/pull/5266

balacij:
Closes #5268 
Builds on #5286

This will 'lower' `drasil-system` quite a bit in the dependency graph!


code/drasil-code/package.yaml
```diff
@@ -33,6 +33,7 @@ dependencies:
 - drasil-makefile
 - drasil-metadata
 - drasil-printers
+- drasil-srs
 - drasil-system
```
JacquesCarette, line 37: If `drasil-code` still depends on `drasil-system`, I doubt there will be a big effect on the dependency graph (yet).

JacquesCarette (approved):

code/drasil-code/package.yaml
balacij: Screenshot 2026-07-10 at 12 06 33 PM

We have a bit more "down and to the right" movement now.

Removing `drasil-code`'s dependency on `drasil-srs` will create a major shift -- #5237.

PR #5289, opened 2026-07-10
debugging: allow usage for any system written with `concretizeAndWriteFiles`

balacij:
Builds on #5287

The debugging tools are very helpful when debugging. However, previously they were only usable when an SRS was in sight.

With this PR, we generate the debugging data with the `ChunkDB` rather than through a `SmithEtAlSRS`.

We can extend this idea further (if we wanted and in a subsequent PR), letting any system also produce any kind of desired 'debugging data files' by building a typeclass:

```haskell
class DebugData sys where
   debugData :: sys -> [FileLayout]
```

and, of course, instantiating it with for each of our system types and using `debugData` in `concretizeAndWriteFiles`


JacquesCarette (approved):
Very nice.

PR #5290, opened 2026-07-10
fix(`dependantsOrErr`): check with `dependants` instead of `find`

balacij:
Previously, this would always fail because `find` will never return a list of `UID`s (because those are not even "chunks," let alone ones we insert in the `ChunkDB`!).


JacquesCarette (approved):
Well spotted!

PR #5292, opened 2026-07-10
CI: Cache .stack-work and .hie

djarabek:
When main is built, a fresh cache is saved which can then be used by CI on PRs. For PRs with small or isolated changes, this can significantly speed up the build time.

See https://github.com/djarabek/Drasil/actions/runs/29031264011 for an example.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -54,6 +54,15 @@ jobs:
             ~/.stack
           key: ${{ runner.os }}-stack-${{ hashFiles('code/stack.yaml') }}
 
+      - name: "Restore stack work cache"
+        uses: actions/cache/restore@v6
+        with:
+          path: |
+            code/.stack-work
+            code/**/.stack-work
+            code/**/.hie
+          key: ${{ runner.os }}-stack-work
+
       - name: "Install dependencies"
         run: make stackArgs="--no-terminal" deps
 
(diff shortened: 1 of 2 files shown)


.github/workflows/Build.yaml
```diff
@@ -54,6 +54,15 @@ jobs:
             ~/.stack
           key: ${{ runner.os }}-stack-${{ hashFiles('code/stack.yaml') }}
 
+      - name: "Restore stack work cache"
+        uses: actions/cache/restore@v6
+        with:
+          path: |
+            code/.stack-work
+            code/**/.stack-work
+            code/**/.hie
```
balacij, line 63: I never thought to cache the `.hie` folders. That's a good idea.

balacij: Should we add this to the haddock job as well?

djarabek: I'm not sure what you mean, it's already in the haddock job as well

balacij: Currently, Deploy.yaml creates the caches and Build.yaml consumes them. But it doesn't look like Deploy.yaml restore its own -haddock caches in its haddock job.

(I was writing a comment on the wrong file, sorry.)

djarabek: The intention is to never restore .stack-work caches on main (Deploy.yaml), to prevent them from growing in size. The speed of PR CI is much more important than that of main. The caches are separate for build and haddock because this gave the best performance in my testing.

balacij (approved):
I forgot to "Approve" this PR. @djarabek and I spoke about this over Discord yesterday. She will draft up some relevant notes in #5296.

JacquesCarette (approved):

PR #5295, opened 2026-07-10
refactor(`package.yaml`s): minimize scope of package dependencies in each `drasil-*` subproject

balacij:
This does two things:

1. Reviews our package dependencies, removing the unnecessary ones.
2. Minimizes the visibility of external package dependencies to only the libraries/tests/executables they are used by.

After this, our dep. graph changes a little bit:

**PRE:**

Screenshot 2026-07-10 at 12 06 33 PM

**POST:**

drasil-all-pkgs-deps

### DIFF
--- a/code/drasil-assets/package.yaml
+++ b/code/drasil-assets/package.yaml
@@ -23,7 +23,6 @@ dependencies:
 - file-io >= 0.1.6
 - template-haskell
 - text
-- drasil-file-handling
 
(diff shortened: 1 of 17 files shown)


djarabek (approved):
Looks good, assuming that the exported dependencies are in fact the smallest possible sets now

JacquesCarette (approved):

PR #5299, opened 2026-07-10
Implement NativeVector instances for Julia

Xinlu-Y:
Contributes to #5059


code/drasil-code/test/Main.hs
```diff
@@ -79,20 +79,21 @@ gProcTestGroup n p =
     [ goldenTest "julia" $ directory [ps|julia|] $ genCodeProc unJLC unJLP p
     ]
 
-gProcMatlabTestGroup :: String ->
-  (forall r vis smt. (ProcProg r vis smt, NativeVector r) =>
+gProcVectorTestGroup :: String ->
```
B-rando1: Can we merge this with `gProcTestGroup`? Since it'll likely be a while before we add a third procedural language, I think it makes sense to combine these tests for now.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -406,14 +406,13 @@ instance Set JuliaCode where
...
 instance NativeVector JuliaCode where
-  vecScale = undefined
-  vecAdd = undefined
-  vecIndex = undefined
-  vecDot = undefined
-  vecMag = undefined
-  vecUnit = undefined
+  vecScale = binExpr multOp
+  vecAdd   = binExpr plusOp
+  vecIndex = jlVecIndex
```
B-rando1: Could this be `listAccess`?

B-rando1 (commented):
Looks good! Just a couple of comments.

code/drasil-code/test/Main.hs
Xinlu-Y: Matlab still has many undefined methods, so only `VectorTest` can generate MATLAB code right now. If I add Matlab to `gProcTestGroup`, `HelloWorld/FileTests/NameGenTest` would crash. I'll merge them once the remaining methods are filled in.

B-rando1: Oh, right. Sounds good.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
Xinlu-Y: Oh..I tried replacing it, but `listAccess` checks the value's `CodeType` and only accepts `List, Set, or Array`
the vector value doesn't carry one of those tags, so it hits error "listAccess called on non-list-type value" at runtime. 
So I'll keep it as-is for now.

B-rando1: I did some digging, because `vecType` is defined as `listType` so there shouldn't be a type error here. I noticed that in the test file, `a` and `b` have type `double` when they should have type `vecType double`. That seems to be where the type error is coming from.

Xinlu-Y: Pushed the fix.

B-rando1 (approved):

JacquesCarette (approved):
Code looks good - but needs a merge.

JacquesCarette (approved):

PR #5301, opened 2026-07-11
chore(`SmithEtAlSRS`): remove unused `_lbldCntnt` field

balacij:
#4972 looks like it did a bit more than we expected. When I rewrote how `LabelledContent` were gathered for the mdbook renderer (for it to know which files it needed to copy over for the mdbook compiler), I didn't realize that the `_lbldCntnt` field was otherwise unnecessary.


JacquesCarette (approved):
Nice!

PR #5307, opened 2026-07-13
Add README.md to .github/workflows documenting our CI

djarabek:
Depends on #5292 as the documentation is written based on those changes.


.github/workflows/README.md
```diff
@@ -0,0 +1,64 @@
...
+This workflow is split up into several jobs, some of which run in parallel:
+
+- Build & Test: Builds the Haskell code and compares the generated output with the stable folder.
+  The generated artifacts are uploaded for use in later jobs.
+- Haddock: Builds the Haddock documentation
+- In parallel, after Build & Test succeeds:
+  - Compile GOOL: Compiles the generated code (both the "gooltests" and the code generated by the examples).
+  - LaTeX: Renders the generated LaTeX code into PDFs.
+  - mdBook: Renders the generated mdBook markdown into HTML.
+  - Jupyter: Renders the generated Jupyter notebooks into HTML.
```
balacij, line 17: Are we able to split these off into a separate workflow file (e.g., called `Artifacts`)? The objective of this would be to deal with the artifacts exclusively. Then we can also rename `Compile GOOL` to `GOOL` or `Code`. If we really wanted to get fancy, we would do something similar to what we do with the document formats, adding one target per language.

balacij (approved):
Thank you! This documentation looks great!

.github/workflows/README.md
djarabek: If we switched around the Makefile to allow choosing between operating on the `stable` versions or the `build` versions this could work. I'm not sure if I want to make that level of changes to the Makefile and scripts though.

balacij: Oh, this was only a question. I don't think we should work on the CI any more than we already have, at least until it becomes a bottleneck again.

JacquesCarette (approved):

PR #5308, opened 2026-07-13
Remove block type family

B-rando1:
Builds on #5305 
Contributes to #5233


code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -15,6 +15,7 @@ import qualified Data.Map as Map (lookup)
 import Control.Monad (liftM2,liftM3)
 import Control.Monad.State (get, modify)
 import Data.List ((\\), intersect)
+import Text.PrettyPrint.HughesPJ (Doc)
```
JacquesCarette, line 18: We should not add more mentions of this. It would be better to add a synonym (you could even call it `MSBlock` !!) that abstracts away `MS (r Doc)`.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Logging.hs
```diff
@@ -5,11 +5,12 @@ module Language.Drasil.Code.Imperative.Logging (
 import Control.Lens ((^.))
 import Control.Lens.Zoom (zoom)
 import Control.Monad.State (get)
+import Text.PrettyPrint.HughesPJ (Doc)
```
JacquesCarette, line 8: Same here, of course.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -17,7 +17,7 @@ import Data.Maybe (maybeToList, catMaybes)
 import Control.Monad (liftM2, zipWithM)
 import Control.Monad.State (get, gets, modify)
 import Control.Lens ((^.))
-import Text.PrettyPrint.HughesPJ (render, parens)
+import Text.PrettyPrint.HughesPJ (Doc, render, parens)
```
JacquesCarette, line 20: Even though `render` and `parens` are a bit of a smell, let's not make it worse by adding explicit `Doc` references.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -726,7 +724,7 @@ jlListSlice
   -> Maybe (SValue JuliaCode)
   -> Maybe (SValue JuliaCode)
   -> SValue JuliaCode
-  -> MSBlock JuliaCode
+  -> MS (JuliaCode Doc)
```
JacquesCarette, line 727: This, on the other hand, can stay, as it will vary. (Same with other occurrences in other language renderers.) We can revisit that aspect of things later.

JacquesCarette (changes requested):

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
B-rando1: That is a good point. It's probably easier to add the type synonym now than later when everything's `Doc`.

Should we keep `MS` in the new type synonym? Since we'll be removing that state soon anyway, I was thinking that exposing the stateful part of the type would help later on.

B-rando1: From talking with Reed, it might be better to keep the `MS` inside the type, since it could go either way whether it helps or hinders the next refactor.

B-rando1:
Closing in favour of #5310

PR #5311, opened 2026-07-13
Made `Body` not a type family

B-rando1:
Builds on #5310 
Contributes to #5233


JacquesCarette (approved):
All these refactors are really bearing fruit now: the code is continually getting simpler.

JacquesCarette:
linter's not happy, but otherwise, it's all good.

JacquesCarette (approved):

PR #5314, opened 2026-07-14
Matlab control flow

Xinlu-Y:
Contributes to the MATLAB backend (control flow, declarations, assignments, and data container operations).


code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -687,3 +700,79 @@ mlPrint newLn f' _ v' = do
...
+mlArrayElem :: SValue MatlabCode -> SValue MatlabCode -> SVariable MatlabCode
+mlArrayElem arr' i' = do
+  i <- intToIndex i'
+  arr <- arr'
+  mkStateVar (render $ RC.value arr) (A.innerType $ return $ valueType arr)
+    (RC.value arr <> parens (RC.value i))
+
+mlListAppend :: SValue MatlabCode -> SValue MatlabCode
+  -> MS (MatlabCode (Doc, Terminator))
+mlListAppend lst' v' = do
```
B-rando1: Could this be simplified using `listSet`?

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -687,3 +700,79 @@ mlPrint newLn f' _ v' = do
...
+               (RC.value lst <> parens (RC.value idx))
+  lvar &= val'
+
+mlIndexOf :: SValue MatlabCode -> SValue MatlabCode -> SValue MatlabCode
+mlIndexOf lst' v' = do
+  lst <- lst'
+  v <- v'
+  d <- int
+  mkVal d (text "find" <> parens (RC.value lst <+> text "==" <+> RC.value v
+    <> text ", 1") <+> text "- 1")
```
B-rando1: Could this be simplified using `funcApp`?

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -366,7 +372,10 @@ instance InternalList MatlabCode where
   listSlice' = undefined
 
 instance InternalListFunc MatlabCode where
-  listAccessFunc = undefined
+  listAccessFunc t v = intValue v >>= ((`funcFromData` t) . mlListAccessFunc)
+
+mlListAccessFunc :: (CommonRenderSym r vis smt md) => r (Value r) -> Doc
+mlListAccessFunc v = parens $ RC.value v
```
B-rando1, line 373: It's too bad a bunch of our polymorphic `List` implementations assume indexing is done with `[]`. I see a few of these are identical except they use `parens` instead of `brackets`.

It might be best to keep them separate though, unless we can find a better way of abstracting over "list indexing syntax". There are many places in GOOL that take a `Doc` combinator so that they can abstract over some bit of syntax, and IMO these functions trade a little bit of code duplication for a much more confusing implementation.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -687,3 +700,79 @@ mlPrint newLn f' _ v' = do
...
+  indent $ RC.body catchB,
+  mlEnd]
+
+mlSubAssign :: SVariable MatlabCode -> SValue MatlabCode
+  -> MS (MatlabCode (Doc, Terminator))
+mlSubAssign vr' v' = do
+  vr <- zoom lensMStoVS vr'
+  v <- zoom lensMStoVS v'
+  stmtFromData (RC.variable vr <+> equals <+>
+    RC.variable vr <+> text "-" <+> RC.value v) Semi
```
B-rando1: This (and `mlAddAssign`) can be simplified using `assign`. E.g. `vr &= valueOf vr - v` or something close to that.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -410,22 +419,22 @@ instance StatementSym MatlabCode (Doc, Terminator) where
...
-  listDecDef = undefined
-  arrayDec = undefined
-  arrayDecDef = undefined
-  constDecDef = undefined
-  funcDecDef = undefined
+  setDec = varDec
+  setDecDef = varDecDef
+  listDec _ = varDec
+  listDecDef = CP.listDecDef
+  arrayDec _ = varDec
```
B-rando1, line 429: Does MATLAB allow you to index out-of-bounds? I know the Julia renderer uses `M.arrayDecAsList` to pad the array with `0`s, but maybe that's not needed with MATLAB.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -410,22 +419,22 @@ instance StatementSym MatlabCode (Doc, Terminator) where
...
+  arrayDecDef = CP.listDecDef
```
B-rando1: Minor nitpick, this could be `listDecDef`, since it's defined as `CP.ListDecDef`

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -463,23 +472,27 @@ instance CommentStatement MatlabCode (Doc, Terminator) where
...
-  switch = undefined
-  ifExists = undefined
-  for = undefined
-  forRange = undefined
-  forEach = undefined
-  while = undefined
-  tryCatch = undefined
-  assert = undefined
+  throw errMsg =
+    mkStmtNoEnd (text "error" <> parens (text ("'" ++ errMsg ++ "'")))
```
B-rando1: You can simplify this a bit with `funcApp`.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -463,23 +472,27 @@ instance CommentStatement MatlabCode (Doc, Terminator) where
...
+  ifExists = M.ifExists
+  for _ _ _ _ = error "MATLAB does not support C-style for loops; use forRange"
+  forRange i initv finalv stepv = forEach i (mlRange initv finalv stepv)
+  forEach = CS.forEach' mlForEach
+  while = C.while id empty mlEnd
+  tryCatch = G.tryCatch mlTryCatch
+  assert condition errorMessage = do
+    cond <- zoom lensMStoVS condition
+    errMsg <- zoom lensMStoVS errorMessage
+    mkStmtNoEnd (text "assert" <> parens (RC.value cond <> text ", " <> RC.value errMsg))
```
B-rando1: This looks like it could be simplified with `funcApp`.

B-rando1 (changes requested):
This is looking really good, I just have a few suggestions to improve the code readability. It's not your fault for using things like `mkVal`, `mkStatementNoEnd`, etc.; a lot of the existing code (some of which I wrote before I knew better 😅) does this, so I'm guessing you were basing your code off that. In my opinion we should only use those "low-level" constructors when we have no other choice, and should use higher-level features like `funcApp` and `(&=)` wherever we can.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
Xinlu-Y: Matlab auto-extends arrays, so I think no need to pre-fill with zeros like Julia.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -336,20 +341,21 @@ instance Reference MatlabCode where
...
+    in funcApp "copy" arrTp [arr]
 
 instance List MatlabCode (Doc, Terminator) where
   listSize = CS.listSize "length"   -- length(v)
   listAdd = undefined
-  listAppend = undefined
-  listAccess = undefined
-  listSet = undefined
-  indexOf = undefined
+  listAppend lst = listSet lst (listSize lst)
```
B-rando1, line 347: Since MATLAB is 1-indexed, wouldn't this need a `+ 1`?

Xinlu-Y:
Hey @B-rando1 , I've made all the changes. Ready for review whenever you have time.

B-rando1 (commented):
I was just looking through it! Code review is a nice break from the refactor work lol

Everything looks much cleaner now. I just noticed one thing, but I think it should be good after that.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
Xinlu-Y: `listSize` returns the length (e.g. 3 for a 3-element array), and `mlListSet` calls `intToIndex` which adds 1, so the final index becomes length(lst) + 1.

B-rando1: Ahh right!

B-rando1 (approved):

JacquesCarette:
There's a conflict now.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
JacquesCarette: "indexing" should probably be a method in some class, so that those details can be dispatched properly.

JacquesCarette (approved):

JacquesCarette:
Merge "failed" in the sense that the build failed after that.

Xinlu-Y:
Fixed.

JacquesCarette (approved):

Issue #5065, opened 2026-06-04
Remove/rework `StateVarSym` from GOOL


B-rando1:
GOOL's variable taxonomy is very weird, and largely implicit as far as the code is concerned.

For most things, we have the `Variable` type family. All renderers instantiate this with `VarData`:

https://github.com/JacquesCarette/Drasil/blob/f2dd17639524501881a9ad38d7302bc55144716f/code/drasil-gool/lib/Drasil/Shared/AST.hs#L154-L156

They have attachment (class- or instance-level), name (for the internal variable map, and sometimes for hacks), type, and doc.

There are many flavours of these: `var` and `const` for both instance-level and function/script-level variables, `classVar` and `classConst` for class-level variables, `classVarAccess` and `instanceVarAccess` for accessing a variable bound to a class or instance, and so on. This is a fairly reasonable design. We could think about adding `FunctionLevel` and/or `ModuleLevel` as other options for `Attachment`, but at least this seems like a good foundation.

But then, specifically for constructing classes, we use `StateVar` instead. They're constructed using these methods:

https://github.com/JacquesCarette/Drasil/blob/d98c8a5859e4678e180e4c85f8b654b9ff34b874/code/drasil-gool/lib/Drasil/GOOL/InterfaceGOOL.hs#L109-L114

Oddly, they also take an `Attachment`. This is definitely wrong, taking both an `Attachment` and an `SVariable` that has an `Attachment`. Indeed, the reason I started writing this issue is that I accidentally discovered we can generate code with two `static` keywords in a variable declaration. So that's interesting. Anyway, most languages instantiate `StateVar` directly as `Doc`, and they seem to only exist for the C++ renderer. It instantiates `StateVar` as `StateVarData`:

https://github.com/JacquesCarette/Drasil/blob/f2dd17639524501881a9ad38d7302bc55144716f/code/drasil-gool/lib/Drasil/Shared/AST.hs#L120-L122

The only unique piece of information here is the `destructStmts` field. It looks like the intention is that when you declare a class, the C++ renderer is supposed to take note of which variables belong to an instance of a class, and create a destructor for the class that calls `free` on any heap-allocated variables belonging to the instance. But (as of 7 years ago: #1945) GOOL doesn't perform any dynamic allocation, and these destructor statements are hard-coded to be empty. So `StateVar` isn't giving us anything, other than holding the `Visibility` of a variable.

[...]

PR #5318, opened 2026-07-15
Removed destructor statements from C++ renderer

B-rando1:
Builds on #5317 
Contributes to #5065 

I'm going to check if the `VisibiltyTag` in `StateVarData` is actually used anywhere. If not, then in a subsequent PR I'll remove it and switch the C++ renderer to use `Doc` for its `StateVar` representation. This will simplify handling `StateVar` for #5233.


B-rando1:
It looks like we can't actually remove the `VisibilityTag` from the C++ renderer, because it uses it to sort public and private members in its class definition. This PR is still a good change, though.

JacquesCarette (approved):

Issue #2857, opened 2021-10-05
In the `code/` folder, many README.md files are outdated


balacij:
As the title says, many of the README.md files our outdated.

Some files are completely outdated and discuss non-existent files, or files that moved elsewhere.

daijingz:
Automatically Selected and started (Not with any permission, I just start working now)

daijingz:
Update 1: In the `/Drasil/code/` repository; the README.md file forgot to explain the newly-added `drasil-codelang` folder. This is been updated on #3959

smiths:
@balacij, I agree that the review process is the way to update our documentation.  I also agree that Haddock should be a higher priority than the `README` file.

Should we think about adding a PR template?  We could include in the template reminders to update the Haddock documentation.

smiths:
I've never used a PR template before, either, but the issue templates have been handy, especially in capstone.

There are plenty of examples online.  There is a list at the bottom of this article on [Github pull request templates](GitHub pull request template).  I think we want something simple, at least to start with, like the first template example:

```
## Describe your changes and why they are necessary

## Issue ticket number and link

## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] I have added Haddock documentation
- [ ] I have followed the Drasil naming conventions
- [ ] If this PR changes stable, I have run `make stabilize`
```

I've made a few modifications from the original source.  You probably have some additional ideas of things we could add to the checklist.  We can't add the point about following naming conventions until we've documented our naming conventions.  Do we want to suggest a local lint as part of the checklist?

JacquesCarette:
Can we semi-Drasil-ify this? Maybe we could extract the `README.md` from the Haskell code itself, once we have the right 'template'?

balacij:
I am surprised: https://pandoc.org/MANUAL.html#general-options . Pandoc supports Haddock as an input format, I'm not sure how well it will work. Maybe someone can look at it next summer?

PR #5319, opened 2026-07-15
refactor(`-printers`): use `drasil-data-formats` HTML IR in renderer

camishibata:
Contributes to #4989


code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
```diff
@@ -26,8 +26,8 @@ newtype HTMLRenderOptions = HTMLRO {
 renderHTML :: HTMLRenderOptions ->  HTML -> Doc ann
 renderHTML opt(HTML heads bodies) =
   vcat ["<!DOCTYPE html>", angles "html",
-    renderHeadSec heads,
-    renderBodySec opt bodies,
+    indent 2 $ renderHeadSec heads,
+    indent 2 $ renderBodySec opt bodies,
```
balacij: Indentation size is a good thing to leave as an option in `HTMLRenderOptions`. The default can be 2.

code/drasil-printers/lib/Language/Drasil/HTML2/CSS.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- | Defines functions to create accompanying .css files for HTML generators.
+module Language.Drasil.HTML2.CSS (
+  genericCSS, stylesheet
+) where
+
+import Text.PrettyPrint (Doc, text, vcat)
```
balacij: Maybe we can convert this into `Prettyprinter`-based `Doc`?

code/drasil-printers/lib/Language/Drasil/HTML2/CSS.hs
```diff
@@ -0,0 +1,146 @@
...
+    ],
+  vcat [
+    text "dd {",
+    text "  margin: 0;}"
+    ]
+  ]
+
+-- | Create the link to the CSS file
+stylesheet :: Text -> HTMLHead
+stylesheet css = Link "stylesheet" (css <> ".css") [Attr "type" "text/css"]
```
balacij: This is a good constructor that probably belongs closer to the other constructors.

code/drasil-printers/lib/Language/Drasil/HTML2/Monad.hs
```diff
@@ -0,0 +1,11 @@
+-- | Defines types needed for HTML printers.
+module Language.Drasil.HTML2.Monad (PrintHTML, unPH) where
+
+import Language.Drasil (RenderSpecial(..), Special(..))
+
+-- | Printing "monad".
+newtype PrintHTML = PH {unPH :: String}
```
balacij: Can we avoid building this?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+mathJaxScript = T.unlines [
+    "MathJax = {",
+    "  loader: {load: ['[tex]/textmacros', 'output/chtml']},",
+    "  tex: {",
+    "    packages: {'[+]': ['textmacros']}",
+    "  },",
+    "  svg: {",
+    "    fontCache: 'global'",
+    "  }",
+    "};"]
```
balacij: It would be nice if we could try to use the new JSON renderer to print the _object_ part of this Javascript snippet. Is that easy enough to do?

If it is, we can also leave the above as more options for the HTML renderer.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+import qualified Language.Drasil.TeX.Print as TeX (pExpr, spec)
+import Language.Drasil.TeX.Monad (runPrint, MathContext(Math), D, toMath, PrintLaTeX(PL))
+
+-- | Referring to 'fence' (for parenthesis and brackeds). Either opened or closed.
+data OpenClose = Open | Close
+
+-- | Generate an HTML document from a Drasil 'Document'.
+genHTML2 :: HTMLRenderOptions -> String -> Document -> PNew.Doc ann
+genHTML2 opts fn doc = renderHTML opts (build fn doc)
+--      second arg should really be of type Filename, but that's not in scope
```
balacij: Can you please add documentation to these arguments?

Should the `String` be added to `HTMLRenderOptions`?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+      HTML.Title (extractText t),
+      Meta [Attr "charset" "utf-8"],
+      Script [] mathJaxScript,
+      Script [Attr "type" "text/javascript", Attr "id" "MathJax-script", Attr "async" "",
+      Attr "src" "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml-full.js"] ""
+      ]
+    bodies = [
+      articleTitle (specToHTML t),
+      author (specToHTML a)
+      ] ++ concatMap loToHTML c
```
balacij: Can we merge `build` into `genHTML2`?

Also, can we move the specific MathJax Javascript to an option?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+extractText (Sp s)              = T.pack $ unPH $ special s
+extractText (Ref (Cite2 n) _ a) = extractText a <> " " <> extractText n
+extractText (Ref _ _ a)         = extractText a
+extractText EmptyS              = ""
+extractText (Quote q)           = "\"" <> extractText q <> "\""
+extractText (Tooltip _ s)       = extractText s
+
+-- | Helper for rendering a 'D' from Latex print.
+printMath :: D -> PLegacy.Doc
+printMath = (`runPrint` Math)
```
balacij: It looks like `printMath` is often used right after `toMath`. Can we collapse this code into those statements?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+extractText (S s)               = T.pack s
+extractText (E e)               = T.pack $ show $ printMath $ toMath $ TeX.pExpr e
+extractText (a :+: b)           = extractText a <> extractText b
+extractText HARDNL              = " "
+extractText (Sp s)              = T.pack $ unPH $ special s
+extractText (Ref (Cite2 n) _ a) = extractText a <> " " <> extractText n
+extractText (Ref _ _ a)         = extractText a
+extractText EmptyS              = ""
+extractText (Quote q)           = "\"" <> extractText q <> "\""
+extractText (Tooltip _ s)       = extractText s
```
balacij: I would call this `printSpec`.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+-- pMatrix :: [[Expr]] -> Doc
+-- pMatrix [] = text ""
+-- pMatrix [x] = text "<tr>" <> pIn x <> text "</tr>\n"
+-- pMatrix (x:xs) = pMatrix [x] <> pMatrix xs
+
+-- Not used since we use MathJax handles this
+-- pIn :: [Expr] -> Doc
+-- pIn [] = text ""
+-- pIn [x] = text "<td>" <> pExpr x <> text "</td>"
+-- pIn (x:xs) = pIn [x] <> pIn xs
```
balacij: Let's just delete any unused code.

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,558 @@
...
+makeDefnHTML ps l =
+  let
+    attrs  = [Attr "id" (extractText l), Attr "class" "defn-table"]
+    refRow = HTML.Row []
+      [THeader [] [RawText "Refname"], TData [] [TextFormat HTML.Bold [] (specToHTML l)]]
+    dataRows = map (\(f, d) -> HTML.Row []
+      [THeader [] [RawText (T.pack f)], TData [] (concatMap loToHTML d)]) ps
+  in
+    [HTML.Table attrs (refRow : dataRows)]
+
```
balacij, line 130: If you think it would be good, you can split this file up too.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+    dataRows = map (HTML.Row [] . map (TData [] . specToHTML)) lls
+    tableNode = HTML.Table attrs (headerRow : dataRows)
+    captionNode = HTML.Paragraph [Attr "class" "caption"] (specToHTML t)
+    wrapperAttrs = [Attr "id" (extractText r)]
+
+-----------------------------------------------------------------
+------------------BEGIN DEFINITION PRINTING----------------------
+-----------------------------------------------------------------
+
+-- | Renders definition tables (Data, General, Theory, etc.).
```
balacij: A goal in #4989 is for it to "know less" about the other packages, so we should remove mention of `(Data, General, Theory, etc.).`.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+
+-- | Helper for rendering list items.
+itemToHTML :: ItemType -> [HTMLBody]
+itemToHTML (Flat s)     = foldRaw $ specToHTML s
+itemToHTML (Nested s l) = foldRaw $ specToHTML s ++ [makeListHTML l]
+
+---------------------
+--HTML bibliography--
+---------------------
+-- **THE MAIN FUNCTION**
```
balacij: This doesn't look like the main function?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+-- | Makes a bilbliography for the document.
+makeBib :: BibRef -> [HTMLBody]
+makeBib bib =
+  [DescriptionList [Attr "class" "reference-list"] (concatMap renderCitation bib)]
+  where
+    renderCitation :: Citation -> [DItem]
+    renderCitation cite@(Cite e _ _) =
+      let (termDoc, detailsDoc) = renderCite htmlBibFormatter cite
+          termHTML = [RawText "[", TextFormat HTML.Bold [] termDoc, RawText "]"]
+      in [DTerm [Attr "id" (T.pack e)] termHTML, DDetails [] detailsDoc]
```
balacij: I would call this `printBib` and then have it return only a single `HTMLBody` because it returns the list node.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+
+-- | HTML specific bib rendering functions
+htmlBibFormatter :: BibFormatter
+htmlBibFormatter = BibFormatter {
+  emph = \x -> [TextFormat Emphasis [] x],
+  spec = specToHTML
+}
```
balacij: Is this strictly necessary? Should this be moved into `HTMLRenderOptions`?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+
+-- | Renders a person's last name.
+rendPersL :: Person -> String
+rendPersL =
+  (\n -> (if not (null n) && last n == '.' then init else id) n) . rendPersLFM
+
+-- | adds an 's' if there is more than one person in a list.
+toPlural :: People -> Text -> Text
+toPlural (_:_) str = str <> "s"
+toPlural _     str = str
```
balacij: I think it would be good to move everything `Citation`-related to a separate file. So, it sounds like splitting this file up might be a good idea.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,558 @@
...
+exprToHTML e              =
+  [RawText $ T.pack $ show $ mjDelimDisp $ printMath $ toMath $ TeX.pExpr e]
+  where mjDelimDisp d = text "\\(" <> d <> text "\\)"  -- PRINTMATH RETURNS OLD DOC
+-- Non-mathjax
+{-
+pExpr (Sqrt e)       = text "&radic;(" <> pExpr e <> text ")"
+pExpr (Div a b)      = fraction (pExpr a) (pExpr b)
+pExpr (Case ps)      = cases ps pExpr
+pExpr (Mtx a)        = text "<table class=\"matrix\">\n" <> pMatrix a <> text "</table>"
+-}
```
balacij: I think you can delete this.

balacij (changes requested):
This is shaping up nicely. You've created a lot of good opportunities for more generation options 😄

code/drasil-data-formats/test/golden/html/tags.html
```diff
@@ -1,67 +1,81 @@
...
+  <body>
+    <div id="main-section">
+      <h1 class="title">tagsHTMLTest</h1>
+      <h2 class="h2">tagsHTMLTest</h2>
+      <h3 class="h3">tagsHTMLTest</h3>
+      <p class="paragraph">
+        Testing paragraph and text formats: <b id="bold">bold, </b><em id="emphasis">emphasis, </em><sub id="subscript">subscript, </sub><sup id="superscript">superscript, </sup><span id="span">span.</span>
+      </p>
+      <ol id="ordered-list">
+        <li>
```
JacquesCarette, line 23: For such short 'li', I would personally be happier if they were all on a single line.

code/drasil-data-formats/test/golden/html/tags.html
```diff
@@ -1,67 +1,81 @@
...
+          Item 2
+        </li>
+        <li>
+          Item 3
+        </li>
+      </ul>
+      <table class="table">
+        <tr class="row">
+          <th class="table-header">Header1</th>
+          <td class="data-cell">
```
JacquesCarette: Same with 'td'

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,227 @@
...
+
+-- | Renders citation as an article style.
+useStyleArtcl :: BibFormatter -> StyleGuide -> (CiteField -> [HTMLBody])
+useStyleArtcl f MLA     = artclMLA f
+useStyleArtcl f APA     = artclAPA f
+useStyleArtcl f Chicago = artclChicago f
+
+-- FIXME: move these show functions and use tags, combinators
+
+-- | Cite books in MLA format.
```
JacquesCarette, line 122: All of these `RawText` followed by a `String` should be pulled out and given names.

And since they are all at the end, maybe we should also have a `foldRaw1` that uses that as its "starting point" instead of doing an at-end append (which is quite inefficient).

code/drasil-printers/lib/Language/Drasil/HTML2/Helpers.hs
```diff
@@ -0,0 +1,203 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Helper functions for creating HTML printers (specifically, HTML tag wrappers).
+module Language.Drasil.HTML2.Helpers (
```
JacquesCarette: Can we try to find a better name than 'Helpers'? Even if that means splitting this file?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,163 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- | Defines all functions needed to print HTML files. For more information on each of the helper functions, please view the [source files](https://jacquescarette.github.io/Drasil/docs/full/drasil-printers-0.1.10.0/src/Language.Drasil.HTML.Print.html).
+module Language.Drasil.HTML2.Print(
```
JacquesCarette: Maybe we should switch to calling this 'rendering' rather than 'printing' -- so this would be `.Render` instead of `.Print`.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,163 @@
...
+
+import Drasil.Data.Formats.JSON (JSON(..), renderJSON, jsonRenderOpts, JSONStyle(Pretty))
+
+import Drasil.Data.Formats.HTML hiding (Title, Row, Bold, ListType, Ordered,
+  Unordered, span, Paragraph, Table, List, Figure)
+import qualified Drasil.Data.Formats.HTML as HTML
+
+-- | Generate an HTML document from a Drasil 'Document'.
+--   Arguments: Rendering options, Bib rendering options, CSS file name, `Document` to be rendered
+genHTML2 :: HTMLGenOptions -> HTMLRenderOptions -> String -> Document -> PNew.Doc ann
```
JacquesCarette: `renderToHTML2` ? And do we need the `2`, given name-spacing?

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
```diff
@@ -0,0 +1,163 @@
...
+      )
+
+-- | Helper for transforming layout objects ('LayoutObj's) into HTML.
+loToHTML :: HTMLRenderOptions -> LayoutObj -> [HTMLBody]
+-- Creates delimeters to be used for mathjax displayed equations
+-- Latex print sets up a \begin{displaymath} environment instead of this
+loToHTML _ (EqnBlock contents) =
+  [RawText $ T.pack $ show $ mjDelimDisp $ renderMath $ TeX.spec contents]
+  where
+    mjDelimDisp d = text "\\[" <> d <> text "\\]"
```
JacquesCarette: this should not be hard-coded here.

code/stable-website/index.html
```diff
@@ -1,850 +1,1050 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <link rel="stylesheet" type="text/css" href="index.css">
-    <title>Drasil - Generate All the Things!</title>
+    <link rel="stylesheet" href="index.css" type="text/css">
+    <title>
```
JacquesCarette: No need for 3 lines, a single is/was better

code/stable-website/index.html
```diff
@@ -1,850 +1,1050 @@
...
-      loader: {load: ['[tex]/textmacros', 'output/chtml']},
-      tex: {
-        packages: {'[+]': ['textmacros']}
-      },
-      svg: {
-        fontCache: 'global'
-      }
-    };
+      MathJax = {
+        "loader": {
```
JacquesCarette, line 9: Here too, too much non-needed vertical space.

JacquesCarette (changes requested):
Ultimately, we'll want the renderers to use the `Reader` monad, so that options passing can become implicit. But it doesn't have to be done in this PR but a subsequent one.

code/drasil-data-formats/test/golden/html/tags.html
balacij: This is a great opportunity to add a rendering option to the HTML renderer! This should not be done in _this_ PR, however.

code/stable-website/index.html
balacij: Strictly speaking, this is what conventional Javascript code would look like (sans the double quotation marks around keys).

This should be a feature of the JSON renderer (which we use here). Specifically, that if an object has only one member, and that member is neither an object with more than one member nor an array with more than a few "short" elements, we can inline the objects and its members. We should probably do this in a separate PR first.

More generally, we can learn from [JSON5](https://json5.org/) for features.

camishibata:
Ready for review @JacquesCarette @balacij

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
```diff
@@ -138,3 +153,9 @@ escapeHTMLText = T.concatMap escapeChar
     escapeChar '\'' = "&#39;"
     escapeChar '/'  = "&#x2F;"
     escapeChar c    = T.singleton c
+
+-- | Merges contiguous `RawText` in one
+normalize :: [HTMLBody] -> [HTMLBody]
+normalize [] = []
+normalize (RawText a : RawText b : rest) = normalize (RawText (a <> b) : rest)
+normalize (x : xs) = x : normalize xs
```
balacij: We should have 3 kinds of `normalize`: `normalizeHTML`, `normalizeBody`, and `normalizeHead`. `normalizeHTML`  would rely on the latter two and would be the only one used (and only used once, after the HTML tree has been created, instead of being sporadically used). We should also try to go a little bit deeper in normalizing an HTML tree. For example, normalizing `TextFormat` internals.

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
```diff
@@ -2,82 +2,83 @@
...
+  TagType (..),
   )
-
-newtype HTMLRenderOptions = HTMLRO {
-  -- | What 'TagType' is each 'CustomTag'?
-  customElementTagTypes :: M.Map CustomTag TagType
+import Prettyprinter
+  (vcat, Doc, angles, dquotes, equals, hcat, hsep, indent, pretty, space)
+
+data HTMLGenOptions = HTMLBO
```
balacij: I think the old name was a bit better. I think we might have miscommunicated on this point in person...

code/stable-website/index.html
```diff
@@ -1,850 +1,543 @@
...
-        <div id="Sec:DebuggingDrasil">
-          <div class="subsection">
-            <h2>Debugging Drasil</h2>
-            <p class="paragraph">
-              Debugging information can be found on the <a href="https://github.com/JacquesCarette/Drasil/wiki/Debugging-in-Drasil">Debugging in Drasil</a> page.
-            </p>
-          </div>
-        </div>
-      </div>
+    <div id="Sec:Introduction" class="section">
```
balacij: Maybe we should switch to using `section` nodes and avoid the class?

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+import Language.Drasil.HTML2.Spec (printSpec, specToHTML, renderMath)
+
+import Drasil.Data.Formats.JSON (JSON(..), renderJSON, jsonRenderOpts, JSONStyle(Minified))
+import Drasil.Data.Formats.HTML hiding (Title, Row, Bold, ListType, Ordered,
+  Unordered, span, Paragraph, Table, List, Figure)
+import qualified Drasil.Data.Formats.HTML as HTML
+
+-- | Generate an HTML document from a Drasil 'Document'.
+--   Arguments: Rendering options, Bib rendering options, CSS file name, `Document` to be rendered
+genHTML :: HTMLGenOptions -> HTMLRenderOptions -> String -> Document -> PNew.Doc ann
```
balacij: Let's have this function only return the `HTML` AST. Whatever code that uses this function should decide on its own `HTML->Doc` renderer.

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+  map (\(i, l) -> LItem (mlrefAttr l) (itemToHTML i)) items
+makeListHTML (Unordered items) = HTML.List HTML.Unordered [Attr "class" "list"] $
+  map (\(i, l) -> LItem (mlrefAttr l) (itemToHTML i)) items
+makeListHTML (Definitions items) = HTML.List HTML.Unordered [Attr "class" "hide-list-style-no-indent"] $
+  map (\(b, e, l) -> LItem (mlrefAttr l) (specToHTML b ++ [RawText " is the "] ++ itemToHTML e)) items
+
+-- | Helper for setting up references as HTML Attributes.
+mlrefAttr :: Maybe Spec -> [Attr]
+mlrefAttr Nothing  = []
+mlrefAttr (Just l) = [Attr "id" (printSpec l)]
```
balacij:
```suggestion
-- | Convert @Maybe Spec@s int ID `Attr`s if the `Spec` exists.
mbIdAttr :: Maybe Spec -> [Attr]
mbIdAttr = maybe [] (\x -> [Attr "id" $ printSpec x]) 
```

(or so...)

code/drasil-printers/lib/Language/Drasil/HTML2/Spec.hs
```diff
@@ -0,0 +1,83 @@
...
+  where
+    mjDelimDisp d = PLegacy.text "\\(" <> d <> PLegacy.text "\\)"
+
+-- | Helper for rendering a 'D' from Latex print.
+printMath :: D -> PLegacy.Doc
+printMath = (`runPrint` Math)
+
+-- | Helper for converting and rendering math expressions to Latex print.
+renderMath :: D -> PLegacy.Doc
+renderMath = printMath . toMath
```
balacij: Is there any way we can avoid these two functions? They look like they might at least be best off in the TeX subfolder?

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+        HTML.Title (printSpec t),
+        Meta [Attr "charset" "utf-8"],
+        Script [] mathJaxScript,
+        Script
+          [ Attr "type" "text/javascript",
+            Attr "id" "MathJax-script",
+            Attr "async" "",
+            Attr "src" (T.pack $ mathJaxSrc rOpts)
+          ]
+          ""
```
balacij: We should create smart constructots for these two kinds of (very common) `script` nodes. That is, `inlineScript` and `externalScript`. The `externalScript` should only specifically require a `src` attribute and then allow arbitrary `Attr`s as well. The `inlineScript` should not allow any attrs.

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+    configJSON = T.pack $ show $ renderJSON
+      (jsonRenderOpts Minified)
+      ( JObject [
+        ("loader",
+        JObject [("load", JArray ["[tex]/textmacros", "output/chtml"])]),
+        ("tex", JObject [("packages", JObject [("[+]", JArray ["textmacros"])])]),
+        ("svg", JObject [("fontCache", "global")])]
+      )
+
+-- | Helper for transforming layout objects ('LayoutObj's) into HTML.
```
balacij: Since we're cleaning up the code as well, we can drop all of the `Helper for` prefixes in our comments.

code/stable-website/index.html
```diff
@@ -1,850 +1,543 @@
...
-    MathJax = {
-      loader: {load: ['[tex]/textmacros', 'output/chtml']},
-      tex: {
-        packages: {'[+]': ['textmacros']}
-      },
-      svg: {
-        fontCache: 'global'
-      }
-    };
+      MathJax = {"loader":{"load":["[tex]/textmacros","output/chtml"]},"tex":{"packages":{"[+]":["textmacros"]}},"svg":{"fontCache":"global"}};
```
balacij: We should make sure to file an issue for this. We need to support more options in the JSON renderer so we can generate the 'right' code. This code is now unreadable. I think the expanded version is better, at least in the interim.

code/stable-website/index.html
```diff
@@ -1,850 +1,543 @@
...
-        <h1></h1>
-        <div id="Figure:Drasil">
-          <figure>
-            <img src="./images/Icon.png" alt="" width="50.0%" >
-          </figure>
-        </div>
-      </div>
+    <div class="author">
+      <h2><a href="https://github.com/JacquesCarette/Drasil">Link to GitHub Repository</a></h2>
     </div>
```
balacij, line 37: It looks like this was happening before as well, but it's very very odd that this is in the "author" `div`. How did this happen?

code/stable-website/index.html
```diff
@@ -1,850 +1,543 @@
...
+        <li>Input Structure<ul class="list">
+          <li>B - Inputs are Bundled in a class</li>
+          <li>U - Inputs are Unbundled</li>
+        </ul></li>
+        <li>Constant Structure<ul class="list">
+          <li>I - Constant values are Inlined</li>
+          <li>WI - Constants are stored With the Inputs</li>
+          <li>B - Constants are stored in variables that are Bundled in a class</li>
+          <li>U - Constants are stored in variables that are Unbundled</li>
+        </ul></li>
```
balacij, line 375: I would have liked to see:

```html
        <li>
           Constant Structure
           <ul class="list">
             <li>I - Constant values are Inlined</li>
             <li>WI - Constants are stored With the Inputs</li>
             <li>B - Constants are stored in variables that are Bundled in a class</li>
             <li>U - Constants are stored in variables that are Unbundled</li>
           </ul>
        </li>
```

The "real" issue here is https://github.com/JacquesCarette/Drasil/pull/5319#discussion_r3619037881 however.

When we render HTML, some nodes very often contain just contain textual content. For example, `li`, `td`, `em`, `i`, `span`, etc. (what goes in this list is up for debate, but that's not a real concern right now). The point is that when we render these nodes, we want to opportunistically inline them when appropriate. That is, we want to inline them as much as possible when their bodies are purely _textual_ in nature as well (that being `RawText`, `TextFormat` (and with only textual inner HTML as well), etc.). Whenever there is non-textual content added, we should render the fully expanded HTML version.

We should add a rendering option to `HTMLRenderOptions` (currently named `HTMLGenOptions`, belonging to `drasil-data-formats`):

```haskell
-- | Mixed content inline policy.
data InlinePolicy = AlwaysBlock | AlwaysInline | InlineTextualContent (Set Tag)

-- | Check if an 'HTMLBody' is "purely texutal" in nature (that is, contains no "block" nodes).
isPureText :: HTMLBody -> Bool
isPureText (RawText _)        = True
isPureText (TextFormat _ bs)  = all isPureText bs
-- etc... = True
isPureText _                  = False
```

The idea here is that we have mixed kinds of content when it comes to rendering in HTML, and when we are rendering textual-focused nodes with no inner content that mix HTML with text, we can render them inline.

code/stable-website/index.html
```diff
@@ -1,850 +1,543 @@
...
     </div>
-    <div id="Sec:Footer">
-      <div class="section">
-        <h1></h1>
-        <p class="paragraph">
-          Copyright (c) Jacques Carette, 2021. All rights reserved. This website is a software artifact generated by Drasil.
-        </p>
-      </div>
+    <div id="Sec:Footer" class="section">
+      <h1></h1>
```
balacij: Are we able to remove the empty h1?

balacij (changes requested):
We're very very close!!

JacquesCarette:
Given @balacij 's rather thorough review, I will wait for an update to do a pass myself.

code/stable-website/index.html
JacquesCarette: Note that we should NOT remove such things via a post facto filter, but rather not generate them to start with.

code/stable-website/index.html
camishibata: This is set here:
https://github.com/JacquesCarette/Drasil/blob/4a5e860b9aaac4532cf09c31821dce60626310ab/code/drasil-website/app/Main.hs#L63-L65
https://github.com/JacquesCarette/Drasil/blob/4a5e860b9aaac4532cf09c31821dce60626310ab/code/drasil-website/lib/Drasil/Website/Body.hs#L125-L126

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
camishibata: Implemented in 329f04e2afe49655b667bc8e9a2add77b428adf5.

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
```diff
@@ -1,83 +1,86 @@
...
+  TagType (..),
   )
-
-newtype HTMLRenderOptions = HTMLRO {
-  -- | What 'TagType' is each 'CustomTag'?
-  customElementTagTypes :: M.Map CustomTag TagType
+import Prettyprinter
+  (vcat, Doc, angles, dquotes, equals, hcat, hsep, indent, pretty, space)
+
+data HTMLRenderOptions = HTMLBO
```
balacij: I think we should keep the previous constructor name of `HTMLRO`?

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Render.hs
```diff
@@ -1,83 +1,86 @@
...
-    renderBodySec opt bodies,
-    angles "/html"]
+renderHTML :: HTMLRenderOptions -> HTML -> Doc ann
+renderHTML opt htmlTree =
+  vcat
+    [ "<!DOCTYPE html>",
+      angles "html",
+      indent (indentationSize opt) $ renderHeadSec opt heads,
+      indent (indentationSize opt) $ renderBodySec opt bodies,
+      angles "/html"
```
balacij: Should we use `wrapBlock` for this?

code/drasil-data-formats/test/Spec/Drasil/Data/Formats/HTML.hs
```diff
@@ -27,11 +26,11 @@ blockquoteTag, inputTag :: CustomTag
 blockquoteTag = customTag "blockquote"
 inputTag      = customTag "input"
 
-testRenderOptions :: HTMLRenderOptions
-testRenderOptions = HTMLRO $ M.fromList [
+testGenOptions :: HTMLRenderOptions
+testGenOptions = HTMLBO (M.fromList [
```
balacij: I think we can also undo these changes?

code/drasil-printers/lib/Language/Drasil/HTML2/Helpers.hs
```diff
@@ -0,0 +1,127 @@
...
+ed = RawText " ed., "
+editedBy = RawText "Edited by "
+
+articleTitle, author :: [HTMLBody] -> HTMLBody
+articleTitle t = HTML.Div [Attr "class" "title"] [Heading H1 [] t]
+author a       = HTML.Div [Attr "class" "author"] [Heading H2 [] a]
+
+-- | Create the link to the CSS file
+stylesheet :: Text -> HTMLHead
+stylesheet css = Link "stylesheet" (css <> ".css") [Attr "type" "text/css"]
```
balacij: These look like good smart constructors should be in the core HTML AST file.

balacij: The rest of the code in this file looks like it belongs in the `Render.hs` file?

code/drasil-printers/lib/Language/Drasil/HTML2/README.md
```diff
@@ -0,0 +1,19 @@
...
+  - Defines helper functions for pretty printing HTML
+
+Monad.hs
+  - Defines the helper monad for printing HTML and certain special renderings
+
+Print.hs
+  - Defines the pretty printer for generating the HTML and CSS documents from the HTML AST representation
+
+README.md
+  - This file
```
balacij: I think we can delete this file. Our subfolder READMEs are drift quickly from being synced with main. The upkeep cost is too high (relative to how often we use them) unless we have it automated -- #2857.

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+genHTML rOpts fn (Document t a c) = HTML heads bodies
+  where
+    heads =
+      [ stylesheet (T.pack fn),
+        HTML.Title (printSpec t),
+        Meta [Attr "charset" "utf-8"],
+        inlineScript mathJaxScript,
+        externalScript
+          (T.pack $ mathJaxSrc rOpts)
+          [ Attr "type" "text/javascript",
```
balacij: I think `type="text/javascript"` can be an attribute always added by `externalScript`?

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,161 @@
...
+      dataRows = map ( \(f, d) -> HTML.Row [] [THeader [] [RawText (T.pack f)],
+        TData [] (concatMap (loToHTML rOpts) d)]) ps
+   in [HTML.Table attrs (refRow : dataRows)]
+
+-----------------------------------------------------------------
+------------------BEGIN LIST PRINTING----------------------------
+-----------------------------------------------------------------
+
+-- | Generates lists in HTML.
+makeListHTML :: ListType -> HTMLBody -- FIXME: ref id's should be folded into the li
```
balacij: Maybe we should call this `buildListHtml` or `genListHtml` instead of `makeListHTML`?

balacij: And we should follow a consistent naming convention.

code/drasil-website/lib/Drasil/Website/Core.hs
```diff
@@ -60,4 +62,7 @@ instance ToFiles DrasilWebsite DrasilWebsiteGenOptions where
       pd = makeDocument printSetting $ dw ^. indexDoc
 
       -- 2. Transform the TDL into HTML.
-      html = genHTML "index" pd
+      html = renderHTML (HTMLBO M.empty 2)
+        $ genHTML
+        (HTMLRO htmlBibFormatter "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml-full.js")
+        "index" pd
```
balacij, line 63: Can you please split this up into two steps?

2. Transform the TDL into HTML.
3. Render the HTML to Doc.

balacij (changes requested):

code/stable-website/index.html
balacij: Reminder about this comment.

code/stable-website/index.html
balacij: Huh, interesting. Let's file an issue about fixing that properly later.

code/stable-website/index.html
balacij: Was this addressed?

camishibata:
Ready for review @balacij

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Core.hs
```diff
@@ -129,6 +134,33 @@ span attrs txt = TextFormat Span attrs [RawText txt]
...
+vol = RawText "vol. "
+pg = RawText "pg. "
+pp = RawText "pp. "
+no = RawText "no. "
+ed = RawText " ed., "
+editedBy = RawText "Edited by "
+
+articleTitle, author :: [HTMLBody] -> HTMLBody
+articleTitle t = Div [Attr "class" "title"] [Heading H1 [] t]
+author a       = Div [Attr "class" "author"] [Heading H2 [] a]
```
balacij: These ones don't look general enough to be in `drasil-data-formats`. I would keep them closer to the renderers that use them.

code/drasil-data-formats/test/golden/html/tags.html
balacij: Was a rendering option added?

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,240 @@
...
+import qualified Drasil.Data.Formats.HTML as HTML (Format(..))
+
+-- | Data type that carries functions that vary
+-- for bib printing
+data BibFormatter = BibFormatter
+  { -- | Emphasis (italics) rendering
+    emph :: [HTMLBody] -> [HTMLBody],
+    -- | Spec rendering
+    spec :: Spec -> [HTMLBody]
+  }
```
balacij: Is `BibFormatter` strictly necessary for the HTML renderer?

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,240 @@
...
+renderCite f (Cite e Article cfs)   = ([RawText $ T.pack e],
+  renderF cfs (useStyleArtcl f)  ++ [RawText (T.pack $ sufxPrint cfs)])
+renderCite f (Cite e MThesis cfs)   = ([RawText $ T.pack e],
+  renderF cfs (useStyleBk    f)  ++ [RawText (T.pack $ sufxPrint cfs)])
+renderCite f (Cite e PhDThesis cfs) = ([RawText $ T.pack e],
+  renderF cfs (useStyleBk    f)  ++ [RawText (T.pack $ sufxPrint cfs)])
+renderCite f (Cite e Misc cfs)      = ([RawText $ T.pack e],
+  renderF cfs (useStyleBk    f))
+renderCite f (Cite e _ cfs)         = ([RawText $ T.pack e],
+  renderF cfs (useStyleArtcl f)) --FIXME: Properly render these later.
```
balacij: What is this TODO note for?

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,166 @@
...
+          (T.pack $ mathJaxSrc rOpts)
+          [ Attr "id" "MathJax-script",
+            Attr "async" ""
+          ]
+      ]
+    bodies =
+      [ articleTitle (specToHTML t),
+        author (specToHTML a)
+      ]
+        ++ concatMap (loToHTML rOpts) c
```
balacij, line 51: This is exactly the kind of code we want to see! Much cleaner!

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
```diff
@@ -0,0 +1,166 @@
...
+    attrs = [Attr "class" (T.unwords $ map T.pack ts)]
+    headerRow = HTML.Row [] (map (THeader [] . specToHTML) l)
+    dataRows = map (HTML.Row [] . map (TData [] . specToHTML)) lls
+    tableNode = HTML.Table attrs (headerRow : dataRows)
+    captionNode = HTML.Paragraph [Attr "class" "caption"] (specToHTML t)
+    wrapperAttrs = [Attr "id" (printSpec r)]
+
+-----------------------------------------------------------------
+------------------BEGIN DEFINITION PRINTING----------------------
+-----------------------------------------------------------------
```
balacij: I think you can remove the block comments like this one.

code/drasil-printers/lib/Language/Drasil/HTML2/Spec.hs
```diff
@@ -0,0 +1,149 @@
...
+  Attr (..), Format (Emphasis, Span, Subscript, Superscript), HTMLBody (..), customTag
+  )
+import qualified Drasil.Data.Formats.HTML as HTML (Format (Bold))
+import qualified Language.Drasil.TeX.Print as TeX (pExpr)
+import Language.Drasil.Markdown.Print (printMath)
+
+-- | Transforms the Sentences ('Spec's) into Text
+printSpec :: Spec -> Text
+printSpec (S s) = T.pack s
+printSpec (E e) = T.pack $ show $ printMath $ TeX.pExpr e
```
balacij: Is that `show` necessary?

balacij (changes requested):

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
camishibata: `Cite` can have the following types:
```
data CitationKind = Article
                  | Book
                  | Booklet
                  | InBook
                  | InCollection
                  | InProceedings
                  | Manual
                  | MThesis
                  | Misc
                  | PhDThesis
                  | Proceedings
                  | TechReport
                  | Unpublished 
```
But `renderCite` only implements an specific rendering for five of them.

code/drasil-printers/lib/Language/Drasil/HTML2/Print.hs
camishibata: Implemented in 705bc5ad002d1864e8d29dcf91b65cc766c448b6

code/stable-website/index.html
camishibata: Fixed in 36288af9e7a1cc3046395aa70d668523c9993f78

code/drasil-printers/lib/Language/Drasil/HTML2/Helpers.hs
camishibata: Splitted and deleted `Helpers.hs` between `Render.hs`, `Spec.hs` and `Citation.hs` in 25090f39bf3f0b52777a07509b3c581d44d157aa

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
camishibata: Implemented in 25090f39bf3f0b52777a07509b3c581d44d157aa

code/drasil-website/lib/Drasil/Website/Core.hs
camishibata: Fixed in 25090f39bf3f0b52777a07509b3c581d44d157aa

code/drasil-printers/lib/Language/Drasil/HTML2/Spec.hs
camishibata: Addressed in 6258ace7fe8956afc0ac5783d2fb80a373b2c3da

code/drasil-printers/lib/Language/Drasil/HTML2/Render.hs
camishibata: Fixed in 6258ace7fe8956afc0ac5783d2fb80a373b2c3da

code/drasil-data-formats/lib/Drasil/Data/Formats/HTML/Core.hs
camishibata: Fixed in 6258ace7fe8956afc0ac5783d2fb80a373b2c3da

code/drasil-data-formats/test/golden/html/tags.html
camishibata: As addressed in https://github.com/JacquesCarette/Drasil/pull/5319#discussion_r3619037881, this might be done in another PR.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
camishibata: `BibFormatter` removed in 9599fc0620cc51c1754851f28e000df410fb484d

balacij (approved):
Latest changes look great to me! Thank you @camishibata !

balacij:
@JacquesCarette This one is ready for review.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,227 @@
...
+
+-- | Makes a bilbliography for the document.
+printBib :: BibRef -> HTMLBody
+printBib bib =
+  DescriptionList [Attr "class" "reference-list"] (concatMap renderCitation bib)
+  where
+    renderCitation :: Citation -> [DItem]
+    renderCitation cite@(Cite e _ _) =
+      let (termDoc, detailsDoc) = renderCite cite
+          termHTML = [RawText "[", TextFormat HTML.Bold [] termDoc, RawText "]"]
```
JacquesCarette, line 41: should abstract out things like `RawText "["` for all single characters

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,227 @@
...
+       in [DTerm [Attr "id" (T.pack e)] termHTML, DDetails [] detailsDoc]
+
+-- | For when we add other things to reference like website, newspaper
+renderCite :: Citation -> ([HTMLBody], [HTMLBody])
+renderCite (Cite e Book cfs)      = ([RawText $ T.pack e],
```
JacquesCarette, line 46: abstract out `RawText $ T.pack`

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,227 @@
...
+-- | Generates citation as an article style.
+useStyleArtcl :: StyleGuide -> (CiteField -> [HTMLBody])
+useStyleArtcl MLA     = artclMLA
+useStyleArtcl APA     = artclAPA
+useStyleArtcl Chicago = artclChicago
+
+-- | Cite books in MLA format.
+bookMLA :: CiteField -> [HTMLBody]
+bookMLA (Address   s) = specToHTML s ++ [colon]
+bookMLA (Edition   s) = [RawText (T.pack (show s ++ sufxer s)), ed]
```
JacquesCarette, line 126: should not mix `Text` and `String` !

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
```diff
@@ -0,0 +1,227 @@
...
+artclAPA :: CiteField -> [HTMLBody]
+artclAPA (Title  s)  = specToHTML s <> [RawText ". "]
+artclAPA (Volume n)  = [emphasis [] (T.pack (show n))]
+artclAPA (Number  n) = [RawText (", (" <> T.pack (show n) <> ") ")]
+artclAPA i           = bookAPA i
+
+-- | Cite articles in Chicago format.
+artclChicago :: CiteField -> [HTMLBody]
+artclChicago i@(Title    _)  = artclMLA i
+artclChicago (Volume     n)  = [RawText (T.pack (show n)), comma]
```
JacquesCarette, line 181: `Text` has ways to show numbers directly.

JacquesCarette (changes requested):

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
balacij: A particularly good one to abstract out. Created `rawText :: Text -> HTMLBody` and `rawText' :: String -> HTMLBody` in a645b00979428553c735246388b93b1b3403cf65.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
balacij: I find this one a bit dubious.

```haskell
lbrak :: HTMLBody
lbrak = RawText "]"
```

How reusable is this? From the POV of someone writing HTML code, I don't see why this would be shipped with the HTML library I'm pulling in. It's not that it's expensive to maintain, it's just that it seems like something oddly specific. It's also as if someone can tell you're going to use it for something hyper-specific, but for general HTML authorship, brackets are not that common. However, if we had `lbrak :: Spec` (prose formatted in a slightly better form than `Sentence` for final rendering), I think that would be completely reasonable!

I think that `Citation` should have been erased earlier than this (into `Spec` sentences) as I mentioned in #5393 (and you [questioned](https://github.com/JacquesCarette/Drasil/pull/5393#discussion_r3677403892) [but I haven't responded yet...]).

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
balacij: Reduced as much as possible now. The few things that are strings are clearly noted through `rawText'`. Those will require fixing things 'up'.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
balacij: I couldn't find one. For now, I pulled in the [`TextShow` library](https://hackage-content.haskell.org/package/http-api-data-0.7/docs/Web-Internal-HttpApiData.html#v:showt) and restricted us to only using it on numbers: https://github.com/JacquesCarette/Drasil/blob/9a91627f7cf8c7eb23c5dc16909dcc0964c62114/code/drasil-utils/lib/Data/Text/Extras.hs#L14-L15

balacij:
There are several more passes we can take at refining the code, but I think we should push that work to other PRs:

1. The "citation" rendering logic can be abstracted further, but I'm currently seeing too many `stable/` `diff`s, so I think this is out of scope for now.
2. The "JSON" and "Markdown" renderers use HTML generation, but they rely on some hacks for escaping twice or just use the HTML when they should use Markdown. For now, I've stripped the old HTML renderer down to only the parts the JSON and Markdown renderers need, and moved them to said namespaces. I think we should work on these in separate PRs but only while we rework their own code as well.
3. Documentation: There's a lot of work that can be done on documentation alone. From ensuring Haddock formatting to writing better descriptions and ensuring consistent formats, I don't think that's worth the effort right now, at least not until more work has been done on #4989.
4. The CSS can be refined, perhaps to make the HTML look prettier as well. Out of scope for this PR.
5. Exploring more generation options. Everywhere there is a stylistic decision, there is an option to be created. Again, out of scope for this PR.

So, I think this is ready for review again. We're reaching a point of diminishing returns on this code, at least without expanding scope.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
JacquesCarette: Do notice that pretty much all pretty-printing libraries define such abbreviations... The point is that these can let you put all of the implementation details behind an abstraction barrier. That then lets you changes the underlying representation more easily.

I do agree that by the time we hit *HTML*, `Citation` should have already been rendered into an abstract layout form.

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
JacquesCarette: `Data.Text.show`

balacij: 🤦 I don't know how I didn't see that.

Fixed in 6a93cecff. Thank you!

JacquesCarette: It is way, way, *way* at the bottom of the `Data.Text` doc!

code/drasil-printers/lib/Language/Drasil/HTML2/Citation.hs
balacij: What do you think about 0faed3fce19a586c6cde19cd507a808725164e01? As we get rid of `String` and `PLegacy`, we will be getting rid of `Text -> String` operations as well.

Edit: Ah, CI failed. I'll fix that first.

balacij: Fixed that work in 6e99b7fd4 (merging in with the previous). We might benefit from having a `drasil-prettyprint` library that exports some tools. The code in 6e99b7fd4 is probably reusable for `drasil-data-formats` as well. Perhaps GOOL too.

JacquesCarette: I like it!

JacquesCarette:
Branch now has conflicts...

balacij:
Merge conflict resolved!

Note: Since this PR is large (37 commits), I opted to rebase. I think this is one scenario where having a linear git history is "cleaner" than not.

JacquesCarette (approved):
The code is good - but there's again conflicts.

balacij:
Nice. Resolved again. Very easy this time -- just needing to update `stable/`.

JacquesCarette:
Sorry, another conflict has since arisen!

balacij:
Merge conflict resolved.

PR #5320, opened 2026-07-15
Changed `StateVar` to parameter with fundep

B-rando1:
Contributes to #5233


JacquesCarette (approved):
Is this the last one?

B-rando1:
There's still `Attachment`, `Program`, and `MethodType`. Getting there though!

PR #5322, opened 2026-07-15
Small progress on Idealized Process

JacquesCarette:
Plug in the Idealized Process page in to the Sidebar. Add some items to Terminology. Take some of the discussion from the PR and add it to the Idealized-Process page itself.


smiths (approved):
I think we should merge this documentation on the idealized process.  I'm sure we will continue to make edits, but this is a nice start.

PR #5324, opened 2026-07-15
Removed `MethodType` type family

### LINKED ISSUE #5233: Removing Type Families from GOOL

B-rando1:
Last one!

Builds on #5323 
Closes #5233

### DIFF
--- a/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CSharpRenderer.hs
+++ b/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CSharpRenderer.hs
@@ -621,7 +621,6 @@ instance VisibilityElim CSharpCode Doc where
   visibility = unCSC
 
 instance MethodTypeSym CSharpCode where
-  type MethodType CSharpCode = TypeData
   mType = zoom lensMStoVS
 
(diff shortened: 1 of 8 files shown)


JacquesCarette (approved):
Wow, this one was (now) easy!

PR #5326, opened 2026-07-15
Removed `NativeVector` from `ProcProg`

B-rando1:
I had noticed that `NativeVector` was added to `ProcProg`. This is handy for now, since both of our procedural languages have native vector support; but long-term and semantically it doesn't make sense to assume that. This PR removes `NativeVector` from `ProcProg` and updates the missing type constraints.

@balacij you had mentioned that vectors in `drasil-code` might need to be reworked. Will this PR make that more difficult? We can close it and come back to it later if that is the case.


JacquesCarette (approved):
Code looks good, but has tons of conflicts now.

balacij:
@B-rando1 Nope, this won't make that work any more difficult. Thank you for asking 😄

B-rando1:
Conflicts resolved.

JacquesCarette (approved):

PR #5333, opened 2026-07-17
Rename "Typical Value" to "Reasonable Value"

djarabek:
This reflects the terminology used in the SRS output.


JacquesCarette:
I'm going to wait on this until @smiths has chymed in on what our intent is, so that we can pick the right terminology.

smiths:
I prefer reasonable value as our terminology.  A reasonable value is one that is physically admissible.  That is all that we promise.  A typical value will be reasonable, but it will also be a common case.  In our examples, we don't know well enough whether values are typical.  That is too high a bar to achieve.  I like the [discussion provided by ChatGPT](https://chatgpt.com/share/6a5fa9f2-6aec-83ea-85fb-7941446b51be).

smiths (approved):
The changes to stable look good to me.  @JacquesCarette still has to look at the changes to the Drasil code.

JacquesCarette (approved):
Got to love generation. 3 tiny, straightforward changes to the code, 40 files changed (correctly). And it is *text* changes - not code in sight.

PR #5335, opened 2026-07-18
Fix inconsistent HTML block tag rendering by replacing cat with vcat

### LINKED ISSUE #1173: Rendering small tags in single line using 'cat'
When `cat` is used instead of `vcat` inside `wrap`, the small tags are rendered in single line. But in some places where the tags can not be rendered in single line, it is splitiing up into multiple lines.

This is the screenshot of tags rendering in single line.
sl

These are the screenshots of the tags which are not rendered in single line.
nsl1

nsl2

nsl3

Purav-Kanda:
Closes #1173

Block-level HTML tags were rendered on a single line when content was small (cat tries horizontal first), then multi-line when content was large. Replacing cat with vcat in wrapGen ensures all block-level tags render consistently with proper indentation. Inline elements (wrap' using hcat) are unaffected.


JacquesCarette:
You should give an explanation when you close a PR. Also, we try hard to not force-push.

Purav-Kanda:
Closed by mistake while cleaning up — sorry for the confusion. Reopened as #5336 with the same fix and a clean commit history.

PR #5336, opened 2026-07-18
Fix 1173 wrap cat vcat

### LINKED ISSUE #1173: Rendering small tags in single line using 'cat'

Purav-Kanda:
In Helpers.hs, wrapGen used cat as the separator function, which collapses
block-level HTML tags (div, p, tr, td, th, ul, ol, li, etc.) onto a single
line when content is short, and only splits them across lines when forced.
This caused inconsistent formatting depending on content length.

Replacing cat with vcat in wrapGen ensures block tags always render with
the opening tag, indented content, and closing tag on separate lines,
regardless of content length.

Also removed the now-unused import of cat from Text.PrettyPrint.

Closes #1173"


JacquesCarette:
Actually, we don't really want things to be displayed vertically all the time. We want the HTML to be laid out more like what a human would do. The real bug (which admittedly is not so clear on #1173 ) is that the `<em>` tag is on multi-line! It fits on one line just fine, so it should be.

Purav-Kanda:
Reworked based on your feedback. The fix is now targeted only at inline HTML tags (em, sup, sub, bold) rather than changing block-level rendering.
The approach: a new renderInline helper uses renderStyle (style { mode = OneLineMode }) to collapse any multi-line Doc to a single line before wrapping in the HTML tag. This prevents <em> from going multi-line when wrapping TeX matrix expressions (e.g. bmatrix), which was the root cause of #1173.
wrapGen is reverted back to cat — block-level tags are unaffected. Only the Projectile SRS stable file changes, since it's the only example with matrix expressions inside inline tags.

Purav-Kanda:
Sorry for the messy commit history — went through a few iterations getting the stable files right. The net change is just the 2 files in "Files changed".

balacij:
Note that we are rewriting the HTML renderer in #5319.

We can't fix this issue as it stands right now without that work because `em` nodes should _strictly_ be inlined when its only inner HTML content is purely text. If there is anything other than text (i.e., an HTML block), it should be broken up.

Writing a (golden) test case for our new HTML type that checks #1173 is handled correctly, would be a great alternative way to close #1173.

Thank you for highlighting this issue @Purav-Kanda !

Purav-Kanda:
Thanks for the context! I'll close this in favor of the work in #5319. I'll look into writing a golden test case for the new HTML type that covers the #1173 case.

PR #5339, opened 2026-07-20
Made `stateVar` require default values

### LINKED ISSUE #2200: Alternative to typeDfltVal for GOOL's Swift renderer
I implemented a function in GOOL's swift renderer, `typeDfltVal`, which returns a default value based on a given type. Using default values in this way is something we would like to avoid. The reason for this function is because all Swift state variables require an initial value, but GOOL's `stateVar` function does not accept a value as a parameter because other, non-Swift languages, do not need initial values. We should think about whether we can remove `typeDfltVal` and use some other solution. See further related discussion here: https://github.com/JacquesCarette/Drasil/pull/2197#commitcomment-40209217

B-rando1:
Closes #2200 

I wasn't sure where to put `typeToDefaultValue`. `ConceptMatch` sort of makes sense, but it was matching two different types of concepts before. @balacij any better ideas?

### DIFF
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/ConceptMatch.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/ConceptMatch.hs
@@ -1,6 +1,6 @@
 -- | Contains functions related to the choice of concept matches.
 module Language.Drasil.Code.Imperative.ConceptMatch (
-  chooseConcept, conceptToGOOL
+  chooseConcept, conceptToGOOL, typeDefaultValue
 ) where
 
 import Prelude hiding (pi)
@@ -9,7 +9,8 @@ import Control.Monad.State (State, modify)
 
 import Drasil.Database (UID)
 import Language.Drasil (Sentence(S), (+:+), (+:+.))
-import Drasil.GOOL (SValue, MathConstant(..))
+import Drasil.GOOL (SValue, MathConstant(..), OOTypeSym, CodeType(..),
+  Literal(..), convTypeOO)
 
 import Language.Drasil.Choices (Choices(..), CodeConcept(..),
     MatchedConceptMap, showChs, Maps(..))
@@ -31,3 +32,17 @@ chooseConcept chs = sequence $ Map.mapWithKey chooseConcept' (conceptMatch $ map
 -- | Translates a 'CodeConcept' into GOOL.
 conceptToGOOL :: (MathConstant r) => CodeConcept -> SValue r
 conceptToGOOL Pi = pi
+
+-- | Gives a default value for a given type
+typeDefaultValue :: (Literal r, OOTypeSym r) => CodeType -> SValue r
+typeDefaultValue Boolean = litFalse
+typeDefaultValue Integer = litInt 0
+typeDefaultValue Float = litFloat 0.0
+typeDefaultValue Double = litDouble 0.0
+typeDefaultValue Char = litChar ' '
+typeDefaultValue String = litString ""
+typeDefaultValue (List t) = litList (convTypeOO t) []
+typeDefaultValue (Array t) = litArray (convTypeOO t) []
+typeDefaultValue (Set t) = litSet (convTypeOO t) []
+typeDefaultValue (Reference t) = typeDefaultValue t
+typeDefaultValue t = error $ "Attempt to get default value for type with none: " ++ show t
--- a/code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
+++ b/code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
@@ -28,7 +28,8 @@ import Drasil.Database (UID, HasUID(..), IsChunk)
 import Language.Drasil (HasSymbol, HasSpace(..),
   Space (Rational, Real), RealInterval(..), Constraint(..), Inclusive (..))
 import Language.Drasil.Code.Imperative.Comments (getCommentBrief)
-import Language.Drasil.Code.Imperative.ConceptMatch (conceptToGOOL)
+import Language.Drasil.Code.Imperative.ConceptMatch (conceptToGOOL,
+  typeDefaultValue)
 import Language.Drasil.Code.Imperative.GenerateGOOL (auxClass, fApp, fAppProc,
   ctorCall, genModuleWithImports, genModuleWithImportsProc, primaryClass)
 import Language.Drasil.Code.Imperative.Helpers (convScope)
@@ -66,7 +67,7 @@ import Drasil.GOOL (Label, MSBody, MSBlock, SVariable, SValue, CSStateVar,
   OOStatement)
 import qualified Drasil.GOOL as S (Set(..)) -- TODO [Brandon Bosman, 07/09/2026]: Merge this with OO
 import qualified Drasil.GOOL as OO (SFile)
-import qualified Drasil.GOOL as C (CodeType(List, Array))
+import qualified Drasil.GOOL as C (CodeType(..))
 import Drasil.GProc (ProcProg, NativeVector(..))
 import qualified Drasil.GProc as Proc (SFile)
 import Drasil.System (systemdb)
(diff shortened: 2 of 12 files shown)


balacij:
`ConceptMatch.hs` is for a different kind of "concept" mapping.

Why not keep this in `drasil-gool`? This seems like something that GOOL should be able to automatically handle on its own internally, no?

For example, if a user (of GOOL) is defining a variable with no assignment (`stateVar`), GOOL can automatically default to a specific value (i.e., `stateVar ... = stateVarDecDef ... x` for Swift [and probably C too]).

B-rando1:
In his [comment](https://github.com/JacquesCarette/Drasil/issues/2200#issuecomment-1530322958), @JacquesCarette made the argument that we shouldn't assume that one particular value is always safe to use as a default value for a given type. There are also special cases like objects, where "default value" is much more subjective, that GOOL is less suited to handle. Pushing the decision up the stack to `drasil-code` gives us more control.

It does work as is (currently things work the way you suggested in the last paragraph), but if we ever wanted to have a class that holds a reference to another class, then the current implementation would throw an error because the Swift renderer doesn't know what to do in that case.

balacij:
Aside: for the default `char`, [`\u0000`](https://www.fileformat.info/info/unicode/char/0000/index.htm) might be more appropriate.

> In his https://github.com/JacquesCarette/Drasil/issues/2200#issuecomment-1530322958, @JacquesCarette made the argument that we shouldn't assume that one particular value is always safe to use as a default value for a given type. There are also special cases like objects, where "default value" is much more subjective, that GOOL is less suited to handle.
> 
> It does work as is (currently things work the way you suggested in the last paragraph), but if we ever wanted to have a class that holds a reference to another class, then the current implementation would throw an error because the Swift renderer doesn't know what to do in that case.

I continue to return to the question: what is GOOL and what do we want from it? Yes, it is an overlanguage, but what is it we're really defining? A specification would go a long way here.

In the case of declaring any variable (whether they be "state" variables or not), having a default value for every type (primitive or not) is not unheard of. For example:

* [C# has defaults](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values), and
* [Java has defaults](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html).

However:

* C/C++ does not (up to compilers),
* with python, the question does not make sense, and
* with Swift, correcting ourselves (in what appears to be the original issue that brought all of this up), [variables can be initialized without assignments](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/) (see https://www.kodeco.com/books/swift-cookbook/v1.0/chapters/1-declare-variables-in-swift for more examples, the rule however is that **variables must be assigned before accessed** and that class types, when constructed, must be fully initialized, so **members of a class must be initialized by the end of the initializer/constructor**).

Personally, I have no issue with C#/Java's rule but I have a slight preference to Swift's policy because I prefer explicitness. The default values that C# and Java have sensible rationale, _but_ they do make code a bit harder to read. For that reason, I find Swift's approach best.

Regarding the subjectivity of default values, specifying what happens through the documentation of GOOL is probably best (both in documentation of `stateVar` and the documentation of the various typeclasses that correspond with specific GOOL-supported types). That way, we have no more ambiguity. We should definitely avoid the C/C++ approach.

Aside on GOOL: how do we view `null`? It's an idiomatic thing to use in C#, Java, and other languages (e.g., Go, whether we like it or not). Do we want to support generating it with GOOL? For Swift, we'd need to use optionals, for example, but for the rest (and for non-primitive types, excl. Python), `null` is native.

Taking inspiration from Swift for a variable initialization-before-usage rule in GOOL is probably best from an idiomatic POV. I don't know if that's possible to implement right now, unfortunately.

> Pushing the decision up the stack to drasil-code gives us more control.

The choice of specific default values can greatly impact program semantics. I don't know if we should ever think of this as a choice. We can think of showing/hiding the default values as a good generation choice, however.

balacij:
Swift's rule regarding variable initialization is very sensible. Why don't we just impose that on GOOL instead of having default values? I don't understand the initial problem/solution...

The rules we can impose on GOOL (perhaps not right now):

1. Before any variable is accessed, it must have been assigned.
2. By the end of any constructor of a class, all variable members must have been assigned to.

Where is a GOOL-internal default value for each GOOL-supported type necessary?

code/drasil-code/lib/Language/Drasil/Code/Imperative/ConceptMatch.hs
```diff
@@ -31,3 +32,17 @@ chooseConcept chs = sequence $ Map.mapWithKey chooseConcept' (conceptMatch $ map
 -- | Translates a 'CodeConcept' into GOOL.
 conceptToGOOL :: (MathConstant r) => CodeConcept -> SValue r
 conceptToGOOL Pi = pi
+
+-- | Gives a default value for a given type
+typeDefaultValue :: (Literal r, OOTypeSym r) => CodeType -> SValue r
```
JacquesCarette, line 37: I would rather do things such that this function would have no need to exist. It's a "code smell" that someone is generating code that requires a 'default value'.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -66,7 +67,7 @@ import Drasil.GOOL (Label, MSBody, MSBlock, SVariable, SValue, CSStateVar,
   OOStatement)
 import qualified Drasil.GOOL as S (Set(..)) -- TODO [Brandon Bosman, 07/09/2026]: Merge this with OO
 import qualified Drasil.GOOL as OO (SFile)
-import qualified Drasil.GOOL as C (CodeType(List, Array))
+import qualified Drasil.GOOL as C (CodeType(..))
```
JacquesCarette, line 70: We should use `..` only when we know for sure that we want all the constructors. Being explicit is usually better.

code/drasil-code/test/GOOL/Observer.hs
```diff
@@ -32,7 +32,7 @@ selfX = instanceVarSelf x
 -- | Helper function to create the class.
 helperClass :: (ClassSym r vis smt md svr att, IOStatement r smt, Literal r,
   OOVariableValue r) => SClass r
-helperClass = buildClass Nothing [stateVar public instanceLevel x]
+helperClass = buildClass Nothing [stateVar public instanceLevel x (litInt 0)]
```
JacquesCarette, line 35: Why is a starting value of `0` meaningful?

code/drasil-code/test/HelloWorld.hs
```diff
@@ -377,7 +377,7 @@ helloTryCatch = tryCatch (oneLiner (throw "Good-bye!"))
 
 helloWorldClass :: (OOProg r vis smt md svr att prg) => SClass r
 helloWorldClass = extraClass "TestClass" Nothing
-  [stateVar public instanceLevel (var "a" int)]
+  [stateVar public instanceLevel (var "a" int) (litInt 0)]
```
JacquesCarette, line 380: Why is a starting value of `0` meaningful?

code/drasil-code/test/OOVector.hs
```diff
@@ -10,7 +10,7 @@ ooVector = prog "OOVector" "" [fileDoc (buildModule "OOVector" []
 
 vectorClass :: OOProg r vis smt md svr att prg => SClass r
 vectorClass = docClass "Vectors of doubles and common vector-related operations." $
-  extraClass "Vector" Nothing [stateVar private instanceLevel localV]
+  extraClass "Vector" Nothing [stateVar private instanceLevel localV (litArray double [])]
```
JacquesCarette, line 13: Isn't an empty array the opposite of meaningful?

JacquesCarette (changes requested):

JacquesCarette:
> how do we view null?

I'm fine with *generating* `null`. I don't want `null` in the interface to GOOL.

B-rando1:
Jason and I were on a call talking about this, and I'm no longer sure if this is really the correct approach.

In the original [comment](https://github.com/JacquesCarette/Drasil/pull/2197#issue-646719362) (second bullet point) Brooks says that class variables need an initial value. @balacij looked it up and apparently that's not true; variables just need to be initialized before they're used. That seems to change the question quite a bit, since choosing "where do we decide what default values are used" assumes that we need default values. Like Jason said in his [comment](https://github.com/JacquesCarette/Drasil/pull/5339#issuecomment-5026876465), we could instead have GOOL enforce Swift's "initialize before accessing" rule.

JacquesCarette:
Right - so the code we *generate* should always 'initialize before access'. The easiest way to insure that is for the level above to provide the initial value at the same time as the declaration information.

balacij:
So I guess we're closing this PR then?

B-rando1:
Yeah, it's too simplistic of an approach. (for future reference, see my [comment](https://github.com/JacquesCarette/Drasil/issues/2200#issuecomment-5038430003) for why)

PR #5340, opened 2026-07-21
chore: refine imports in `Language.Drasil.Document.SentenceCombinators`

balacij:
I just need the [module dep. graph](https://jacquescarette.github.io/Drasil/graphs/drasil-lang.pdf) to be a bit clearer.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
+++ b/code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
@@ -23,13 +23,19 @@ import Control.Lens ((^.))
 import Data.Decimal (DecimalRaw, realFracToDecimal)
 import Data.List (transpose)
 
-import Language.Drasil (ConceptChunk, DefinesQuantity(defLhs) , UnitDefn, MayHaveUnit(..)
-  , HasUnitSymbol(usymb), Quantity, Concept, Definition(defn), NamedIdea(..)
-  , HasShortName(..) , short, atStart, titleize, phrase, plural
-  , ModelExpr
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
Whoa, I never noticed that something inside `drasil-lang` used `Drasil.Lang` ! That is definitely not good. I wonder when I let that slip through. Sigh.

JacquesCarette:
The only upside is that that particular dependency graph made is super clear that the whole `Language.Drasil.Document` graph was separate from the rest of `drasil-lang` so that it could be moved elsewhere quite easily. Time for a `drasil-document` package?

balacij:
> Whoa, I never noticed that something inside drasil-lang used Drasil.Lang ! That is definitely not good. I wonder when I let that slip through. Sigh.

I think one of them was my fault. Not a big deal. We're fixing it now. I also just made #5341.

> The only upside is that that particular dependency graph made is super clear that the whole Language.Drasil.Document graph was separate from the rest of drasil-lang so that it could be moved elsewhere quite easily. Time for a drasil-document package?

I would really like to do that. We need a shallow file analysis on the files related to "documents" in `drasil-lang` first. I'm fairly confident there are a few other files that belong in the `Language.Drasil.Document` namespace, too.

Additionally, if we're not careful, we will end up with a cyclic dependency between `Sentence` (`drasil-lang`) and this prospective new package. That is, when we switch the `Ref` constructor of `Sentence` from `UID` based chunk references to `UIDRef`-based ones.

JacquesCarette:
There were 5-6 files at the top of the dependency graph that could freely be moved.

I agree that some other parts are "held captive" under `Sentence` right now. So the thing to do would be not move them yet!

Issue #5328, opened 2026-07-16
GOOL: Remove `repr` Types


B-rando1:
I was talking to @TOTBWF a couple days ago, and he suggested that as a next step with the GOOL refactor we remove the `repr` wrapper types from the languages. They were originally used for the type families to dispatch off of, but with fundeps we can use other methods. The benefit of this is that apparently `DerivingVia` isn't able to work under the `newtype`, so removing it should let us simplify some of our existing implementations.

I'm not completely sure how to start, but hopefully I can talk to Reed about it more soon.

JacquesCarette:
The `repr` type was going to be most useful if we ever went to a "typed GOOL" and wanted to use tagless techniques to get there. But we can probably use other means to do so.

B-rando1:
I read through #1869, it sounds like the lack of types in `Expr` was the primary problem at the time. I'm not sure if that's still an issue, but there was hope of eventually getting typed GOOL to work. I don't know if it'll be something I get to with my work, but I think that typed GOOL is a worthwhile goal.

I don't have the firmest grasp of the subtleties of tagless, but but we should try to leave the door open to doing typed GOOL in some way.

JacquesCarette:
The branch is still around, as far as I know.

B-rando1:
I had a quick call with Reed about this. The path forward seems much clearer now. Instead of having higher-kinded `newtype`s for each language, each language can wrap `ProgData` in a newtype; and then that can drive the functional dependencies. This will let us remove a lot of the boilerplate from our implementations, and make things like `LoggingFor` much cleaner and simpler to write.

We also talked about opening the door to a "Typed GOOL". His opinion is that GOOL isn't well-suited to the Typed-Tagless approach, I think because Typed-Tagless works best when the host and target languages' type systems are mostly aligned. He believes that a better idea would be to implement a generation-time type-checker.

JacquesCarette:
I agree that trying to convince Haskell's type-checker to type GOOL would involve a lot of gymnastics that might be more 'acrobatic' than we really want. So that a type-checker would likely be quite a bit easier.

Having said that, we can certainly help that type-checker *a lot* by having things coming from Drasil be richly annotated (so that we indeed do mostly checking and very little inference).

B-rando1:
I'm still working to chart a pathway for this refactor. There seem to be a few intermediate steps we'll have to take. With some help from Reed, I got `hls` to be happy with `BodySym` by changing it from:
```hs
[...]

PR #5342, opened 2026-07-21
Removed `SFile`

B-rando1:
Contributes to #5328


JacquesCarette (approved):
Good thing that I know that what looks like a "local worsening" of this code is for a good cause that will later cause great global improvements!

PR #5351, opened 2026-07-21
refactor: make `Quantity` a constraint alias instead of a unique typeclass

balacij:
Reduces code size a little bit, and gets rid of empty typeclass instances.


JacquesCarette (approved):
Well spotted. I think that when we started Drasil, GHC's then-current version didn't have that feature (or it didn't work nicely).

PR #5363, opened 2026-07-23
refactor(`genMDBook`): make process clearer by reorganizing code

balacij:
The `mdBook` generation was a little bit confusing. The code was split up into too many functions and that ended up making the code difficult to read.


JacquesCarette (approved):
lol, "too many functions" is so rarely a problem!

But sure, since all that stuff was private anyways, the new code is indeed clearer.

PR #5364, opened 2026-07-23
refactor(`layLabelled`): use `layUnlabelled` when no label-creation is possible

balacij:
Code reuse!


JacquesCarette (approved):
Well spotted.

PR #5367, opened 2026-07-23
Split up `IOStatement`

### LINKED ISSUE #5362: GOOL: Split up `IOStatement`
As discussed in #5356, `IOStatement` is quite large and unfocused. We should split it up into multiple classes, based on the spaces between the methods where it is defined.

B-rando1:
Closes #5362 

The names I chose are `PrintConsole`, `ReadConsole`, `FileHandling`, `PrintFile`, and `ReadFile`. Hopefully they're helpful names; let me know if any of them could be better.


JacquesCarette (approved):
Those names definitely work for now. We might figure out better names later when/if we end up breaking these into different pieces.

PR #5370, opened 2026-07-23
Removed `Visibility` from shared interface

### LINKED ISSUE #5360: Remove `vis` type parameter from `MethodSym`
This is continued from [comment](https://github.com/JacquesCarette/Drasil/pull/5356#discussion_r3634648367)s in #5356. Right now, `function` takes a `vis`, i.e. public/private, parameter. This is only used by the Java and C# renderers, which translate `function` into class-level methods. I don't think visibility really has anything to do with functions, and we shouldn't include it just because 2 of our languages support it.

This should also let us move `VisibilitySym` into InterfaceGOOL, which would be nice.

B-rando1:
Closes #5360 

Posting this because it's already done, but I'm not actually sure if this is the way to go.

I noticed right at the end (from changes to stable) that Swift actually does have `private` functions. That got me thinking, and I realized that with languages that have them, explicit export lists are basically a way of marking functions as public. And there are definitely lots of functions, especially in procedural languages, that are meant for use only within their file. So it might be good to keep visibility in `MethodSym`, and refine its use a little.


B-rando1:
(Continuing to think about this)

We should answer the question of whether function "visibility" is the same as class member "visibility". `private` class members are not accessible outside the class, even if they're within the same file. `private` functions, on the other hand, are always accessible within the file/module they're defined in.

It's a bit tricky because from what I've seen, procedural languages are much less consistent about if and how you can restrict imports between modules, and what a "module" is, than OO languages are about classes and accessing their members. But it _might_ make more sense to have separate concepts for "class member visibility" and "module member importability".

JacquesCarette:
I agree that we should split those concepts. When it is that easy to verbally separate them, it's a solid hint that they mean something different.

B-rando1:
Closing. There's clearly much more subtlety to be discussed in #5360, and the solution will look different.

PR #5368, opened 2026-07-23
Refactor unwrap into chWithUnit, unitInParen, unitSym (closes #5354)

Purav-Kanda:
Add unitSym, unitInParen, chWithUnit combinators to SentenceCombinators. Replace all external uses of unwrap in swhs and swhsnopcm. Remove unwrap from exports.

Issue #5354, opened 2026-07-21
Refactor all code of general form `phrase tempInit, ch tempInit, sParen (unwrap $ getUnit tempInit)` with a (new) sentence combinator that properly introduces variable symbols with their units


balacij:
`phrase tempInit, ch tempInit, sParen (unwrap $ getUnit tempInit)` does 3 things:

1. Introduces `tempInit` (the term; `phrase tempInit`).
2. Introduces `tempInit` (the symbol; `ch tempInit`).
3. Introduces the unit of `tempInit` (`sParen (unwrap $ getUnit tempInit)`).

We should create a smart constructor that introduces variables nicely (opportunistically showing units where appropriate) and then we should delete the `unwrap` function (because there is a better design in all its usage):

https://github.com/JacquesCarette/Drasil/blob/aafe2c45651c0a8a759bb0b9e6b9606e71e13855/code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs#L152-L155

Search for `unwrap` in the codebase:

```console
λ rg "unwrap" -ths
drasil-lang/lib/Language/Drasil/People.hs
8:  , Person, People, Conv(..) --This is needed to unwrap names for the bibliography

drasil-gool/lib/Drasil/GOOL/LanguageRenderer/SwiftRenderer.hs
965:  let unwrap = if getCodeType t `elem` swiftReadableTypes &&
967:  unwrap $ mkStateVal (pure t) (R.castObj (renderType t) (RC.value v))

drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs
672:  -- this (which wouldn't duplicate state transitions) would be to unwrap the

drasil-example/swhs/lib/Drasil/SWHS/IMods.hs
81:   sParen (unwrap $ getUnit tempW), S "where", eS (exactDbl 0), sParen (unwrap $ getUnit tempW) `S.and_`
82:   eS (exactDbl 100), sParen (unwrap $ getUnit tempW) `S.are` D.toSent (pluralNP (NP.the ((melting `and_`
294:   phrase temp, S "at", phrase time, ch time, sParen (unwrap $ getUnit tInitMelt) `sC`
295:  ch tempW `S.andThe` phrase tempInit `sC` ch tempInit, sParen (unwrap $ getUnit tempInit)],
298:  unwrap (getUnit tempW), sParen $ refS assumpWAL `sC` refS assumpAPT]]
[...]

PR #5375, opened 2026-07-24
Follow-up: fix unitInParen and add phraseWithUnit

Purav-Kanda:
Follow-up to #5368 addressing balacij's review comments.

- Rewrite chWithUnit and phraseWithUnit to check whether a unit exists before formatting, instead of always appending unit parens — fixes the empty () output when no unit is present
- Remove unitInParen entirely (no longer needed now that chWithUnit/phraseWithUnit handle the unit check themselves)
- Add new introduceVar combinator: phrase x, ch x, (unit) — introduces a variable's name, symbol, and unit together in one call
- Keep unitSym exported (still used internally by chWithUnit/phraseWithUnit, and directly by call sites that only need the unit alone)
- Update all call sites in swhs/IMods.hs and swhsnopcm/IMods.hs to use the new combinators, fixing the one spot where phraseWithUnit was missing the symbol entirely

Relates to #5354


code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
```diff
@@ -147,14 +147,18 @@ bulletNested t l = Bullet (zipWith (\h c -> (Nested h c, Nothing)) t l)
 unitSym :: MayHaveUnit a => a -> Sentence
 unitSym = maybe EmptyS (Sy . usymb) . getUnit
 
--- | Formats a unit in parentheses. Outputs "(unit)".
+-- | Formats a unit in parentheses. Outputs "(unit)", or 'EmptyS' if no unit exists.
 unitInParen :: MayHaveUnit a => a -> Sentence
-unitInParen = sParen . unitSym
+unitInParen x = maybe EmptyS (sParen . Sy . usymb) (getUnit x)
```
balacij, line 152: This is definitely an improvement. The next thing we should do is look at the sites that use this and see if we can get around needing this entirely. If we can avoid using `EmptyS`, we should.

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
```diff
@@ -147,14 +147,18 @@ bulletNested t l = Bullet (zipWith (\h c -> (Nested h c, Nothing)) t l)
...
 
 -- | Outputs "ch x (unit of x)". For introducing a quantity with its symbol and unit.
 chWithUnit :: (Quantity a, MayHaveUnit a) => a -> Sentence
 chWithUnit x = ch x +:+ unitInParen x
 
+-- | Outputs "phrase x (unit of x)". For introducing a concept with its name and unit.
+phraseWithUnit :: (NamedIdea a, MayHaveUnit a) => a -> Sentence
+phraseWithUnit x = phrase x +:+ unitInParen x
```
balacij, line 160: This is a really good constructor to have. Are you sure it can't be used more in the codebase? Are we able to remove `unitSym` now?

It would be good to look how and where we introduce variables in `Sentence`s more generally and see if we would also want a constructor that mixes both `chWithUnit` and `phraseWithUnit`. For sentences where we specifically intend to _introduce_ a variable, we should always be rendering its units if they exist. So, if we often introduce variables in `Sentence`s, then we might not want `chWithUnit` nor `phraseWithUnit` but an `introduceVar` constructor that forms something along the lines of `<term>, <symbol>, (<unit>)`.

balacij (changes requested):

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
Purav-Kanda: On unitInParen: Agreed — will look at the call sites. All current uses are on quantities with known units, so I'll add a unitInParenDef :: UnitDefn -> Sentence that takes a UnitDefn directly and replace the call sites. unitInParen can remain for uncertain cases.

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
Purav-Kanda: On introduceVar: Great idea — I'll add introduceVar :: (NamedIdea a, Quantity a, MayHaveUnit a) => a -> Sentence that produces <term>, <symbol>, (<unit>). Will check if chWithUnit and phraseWithUnit can then be replaced or removed. unitSym can be removed by inlining into fmtU.

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
balacij: It would be best if we could minimize use of `unitInParenDef` use as well, but I agree, that's _definitely_ a good move forward.

To be clear, we want to remove `unitInParen` entirely as well. If you have a list of the uncertain cases, we can hop on a call and talk about them briefly?

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
balacij: You're free to suggest alternative variants of `introduceVar` as well!

What is `fmtU`? We really want to avoid using `EmptyS`, so inlining doesn't necessarily contribute to #5354, it just moves the issue to another constructor. What #5354 is really about is making the code more meaningful, showing the rationale for showing certain things together in a `Sentence`. In this sense, `introVar`/`introduceVar` is great because its reading shows clear intent -- "we're going to introduce a variable, so that means showing its important components." `fmtU` doesn't really indicate anything meaningful.

(If you'd like to hop on a quick call to chat about this, just let me know -- since you're a McMaster student, you should be able to message me on Teams.)

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
Purav-Kanda: Agreed on removing unitInParen entirely where possible. I went through the codebase for its call sites and split them into two groups:

Definite unit (chunk is guaranteed to have a UnitDefn) — these can move to a new unitInParenDef :: UnitDefn -> Sentence that takes the unit directly, no Maybe, no EmptyS.
Uncertain (unit may not exist) — [list these once you've run the grep below]

If the uncertain list ends up small or empty, we can drop unitInParen entirely and only keep unitInParenDef. I know you're busy, so rather than waiting on a call I'll push the actual list as a comment here — let me know if the split doesn't match what you had in mind and we can adjust from there.

code/drasil-lang/lib/Language/Drasil/Document/SentenceCombinators.hs
Purav-Kanda: `fmtU` takes an already-formatted `Sentence` (e.g. a formatted uncertainty value like "5 ± 0.1") and appends a unit symbol to it:

```haskell
fmtU :: MayHaveUnit a => Sentence -> a -> Sentence
fmtU n u = n +:+ unitSym u
```

It operates on raw `Sentence`s, not full chunks — so it's solving a different problem than `introduceVar` (which needs the whole chunk to pull term+symbol+unit together) and I don't think it can be replaced by it. That also means `unitSym` needs to stay exported, since `fmtU` depends on it directly — inlining it into `fmtU` would just move the same `EmptyS` logic one layer deeper, not remove it.

For `introduceVar` itself, here's what I'm proposing:

```haskell
introduceVar :: (NamedIdea a, Quantity a, MayHaveUnit a) => a -> Sentence
introduceVar x = phrase x `sC` ch x `sC` unitInParen x
```

Most of the current `chWithUnit`/`phraseWithUnit` call sites look like they're actually building this full triple, so those can become `introduceVar`. I'll keep `chWithUnit`/`phraseWithUnit` only where there's a genuine two-part use left over after the audit.

code/drasil-example/swhs/lib/Drasil/SWHS/IMods.hs
```diff
@@ -291,8 +291,8 @@ htWtrNotes :: [Sentence]
 htWtrNotes = map foldlSent [
   [S "The above", phrase equation `S.is` S "derived using", refS sensHtE],
   [D.toSent (atStartNP (NP.the (change `in_`temp))) `S.isThe` S "difference between the",
-   phrase temp, S "at", phrase time, ch time, unitInParen tInitMelt `sC`
-  ch tempW `S.andThe` phrase tempInit `sC` chWithUnit tempInit],
+   phrase temp, S "at", phrase time, ch time, sParen (unitSym tInitMelt) `sC`
```
balacij, line 294: `phrase time, ch time, sParen (unitSym tInitMelt)` -- Can't we use introduceVar here too?

balacij (changes requested):
Looking good. We might still want a function for `sParen . unitSym`, but it's still odd, and I don't think we should need to. I think we might want something like:

```haskell
exprWUnit :: PExpr -> UnitDefn -> Sentence
exprWUnit = ...
```

and/or maybe:

```haskell
matchingUnitOf :: HasUnit c => PExpr -> c -> Sentence
matchingUnitOf = ...
```

What is your analysis?

code/drasil-example/swhs/lib/Drasil/SWHS/IMods.hs
Purav-Kanda: no, introduceVar doesn't actually work here. That line is borrowing tInitMelt's unit to describe time, not time's own unit, so introduceVar time would try to pull a unit off time itself, which isn't the same thing (and might not even exist/compile). This spot isn't really "introduce one variable fully," it's "show this expression paired with a unit borrowed from something else" which is a different job

Purav-Kanda:
On the two options you suggested ,I think matchingUnitOf fits better than exprWUnit for this specific case. We already have tInitMelt sitting right there in scope, so matchingUnitOf time tInitMelt lets it grab the unit for us directly instead of us manually pulling unitSym tInitMelt out ourselves like the current code does.

code/drasil-example/swhs/lib/Drasil/SWHS/IMods.hs
balacij: Screenshot 2026-09-21 at 2 10 45 PM

Using the unit for "time" looks more correct to me anyways. I think we can use some variant of `introduceVar`.

balacij: Though the phrasing of the note is a bit awkward:

> The change in temperature is the difference between the temperature of the water, $T_W$ ($\deg{}C$), at time $t$ ($s$) and the initial temperature, $T_{\text{init}}$ ($\deg{}C$).

Maybe?

PR #5377, opened 2026-07-24
Flatten nested `ConstrConcept` in `UncertQ`

djarabek:
Part of https://github.com/JacquesCarette/Drasil/discussions/5330

Some follow up work should be done to review the constructors and their usages. A number of place use the form `uq (constrainted' ...` which should be simplified to a new constructor which doesn't use `ConstrConcept` at all.


JacquesCarette (approved):
Well, that was easy! Hurray for the pervasive use of 'classy lenses'!

Issue #5285, opened 2026-07-09
Integrate `OOVector` from #4909 into `drasil-code`s code generator


balacij:
> As a test of GOOL, I am fine with this. I would not want it to make it 'in' to GOOL or Drasil -- then I would have more comments about the design.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/4909#pullrequestreview-4631565019_
            
There are a few "clean up" hurdles to overcome, but the objective is as follows: to integrate `OOVector` as a _naive_ vectors implementation scheme in the code generator. This should _definitely_ be an option.

PR #5378, opened 2026-07-25
Removed `SharedStatement`

### LINKED ISSUE #5359: Remove `SharedProg` and `SharedStatement`?
In #5356, @JacquesCarette said we should remove `SharedProg` and `SharedStatement`. To be honest, I don't understand what we would gain from that. It would make some of our constraints enormous, and I don't see what we would gain from more precise constraints. @JacquesCarette could you provide a rationale for removing them?

B-rando1:
Closes #5359 

Btw is there a way to get HLS to automatically fill-in type holes? It's handy that it infers their type, but annoying that I still have to type them out.


code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
```diff
@@ -179,7 +194,20 @@ genDerivedCallProc = do
...
+  ::
+    ( MathConstant r
+    , VariableValue r
+    , BooleanExpression r
+    , Comparison r
+    , NumericExpression r
+    , ValueExpression r
+    , NativeVector r
+    , Argument r
+    , List r stmt
```
JacquesCarette, line 206: The great thing about this change is that it highlights that the only reason we need `stmt` here at all is `List` -- and why does `List` need it? [Something for a different PR.]

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
```diff
@@ -189,7 +217,21 @@ genConstraintCallProc = do
...
 genCalcCallProc
-  :: (NativeVector r, SharedStatement r stmt, TypeElim r)
+  ::
+    ( MathConstant r
+    , VariableValue r
+    , BooleanExpression r
+    , Comparison r
+    , NumericExpression r
+    , ValueExpression r
+    , DeclStatement r stmt
```
JacquesCarette, line 227: Whereas here, the dependency on `DeclStatement` more meaningfully depends on `stmt`

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -922,7 +978,16 @@ publicFuncProc n t desc ps r b = do
...
 privateFuncProc
-  :: (SharedProg r vis stmt mthd)
+  ::
+    ( OO.Literal r
+    , VariableValue r
+    , DeclStatement r stmt
+    , FileHandling r stmt
+    , PrintFile r stmt
+    , BodySym r stmt
+    , SharedProg r vis stmt mthd
```
JacquesCarette: Why is `SharedProg` still here?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
```diff
@@ -938,7 +1003,16 @@ privateFuncProc n t desc ps r b = do
...
 genMethodProc
-  :: (SharedProg r vis stmt mthd)
+  ::
+    ( OO.Literal r
+    , VariableValue r
+    , DeclStatement r stmt
+    , FileHandling r stmt
+    , PrintFile r stmt
+    , BodySym r stmt
+    , SharedProg r vis stmt mthd
```
JacquesCarette: and here (and more below)?

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
```diff
@@ -664,7 +665,8 @@ genOutputFormat = do
 
 -- | Generates a controller module.
 genMainProc
-  :: (ProcProg r vis stmt mthd prg, NativeVector r) => GenState (FS (r File))
+  :: (NativeVector r, ProcProg r vis stmt mthd prg)
```
JacquesCarette, line 668: Why does `genMainProc` always require `NativeVector` ?

JacquesCarette (commented):
Nothing wrong per se, just a bunch of questions before I'm ready to approve.

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
B-rando1: `List` needs it for operations like `listAdd`, `listAppend`, etc.

We could split `List` in half (or more) so that `stmt` isn't needed for most operations.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs
B-rando1: That's the next step. Removing `SharedStatment` was enough work that I decided to make it a checkpoint.

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
JacquesCarette: We should have an OO/imperative `List` API (which needs `stmt`) and a functional one (that doesn't).

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
B-rando1: Because both of our current procedural languages support `NativeVector`, we have `drasil-code` wired to use `NativeVector` for translating vector operations with procedural languages.

It's not ideal, but other than renaming it to `genMainProcNativeVector` (which is awful IMO) to be more precise, I don't know what we can do right now. Once we've worked out #5285 it _might_ make sense to have an overarching `Vector` typeclass that dispatches to `NativeVector` and various others; but I think it's too early to tell if that's the right way.

JacquesCarette: Supporting `NativeVector` does not mean that all valid programs *must* as well. In other words, it feels like there must be a `NativeVector` constraint somewhere "low" in the hierarchy which is not warranted.

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
JacquesCarette: In other words:
1. please file an issue about this, and
2. do indeed split the current `List` API into pieces.

B-rando1:
I did not realize I was still on this branch when I pushed my changes removing `SharedProg` (I seem to have a knack for that :sweat_smile:)

It might be fine, I'll just update "Contributes to" to "Closes"

code/drasil-code/lib/Language/Drasil/Code/Imperative/FunctionCalls.hs
B-rando1: Moved to #5380

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
B-rando1: I believe it mostly comes from `convExprProc`, e.g. here:
https://github.com/JacquesCarette/Drasil/blob/57d0ee44733d06f58f0d2b779e87feb6ae685e57/code/drasil-code/lib/Language/Drasil/Code/Imperative/Import.hs#L1105

Because we use `NativeVector` down there, everything above `convExprProc` needs `NativeVector` as well.

JacquesCarette (approved):

code/drasil-code/lib/Language/Drasil/Code/Imperative/Modules.hs
JacquesCarette: Ah, indeed, this is due to the design of `Expr` which is much too "flat". All the way back in `drasil-lang`. That will be more work to disentangle, and definitely not yours to worry about.

Issue #5160, opened 2026-06-15
Switch from `stack` to pure `cabal`


balacij:
> In 2026, it is probably wise to switch to cabal to unify all of our configurations in a single project. It has further improved support for multi-project setups and we can continue to use stackage package lists. However, it lacks module autodiscovery (which is a big negative in my opinion because of how many files we have). There are some tools (such as https://github.com/kowainik/autopack) out there to help automate module discovery that hpack would normally do for us, but it just doesn't look like it's worth the troubles right now. 

 _Originally posted by @balacij in [#3395](https://github.com/JacquesCarette/Drasil/issues/3395#issuecomment-4568008015)_

One thing I missed in my original comment is that there are tools that can do module autodiscovery!! For example, [cabal-gild](https://github.com/tfausak/cabal-gild). We should look into how much of a hassle it would be to have users install `cabal-gild`, and how much of the env setup process we can just have done automatically through the `Makefile`.

We've been discussing this in passing for a while now and it looks like we should pull the trigger and do it, but, it's not quite a major priority. That being said, if @djarabek is working on the Makefile and the CI and finds that this would be very beneficial and can be done within a reasonable amount of time (i.e., 1 day for everything), we can invest some time into this.

djarabek:
Stack also has an issue where building a test suite rebuilds the lib code as well, see https://github.com/JacquesCarette/Drasil/issues/5037#issuecomment-4693843789. Modern versions of Cabal do not appear to have this issue.

PR #5379, opened 2026-07-26
meta: update Stackage LTS to 25.41 and Haskell language spec. to GHC2024

### LINKED ISSUE #5372: Should we upgrade to GHC2024?
[GHC2024](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/control.html#extension-GHC2024) contains a number of helpful default flags over Haskell2010 that we often use.

We currently use (`<N> <FLAG>` -- i.e., each `FLAG` $N$ times):

```console
λ rg "LANGUAGE ([A-Za-z]+)" -ths -o -r '$1' --no-filename | sort | uniq -c | sort -nr
  59 PostfixOperators
  28 FlexibleContexts
  25 TemplateHaskell
  20 GADTs
  19 MultiParamTypeClasses
  15 TypeFamilies
  14 QuasiQuotes
  14 FlexibleInstances
  10 OverloadedStrings
   8 RankNTypes
   8 FunctionalDependencies
   6 TupleSections
   5 LambdaCase
   5 InstanceSigs
   4 PatternSynonyms
   3 TemplateHaskellQuotes
   3 ScopedTypeVariables
   3 RecordWildCards
   3 NamedFieldPuns
   2 DeriveLift
   2 DeriveFunctor
   1 UndecidableInstances
   1 TypeApplications
   1 tag
   1 Rank
   1 pragma
   1 NoMonomorphismRestriction
   1 KindSignatures
   1 GeneralizedNewtypeDeriving
   1 ExistentialQuantification
   1 DerivingVia
   1 DeriveGeneric
   1 DeriveAnyClass
   1 DataKinds
   1 ConstraintKinds
```

Of these, the following are enabled by default with GHC2024:

* PostfixOperators
* FlexibleContexts
* GADTs
* MultiParamTypeClasses
* FlexibleInstances
* RankNTypes
* TupleSections
* LambdaCase
* InstanceSigs
* ScopedTypeVariables
* NamedFieldPuns
* DeriveLift
* DeriveFunctor
* TypeApplications
* KindSignatures
* GeneralisedNewtypeDeriving
* ExistentialQuantification
* DerivingGeneric
* DataKinds
* ConstraintKinds

The only ones that are not in the list are mostly the ones I would find questionable to enable by default:

* TemplateHaskell
* TypeFamilies
* QuasiQuotes
* OverloadedStrings
* FunctionalDependencies (I don't see this one as questionable, really? Since @B-rando1 is going to be using this a lot, it might be good to add it to the `package.yaml` of `drasil-gool`)
* PatternSynonyms
* RecordWildCards
* NoMonomorphismRestriction
* DerivingVia
* DerivingAnyClass

_Some_ other notable ones that would be enabled by default:

* ImportQualifiedPost: This is a nice readability touch.
* ConstrainedClassMethods
* DerivingTraversable

The only issue with upgrading our default language is that we would also need to upgrade our minimum GHC version from GHC 9.6.7 to at least GHC 9.10.1. GHCup currently recommends GHC 9.10.3 as the current LTS and is also supported by HLS.

balacij:
Closes #5372 

As a result, our GHC compiler version is updated to GHC 9.10.3.

One caveat: I had to enable `-Wwarn=x-partial` in our CI because we have a number of warnings of partial functions used (`tail` and `head`).


.github/workflows/Lint.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         if: ${{ steps.filter.outputs.hs == 'true' }}
         uses: haskell-actions/hlint-setup@v2
         with: # DO NOT MODIFY THE BELOW LINE MANUALLY -- code/scripts/update_default_stackage.sh is in charge of this!
-          version: '3.6.1' # HLINT VERSION TIED TO CURRENT TARGET (LTS-22.44)
```
JacquesCarette, line 38: The comment above says that this shouldn't be modified manually?

code/drasil-srs/stack.yaml
```diff
@@ -31,6 +31,7 @@ resolver: lts-22.44
 packages:
 - .
 - ../drasil-code
+- ../drasil-data-formats
```
JacquesCarette, line 34: Why the new dependency, and why this PR?

JacquesCarette (changes requested):
A couple of things to look at first. And I want to also see an issue raised about `-Wwarn=x-partial` before merging.  We should not use `head` and `tail` !

.github/workflows/Lint.yaml
balacij: Yes, it was modified through the `code/scripts/update_default_stackage.sh` script. It's not strictly a hard rule in anyway, it's really just to encourage folks into using the script. That (a) makes their lives easier and (b) ensures that all necessary yaml files are updated all at once.

code/drasil-srs/stack.yaml
balacij: It's an indirect dependency. `stack build` was not working on `drasil-srs` (to test: `cd` into `drasil-srs` and run `stack build`, it will fail due to the missing dependency). Nobody really tests Drasil's individual packages through their individual `stack.yaml` files, only using the global one. If anyone tried, a few packages fail.

balacij: #5160 will deal with that issue when it's done.

balacij:
Created an issue for that -- #5382 :)

JacquesCarette (approved):

JacquesCarette:
Oops: this broke the build: "make graphs" which uses `stack install dotgen graphmod` failed.

JacquesCarette:
I restored the branch -- should I also revert?

balacij:
Moved to #5384 -- That should be enough to fix it, but we'll see!

Issue #5382, opened 2026-07-27
Avoid `head`/`tail` use


balacij:
`head` and `tail` are "unsafe"/discouraged functions in the Haskell world because they `error` programs out instead of returning "safe" failure values (`Nothing`). For each of our usecases in Drasil, we should try as best as we can to avoid them. There are a few ways we can do this:

1. Using `NonEmpty` lists if we always expect a list to be non-empty.
2. Rewriting code to avoid the `head` call entirely.
3. Incorporating more `Maybe`s or `Either`s to return soft errors instead of hard, program-halting `error`s.

The closure requirement for this issue will be to undo the changes to the CI and Makefile added in #5379 (i.e., removing the `-Wwarn=x-partial` added GHC options to our build/test arguments):

As of writing, see `rg "\b(head|tail)\b" -ths`:

```console
drasil-lang/lib/Language/Drasil/Data/Citation.hs
151:    a = if not (null as) then head as else error "No author found"
157:    y = if not (null ys) then head ys else error "No year found"
163:    t = if not (null ts) then head ts else error "No title found"

drasil-theory/lib/Theory/Drasil/DifferentialModel.hs
133:  | size == 1 = formASingleODE (head (d ^. coefficients)) unknownVec (d ^. dmConstants)
228:transUnknowns = tail
233:  | head es == exactDbl 1 = mapNeg $ tail es
234:  | otherwise = mapNeg $ tail $ map ($/ head es) es
251:addIdentityValue n es = fst splits ++ [exactDbl 1] ++ tail (snd splits)
270:        transEs = addIdentityCoeffs [transCoefficients $ head (dm ^. coefficients)] (length transUnks) 0
271:        transConsts = addIdentityConsts [head (dm ^. dmConstants) `divideConstant` head (head (dm ^. coefficients))] (length transUnks)

drasil-printers/lib/Language/Drasil/HTML/Helpers.hs
37:headTag   = wrap "head" []

drasil-website/lib/Drasil/Website/Analysis.hs
136:  \follows through so that dependent types are placed at the tail. Usually, this means that those types at the tail may contain the type at the tip of the arrow."
142:  \tail specifies the type that is an instance of the tip's class. For clarity in analyzing the structure, classes defined in the graph's package are coloured magenta, \
190:  \Each package and its dependencies are displayed in the form of a graph, with the tail of the arrow being the dependent module, \
[...]

PR #5383, opened 2026-07-28
remove uses of head, using pattern-matching instead.

JacquesCarette:
Still hard-error if required author/year/title not found.

Contributes to #5382 .


balacij (approved):
Certainly one of the most tricky to deal with amongst our head/tail uses.

Issue #5376, opened 2026-07-24
Split up `StatementSym`


B-rando1:
@JacquesCarette made a [comment](https://github.com/JacquesCarette/Drasil/pull/5356#discussion_r3634679482) in #5356 that the `StatementSym` hierarchy needs to change. Currently, we have a bunch of `XStatement` typeclasses that require `StatementSym` as a constraint. This sort of makes sense for the _types_ that `StatementSym` collects, but not for the methods. It has three: `valStmt`, `emptyStmt`, and `multi`. The case could be made that `emptyStmt` is a "fundamental" statement and `multi` is a sort of "statement combinator", so it _might_ make sense for them to stay inside `StatementSym` (though not necessarily). `valStmt` is clearly "a type of statement", not a fundamental one; so it should definitely be moved into its own typeclass.

If we want our constraints to be as precise as possible, the decision of where `emptyStmt` and `multi` should go depends on which typeclasses depend on them. Both of them likely show up in at least some of the implementations of most of the `XStatement` typeclasses, but I wouldn't say they're fundamental to any of them. Or if they are, then `emptyStmt` and/or `multi` should be constraints only on the typeclasses that require them.

Based on this, I think we should split `StatementSym` into four pieces: an empty `StatementSym` into four pieces: a now empty `StatementSym` that defines the `stmt` type parameter with its functional dependency, `MultiStatement`, `EmptyStatement`, and `ValueStatement`. We could also remove `StatementSym` altogether and define the functional dependency on every `XStatement` typeclass, but that seems less efficient to me.

JacquesCarette:
I agree that split into four is the likely best move.

B-rando1:
One question, will we be moving away from [OOStatement](https://github.com/JacquesCarette/Drasil/blob/66c66149626279fdaebe015bcdd51f8a9b378822/code/drasil-gool/lib/Drasil/GOOL/InterfaceGOOL.hs#L53-L62)? I need to decide if I should add the new typeclasses to it.

JacquesCarette:
Yes, we'll want to move away from that one too.

B-rando1:
As we discussed offline, it's probably better to remove `StatementSym` and only have one typeclass per method of `StatementSym`.

PR #5385, opened 2026-07-28
Split up `StatementSym`

B-rando1:
Contributes to #5376 

I'll make a follow-up PR removing `OOStatement` to close the issue.


B-rando1:
OK, I understand why these extra constraints seem to come out of nowhere in the diff. There's a bunch of other typeclasses that require `StatementSym` as a constraint, and that constraint wasn't changed. This means that before, if any of those other typeclasses were a constraint, then we got `emptyStmt`, `multi`, and `valStmt` "for free". Since those methods are now split off, they have to be explicitly mentioned as constraints.

This isn't super clear from the diff, and I'm realizing I probably did things exactly backwards: I should have started by removing `OOStatement`, then removing `StatementSym` as a constraint from other typeclasses (adding functional dependencies to `stmt` as needed), and then splitting up `StatementSym`. I hate to redo work, but it might be a good sanity check that I haven't accidentally introduced any redundant constraints.

B-rando1:
Closing in favour of #5396

PR #5390, opened 2026-07-28
Removed `OOStatement`

### LINKED ISSUE #5376: Split up `StatementSym`

B-rando1:
Closes #5376 
Builds on #5385


B-rando1:
Closing in favour of #5392

PR #5392, opened 2026-07-28
Removed `OOStatement`

B-rando1:
Contributes to #5376 

The only case of constraints appearing out of nowhere is `GetSet`, because it was part of `OOStatement` and I didn't add it to `OOProg`. I wasn't sure if it's essential enough to be included there, or if it's better left out like we did with `StrategyPattern` and `ObserverPattern`.


JacquesCarette:
`GetSet` is a kind of pattern too, so should be treated like other patterns.

JacquesCarette (approved):

PR #5393, opened 2026-07-28
docs: `drasil-printers` shallow file analysis

balacij:
Not really "shallow."

Contributes to #4989 .

### DIFF
--- a/People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
+++ b/People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
@@ -1,5 +1,19 @@
 # `drasil-printers` Shallow File Analysis
 
+## Summary
+
+[#4989](https://github.com/JacquesCarette/Drasil/issues/4989) is a great project and this file analysis affirms a considerable portion of the intended work outlined in that project. To continue however (I will not reiterate major work mentioned in that issue; most important, relevant one being about using intermediate output representations before rendering to `Doc`s), we should also:
+
+1. [ ] Work on erasing (rendering) `Citation`s earlier (at the level of the document language currently residing in `drasil-lang`) before it is converted into the `LayoutObj`/`Document` language residing in `drasil-printers`. This will require us to upgrade the `Spec` language with more features for text formatting (bold, italics, etc.).
+2. [ ] Work on erasing `Reference`s earlier too.
+3. [ ] Remove the `genJupyterLessonPlan` vs `genJupyterSRS` difference. We should opt for `genJupyterLessonPlan` short-term and wait for [#2346](https://github.com/JacquesCarette/Drasil/issues/2346) to 'fix' the results of the erasing `genJupyterSRS` (which generates a single cell) versus `genJupyterLessonPlan` (which renders every `LayoutObj` in its own cell).
+4. [ ] Move the mdBook generator _out_ of `drasil-printers`. `mdBook` _relies_ on `drasil-printers`, but there is much more knowledge that goes into building an mdBook project than that should be captured in `drasil-printers`.
+5. [ ] Move `drasil-printers/lib/Language/Drasil/Plain/Print.hs`, removing mention of `LayoutObj` and `Spec`, opting for directly rendering to `String`, so as to allow us to move the `Drasil.Code` modules in `drasil-lang` to be moved to `drasil-code`.
+6. [ ] Move `mdbook` to its own package. This is a package that should _depend_ on `drasil-printers` for `LayoutObj` and Markdown rendering, and `drasil-gool` for code rendering.
+7. [ ] We also need to be able to encode arbitrary code blocks in our `LayoutObj` language. This should not rely on `drasil-gool` nor `drasil-code` at all. It truly needs to be arbitrary with an optional tag for what language is used (if a real language was used).
+8. [ ] (Copied from later section: `/Import/**.hs`) `drasil-printers` should only be dependent on only one kind of package: intermediate document representation formats, including HTML, TeX, Markdown, `Doc`, etc. Even `Doc` (e.g., `prettyprinter`) is a stretch. Specific `Doc`s should be producible through rendering the HTML/TeX/Markdown/etc. External packages that need to typeset scientific documents should rely on `drasil-printers`' public API which also provides renderers for specific languages (e.g., HTML/TeX/Markdown/etc.).
+9. [ ] With (8), `drasil-printers` is an imprecise name. We should refine that. It would contain exclusively a generic scientific document typesetting (not 'writing'/'laying out') language and renderers for the generic scientific typesetting language to arbitrary supported document formats (TeX, HTML, Markdown, and Jupyter). "Project output formats" (such as mdBook) should rely on `drasil-printers` for running generic scientific document layout programs (i.e., for converting `Section`s into `LayoutObj`s/finally typesetting scientific documents). In this sense, `drasil-printers` is both an overlanguage (over HTML/TeX/etc.) and intermediate representation (for the generic scientific document layout language).
+
 ## `drasil-printers/lib/Language/Drasil/Config.hs`
 
 Contains two kinds of options:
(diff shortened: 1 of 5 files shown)


People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
+# `drasil-printers` Shallow File Analysis
+
+## Summary
+
+[#4989](https://github.com/JacquesCarette/Drasil/issues/4989) is a great project and this file analysis affirms a considerable portion of the intended work outlined in that project. To continue however (I will not reiterate major work mentioned in that issue; most important, relevant one being about using intermediate output representations before rendering to `Doc`s), we should also:
+
+1. [ ] Work on erasing (rendering) `Citation`s earlier (at the level of the document language currently residing in `drasil-lang`) before it is converted into the `LayoutObj`/`Document` language residing in `drasil-printers`. This will require us to upgrade the `Spec` language with more features for text formatting (bold, italics, etc.).
```
JacquesCarette, line 7: I don't think the `Spec` language should have text formatting primitives. Those are very much layout stuff. `Spec` should be more semantic.

Why do you think this should be done that early?

Could you rephrase this in terms of "Problem => Solution" instead?

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+2. [ ] Work on erasing `Reference`s earlier too.
```
JacquesCarette, line 8: Same "Problem => Solution" needed, as I don't think this is necessarily a good idea!

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+3. [ ] Remove the `genJupyterLessonPlan` vs `genJupyterSRS` difference. We should opt for `genJupyterLessonPlan` short-term and wait for [#2346](https://github.com/JacquesCarette/Drasil/issues/2346) to 'fix' the results of the erasing `genJupyterSRS` (which generates a single cell) versus `genJupyterLessonPlan` (which renders every `LayoutObj` in its own cell).
+4. [ ] Move the mdBook generator _out_ of `drasil-printers`. `mdBook` _relies_ on `drasil-printers`, but there is much more knowledge that goes into building an mdBook project than that should be captured in `drasil-printers`.
```
JacquesCarette, line 10: This would need more precision (i.e. what aspects belong here, what aspects belong elsewhere, and why). I do think I agree, but...

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+5. [ ] Move `drasil-printers/lib/Language/Drasil/Plain/Print.hs`, removing mention of `LayoutObj` and `Spec`, opting for directly rendering to `String`, so as to allow us to move the `Drasil.Code` modules in `drasil-lang` to be moved to `drasil-code`.
```
JacquesCarette, line 11: Again, all mixed up. For one, it should probably be `Text` and not `String`. Second, it should be clearer that the intent is *no layout*, i.e. one long line.

Nevertheless, it is still a renderer, and should be treated as such.

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+6. [ ] Move `mdbook` to its own package. This is a package that should _depend_ on `drasil-printers` for `LayoutObj` and Markdown rendering, and `drasil-gool` for code rendering.
```
JacquesCarette, line 12: and probably should not be called `mdbook` as that is a *format*. If it is a 'twin' to `drasil-srs`, then that makes sense. But it needs to have a 'semantic' name.

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+7. [ ] We also need to be able to encode arbitrary code blocks in our `LayoutObj` language. This should not rely on `drasil-gool` nor `drasil-code` at all. It truly needs to be arbitrary with an optional tag for what language is used (if a real language was used).
```
JacquesCarette, line 13: Right - our architecture should allow "arbitrary block of stuff formatted by a foreign entity". The caller would then be responsible to supply an appropriate renderer, thus decoupling things.

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+On the bibliography formats:
+
+1. MLA and APA are "Works Cited"-style bibliographies (i.e., only showing cited references).
+2. Chicago is a true bibliography (showing more than cited references).
+
+So, that is to say we should
+
+- [ ] Allow Chicago formatting to contain an extra list of references that will always be displayed in the bibliography (even if not cited within the document).
+
+However, a question arises: why does `drasil-printers` know about references at all? These should have been rendered earlier (at the level of the semantic document language being rendered into the `LayoutObj` language that `drasil-printers` defines).
```
JacquesCarette, line 33: I agree that someone else should probably be doing the conversion to layout.

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+
+- [ ] References and citations should have been "erased" (rendered) earlier.
+
+Finally:
+
+- [ ] All options should be routed through a record.
```
JacquesCarette, line 39: Premature implementation decision?

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+
+### `drasil-printers/lib/Language/Drasil/Printing/Helpers.hs`
+
+Contains a number of various "helper" functions. This file primarily sits atop `Text.PrettyPrint`.
+
+Idea: create `drasil-prettyprinter` that exports various tools that `drasil-gool` and `drasil-printers` could both benefit from. Perhaps also reexporting the entire `Prettyprinter` library at the same time if we feel like we can add enough to the core library.
+
+### `drasil-printers/lib/Language/Drasil/Printing/LayoutObj.hs`
+
+Contains a lot more than `LayoutObj`:
```
JacquesCarette, line 198: Good point. On the other hand, this is also quite close to what `pandoc` does. So it is a general "layout" language (for whole documents).

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+   | CodeBlock Contents                                              -- ^ Code block.
+   | HDiv Tags [LayoutObj] Label                                     -- ^ Holds tags, more contents, and a label.
+   | Cell [LayoutObj]
+   -- this shouldn't be here, it should have been expanded.
+   | Bib BibRef                                                      -- ^ Bibliography section.
+```
+
+There is no clear reason why the majority of these exist in this file.
+
+* [ ] The "large" ones (`Document`, `Project`, `File`) should be moved to another file. However we should dig deeper and see if these _should_ be necessary. They have a code smell to them in this package.
```
JacquesCarette, line 248: Again, jumping to a solution too soon. Need to figure out what the problem is first. (But I agree that `File` is odd).

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+* [ ] `Author` should be removed. Definitely information that should have been erased earlier.
```
JacquesCarette, line 249: Take a look at `pandoc` before deciding this.

People/Jason/drasil-analysis/07-20-2026-drasil-printers.md
```diff
@@ -0,0 +1,408 @@
...
+* [ ] `RefMap` appears to only be used in the mdBook generator. It should be moved there because the "reference" system also needs to be rebuilt and we don't want to encourage reusing this type.
+* [ ] `HDiv` should be investigated. It does not look like a good constructor in a generic object layout language.
```
JacquesCarette, line 251: Don't be fooled by the name! It might be a wonderful 'block' constructor.

JacquesCarette (changes requested):

JacquesCarette:
Way, way too deep. The point was to get an inventory of assets. I understand also making notes of other things as you go -- but that should have been split into a different 'notebook', as it were. And some of the offered solutions don't match a clear problem.

PR #5386, opened 2026-07-28
Add MATLAB to GOOL procedural test suite and implement renderer methods

Xinlu-Y:
The generated MATLAB code has some known issues that I plan to fix in follow-up work, including `else if` vs `elseif`, uninitialized temp variables in list slicing, and incorrect use of `~=` for string comparison.

PR #5400, opened 2026-07-31
Fix MATLAB filename to match entry function name

Xinlu-Y:
contribute to #5386


code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -374,6 +374,16 @@ mlCellWrap :: CodeType -> Doc -> Doc
 mlCellWrap String = braces
 mlCellWrap _      = parens
 
+mlExtractFuncName :: MatlabCode MethodData -> String
+mlExtractFuncName m = extractName $ lines $ render $ mthdDoc $ unMLC m
```
JacquesCarette: We should definitely not do this, in none of our backends (i.e. render to a `String` then manipulate it -- neither the rendering nor the manipulation.)

The right way to do this is to make sure that `MethodData` carries the function name.

JacquesCarette (changes requested):

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
Xinlu-Y: It's fixed.

JacquesCarette (approved):

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/JavaRenderer.hs
B-rando1, line 686: If most of the renderers aren't using the new field of `MethodData`, we should probably create a new `MethodData` for them that doesn't have the unused field.

B-rando1:
Just getting myself back up to speed on the last few days of work, what exactly was the issue and solution here? Does MATLAB require the filename to match the name of the first function in the file?

From looking at the generated code, I see that in [VectorTest.hs](https://github.com/JacquesCarette/Drasil/blob/dbd5860df5b47d4e4c74abe822687157420e941e/code/drasil-code/test/VectorTest.hs#L16-L25) the module name is still `VectorTest`, and the Julia version is still [VectorTest.jl](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/test/golden/VectorTestProc/julia/VectorTest.jl); but the MATLAB version is [vectorOps.m](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/test/golden/VectorTestProc/matlab/vectorOps.m). Is the MATLAB renderer silently changing the file name from what the GOOL programmer wrote? I think at the very least this should be made more explicit, and ideally we would have a user-facing mechanism to keep the names in sync.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -293,24 +293,24 @@ instance Comparison MatlabCode where
...
-  libFuncAppMixedArgs = CS.extFuncAppMixedArgs
+  extFuncAppMixedArgs _ = G.funcAppMixedArgs
+  libFuncAppMixedArgs _ = G.funcAppMixedArgs
   lambda = undefined
-  notNull = undefined
+  notNull v = (?!) $ funcApp "isempty" bool [v]
 
 instance RenderValue MatlabCode where
-  inputFunc = undefined
+  inputFunc = mkStateVal string (text "input" <> parens (text "'', 's'"))
```
B-rando1, line 308: This looks like it could be replaced with `funcApp`

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -698,8 +716,70 @@ mlPrint newLn f' _ v' = do
...
+    <+> text "+ ~" <> parens (RC.value c) <+> text ".*" <+> parens (RC.value v2)))
+
+mlCast :: VS (MatlabCode TypeData) -> SValue MatlabCode -> SValue MatlabCode
+mlCast t' v' = do
+  t <- t'
+  v <- v'
+  let vTp = getCodeType $ valueType v
+      tTp = getCodeType t
+      vDoc = RC.value v
+      mlCast' String Integer = text "str2double" <> parens vDoc
```
B-rando1, line 736: These could probably be `funcApp` too, if you change the types of the `let` statements around a bit.

Xinlu-Y:
> Just getting myself back up to speed on the last few days of work, what exactly was the issue and solution here? Does MATLAB require the filename to match the name of the first function in the file?
> 
> From looking at the generated code, I see that in [VectorTest.hs](https://github.com/JacquesCarette/Drasil/blob/dbd5860df5b47d4e4c74abe822687157420e941e/code/drasil-code/test/VectorTest.hs#L16-L25) the module name is still `VectorTest`, and the Julia version is still [VectorTest.jl](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/test/golden/VectorTestProc/julia/VectorTest.jl); but the MATLAB version is [vectorOps.m](https://github.com/JacquesCarette/Drasil/blob/main/code/drasil-code/test/golden/VectorTestProc/matlab/vectorOps.m). Is the MATLAB renderer silently changing the file name from what the GOOL programmer wrote? I think at the very least this should be made more explicit, and ideally we would have a user-facing mechanism to keep the names in sync.

Yes, MATLAB requires the filename to match the name of the first function in the file. If the file is called `VectorTest.m` but the first function is `vectorOps`, MATLAB will throw an error when you try to call `vectorOps`.

You're right that the renderer is silently renaming the file. The current implementation detects when there's no main function, and overrides the module name with the first function's name (via `mthdName`). This works, but it's not ideal since the GOOL programmer has no visibility into the rename.

A better approach might be to have the test code use matching names in the first place (e.g. `buildModule "vectorOps"` instead of `buildModule "VectorTest"`), or to emit a warning when the module name and function name don't match. I'm open to suggestions on how to make this more explicit.

Issue #5404, opened 2026-08-05
`ConceptDomain` is used to provide type information for `ConceptInstance`, not general domain information


djarabek:
`ConceptDomain` is a typeclass with a single method, `cdom`: https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-lang/lib/Language/Drasil/Classes.hs#L56-L59

`cdom` returns a list of `UID`s, which (according to the documentation) refer to chunks that describe the domain the concept is in.

`ConceptDomain` is part of the `Concept` type, so every chunk that is a `Concept` is an instance of `ConceptDomain`.

However, the only instance there the domain is used is `ConceptInstance`. There are only a few places where `cdom` is called, and in all of these situations, the chunk type is `ConceptInstance`, and `sDom` is applied to the result of `cdom`. `sDom` requires exactly 1 domain to be present, and returns it: https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-lang/lib/Language/Drasil/Chunk/Concept/Core.hs#L23-L27

This can also be seen in the `cic` constructor, which takes in a single `Concept` as the domain, and constructs a singleton list from it: https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-lang/lib/Language/Drasil/Chunk/Concept.hs#L83-L90

This 'domain' is in fact used to encode type information. `ConceptInstance`s are used for assumptions, unlikely and likely changes, functional and non functional requirements and goal statements, with the domain used to specify which type it is. The available types are declared in `drasil-metadata`: https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-metadata/lib/Drasil/Metadata/Documentation.hs#L219-L228

In the SRS generation code, these types are hardcoded, for instance:
https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-srs/lib/Drasil/SRS/DocDecl.hs#L132-L134
where `fromConcInsDB` is (note the usage of `sDom`):
https://github.com/JacquesCarette/Drasil/blob/1d60c29ca676864b6189d9f825845470d2e98415/code/drasil-srs/lib/Drasil/SRS/DocDecl.hs#L154-L156

[...]

PR #5405, opened 2026-08-05
Move `ConceptInstance` to Document namespace

djarabek:
As discussed in #5404, `ConceptInstance` is used only for SRS purposes, so it has been moved to reflect that.


balacij (approved):

balacij:
This will be very good for the [`drasil-lang` module dependency graph](https://jacquescarette.github.io/Drasil/graphs/drasil-lang.pdf)! The lines for the new 'document' package will be clearer.

> As discussed in https://github.com/JacquesCarette/Drasil/issues/5404, ConceptInstance is used only for SRS purposes, so it has been moved to reflect that.

To refine this a little bit, the rationale to move it to the `.Document` namespace is that it is usable in non-SRS scenarios as well. That being said, as you also recently mentioned in https://github.com/JacquesCarette/Drasil/issues/5404, we probably need to rework/remove this type altogether eventually.

JacquesCarette (approved):
Good "refactor step 1". The name is now horrible, but this is a necessary move.

PR #5406, opened 2026-08-05
Introduce Markdown AST and renderer

camishibata:
Contributes to https://github.com/JacquesCarette/Drasil/issues/4989
Depends on #5319


code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Core.hs
```diff
@@ -0,0 +1,36 @@
...
+  | Link URL [Markdown]
+  | Image Source [Markdown] (Maybe Text)
+  | List ListType [ListItem]
+  | Table [[Markdown]] [[Markdown]] (Maybe [Markdown]) (Maybe Text)
+  | Paragraph [Markdown]
+  | Bold [Markdown]
+  | Italic [Markdown]
+  | RawText Text
+  | Line
+  deriving (Eq, Show)
```
balacij, line 23: Can you please add Haddock documentation to each constructor and argument for clarity?

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Core.hs
```diff
@@ -0,0 +1,36 @@
...
+  ( -- * Markdown
+    Markdown(..), URL, Source, ListType(..), ListItem
+  )
+where
+
+import Data.Text (Text)
+
+data Markdown =
+    Heading Int (Maybe Text) [Markdown]
+  | Div Text [Markdown]
```
balacij: I think we should allow more than just `div`s in the Markdown AST. It would be nice if we could place arbitrary `HTML` (your new HTML AST) here too. We should _not_ allow the HTML AST to escape back into Markdown land. That is for a different kind of Markdown flavour that we should not support for now. Let's go for the simplest variant for now.

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
```diff
@@ -0,0 +1,114 @@
...
+import qualified Prettyprinter as PNew (Doc)
+import Data.List (transpose)
+import Data.Text (Text, pack, singleton)
+import qualified Data.Text as T (concatMap)
+
+-- | Options for rendering 'Markdown'.
+data MDRenderOptions = MDRO {
+    mdFlavour :: MDFlavour,
+    tableStyle :: TableStyle
+}
```
balacij, line 21: Two more options: preferred characters for bold and italicized text. `_` or `*`.

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
```diff
@@ -0,0 +1,114 @@
...
+
+-- | Markdown Flavour: Markdown extension/variant
+data MDFlavour = Pandoc
+  deriving (Eq, Show)
```
balacij, line 25: Instead of supporting flavours of Markdown as options, which would mean we would need one Markdown AST that covers all possible flavours, we should have separate namespaces for each variant, especially because each flavour may have its own host of options (too many things to consider all at once, such as Pandoc which has extensions for its own Markdown).

Instead of rendering `Div` with `:::: {#id}`, we should escape into HTML land and use a raw `div`. So, this comment is in agreement with my earlier one about `Div`, and also should simplify some of this code?

balacij (changes requested):
I think you should base this PR off of the HTML one. Having that AST in scope of the Markdown one will be good.

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
camishibata: So each flavor should have its own AST and `render.hs`?

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
camishibata: Added in 9a67ff8903a52099dea41219e6b0189397f41dc5

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Core.hs
camishibata: Added in 9a67ff8903a52099dea41219e6b0189397f41dc5

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
camishibata: About `Div`, I thought `:::: {#id}` would be a nice feature to prioritize pure Markdown

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
balacij: Perhaps we should split it in two options? And we should add documentation.

code/drasil-data-formats/lib/Drasil/Data/Formats/MD/Render.hs
balacij: I don't believe `:::: {#id}` is a standard Markdown feature. It looks like it's normally part of extended feature sets.

PR #5407, opened 2026-08-05
Remove unneeded `ConceptDomain` instances

djarabek:
Depends on #5405

As discussed in #5404, `ConceptDomain` is used to provide type information for `ConceptInstance`s rather than general domain information.


JacquesCarette (approved):
We're losing functionality... that we didn't use.

PR #5410, opened 2026-08-06
Removed use of `tail` in Swift renderer

B-rando1:
Contributes to #5382 

In my [comment](https://github.com/JacquesCarette/Drasil/issues/5382#issuecomment-5207183858) I was being led astray by the existing implementation. `TypeData` _already_ gives a sort of AST for nested types, and we just had to make use of it. The fact that the original implementation worked at all is due to a major hack/coincidence, that in Swift the Set type is denoted by wrapping the inner type with brackets.


B-rando1:
Elaborating on how this code got here in the first place, Noah added it when he added sets to GOOL 2 years ago. He wasn't very familiar with GOOL and must not have realized that there's a better way to extract the inner type of a composite type; and we didn't catch this hack.

JacquesCarette (approved):

PR #5414, opened 2026-08-07
feat(SystemMeta): Create `ProjectName` chunk

### LINKED ISSUE #4884: Need new chunk: `ProgramName`

balacij:
Closes #4884

Is this PR ready to be merged? No. Feedback? Yes please!

This PR creates a new chunk type, `ProjectName`, intending to fully replace the `CI` carried in `SystemMeta`.

Each project we intend to codify in Drasil has basic meta-information:

* A title and abbreviation.
* Authors list.
* A description (split up with background, scope, purpose, and motivation information).
* Background knowledge (a `ChunkDB`).

The first point is the concern of this PR. Currently, we use `CI`s to encode them. For example, with GlassBR:

https://github.com/JacquesCarette/Drasil/blob/f52e7b12b127e3911a24e1c600336e264a19633d/code/drasil-example/glassbr/lib/Drasil/GlassBR/MetaConcepts.hs#L7-L8

The majority of our case studies follow the above pattern of duplication where duplicate a "term" and "abbreviation."

These are used in:

1. SRS documents (referring to "GlassBR," the software requirements specification).
2. Makefile targets: refering to the GlassBR binary.
3. Java package names: "GlassBR" is our main package. Aside: is this strictly necessary? We don't generate a folder for any other language.
4. Artifact folder names: a lower-cased version of the abbreviation (i.e., "glassbr"). This is common for many code repository names, which is the convention we follow. This convention is only up to specific circles (e.g., go folks use `go-` prefixes, rust folks use `-rs` suffixes, and some folks use CamelCase).

SSP is an outlier in its "term," which contains more meaningful information:

https://github.com/JacquesCarette/Drasil/blob/f52e7b12b127e3911a24e1c600336e264a19633d/code/drasil-example/ssp/lib/Drasil/SSP/MetaConcepts.hs#L7-L8

What it looks like we really mean by "term" when it comes to discussing systems is "title." (A title is a kind of term, so using less specific terminology is not terrible, but we should clarify that in documentation. Something along the lines of 'term means the fully expanded identifier, which may be a title, full name, etc.'.)

**Assumption**: We need to give a full title to each of our case studies. For example, some ideas:

* Projectile: Projectile Motion Analysis Program
* ProjLsnPlan: Lesson Plan for Analyzing Projectile Motion
* PD Controller: Automatic Process Controller Analysis Tool

Note: For now, I do not "fix" any of the titles.

This PR creates `ProjectName` which contains all 3 different kinds of "names" we might refer to a project by, each commonly understood as referring to different aspects:

* Title.
* Abbreviation.
* Repository name.

Reminder: The goal of this PR is to fully replace the `CI` with `ProjectName`.

This PR stops short of its complete goal, currently working towards replacing `CI`. Posting now for early feedback.


code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
```diff
@@ -180,6 +180,21 @@ instance HasSystemMeta DrasilState where
...
+  chunkRefs x = chunkRefs (x ^. dsCodeSpec)
+
+instance NamedIdea DrasilState where
+  term = dsCodeSpec . term
+
+instance Idea DrasilState where
+  getA x = getA (x ^. dsCodeSpec)
+
+instance CommonIdea DrasilState where
+  abrv x = abrv (x ^. dsCodeSpec)
```
balacij: Please ignore the (1) reused UIDs, (2) added `HasUID` instances entirely. They are required if we want to instantiate `NamedIdea`, which is helpful.

balacij:
Some work:

1. Move away from `progName :: CI` usage to `projName :: ProjectName`. This will require making some code that previously required `CI`s specifically polymorphic.
2. Remove `sysName :: CI` field from `SystemMeta`.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
JacquesCarette: Why on earth would we need these instances for `DrasilState`?! This seems really wrong!

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
```diff
@@ -204,7 +219,7 @@ modExportMap cs chs@Choices {
     ++ getExpInputFormat prn chs extIns
     ++ getExpCalcs prn chs (cs ^. execOrder)
     ++ getExpOutput prn chs (cs ^. outputs)
-  where prn = cs ^. programName
+  where prn = abrv cs
```
JacquesCarette: I don't think we want to overload the ill-defined `abrv` even more. This should perhaps be `projectName` or something like that (rather than `programName`), but not `abrv`.

code/drasil-example/bss/lib/Drasil/BinaryStar/Body.hs
```diff
@@ -197,9 +198,12 @@ stdFields = [DefiningEquation, Description Verbose IncludeUnits, Notes, Source,
 -- System Information --
 ------------------------------
 
+projName :: ProjectName
```
JacquesCarette: Good!

code/drasil-system/lib/Drasil/System/Core.hs
```diff
@@ -1,14 +1,17 @@
+{-# LANGUAGE FlexibleInstances, UndecidableInstances #-}
```
JacquesCarette: We should not use `UndecidableInstances` unless we're doing evil trickery (which we don't want to do).

code/drasil-system/lib/Drasil/System/ProjectName.hs
```diff
@@ -0,0 +1,64 @@
...
+  where
+    frmtr = map toLower . replaceAll (not . isAlphaNum) '-'
+
+instance HasUID ProjectName where
+  uid = pnUID
+
+instance NamedIdea ProjectName where
+  term = title
+
+instance Idea ProjectName where
```
JacquesCarette: These next 2 instances are things we don't want. I think we need new classes to describe the information that's in a `ProjectName`, since they mean something different.

JacquesCarette (changes requested):
Many of the code improvements are good, but there's some conceptual mixup that needs untangled.

code/drasil-system/lib/Drasil/System/Core.hs
balacij: Yeah, I was trying to do some evil trickery with default instances, but it's not worth the effort. Removed in f3caf399a8c5fa41752ead6d08586caab3891018.

code/drasil-system/lib/Drasil/System/ProjectName.hs
balacij: We have `HasProjectName` now in f3caf399a8c5fa41752ead6d08586caab3891018. We also remove the instances of `NamedIdea`, `Idea`, and `CommonIdea` from `ProjectName` and everything else I added it to (e.g., `DrasilState`...).

balacij:
Welp, I tried to use Copilot to review, but I've apparently reached my quota. What quota? Not sure. I don't even use Copilot!

code/drasil-system/lib/Drasil/System/ProjectName.hs
```diff
@@ -0,0 +1,66 @@
...
+--
+-- Abbreviation:
+-- 1. Non-empty
+-- 2. No special whitespace (\t, \r, \n, \f, \v)
+-- 3. <64 chars?
+--
+-- Repo Name:
+-- 1. Non-empty
+-- 2. Valid chars: lowercase alphanumeric or '-' only
+-- 3. <64 chars?
```
balacij: The only question I have before I continue working on this is about the validation policy. How long can an human-readable abbreviation really be? What should be our validation policy for the preferred repo name/slug?

JacquesCarette: The guidelines as you have them above seem sensible to be as a starting point.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -741,10 +741,6 @@ <h2>Abbreviations and Acronyms</h2>
                   <td>SRS</td>
                   <td>Software Requirements Specification</td>
                 </tr>
-                <tr>
```
JacquesCarette, line 744: Hmm, we should not lose this abbreviation. This means our abbreviation gathering system might need enhancement.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -768,16 +764,16 @@ <h2>Abbreviations and Acronyms</h2>
       <div class="section">
         <h1>Introduction</h1>
         <p class="paragraph">
-          Due to increasing costs, diminishing availability, and negative environmental impact of fossil fuels, the demand is high for renewable energy sources and energy storage technology. Solar water heating systems incorporating phase change material (<span title="phase change material">PCM</span>) use a renewable energy source and provide a novel way of storing energy. Solar water heating systems incorporating PCM improve over the traditional solar water heating systems because of their smaller size. The smaller size is possible because of the ability of <span title="phase change material">PCM</span> to store thermal energy as latent heat, which allows higher thermal energy storage capacity per unit weight.
```
JacquesCarette, line 771: I think the previous plural was correct, so this too needs investigated.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -796,7 +792,7 @@ <h2>Scope of Requirements</h2>
           <div class="subsection">
             <h2>Characteristics of Intended Reader</h2>
             <p class="paragraph">
-              Reviewers of this documentation should have an understanding of heat transfer theory from level 3 or 4 mechanical engineering and differential equations from level 1 and 2 calculus. The users of <span title="solar water heating systems incorporating PCM">SWHS</span> can have a lower level of expertise, as explained in <a href="#Sec:UserChars">Sec:User Characteristics</a>.
+              Reviewers of this documentation should have an understanding of heat transfer theory from level 3 or 4 mechanical engineering and differential equations from level 1 and 2 calculus. The users of <span title="Solar Water Heating Systems Incorporating PCM">SWHS</span> can have a lower level of expertise, as explained in <a href="#Sec:UserChars">Sec:User Characteristics</a>.
```
JacquesCarette: Need to discuss with @smiths if the capitalization of Incorporating is what we want.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -807,7 +803,7 @@ <h2>Organization of Document</h2>
...
+              The <a href="#Sec:GoalStmt">goal statements</a> are refined to the theoretical models and the <a href="#Sec:TMs">theoretical models</a> to the <a href="#Sec:IMs">instance models</a>. The instance models to be solved are referred to as <a href="#IM:eBalanceOnWtr">IM:eBalanceOnWtr</a>, <a href="#IM:eBalanceOnPCM">IM:eBalanceOnPCM</a>, <a href="#IM:heatEInWtr">IM:heatEInWtr</a>, and <a href="#IM:heatEInPCM">IM:heatEInPCM</a>. The instance models provide the ordinary differential equations (<span title="ordinary differential equation">ODE</span>s) and algebraic equations that model solar water heating system incorporating phase change material. <span title="Solar Water Heating Systems Incorporating PCM">SWHS</span> solves these <span title="ordinary differential equation">ODE</span>s.
```
JacquesCarette: The expansion of `PCM` here seems gratuitous?

code/stable/swhsnopcm/src/csharp/README.md
```diff
@@ -9,7 +9,7 @@
 ## What 
  
 
-> Background: Solar Water Heating System with no Phase Change Material provide a novel way of storing energy..
+> Background: Solar Water Heating System With No Phase Change Material provide a novel way of storing energy..
```
JacquesCarette: this capitalization change is definitely odd.

code/stable/swhsnopcm/SRS/PDF/SWHSNoPCM_SRS.tex
```diff
@@ -62,7 +62,7 @@ \subsection{Table of Units}
 \end{longtblr}
 \subsection{Table of Symbols}
 \label{Sec:ToS}
-The symbols used in this document are summarized in the \hyperref[Table:ToS]{Table of Symbols} along with their units. The choice of symbols was made to be consistent with the heat transfer literature and with existing documentation for solar water heating systems with no phase change material. The symbols are listed in alphabetical order. For vector quantities, the units shown are for each component of the vector.
+The symbols used in this document are summarized in the \hyperref[Table:ToS]{Table of Symbols} along with their units. The choice of symbols was made to be consistent with the heat transfer literature and with existing documentation for solar water heating systems. The symbols are listed in alphabetical order. For vector quantities, the units shown are for each component of the vector.
```
JacquesCarette, line 65: why was the "with no phase change material" dropped? This too seems like a mistake.

JacquesCarette (changes requested):

code/stable/swhs/SRS/HTML/SWHS_SRS.html
smiths: I used to think we needed capitalization in this case, but I looked into it recently and I was wrong.  The [grammar rule](https://www.aje.com/arc/editing-tip-capitalization-when-defining-abbreviations) is that when acronyms/abbreviations are expanded the words are only capitalized if they are proper nouns.

code/stable/swhsnopcm/src/csharp/README.md
smiths: Agreed!

code/stable/swhsnopcm/SRS/PDF/SWHSNoPCM_SRS.tex
smiths: I think "phase change material" was dropped because the existing documentation that is implicitly being referenced is the solar water heating system WITH phase change material, which we simply call the solar water heating system.  Is it an option to reference the other document?  If we point to the swhs (with PCM) documentation, then it is not longer ambiguous what existing documentation we are talking about.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: In #4363, we decided to remove system names from the automatically created tables of abbreviations. I wrote 9db8e738fd1e9340d954285674d1d5cc6d7a04c4 to ensure that happened. It relies on the `UID` of the chunk declared the "program name." For SWHS, we previously had two chunks that did this:

https://github.com/JacquesCarette/Drasil/blob/e537ae7473df2fa2d2378d360d0dcedf4decc8b9/code/drasil-example/swhs/lib/Drasil/SWHS/MetaConcepts.hs#L7-L15

However, only the latter one was used as the "program name:"

https://github.com/JacquesCarette/Drasil/blob/e537ae7473df2fa2d2378d360d0dcedf4decc8b9/code/drasil-example/swhs/lib/Drasil/SWHS/Body.hs#L60-L62

Hence, SWHS was never filtered out of the table of abbreviations and acronyms.

So, this change is technically in line with the other case studies. Also note: we specifically avoided instantiating `NamedIdea` and `Idea` for `ProjectName`. So this is a different 'kind' currently (in the ontology) as the other others.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: Thanks! Fixed in de95d30b8161af22bcec320be616e063afc21a47

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: This particular "Solar Water Heating Systems Incorporating PCM" is the project title. Isn't that an exception to this rule?

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: Fixed in ff47a2e2c

code/stable/swhsnopcm/src/csharp/README.md
balacij: Fixed in 0594fcddb

code/stable/swhs/SRS/HTML/SWHS_SRS.html
smiths: You are correct.  The project title is a proper noun and should be capitalized as it appears in the original title of the project.  We aren't introducing the abbreviations SWHS here, we are referring to the project title.

code/stable/swhsnopcm/SRS/PDF/SWHSNoPCM_SRS.tex
balacij: This code is generated as a result of the following line:

https://github.com/JacquesCarette/Drasil/blob/0594fcddb8c1b2a719bb596bfe60c8a5988b8c26/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/Body.hs#L94

Previously, this code wrote:

https://github.com/JacquesCarette/Drasil/blob/e537ae7473df2fa2d2378d360d0dcedf4decc8b9/code/drasil-example/swhsnopcm/lib/Drasil/SWHSNoPCM/Body.hs#L92

Now, using `progName` here appears to be an operational hack to get the right output, but using `progName :: CI` (SWHSNoPCM's name) would mean the document is specifically referring to itself (which is a meaningless statement). So what I did was have it refer to more broader SWHS literature. However, I used the `swhs :: IdeaDict`, not the `projName :: ProjectName` from SWHS because `Doc'` expects a `NamedIdea` (which we explicitly forbid `ProjectName` from instantiating):

https://github.com/JacquesCarette/Drasil/blob/0594fcddb8c1b2a719bb596bfe60c8a5988b8c26/code/drasil-srs/lib/Drasil/SRS/DocumentLanguage/Core.hs#L82-L91

Was the right idea to refer to the project name of SWHS? I didn't think so, because that project name was meant to refer to the specific SWHS instance, not the broader set of SWHSNoPCM-related documents. So, the `IdeaDict` sounded the locally most appropriate for now. I also opened the scope a bit -- the "no PCM" bit sounded unnecessary to me. I assume one is a subset of the other, or at least that if I knew one, I could probably understand >75% of the symbols in the other. But I could be wrong! I overlooked this change in my review.

From an ontological view, referring to an `IdeaDict` seems wrong, but so does a `ProjectName`. I don't think we want to strive for perfection here, however. That can be done in another PR (preferably after the `drasil-code`-related work has finished).

smiths: I agree with opening the scope to refer to solar water heating systems in general, thus dropping the no PCM.  The PCM system has everything that the noPCM version has, but the PCM version adds phase change material.  If the volume of the phase change material is set to 0, then the calculations should be the same as the noPCM version.  You are correct @balacij that much of the symbols and theories are in common between the two.

code/drasil-code/lib/Language/Drasil/Code/Imperative/DrasilState.hs
balacij: Dealt with in 33d0f1d5f560c18ef21b35627d9e02aa702030da

code/drasil-system/lib/Drasil/System/ProjectName.hs
balacij: Thanks! Implemented in ~~5739345e5c97cfb1658ba10ff0c80ffd8a5216e2~~ [a5743a1](https://github.com/JacquesCarette/Drasil/commit/a5743a196520c5d9a7e394603d9d4c65abd9f831)

balacij:
Ready for review again, @JacquesCarette @smiths.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
JacquesCarette: Yes, it was a mistake that SWHS was not filtered out before -- but it is still a mistake that it is not in the table, as it really is an abbreviation!

code/stable/swhs/SRS/HTML/SWHS_SRS.html
JacquesCarette: The question was specifically about "Incorporating", which used to be lower-case but now is upper case. Is that right?

code/stable/swhsnopcm/SRS/PDF/SWHSNoPCM_SRS.tex
JacquesCarette: Perfect.

JacquesCarette (commented):
All the code looks good - just those two outstanding things on the changes to stable.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
smiths: I think for the name of a project, the authors of the project's name get to decide on the capitalization.  When it is referred to by name, the same capitalization should be used each time.  Consistency is what matters.  That is one of the selling features of Drasil; we are consistent by construction.  :smile:

smiths:
Once the comments from @JacquesCarette are addressed, I'm okay with the changes to stable.

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: Fixed in ec8e8549a7ea156435d42f54c0769f0d0f0326ba

code/stable/swhs/SRS/HTML/SWHS_SRS.html
balacij: Fixed in 4a2a15d8016797b41e18eeca63f903436fc957b1

JacquesCarette (approved):

PR #5416, opened 2026-08-07
refactor(srs-gen): remove intro section 'document-reading entry' field in Organization of Document section generator

balacij:
Contributes to #5414

This field was used to indicate which section a reader can/should start reading the document from. Note that GlassBR and PDController mentioned "Data Definitions" section but hotlinked to the "Instance Models" section. Assuming those two are mistakes (whether it be a copy-pasting or choosing data definitions incorrectly), all case studies mention beginning with the "Instance Models" section first. Any SmithEtAl-formatted SRS should have an IMs section, therefore defaulting to it all the time is fine.

Alternatively, we can add an option mentioning the user can begin from the table of outputs (when relevant).


smiths (approved):
The changes to stable look good to me.  Thank you @balacij for catching the error in GlassBR and PDController.

balacij:
Thank you!

Note: I just had to rebase this PR because GitHub re-signed my commits as part of their new 'PR stack rebase' feature. My mistake.

JacquesCarette (approved):

PR #5419, opened 2026-08-10
Made `file` be a parameter

B-rando1:
Contributes to #5328 

Once every part of GOOL's current "AST" is parameterized, we can move to the next step of removing the repr type and changing the fundeps to be `r -> ...` to `<larger language feature repr> -> <smaller language feature repr>`.


B-rando1:
Making this a draft in favour of #5422

JacquesCarette:
I've merged #5422 now, so I guess this should be closed?

PR #5425, opened 2026-08-10
refactor(srs-gen/`PhySysDesc`): automatically retrieve program name from `SmithEtAlSRS` instead of manually passing around the program name through `SRSDecl`

balacij:
Contributes to #5414 

This PR improves how "program names" are reused, moving away from manually reuse towards automatic reuse in the SRSDecl declaration.

### DIFF
--- a/code/stable/swhs/SRS/HTML/SWHS_SRS.html
+++ b/code/stable/swhs/SRS/HTML/SWHS_SRS.html
@@ -916,7 +916,7 @@
               <div class="subsubsection">
                 <h3>Physical System Description</h3>
                 <p class="paragraph">
-                  The physical system of <span title="solar water heating system">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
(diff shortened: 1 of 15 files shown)


code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -916,7 +916,7 @@ <h3>Terminology and Definitions</h3>
               <div class="subsubsection">
                 <h3>Physical System Description</h3>
                 <p class="paragraph">
-                  The physical system of <span title="solar water heating system">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
+                  The physical system of <span title="solar water heating systems incorporating PCM">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
```
samm82, line 919: Other than this potentially undesired side effect, I love this change!

JacquesCarette: But the uniformity is well worth it. If we want to be able to use 'short forms', we should have a system for that, rather that having back doors that let us hack it in!

JacquesCarette (approved):

PR #5427, opened 2026-08-11
 Add `listLast` to fix redundant `- 1 + 1` in generated code

### LINKED ISSUE #5409: Redundant - 1 + 1 in generated code for 1-indexed languages
While working on the MATLAB renderer ([#5386](https://github.com/JacquesCarette/Drasil/pull/5386#discussion_r3714315244)), I noticed that generated code for 1-indexed languages has a redundant pattern when accessing the last element of a list:

`arr(length(arr) - 1 + 1)`

This could just be `arr(length(arr))`, since `- 1` and `+ 1` cancel each other out.

Right now, when we compute `listSize - 1`, nobody remembers that we subtracted 1. So when `intToIndex'` adds 1 later, it can't cancel them out. We could fix this by storing the offset (e.g. -1) as part of the expression data, so that a later +1 can detect and cancel it. This would fix the problem everywhere, but it means changing `ValData`, which is used across all languages, so the change could be big.

Or instead of computing `listSize - 1` and then converting back, add something like `listLast` to the GOOL interface that directly accesses the last element without going through the index translation. This is more targeted, but we'd need to find all the places that access the last element and switch them to use the new operation.

I'd like to discuss which direction makes more sense before starting.

Xinlu-Y:
This PR adds a `listLast` operation to the `List` typeclass so each language can handle it in its own way. MATLAB uses `arr(end)` or `arr{end}` depending on the element type, Julia uses `arr[end]`, and Python uses `arr[-1]` since that's more idiomatic. The other languages (Java, C#, C++, Swift) just fall back to `listAccess v (listSize v - 1)` since they don't have this problem.

Closes #5409


code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -383,6 +383,10 @@ instance Array JuliaCode where
 instance List JuliaCode where
   listSize = CS.listSize jlListSize
   listAccess = G.listAccess
+  listLast v = do
+    v' <- v
+    let t = A.innerType $ return $ valueType v'
+    mkStateVal t (RC.value v' <> brackets (text "end"))
```
B-rando1: Shouldn't `end` have type `int`?

This function could probably be simplified to something like:
```hs
let endVal = mkStateVal int jlEnd -- jlEnd is already defined
in listAccess v endVal
```

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -337,6 +338,11 @@ instance Array MatlabCode where
 instance List MatlabCode where
   listSize = CS.listSize "length"   -- length(v)
   listAccess = G.listAccess
+  listLast v = do
+    v' <- v
+    let t = A.innerType $ return $ valueType v'
+        innerCt = getInnerType $ cType $ unMLC $ valueType v'
+    mkStateVal t (RC.value v' <> mlCellWrap innerCt (text "end"))
```
B-rando1: I'm not sure if you can get rid of `innerCt`, but you might be able to simplify this function similarly to the Julia renderer.

B-rando1 (changes requested):
Looking good, just a couple of comments.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
Xinlu-Y: But `listAccess` internally calls `intToIndex`, which adds + 1 ?

B-rando1: Ahh right, I forgot about that!

B-rando1: Hmm, maybe wrapping the whole thing in `mkStateVal` like what you have is the best option then. We _could_ make a helper function `listAccessMaybeAdjustIndex` (but with a less awful name :smile:) that takes a boolean for whether it should do the `+ 1`, but I'm not sure if that would actually be better. What are your thoughts?

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/JuliaRenderer.hs
```diff
@@ -383,6 +383,10 @@ instance Array JuliaCode where
 instance List JuliaCode where
   listSize = CS.listSize jlListSize
   listAccess = G.listAccess
+  listLast v = do
+    v' <- v
+    let t = A.innerType $ return $ valueType v'
```
B-rando1:
```suggestion
    let t = innerType $ return $ valueType v'
```

It's probably better to use the Julia-specific `innerType` (which happens to be `A.innerType`).

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -337,6 +338,11 @@ instance Array MatlabCode where
 instance List MatlabCode where
   listSize = CS.listSize "length"   -- length(v)
   listAccess = G.listAccess
+  listLast v = do
+    v' <- v
+    let t = A.innerType $ return $ valueType v'
```
B-rando1:
```suggestion
    let t = innerType $ return $ valueType v'
```

B-rando1 (changes requested):
Just two tiny suggestions.

B-rando1:
The one other thought I had, the `listLast` approach is specialized only for accessing the last element of a list. Which might be fine. But it won't help, for example, if someone wanted the _second_-last element of a list.

I'm not sure if we could do this while generating nice code in all languages, but it feels like there's a more general version of `listLast`. E.g. Python lets you do `a[-n]` for the `n`th element from the end. It could be nice to recreate that with `listAccess`, but this would have the same problem as `listSlice` where we'd need to generate a bunch of runtime-checks if the value of the index isn't known statically. Alternatively, we could introduce `listAccessFromEnd`. That actually could work nice. It should be easy to make `listAccessFromEnd list (litInt 0)` reduce to be identical to `listEnd list` in this PR.

Sorry @Xinlu-Y for not thinking of this option until after your PR. Do you think some sort of `listAccessFromEnd` would be worth it?

Xinlu-Y:
> The one other thought I had, the `listLast` approach is specialized only for accessing the last element of a list. Which might be fine. But it won't help, for example, if someone wanted the _second_-last element of a list.
> 
> I'm not sure if we could do this while generating nice code in all languages, but it feels like there's a more general version of `listLast`. E.g. Python lets you do `a[-n]` for the `n`th element from the end. It could be nice to recreate that with `listAccess`, but this would have the same problem as `listSlice` where we'd need to generate a bunch of runtime-checks if the value of the index isn't known statically. Alternatively, we could introduce `listAccessFromEnd`. That actually could work nice. It should be easy to make `listAccessFromEnd list (litInt 0)` reduce to be identical to `listEnd list` in this PR.
> 
> Sorry @Xinlu-Y for not thinking of this option until after your PR. Do you think some sort of `listAccessFromEnd` would be worth it?

Great idea! I will go ahead and implement it.

B-rando1:
Sounds good, thanks! Let me know if you want to discuss the semantics of this new keyword at all.

Xinlu-Y:
> Sounds good, thanks! Let me know if you want to discuss the semantics of this new keyword at all.

Thanks, `listAccessFromEnd` is implemented and pushed!

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -321,6 +321,11 @@ class (IndexTranslator r) => List r where
   -- | Gets the value of an index of a list.
   --   Arguments are: List, Index
   listAccess :: SValue r -> SValue r -> SValue r
+  -- | Gets an element from the end of a list.
+  --   Arguments are: List, Offset (0 = last, 1 = second-to-last, ...)
+  listAccessFromEnd :: SValue r -> SValue r -> SValue r
+  -- | Gets the last element of a list.
+  listLast   :: SValue r -> SValue r
```
B-rando1: Since all the implementations are the same, we can probably split this off into a function, similar to `listSlice` or `funcApp`.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -337,6 +339,17 @@ instance Array MatlabCode where
...
   listAccess = G.listAccess
+  listAccessFromEnd v n = do
+    v' <- v
+    n' <- n
+    let t = innerType $ return $ valueType v'
+        innerCt = getInnerType $ cType $ unMLC $ valueType v'
+        idx = case RC.valueInt n' of
+          Just 0  -> text "end"
+          Just i  -> text "end" <+> text "-" <+> integer i
+          Nothing -> text "end" <+> text "-" <+> RC.value n'
```
B-rando1: Is there anything different in the output between the `Just i` and `Nothing` branches? We'll need to test it, but we might be able to merge them with a `_` branch. Same for Julia.

code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/SwiftRenderer.hs
```diff
@@ -447,6 +447,8 @@ instance Array SwiftCode where
 instance List SwiftCode where
   listSize = C.listSize' swiftListSize
   listAccess = G.listAccess
+  listAccessFromEnd v n = listAccess v (G.smartSub (listSize v #- litInt 1) n)
```
B-rando1: We should have a test case to confirm that `smartSub` does what we want for a non-zero `n`. I believe that here because the integer value of `listSize v` isn't known statically, the outer `smartSub` won't be able to constant-fold.

We might have to reorganize it to `(listSize v #- (G.smartAdd (litInt 1) n))` so that `smartSub` can do its job.

code/drasil-gool/lib/Drasil/Shared/InterfaceCommon.hs
```diff
@@ -321,6 +321,11 @@ class (IndexTranslator r) => List r where
   -- | Gets the value of an index of a list.
   --   Arguments are: List, Index
   listAccess :: SValue r -> SValue r -> SValue r
+  -- | Gets an element from the end of a list.
+  --   Arguments are: List, Offset (0 = last, 1 = second-to-last, ...)
```
B-rando1, line 327: Good that you made the indexing scheme explicit! I agree that 0-indexed from the end makes the most sense.

B-rando1 (changes requested):
Looks like good progress! Just a few comments.

JacquesCarette:
Love the conversation!

B-rando1 (changes requested):
I'm happy with this. The last thing I'd like to see is a test for `listAccessFromEnd (valueOf myOtherList) (litInt 1)`. Just to make sure that constant-folding works there as well.

PR #5431, opened 2026-08-12
Removed `MethodSym` as a constraint on `ModuleSym`

B-rando1:
Contributes to #5328 

Decoupling the typeclasses seems to have two benefits: it reduces the number of type variables on most type classes, and it's necessary to do before we remove `repr`.


JacquesCarette (approved):

JacquesCarette:
Arg, I should not have done the 'rebase' myself, now I can't merge! @balacij , care to do the honours?

balacij (approved):

PR #5438, opened 2026-08-14
Introduce `ReasonableValue` ADT and remove `MayHaveRationale`

djarabek:
The rationale is now captured in the ADT, rather than being a separate field with a separate typeclass.

### DIFF
--- a/code/drasil-lang/lib/Language/Drasil/ReasonableValue.hs
+++ b/code/drasil-lang/lib/Language/Drasil/ReasonableValue.hs
@@ -1,0 +1,30 @@
+{-# Language TemplateHaskell #-}
+
+module Language.Drasil.ReasonableValue (
+  -- * Reasonable Value
+  ReasonableValue,
+
+  -- ** Constructors
+  reasonableValue,
+
+  -- ** Lenses
+  reasV, rationale
+) where
+
+import Control.Lens (makeLenses)
+
+import Language.Drasil.Expr.Lang (Expr(..))
+import Language.Drasil.Sentence (Sentence)
+
+-- | Represents a reasonable value ('Expr') with an optional
+-- rationale ('Maybe Sentence').
+--
+-- A reasonable value is an example of a value that is physically
+-- admissible.
+data ReasonableValue = RV { _reasV     :: Expr
+                          , _rationale :: Maybe Sentence
+                          }
+makeLenses ''ReasonableValue
+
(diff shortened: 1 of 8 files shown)


code/drasil-lang/lib/Language/Drasil/ReasonableValue.hs
```diff
@@ -0,0 +1,30 @@
...
+
+-- | Represents a reasonable value ('Expr') with an optional
+-- rationale ('Maybe Sentence').
+--
+-- A reasonable value is an example of a value that is physically
+-- admissible.
+data ReasonableValue = RV { _reasV     :: Expr
+                          , _rationale :: Maybe Sentence
+                          }
+makeLenses ''ReasonableValue
```
balacij, line 27: Should situations where `ReasonableValue`s don't exist come with an explanation of why? Do we have any `Nothing`s for reasonable values?

djarabek: The current implementation didn't allow a rationale without a reasonable value, so I stuck with that (though I see the argument for allowing it). There are currently a bunch of things with no reasonable value.

smiths: If we can't come up with a reasonable value, then there should definitely be an explanation.  However, is there ever a case where we can't come up with a reasonable value?  Our definition of reasonable is a value that satisfies the constraints.  The bar for a reasonable value isn't as high as the bar for a typical value.  I can't thin of a case where we couldn't think of a reasonable value.

smiths: > The current implementation didn't allow a rationale without a reasonable value, so I stuck with that (though I see the argument for allowing it). There are currently a bunch of things with no reasonable value.

My guess is that we don't have reasonable values because we were lazy, not because it is impossible to come up with reasonable values.  Laziness is to be expected though, and reasonable values aren't always all that interesting.  Allowing a rationale without a reasonable value wouldn't be wrong, but I'n not sure if there would ever be a rationale other than "it isn't a priority to come up with a reasonable value right now."  :smile:

balacij: It sounds like we should try to remove the option to not have a reasonable value then! :)

JacquesCarette (approved):

Issue #2895, opened 2021-11-16
Inlining Exprs in `Sentence`s without Expr data constructors


balacij:
Resulting as a subtask from #2893, which has a bit of discussion of the context

>  I think the trick would be to have `Sentence` have a parameter or two. Or at least `SentenceF` would, and then `Sentence` for external consumption would "close over" particular parameters of `SentenceF`.

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/issues/2893#issuecomment-968295439_

JacquesCarette:
The point would be that there would be a 'hole' in `SentenceF` that could be filled by anything. That way `SentenceF` would be parametric in that, and would not **depend** on `Expr`. When we create `Sentence`, which could be in an entirely different package, it can fill the hole of `SentenceF` with `Expr`.

That should let us decouple various things (and reveal bad assumptions too).

balacij:
In light of more recent discussions on chunks/concepts, display knowledge, etc., let's take a step back and revisit the requirements.

`Sentence` is our DSL that lets us write things, where we direct attention towards things, relate things, etc. 

At times, through `Sentence`, we need to refer to theories, specific theories, mathematical expressions, expressions, the list of months, and/or specific months.

Let's consider the following "`Sentence`": "In January, I wrote my Comp II document which contained a section on theories. I used Projectile's landing position instance model (a theory) as a running example in the document, though I had to slightly modify the formula it defines ($p_{land}=...$) to fit within page margins."

How should this be encoded in Drasil? Not necessarily with Haskell code, but _maybe_ with pseudocode.

For starters, let's talk about the important words would be encoded:
1. Projectile: a `SystemInformation` chunk (we don't currently consider `SystemInformation` a chunk, but I think we should),
2. January: a [`Month`](https://github.com/JacquesCarette/Drasil/blob/22169038db12adb1c686ffa3cda229ceaa5e5537/code/drasil-lang/lib/Language/Drasil/Data/Date.hs#L10),
3. Comp II: a `ConceptChunk`,
4. document: a `ConceptChunk`,
5. a section: a `ConceptChunk`,
6. landing position: an `InstanceModel`,
7. $p_{land}=...$: an `Expr`.

Next, let's look at how they would be referenced under the current `Sentence`-scheme:
1. The chunk names: `phrase landingPos` or `sy landingPos`, for example, should write `landing position` or $p_land$, respectively, depending on whether we want to reference the English designator or the math designator.
[...]

PR #5440, opened 2026-08-14
refactor: move `Quantity` constraint->`Sentence`-fold combinator to sole usecase area (`drasil-srs`)

balacij:
Contributes to JacquesCarette/Drasil#2895

This de-couples 'quantities' just a bit from `Sentence`.


samm82 (approved):
This looks good and makes sense to me!

JacquesCarette:
There is now a conflict. And it feels like it should be re-based, as I'm seeing lots of changes that I know are merged.

balacij:
Those PRs were merged into other PRs and then ultimately here. So the commits are not on `main`. Originally I intended for these PRs to be part of the 'stack' feature, but it doesn't seem like that works too well yet.

JacquesCarette (approved):

PR #5441, opened 2026-08-14
refactor(`drasil-lang`): move `Citation` to `Language.Drasil.Document` namespace

balacij:
This is a follow-up to #5089 and related PRs, moving code around according to the `drasil-lang` module dependency graph.


JacquesCarette (approved):
I really don't think that citations belong with 'document' -- that's merely where they get rendered! 

Nevertheless, the code is definitely improved by this, and 'document' is probably less bad than being sort of 'global' to the Drasil namespace. So I'm merging, but this will move again!

PR #5443, opened 2026-08-15
refactor(`drasil-lang`): move document "label"-related data types to `.Document` namespace

balacij:
Similar to #5441, this is a follow-up to #5089 and related PRs, moving code around according to the `drasil-lang` module dependency graph.


JacquesCarette (approved):
Unlike for citations, it is probably true that all our labels are document related and are all for 'printing' purposes, and don't have a meaning outside of documents.

Issue #4038, opened 2025-04-28
`ShortName`: Redesign and move to `drasil-docLang` or `drasil-printers`


balacij:
[https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-lang/lib/Language/Drasil/ShortName.hs#L8-L28

https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-lang/lib/Language/Drasil/ShortName.hs#L13-L18](https://github.com/JacquesCarette/Drasil/blob/5705da6c270bd1839801037c71a5e8466072e655/code/drasil-lang/lib/Language/Drasil/ShortName.hs#L6-L18)

A `ShortName` is the human-readable reference address/symbol of a displayed chunk. For example, for an InstanceModel, it is of the form "IM:X," and for a bibtex entry, it is the key itself. Both are immediately shown in the rendered SRS documents. This is not fundamentally important knowledge to writers of theories and it is not a decision that should be imposed at time of creating a theory (though perhaps at time of declaring which theories are IMs/TMs/DDs/GDs).

What we really care about when it comes to `ShortName`s is having a strategy for giving display names for certain chunks we render in the SRS. That is important _after_ the theories are created. In the code itself, when making theories, we might do these two actions at the same time, but that would come in some sort of `WrappedTheory` (which might end up being our existing IM/TM/GD/DD containers!) rather than a pure theory itself (which we want to move IMs/TMs/GDs/DDs to).

Should `HasShortName` and `ShortName` stay the way they are? I don't think so. I think `HasShortName` should be changed to something along the lines of `ReferenceNameStrategy`.

This might also be something we redesign entirely at the same time as the referencing system.

Also, `ShortName` should be wrapping an `NP` rather than a `Sentence`.

JacquesCarette:
I agree with the analysis (of what a `ShortName` is). And thus that maybe it is information that should be given to various items at the *use site* instead of at the *definition site*.

Maybe we need "semantic theories" and "displayable theories"? A displayable theory would wrap a semantic one -- indeed, like your proposed `WrappedTheory`).

I would tend to not wait. Fixing this now will still be an improvement, even if it gets further redesigned later.

PR #5444, opened 2026-08-15
refactor(`drasil-lang`): move document `ShortName` to `.Document` namespace

balacij:
Contributes to JacquesCarette/Drasil#4038

Similar to #5441 and #5443, this is a follow-up to #5089 and related PRs, moving code around according to the `drasil-lang` module dependency graph.


JacquesCarette (approved):
Much more mixed about this. I'd love to know what "short name" means... if even as a purely operational definition (i.e. how we use it). I think I used to know this...

We so desperately need a Drasil Glossary!

Issue #3691, opened 2024-02-14
'Theories': Shift association with IM/DD/TM/GD terms to 'printing' level


balacij:
Also, rename `ModelKind(s)` to `TheoryKind(s)`.

Right now, theories are heavily associated with IMs/DDs/TMs/GDs. We want to remove that association in preparation for work that @smiths is doing on Projectile. However, we want to be able to also keep the same functionality by shifting grouping association to the display-knowledge level.

PR #5445, opened 2026-08-15
refactor(`drasil-lang/../Document`): move to new package: `drasil-document-writing`

balacij:
Contributes to #2885 

With #5444, all modules related to "document writing" are moved to the `Language.Drasil.Document` namespace. Hence, `drasil-lang` is ripe for splitting off another package: `drasil-document-writing`.

I haven't yet changed the name of the namespace in `drasil-document-writing` yet, but that is a good next step. However, I will wait for a review of this PR to confirm the new package name. Idea for the new namespace: `Drasil.Document.Writing`. However, when `drasil-printers`' AST is eventually rebuilt, we might consider also moving it to the same package. In which case, we might move it to a `Drasil.Document.Rendering` subpackage. So, perhaps `drasil-document-writing` is an inappropriate name. Or maybe this is an argument for having two separate packages entirely really. That would be closer to following the Parnas "one secret per package" rule.

Next steps for this general area of work would involve (in no particular order):

* Refactoring `Sentence` (because we currently still cheat with getting `Reference`s inserted into them as components) to enable extensibility (#2895) -- this might become an argument for moving `Sentence` into `drasil-document-writing` as we work through the task... (I'll say I doubt it for now).
* Inverting the relationship packages have with `drasil-printers`.
* Merging the rendering AST and renderers in `drasil-printers` into `drasil-document-writing`?


code/drasil-theory/package.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
 - base >= 4.7 && < 5
 - lens
 - drasil-database
+- drasil-document-writing
```
JacquesCarette, line 24: This is strange to me: why does `drasil-theory` depend on `drasil-document-writing` ?

code/drasil-metadata/stack.yaml
```diff
@@ -28,10 +28,11 @@ resolver: lts-24.51
 #   - wai
 packages:
 - .
-- ../drasil-database/
-- ../drasil-lang/
-- ../drasil-utils/
+- ../drasil-database
+- ../drasil-document-writing
```
JacquesCarette, line 32: same here: why?

JacquesCarette (changes requested):
I'm fine with `drasil-document-writing` for now, we can rename it when its true purpose becomes clear.

However, it does seem to be depended on by packages that I think don't actually use any of it. So that's curious.

code/drasil-theory/package.yaml
balacij: IMs, GDs, TMs, and DDs are all display-focused knowledge. They implement typeclasses defined in `drasil-document-writing` (i.e., `HasRefAddress` and `Referable`). I believe these types should be moved towards `drasil-srs` so we can remove this dependency. We have #3691 for that. I expect us to to work on it more deeply after the GHRC-related work, unless said work becomes blocked on this.

code/drasil-metadata/stack.yaml
balacij: It contains `Citation`s (defined in `drasil-document-writing`). I don't know if these citations should really be in `drasil-metadata`, however. They look like they belong in `drasil-srs` (because this package directly uses some of them) or `drasil-data`.

https://github.com/JacquesCarette/Drasil/blob/52f57485bd3d54e36567c10c3a8e5423da6e624a/code/drasil-metadata/lib/Drasil/Metadata/Citations.hs#L13-L52

code/drasil-theory/package.yaml
JacquesCarette: IMs, GDs, etc are all hacks-for-theories. So some of the information belongs in `drasil-theory` while some other might be display only, and maybe that does belong in `drasil-srs`.

Having said that, references and referencing has noting to do with documents! They are a very basic idea that lets you cross-link "stuff". As a hack, I'd be fine if `HasRefAddress` and `Referable` were temporarily defined in `drasil-database`... and if that's not possible (likely) then back to `drasil-lang` they go!

code/drasil-metadata/stack.yaml
JacquesCarette: All data that Drasil uses itself should be in `drasil-metadata`. `drasil-data` is user-facing data, and nothing in Drasil should use `drasil-data`; `drasil-examples` is not *in* Drasil...

`Citation` is a fundamental concept that transcends documents. Like `Person` does. It belongs somewhere where we define such things. That's currently `drasil-lang`. I'm fine with splitting off a `drasil-concept` package from  `drasil-lang` (which would mostly define `data` types in Haskell)

code/drasil-theory/package.yaml
balacij: Yes, I agree. That will be valuable "next steps" once we continue the shallow file analysis.

`HasRefAddress`, `Referable`, `Reference`, and `Citation` are very heavily influenced by LaTeX and are exclusively used in rendering in documents. `HasRefAddress` and `Rererable` help find pretty-printable inline links to other 'display-oriented', rendered chunks (e.g., `ConceptInstance`) and external links. [`Reference` and `Citation` are backwards.](https://github.com/JacquesCarette/Drasil/blob/main/People/Jason/drasil-analysis/07-15-2026-References.md)

This package does not need to be the final home to anything moved to it, but splitting off code from `drasil-lang` will simplify future analysis and work. As a kind of "bad but locally 'explainable code' maxima," moving this code to `drasil-document-writing` (/`-building`) makes sense. A revised design, clear problem, and clarified terminology would go a long way before we try to settle a final location for any of the code, however, of course.

code/drasil-metadata/stack.yaml
balacij: I agree that "references to source materials" are an important concept that transcends document _writing_ (not documents). It definitely does not ultimately belong in this package. I would be happy to move it to a new package, but maybe that would be best in a new PR after some more analysis? At the same time we do that, I would like to see things renamed.

PR #5447, opened 2026-08-17
chore(`-printers`): remove unused `HARDNL` constructor in `Spec`

balacij:
"Hard newline" was never used. According to its comments, it was only meant to be temporary. We can (and should) remove it before there is any usage of it.


JacquesCarette (approved):
Nice catch.

PR #5423, opened 2026-08-10
refactor(srs-gen/`IntroSec`): reuse "program name" information (rather than feeding in from `SRSDecl`)

balacij:
Contributes to #5414

Closes #3529

Note: this one has `stable/` changes!

This PR reuses the "program name" field of the SRS encoding so as to avoid having users frequently readdress the program name `CI`. As a result, some `Sentence`-hacking was uncovered in both SSP and SWHS[NoPCM]: adding in an extra note, explaining that the projects were based on manually-created artifacts.

This PR makes the introduction section of each program a bit more uniform ~~and ensures program name abbreviations are introduced next to the first mention of the full program title~~, ensuring the first sentence of the second paragraph of the introductory paragraphs in the "Introduction" sections of the SRS is of the form: `The following section provides an overview of the software requirements specification (SRS) for <NAME>.`, where `<NAME>` is always the short form name of the project. For example, [BSS](https://jacquescarette.github.io/Drasil/examples/bss/SRS/HTML/BSS_SRS.html#Sec:Intro) and [DblPend](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/HTML/DblPend_SRS.html) use the "long form" names in title case, [GamePhysics](https://jacquescarette.github.io/Drasil/examples/gamephysics/SRS/HTML/GamePhysics_SRS.html#Sec:Intro) and [GlassBR](https://jacquescarette.github.io/Drasil/examples/glassbr/SRS/HTML/GlassBR_SRS.html#Sec:Intro) use the short form, [SWHSNoPCM](https://jacquescarette.github.io/Drasil/examples/swhsnopcm/SRS/HTML/SWHSNoPCM_SRS.html#Sec:Intro) uses a lowercase expansion of SWHSNoPCM. In a subsequent PR/issue, we would refine the introduction sections. There is considerable duplication between them (e.g., each contain introduction sentences introducing the short forms, linking to the "original, manually created version," with duplicated Haskell code, but all have unimportant wording variations).

PR #5460, opened 2026-08-20
fix(pdcontroller): expand project title

balacij:
Targets #5423

@smiths and I briefly discussed the expansion of "PDController." Assuming we want the expanded version as the title of the software project, this PR expands the project ~~name~~ title from "PD Controller" to "Proportional Differential Controller."

(Note: I'm not sure if that _is_ what we want, so feel free to close this PR if not.)


smiths (commented):
I don't think we want the name of the project to be Proportional Differential Controller.  That is a very long name.  When I read where this name appears in the changed files, it doesn't feel right.  The name of the project as PDController makes sense to me.  That is what people will call it when the refer to it.  Explaining where the P and D come from in the name would be good, but I don't think we should treat this as an abbreviation.  

We can make an analogy to Drasil.  The name of our project is Drasil, but we will also want to explain the name Drasil in our documentation.  It isn't an acronym, but it does have an origin.  I think of PDController like that.  The proper name of the project is PDController, not an expanded form, just like the proper name of Drasil is Drasil, not the explanation of how we came up with the name.

We can see what @JacquesCarette says, but my instinct is to not accept this PR.

balacij:
We previously had set both the project title and abbreviation to `PD Controller`. What this PR does is change the title, not the abbreviation. Perhaps the title can be expanded further even. The downstream effects are caused by the SRS code generator. Perhaps what this PR means is that for those areas, we should be writing the abbreviation instead of the full title (assuming a "title" is something we want). This is a change that would then affect the other case study software artifacts.

What kinds of names more broadly do (software) projects have? With `ProjectName` (#5414), I assumed 3:

1. Titles ("descriptive name" alternatively)
2. Abbreviations (primary "name" used in normal conversation)
3. Repository names (preferred repository slug/name)

Related: #5423

JacquesCarette:
> What kinds of names more broadly do (software) projects have? With ProjectName (https://github.com/JacquesCarette/Drasil/pull/5414), I assumed 3:
>
> 1.    Titles ("descriptive name" alternatively)
> 2.    Abbreviations (primary "name" used in normal conversation)
> 3.    Repository names (preferred repository slug/name)

I don't like your labels... your long forms in parens are much better.

1. Descriptive Name
2. Name
3. Repository Name

where the need for 3 is because we want more freedom in 'Name' than we would be afforded in Name.

We could split 2 into a. Official Name and b. Informal Name.

smiths:
I agree with the response from @JacquesCarette [above](https://github.com/JacquesCarette/Drasil/pull/5460#issuecomment-5364318115).  The idea of having an official name and an informal name also makes sense to me.  It wouldn't always be relevant, but that would give us additional flexibility.

balacij:
This work just needs to be redone. There's no point in trying to salvage anything here because of how small the diff is.

PR #5463, opened 2026-08-21
demo: cabal build

djarabek:
Notes:
- `cabal build all`, `cabal test all` and `cabal run <example>` all work
- the `.cabal` files are currently just the hpack ones generated by stack
- `code/scripts/patch_stackage_cabal.sh` downloads and patches the appropriate cabal config file from stackage. Currently the only thing that needs to be patched is removing the constraint on the `template` package as we have a local package named template.

Contributes to: #5160


balacij:
One thing I was really hoping for with #5160 was to merge the `.cabal` files. That way we can use [cabal stanzas](https://vrom911.github.io/blog/common-stanzas) to further deduplicate the `.cabal` files. I think we should focus on that.

EDIT: So, I don't think we should pursue this PR.

PR #5464, opened 2026-08-21
Demo: UID namespacing playground

djarabek:
By uncommenting the various `*Ns` functions, you can play around with what happens when UIDs of certain chunk types are guaranteed to be unique.


JacquesCarette:
Do you have an example run stashed away somewhere?

Issue #4956, opened 2026-05-06
GOOL: fix lambdas and function types for C++ and Java


B-rando1:
TIL that GOOL lambda expressions are never used, other than within some of the renderers as part of implementing other expressions.

I added a test to `HelloWorld.hs` that creates a lambda and stores it in a variable, then calls it and prints the output. Lo and behold, it's broken for C++ and Java. There are a few issues:

### Empty type for lambda variables

In Java and C++, declaring a variable with a function type prints nothing for the type. E.g. in C++ you get `myLambda = ...` instead of `auto myLambda = ...`.

- This is fairly easy to fix in C++ (at least for declaring variables): from a quick look around, the most idiomatic way to type a lambda abstraction is to give it the `auto` type and let the compiler figure it out.
- In Java this would be a bit of a pain. Lambdas need to have a type that is a "Functional Interface", basically a class with a single method that is instantiated by the body of the lambda. There are a few preset interfaces in [java.util.function](https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html), but the specific interface to use depends on how many inputs and outputs the lambda has, and there aren't interfaces for arbitrary numbers of input parameters. There are two possible routes we could take to work through this:
  - The more idiomatic way is to use the best interface for our needs. If the interface is in `java.utils.function`, we use that. If it isn't, then we generate an interface for it. This should technically be possible in GOOL, but it would be a non-trivial project. I don't think it would be worth the effort, to be honest.
  - The easier way is to always use the single-input interface, and pass it an Object array. There would still be a few hurdles to overcome, but it should be possible with a bit of effort.

### Syntax errors with defining lambda expressions

C++ adds an `->` between the parameters and body of the lambda, which is not supposed to be there. We can just take that out.

### Syntax errors with calling lambda expressions

- It's a bit of a hack to even call lambdas at at all - `funcApp` takes a `String` for the function, not a `Value`.
- Specifically in Java, we can't call lambdas as a function, instead we need to call `.apply()`.
- Probably the best solution would be to create a new `lambdaApp` structure in GOOL.

## Conclusion

After looking more closely at lambdas in GOOL, they seem rather odd. They're part of the external interface, but they're treated as if they're part of the internal interface - helper functions that are only used by implementations of the external interface. If they're used in a GOOL file, then the Java and C++ renderers silently generate broken code. Clearly, this is a problem.

Here are my suggested action items:

[...]

Issue #3970, opened 2024-09-20
Implementing `setUnion` in C++.


Xinlu-Y:
https://github.com/JacquesCarette/Drasil/blob/42e4ca133694e80256d42ee388aa19cfaa17cbfb/code/drasil-gool/lib/Drasil/GOOL/LanguageRenderer/CppRenderer.hs#L1390-L1394
At the moment, `setUnion` is not implemented in C++ yet. The standard way to compute a set union is via `std::set_union`, but this function does not modify sets in-place. Instead, it computes the union of two sets and stores the result in a third set.

Is there a recommended way in C++ to perform an in-place set union similar to Julia's `union!` or Java's `addAll`?
I found that using the `insert` method (e.g., a.insert(b.begin(), b.end());) can achieve an in-place union by adding elements from one set to another directly. To implement this, I need to iterate over the second set and insert each element into the first set. How can I achieve this iteration and insertion using a loop, such as `ForEach`, to go from `b.begin()` to `b.end()` and add the elements into `set a`?

JacquesCarette:
I don't think it is a strong requirement to do set union in-place. I think `std::set_union` is just fine, and can be replaced at a later time with more imperative code if the need truly arises.

balacij:
Aside: out of curiosity, why does the definition of `contains` refer to something called `containsInt`? Are we currently only able to handle membership checking for integers?

C++ `std::set_union` has an interesting [definition](https://en.cppreference.com/w/cpp/algorithm/set_union). Some notes:
1. it expects that we give it two iterators for each set: one iterator starting from the first element (i.e., `begin`), and another that starts just past the last element (i.e., `end`),
2. the return value is an output iterator that points an imaginary element past the last element in the new set,
3. it expects an 'output iterator' (i.e., the data writer). I assume we'll need to use the same one for a `set`, specifically a `std::inserter` because that one preserves order of the set (note: sets in C++ are ordered). However, in order for us build an `inserter`, we need references to the set that we're inserting into. So, it looks like we need to have a variable pre-defined for each set union, so that we can insert into it. (Am I getting that right, @Xinlu-Y ?)

Some comments on each:
1. The syntax is going to be noisy.
2. We will struggle to generate code for expressions like $e \in (P \cup Q)$ because the hypothetical 'return' of $P \cup Q$ (as translated to `set_union`) will not be the set itself, but an iterator that points to one past the last element in the set.
3. Again, the syntax is going to be noisy.

[...]

PR #5467, opened 2026-08-25
Implement missing methods in MATLAB renderer

Xinlu-Y:
Filled in all remaining `undefined` methods in `MatlabRenderer`. Also added a test block for set ops and lambda, and verified the generated MATLAB.

With this PR, `MatlabRenderer` has all its methods implemented.

### DIFF
--- a/code/drasil-code/test/HelloWorld.hs
+++ b/code/drasil-code/test/HelloWorld.hs
@@ -13,7 +13,7 @@ import Drasil.GOOL (Class, SVariable, CS, MS, OOProg, BodySym(..),
   ValueExpression(..), extFuncApp, newObj, Reference(..), Array(..), List(..),
   ListStatement(..), InternalList, MethodSym(..), OOMethodSym(..), objMethodCall,
   classMethodCall, initializer, OODeclStatement(objDecDef), Set(..),
-  ParameterSym(..))
+  ParameterSym(..), BinderSym(..))
 import qualified Drasil.GOOL as OO (GSProgram, ProgramSym(..), FileSym(..),
   ModuleSym(..))
 import Drasil.GProc (ProcProg)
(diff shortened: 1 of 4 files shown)


code/drasil-code/test/HelloWorld.hs
```diff
@@ -139,6 +139,34 @@ helloInitVariables = block [comment "Initializing variables",
   assert (contains (valueOf (var "s" (setType int))) (litInt 7))
     (litString "Set s should contain 7")]
 
+helloProcTests
```
B-rando1, line 142: These tests are good additions! It makes sense that we can't add them as tests to all the generators, because of #3970 and #4956. With that said, these aren't "procedural" concepts. We'd like them to work for all of our target languages one day, and IIRC they each currently fail for only one language. Because of that, I would prefer if we split these into separate functions and added them as tests to as many languages as we can.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
```diff
@@ -118,8 +121,8 @@ instance RenderFile MatlabCode FileData ModData where
   fileFromData = A.fileFromData (onCodeValue . fileD)
 
 instance ImportSym MatlabCode where
-  langImport = undefined
-  modImport = undefined
+  langImport _ = toCode empty
+  modImport _ = toCode empty
```
B-rando1, line 125: Does MATLAB not require imports of local files or external libraries?

B-rando1 (commented):
This looks like good work! It might be worth rearranging the new tests a little bit, so that we can use them with as many languages as we can. I'm not sure what the best way to split things up would be. It might be easiest to split up `helloWorldOO` and have different versions for different languages, depending on what they support. Or maybe there's a better way, let me know what you think.

code/drasil-code/test/HelloWorld.hs
Xinlu-Y: Makes sense! I'll split them into separate functions and add them to the OO tests too.

code/drasil-gool/lib/Drasil/GProc/LanguageRenderer/MatlabRenderer.hs
Xinlu-Y: MATLAB doesn't need explicit imports for the code we generate: helper functions are just `.m` files in the same directory and can be called directly, and built-in functions are always available. MATLAB does have an import statement for package namespaces, but we don't use packages in the generated code.

balacij: Doesn't that imply that for code we should be able to generate (e.g., code that references external libraries or locally built packages), that this generator would produce bad output? For example, if we generated a local folder (a "namespace") and we tried to use code from said folder (or a purely external library), that no import statements would be created?

If we had a local "namespace," does code in said namespace not still need to use imports for things within its own namespace? Or does it also automatically import things within any file in its namespace?

PR #5475, opened 2026-08-27
refactor(`ConceptChunk`s): remove unused `ConceptDomain` lists

balacij:
Contributes to #5404

This aligns with earlier analysis: <https://github.com/JacquesCarette/Drasil/issues/5404#issuecomment-5196276774>


JacquesCarette (approved):
I hate removing this kind of knowledge... but it is getting in our way. So okay to remove now, knowing it will eventually come back.

Issue #5474, opened 2026-08-27
#5472 analysis: subtle bug in `insertAllOutOfOrder13` and logical issue with `HasChunkRefs` (ignores types)


balacij:
See https://github.com/JacquesCarette/Drasil/pull/5472's patch:

```
physicalcon = [gaseous, liquid, solid, ctrOfMass, dimension, flexure]
```

to:

```
physicalcon = [gaseous, liquid, solid, ctrOfMass, density, specWeight, mass,
  len, dimension, vol, flexure]
```

However, this resulted in the CI failing with a seemingly bizarre, unrelated error:

```console
swhsnopcm: Failed to find chunk `volume` (expected type: DefinedQuantityDict)
CallStack (from HasCallStack):
  error, called at lib/Drasil/Database/ChunkDB.hs:88:26 in drasil-database-0.1.1.0-8Pro8mqs7mGibDJOcvy8m:Drasil.Database.ChunkDB
```

Huh?! For this error to have occurred, it means one of two things:

1. `volume` is not in the `ChunkDB`. **FALSE**. The PR specifically added the `ConceptChunk`.
2. A reference to `volume` was expected to be of a different type (bing bing bing). In the PR, we inserted a `ConceptChunk` but there is a reference to another chunk with `UID = volume` but of type `DefinedQuantityDict`. This is the chunk resolution that fails.

So, what does this mean? Two issues:

1. Our method of inserting chunks en masse (`insertChunksOutOfOrder13`) has a subtle bug: when two lists carrying elements sharing the same `UID` are inserted, only the second one is actually inserted in the `ChunkDB`. It happens that with the original PR, the `ConceptChunk` was overriding the `D-Q-D-` (!).
   https://github.com/JacquesCarette/Drasil/blob/0ea5a1821724fb72e677ad0162ac9bec4a6175e2/code/drasil-database/lib/Drasil/Database/ChunkDB.hs#L211-L218
2. Our method of checking chunk references is weak because it does not also check the types of things associated. The error that occurred should have happened much earlier: when constructing the `ChunkDB`!

balacij:
For (2), I think a suitable solution is to switch:

```haskell
newtype UnitypedUIDRef = UnitypedUIDRef UID

...

class HasChunkRefs a where
  chunkRefs :: a -> S.Set UID
```

to

```haskell
data UnitypedUIDRef = UnitypedUIDRef UID TypeRep

...

class HasChunkRefs a where
[...]

PR #5478, opened 2026-08-30
fix(`ChunkDB`): catch duplicate `UID`s during mass chunk insertion

balacij:
Contributes to #5474 (but will fail the CI until we fix the existing issues)

`insertAllOutOfOrder13` previously used `fromList` when building the incoming chunk map, which silently discarded chunks sharing the same `UID`. Now using `fromListWith` to detect and report `UID` conflicts.


JacquesCarette (approved):
Definitely an improvement!

balacij:
Status update: To get the CI passing, we need to deal with `GlassBR`. #5482 is one proposed hacky fix for now.

JacquesCarette (approved):

PR #5482, opened 2026-08-31
hacky-fix(glassbr): Avoid double-inserting quantities

balacij:
Contributes to #5478

This changes `stable/` (semantically erroneously) but technically correctly because we have two separate concepts of annealed, fully tempered, and heat strengthened glass and each have an abbreviation. For each, we have a `CI` (which has a noun phrase and an abbreviation) and a `ConceptChunk` (which has a noun phrase, an abbreviation, and a description). The reason why we have both is because we want to have a Haskell-type-level guarantee that a the chunk representing these types of glass _have an abbreviation_. The `ConceptChunk` does not strictly guarantee that. Only that it _might_ have one. The `CI`, however, does.

So, what's the solution? We can either "cheat" by filtering out duplicate abbreviations before building the Table of Abbreviations and Acronyms even if they come from chunks with different `UID`s, or we can revisit the ontology. The first is a hack that can mask other logical issues (e.g., same abbreviations but actually-different-concepts) and the second is time-consuming (an issue I'd rather punt to later). I would personally prefer having the duplicate abbreviations for now until we have (2) dealt with correctly.

Along the way, we also remove a duplicate chunk instance (glass / idglass).


smiths:
This seems very odd to me.  Why do we have to have two separate concepts of annealed etc.?  Why can't we have one concept?  Why not just use `CI`?  If they truly are different concepts, when they should have different abbreviations.

samm82:
Echoing Dr. Smith's comment here, but in favour of keeping the `ConceptChunk` (since it has more information: the description, etc.) - my big question is _why_ do "we want to have a Haskell-type-level guarantee that the chunk representing these types of glass _has an abbreviation_"? (NB: There's a good chance I'm missing important context here, but I did my best to check related issues and search the repo. 😅)

I probably haphazardly started on the "time-consuming" second solution Jason mentioned in the original issue, but out of curiosity, I was able to (locally) completely remove the `CI` versions of these glass types by replacing `abrv` with `(fromMaybe "") . getA`. I realize this is probably hacky, but I think the crux of the issue here is that we "build a mapping" between these glass types and their associated glass type factors manually:

https://github.com/JacquesCarette/Drasil/blob/c621c01fbe96a04423e259cacc89dc5e56b426a7/code/drasil-example/glassbr/lib/Drasil/GlassBR/Concepts.hs#L189-L190

This gives us the desired abbreviations, but later we want these abbreviations as strings so we can build an expression out of them: 

https://github.com/JacquesCarette/Drasil/blob/c621c01fbe96a04423e259cacc89dc5e56b426a7/code/drasil-example/glassbr/lib/Drasil/GlassBR/Unitals.hs#L116-L117

I was originally (and naively) going to investigate redefining `glassType` to only contain the mappings between value and abbreviation (assuming that it doesn't use any other information about the glass types elsewhere), but I haven't started yet since a better fix would probably be to encode the relation between these glass types and their factors directly, potentially by defining them as `RelationConcept`s or `EquationalModel`s.

https://github.com/JacquesCarette/Drasil/blob/c621c01fbe96a04423e259cacc89dc5e56b426a7/code/drasil-theory/lib/Theory/Drasil/RelationConcept.hs#L15
https://github.com/JacquesCarette/Drasil/blob/c621c01fbe96a04423e259cacc89dc5e56b426a7/code/drasil-theory/lib/Theory/Drasil/ModelKinds.hs#L35

Just my (potentially ill-informed) thoughts that I thought I'd leave here in case they inspire anything! 😁

smiths:
Thanks @samm82.  I am fine with using `ConceptChunk` instead of `CI`, if that is better.  I'm just hoping to have one place where we capture the information, and only one version of each abbreviation in our tables.  :smile:

balacij:
@samm82 Thank you for looking into this! The only other relevant discussion is https://github.com/JacquesCarette/Drasil/pull/5148#discussion_r3408969000 , where we discussed the exact code you referenced.

Switching to the `ConceptChunk` would mean that we replace the `abrv` usage above with a `fromMaybe`. Personally, I would prefer using `fromMaybe` in the code over having the duplicate abbreviations appear in the generated artifacts.

Note that once #5414 is merged, a lot of `CI`s will be deleted. In which case, we can (and should) review what `CI`s are left over and whether they should be `CI`s or not. For example, the above found glass variants and recent discussion on #3472.

Aside: I guess this isn't really a 'partial' fix more than it is a 'hacky fix'.

JacquesCarette:
`fromMaybe` is a super hacky code level fix that isn't trying to figure out what the semantics here ought to be.

We're probably being too 'literal' / display-oriented when we talk 'abbreviation'. That's the phenomenological view of the output in our artifacts. We need to figure out the *meaning* behind those abbreviations, and then create a chunk type that captures that meaning.

So I'm not going to merge this as is, it needs more thinking.

balacij:
I agree that neither this PR nor the two other proposed hacks are suitable, but this is a blocker for #5478. Are we willing to accept a backwards step here? If so, are any of the following fine?

1. Accepting this PR as it is. We accept that we now generate a bad Table of Abbreviations and Acronyms that contains duplicate short forms.
2. (1) but with a visual-hack-fix that filters out duplicate short forms if they are associated with the same noun phrase. If we do this, we should at least output a warning message that this happened. We will need an `Eq` instance for `NP`.
3. Removing the `CI`s, referring to the abbreviations from the `ConceptChunk`s and using `fromMaybe`.

code/stable/glassbr/SRS/Jupyter/GlassBR_SRS.ipynb
```diff
@@ -147,13 +147,16 @@
         "|:--- |:--- |\n",
         "|A|Assumption|\n",
         "|AN|Annealed|\n",
+        "|AN|Annealed|\n",
```
JacquesCarette, line 150: I was almost leaning towards accepting this hacky "fix" until I noticed double entries... and that's too much!

JacquesCarette (changes requested):

samm82:
> 3. Removing the `CI`s, referring to the abbreviations from the `ConceptChunk`s and using `fromMaybe`.

If we choose to go with this option, I have a commit I can clean up and push, but only if that won't make things more complicated than they already are in trying to figure out how to address this PR! 😅

balacij:
Thank you @samm82 ! It actually turns out we can remove the references to the `CI`s that have them appear in the ToAA too (creating the duplicate entries). It's not a "real" fix to the greater ontological issue, but it at least works for us with the greater goal of getting #5478 merged.

code/stable/glassbr/SRS/Jupyter/GlassBR_SRS.ipynb
balacij: Found out we can remove the cause of the duplicate entries directly (by removing the references to the `CI`s): 75fd129147c37cb2422427a0370c41ccb7d809f5

JacquesCarette (approved):

PR #5486, opened 2026-09-04
REDO #5425: refactor(srs-gen/`PhySysDesc`): automatically retrieve program name from `SmithEtAlSRS` instead of manually passing around the program name through `SRSDecl`

balacij:
This PR is a duplicate of #5425, created because it was originally merged into #5423 but #5423 is currently blocked. So, we can redo that PR targeting `main`. Original PR description:

> Contributes to #5414 
>
> This PR improves how "program names" are reused, moving away from manually reuse towards automatic reuse in the SRSDecl declaration.

### DIFF
--- a/code/stable/swhs/SRS/HTML/SWHS_SRS.html
+++ b/code/stable/swhs/SRS/HTML/SWHS_SRS.html
@@ -916,7 +916,7 @@
               <div class="subsubsection">
                 <h3>Physical System Description</h3>
                 <p class="paragraph">
-                  The physical system of <span title="solar water heating system">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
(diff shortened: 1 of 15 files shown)


code/stable/swhs/SRS/HTML/SWHS_SRS.html
```diff
@@ -916,7 +916,7 @@ <h3>Terminology and Definitions</h3>
               <div class="subsubsection">
                 <h3>Physical System Description</h3>
                 <p class="paragraph">
-                  The physical system of <span title="solar water heating system">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
+                  The physical system of <span title="solar water heating systems incorporating PCM">SWHS</span>, as shown in <a href="#Figure:Tank">Fig:Tank</a>, includes the following elements:
```
samm82, line 919: Assuming we're good with this change (which I seem to remember being the case previously) this all looks good to me!

JacquesCarette (approved):

PR #5490, opened 2026-09-08
Demo: GHC plugin for working with typeclass graphs

B-rando1:
This PR shows the plugin I developed to track typeclass dependencies in GOOL. This is useful so that I can more easily determine the best order for removing typeclass dependencies, which is necessary for #5328. I'm making this PR because I'm not sure what I should do with it, now that it works for my current needs.

I'll start by describing the current workflow: if you run `make gool_build`, it'll build and run the plugin while building GOOL like normal. That'll generate `facts.pl` in the drasil-gool directory. Because the plugin appends to the file (needed since the plugin runs per-module), you should empty/delete `facts.pl` if you re-compile GOOL. `facts.pl` has two relations on typeclasses: `base(A)` says that `A` references TypeData in its method signatures, and `parent(X, Y)` means that `Y` is a constraint on `X`.

`make_graph.pl` reads `facts.pl` and uses it to build and output the graph. You can run `swipl drasil-gool/make_graph.pl` and then run `write_graphviz('drasil-gool/graph.dot')`. to generate the graphviz file, then `dot -Tsvg drasil-gool/graph.dot -o drasil-gool/graph.svg` to export to svg.

I showed this to @balacij on Friday, and he suggested that we might want to replace some of our other meta-analysis tools with something like this. If we want to do that, though, there are a few aspects of this approach that could probably be improved:
- The use of Prolog: Prolog is really nice for quickly and easily processing the output of the plugin, but it's another dependency for our CI if we want to use it there.
- Because it's a GHC plugin, running it requires building Drasil (or at least parts of it).

In theory, it should be possible to write a separate Haskell script that achieves the same functionality as this, using our generated `.HIE` files instead of the GHC compiler. This would remove both of the main downsides of the current approach, by keeping everything in a separate Haskell file. If we want to do more meta-analysis on Drasil using this approach, then it's probably worth me taking the time to figure out the API for working with `.HIE` files.

Alternatively, we could leave the plugin as-is, since it works for my current needs. We can put the plugin in my personal Drasil folder, and leave the other file changes for me to set locally whenever I need to run it.


People/Brandon-Bosman/typeclass-graph-plugin/src/TypeclassGraphPlugin.hs
```diff
@@ -0,0 +1,70 @@
...
+  :: [CommandLineOption]
+  -> ModSummary
+  -> TcGblEnv
+  -> TcM TcGblEnv
+inspectClasses _ _ gblEnv = do
+  let classes = typeEnvClasses $ tcg_type_env gblEnv
+      classGraph = map (\cls -> (cls, superClasses cls)) classes
+      baseClasses = filter (classDependsOnTyConName "TypeData") classes
+      output = (unlines ((map printNode classGraph) ++ (map printBaseNode baseClasses)))
+  liftIO $ appendFile ("./facts.pl") output
```
balacij: It looks like the CI failed due to a race condition caused by this line (`appendFile`, specifically) and `-j2`. I often use `-j6` locally. I'm not sure how much `-j2` vs `-j1` helps the CIs speed -- that might be worth measuring and seeing if `-j1` would be a fine quick-fix.

Another quick-fix is to split this up into many files (e.g., `facts/<MODULE>/<CLASS>.hs` that we merge later.

B-rando1: Right, makes sense. Another reason that using the HIE files is probably a better way to do this.

balacij: It'd be a shame to delete this code on that basis, however. Since your needs of this code are satisfied, maybe we can remove the `package.yaml` changes from GOOL and move the Prolog code to your folder? Easy path to merging this code. That is, unless you wanted to invest the time into working the HIE files.

B-rando1: Hmm, maybe that's a good option for now. I am planning to look into the HIE files at some point, but it would be nice to have a more flexible deadline for that.

balacij (approved):

B-rando1:
@JacquesCarette could you take a look at this when you get a chance? It would be helpful to have on the main branch so that I can easily rerun the analysis when I make changes.

JacquesCarette (approved):

PR #5495, opened 2026-09-10
chore(DblPend): remove TMs duplicated as DDs

### LINKED ISSUE #5494: DblPend: Delete "acceleration" and "force" DDs duplicating TMs
DblPend has:

* [TM:acceleration](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/HTML/DblPend_SRS.html#TM:acceleration)
* [TM:velocity](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/HTML/DblPend_SRS.html#TM:velocity)
* [DD:accelerationGDD](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/HTML/DblPend_SRS.html#DD:accelerationGDD)
* [DD:forceGDD](https://jacquescarette.github.io/Drasil/examples/dblpend/SRS/HTML/DblPend_SRS.html#DD:forceGDD)

However, the DDs don't appear to be referenced anywhere. They look like they can be cleanly deleted (or they should be, rather; if an issue somehow arises, please report it!).

balacij:
Closes #5494


code/drasil-example/dblpend/lib/Drasil/DblPend/Expressions.hs
```diff
@@ -9,13 +10,19 @@ module Drasil.DblPend.Expressions (
...
+import Data.Drasil.Quantities.PhysicalProperties (mass)
+
 import Drasil.DblPend.DataDefs (positionXEqn_1)
 import Drasil.DblPend.Unitals (lenRod_1, lenRod_2, massObj_1, massObj_2,
   xVel_1, yVel_1, xAccel_1, yAccel_1, angularAccel_1, angularAccel_2,
   tension_1, tension_2, angularVel_1, angularVel_2,
   pendDisAngle_1, pendDisAngle_2)
 
+forceExpr :: PExpr
+forceExpr = vScale (sy mass) (sy acceleration)
```
JacquesCarette, line 24: Why does DblPend need to define this itself? Is this not in `drasil-data` somewhere already?

code/stable/dblpend/SRS/mdBook/src/SecDDs.md
```diff
@@ -85,35 +85,3 @@ This section collects and defines all the data needed to build the instance mode
...
 |Source     |--                                                                                                                                                                                                                                                                                                                                                           |
 |RefBy      |[GD:velocityY2](./SecGDs.md#GD:velocityY2)                                                                                                                                                                                                                                                                                                                   |
-
-<div align="center">
-
-## Acceleration {#DD:accelerationGDD}
-
-</div>
-
-|Refname    |DD:accelerationGDD                                                                                                                                                                                                                                              |
```
JacquesCarette, line 95: I understand why this was deleted -- but acceleration and force are still relevant to DblPend. So they should "show up" somewhere. Did they show up twice?

JacquesCarette (changes requested):

code/stable/dblpend/SRS/mdBook/src/SecDDs.md
balacij: Yes. The DD duplicated a TM that already existed and displayed within the same document.

code/drasil-example/dblpend/lib/Drasil/DblPend/Expressions.hs
balacij: The one in `drasil-data` uses `$*`:

https://github.com/JacquesCarette/Drasil/blob/e2e739c35680e5fec8326f747129ee19c83f6e6c/code/drasil-data/lib/Data/Drasil/Equations/Defining/Physics.hs#L16

I suppose 'acceleration' here is a vector while the one in `drasil-data` is meant to be a scalar.... Nope:

https://github.com/JacquesCarette/Drasil/blob/e2e739c35680e5fec8326f747129ee19c83f6e6c/code/drasil-data/lib/Data/Drasil/Quantities/Physics.hs#L48

So, I changed that and reused it in `DblPend`, but it created a large `stable/` diff (126933b1ee196eafa8b354acd13ccfec531827df) because `$*` and `vScale` are not rendered to LaTeX in the same way (former: `\,`, latter: ` `).

smiths: I looked at the difference in the rendering in LaTeX.  The top one is the original spaced for F = ma, and the second is the new spacing.

F=ma

I prefer the old spacing, but I don't feel that strongly.  If the change to the new version is desirable for other reasons, I don't think the new spacing should be a blocker.

balacij: I see the value in having the thin space. I naively fixed this in #5510. I say 'naively' because I'm not immediately sure if there are any rules we should be applying for the spacing or not. For multi-letter symbols, having the thin space is crucial. Otherwise, I'm not sure? For example, for a polynomial in standard form with constant symbols, I think I prefer rendering without the extra spaces. But, I don't want to get too into the weeds on this right now. We should add a testing-suite for readability once we rework the LaTeX renderer (#4989).

smiths: Yes, the rendering with the thin space looks better, at least in the examples that I checked.  I agree that we should make this improvement, and that we shouldn't get lost in the weeds.  How things look is important, but saying the right things is more important.  :smile:  We can always go back later and make improvements to the appearance.  I like the idea of eventually assessing the readability and making sure that we respect visual formatting rules.

balacij: This is now resolved in [1a48d3a](https://github.com/JacquesCarette/Drasil/pull/5495/commits/1a48d3a5cc57592d6d15455ffc43d96c21070380)

balacij:
Ready for another review on this one now too!

JacquesCarette (approved):

Issue #5499, opened 2026-09-11
Split NounPhrase out of drasil-lang into drasil-natural-language


Purav-Kanda:
# Split NounPhrase out of drasil-lang into drasil-natural-language

## Scoping summary

Checked every dependency `NounPhrase`/`NPStruct` has outside itself, by reading the source directly, checking real usage across the repo, and a compiler-verified check (temporarily hid the constructor exports and rebuilt; the errors matched exactly what the source-reading/grep had already found).

Findings:
- `Symbol` (the P constructor) is used **opaquely** everywhere — never pattern-matched into, only carried through.       
    , Confirmed zero raw usage of the `P` constructor anywhere outside drasil-lang's own `NounPhrase`-adjacent files.
- `HasChunkRefs` (from `drasil-database`) is an **abstract** dependency (typeclass/interface) — already fine, since                
   `drasil-database` is a package both `drasil-lang` and the new package can depend on independently.

So my conclusion is that  no new shared sub-package is needed to resolve this — just making `NPStruct` (and the types that contain it, `NP` and `CapitalizationRule`) generic  over the symbol type will resolve it cleanly.

## Design

* `NPStruct`, `NP`, `CapitalizationRule`, and `PluralForm` each get a `G`-suffixed, fully generic sibling (`NPStructG a`, `NPG a`, `CapitalizationRuleG a`, `PluralFormG a`), parameterized over the type they embed instead of hardcoding `Symbol`.
* The original names become type aliases pointing at the `Symbol`-filled version (e.g. `type NPStruct = NPStructG Symbol`). Since a type alias is fully transparent, every existing consumer that only references the plain type (`NPStruct`, `NP`, etc.) needs no changes.
* No functional dependency or typeclass changes were needed — an earlier version of this design propagated genericity through `NounPhrase`'s own typeclass and every function signature in `NounPhrase.hs`/`Combinators.hs`; that was discarded in favour of the alias approach above, which needed zero signature changes in either file.
* Follow-up review requested going further: `NPStructG`'s own constructors (`S`, `P`, `:-:`, `:+:`) shouldn't be exported at all, even under the alias. Real constructors are renamed and kept private to `NounPhrase/Core.hs` and the two modules that genuinely need to pattern-match into them (`NounPhrase.hs`, `Development/Sentence.hs`); everywhere else uses smart-constructor functions (`npS`, `npP`, `.-.`, `.+.`) instead.

[...]

PR #5501, opened 2026-09-11
NounPhrase generic split

Purav-Kanda:
## Summary

Contributes to #5499

First step of splitting `NounPhrase` out of `drasil-lang` into its own package: removes the hardcoded `Symbol` dependency from `NPStruct` and the types built on top of it, in `NounPhrase/Core.hs`.

Per the scoping in #5499, `Symbol` entered noun-phrase code at exactly one point — the `P` constructor of `NPStruct` — and was never destructured, only ever transported or copied into `Sentence`. That made it safe to replace with a generic type parameter rather than a shared sub-package.

 **Fallout fix** — a type synonym doesn't carry its underlying constructors, so any import/export list that attached a constructor list directly to the old name silently imported/exported zero constructors. Changed 7 files to import constructors from the `G`-suffixed type instead, keeping the plain alias name wherever the bare type is also used as an annotation:
   - NounPhrase.hs, NounPhrase/Combinators.hs 
   - Development/Sentence.hs
   - Development.hs
   - Chunk/Concept/NamedCombinators.hs
   - Language/Drasil.hs
   - drasil-data/Data/Drasil/Concepts/Math.hs

### DIFF
--- a/code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
+++ b/code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
@@ -9,7 +9,7 @@ import Drasil.Database (mkUid)
 import Language.Drasil hiding (norm, matrix, Sentence(P, S, (:+:)))
 import qualified Language.Drasil as D
 import Language.Drasil.Chunk.Concept.NamedCombinators
-import Language.Drasil.Development (NPStruct(P, S, (:-:)))
+import Language.Drasil.Development (NPStruct, NPStructG(P, S, (:-:)))
 import Language.Drasil.Document (fromSource)
 import Language.Drasil.ShortHands (lX, lY, lZ)
 
--- a/code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
+++ b/code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
@@ -9,7 +9,7 @@ import Drasil.Database (mkUid)
 import Language.Drasil hiding (norm, matrix, Sentence(P, S, (:+:)))
 import qualified Language.Drasil as D
 import Language.Drasil.Chunk.Concept.NamedCombinators
-import Language.Drasil.Development (NPStruct, NPStructG(P, S, (:-:)))
+import Language.Drasil.Development (NPStructG(P, S, (:-:)))
 import Language.Drasil.Document (fromSource)
 import Language.Drasil.ShortHands (lX, lY, lZ)
 
--- a/code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
+++ b/code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
@@ -32,15 +32,15 @@ module Language.Drasil.NaturalLanguage.English.NounPhrase.Combinators (
   -- ** Other Combinators
   with, parensNP,
   -- re-exports
-  NPStruct((:+:), S, P)
+  NPStructG((:+:), S, P), NPStruct
 ) where
 
 import Language.Drasil.NaturalLanguage.English.NounPhrase
     ( NP,
-      CapitalizationRule(CapWords, CapFirst, CapNothing),
+      CapitalizationRuleG(CapWords, CapFirst, CapNothing),
       NounPhrase(phraseNP, pluralNP),
       nounPhrase'', surroundNPStruct)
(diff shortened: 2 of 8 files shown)


code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
```diff
@@ -9,7 +9,7 @@ import Drasil.Database (mkUid)
 import Language.Drasil hiding (norm, matrix, Sentence(P, S, (:+:)))
 import qualified Language.Drasil as D
 import Language.Drasil.Chunk.Concept.NamedCombinators
-import Language.Drasil.Development (NPStruct(P, S, (:-:)))
+import Language.Drasil.Development (NPStructG(P, S, (:-:)))
```
JacquesCarette, line 12: Even in this first pass, none of the `G` versions should be exported / used outside of `drasil-lang`.

code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
```diff
@@ -32,15 +32,15 @@ module Language.Drasil.NaturalLanguage.English.NounPhrase.Combinators (
   -- ** Other Combinators
   with, parensNP,
   -- re-exports
-  NPStruct((:+:), S, P)
+  NPStructG((:+:), S, P), NPStruct
```
JacquesCarette, line 35: We shouldn't be exporting the constructors either, but instead functions that 'alias' the constructors (and can become 'smart constructors' later).

So `:+:` should become a function, and the constructor renamed; `S` and `P` can remain constructors, but have functions that correspond to them exported.

JacquesCarette (changes requested):

code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
Purav-Kanda: Ok so drasil-data and anything else outside drasil-lang only ever sees the plain aliased names, not NPStructG directly.

code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
Purav-Kanda: So swap the exported constructors for plain functions with the same names (renaming the actual :-: constructor internally), so nothing outside can pattern-match on the real shape.

code/drasil-data/lib/Data/Drasil/Concepts/Math.hs
JacquesCarette: Correct

code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
JacquesCarette: Also correct.

Purav-Kanda: Just a small doubt :-: itself can't be reused as a function name (operators starting with : are constructor-only in Haskell), so did you have a specific name in mind for that one, or should I just pick something?

JacquesCarette: Pick something - we'll critique. You can inspire yourself from those we already use.

balacij:
Is `NounPhrase` now able to be split off into `drasil-natural-language`?

Purav-Kanda:
The core generic type (NPStructG) has zero drasil-lang-specific dependencies now, so that part is extractable. But i'm still working on changes requested also there are two things that create a two-way dependency between the packages, so if we move Core.hs as-is: the NPStruct = NPStructG Symbol alias needs Symbol (which lives in drasil-lang), and Development/Sentence.hs's toSent needs the real constructors, not just the smart ones i'm currently working on. So i will scope that as part of another PR.

code/drasil-lang/lib/Language/Drasil/NaturalLanguage/English/NounPhrase/Combinators.hs
Purav-Kanda: Went with npS/npP for the String/Symbol smart constructors, and .-./.+. for no-space/with-space join.

JacquesCarette (approved):

PR #5502, opened 2026-09-11
Removed constraints from a bunch of OO typeclasses

B-rando1:
Contributes to #5328. These are all leaf nodes of the graph I made in #5490, so their constraints need to be removed so that other constraints can be removed, so that eventually we can parameterize `TypeData`, which will allow us to remove `repr`.


JacquesCarette (approved):
Although a few things have gotten bigger (but also more precise), I believe that overall there is more simplification, so this refactor is continually improving things, if a small step at a time.

PR #5504, opened 2026-09-13
refactor(`drasil-makefile`): drop the extra `Make` namespace nesting

### LINKED ISSUE #5470: Move `Drasil.Makefile.Make.*.hs` files into `Drasil.Makefile.*` namespace
Split off from #5243 

The extra `.Make` namespace nesting adds no value. It looks like it was just a mistake.

subodh-thallada:
Resolves #5470

Moved `Drasil.Makefile.Make.*` up to `Drasil.Makefile.*`. `Drasil.Makefile` is the only exposed module, so nothing outside the package referenced the old names and no other packages needed changes.

Also folded the now-empty `Make/` folder README into the parent one (it listed an `Import.hs` that no longer exists).

`stack build drasil-makefile` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


balacij:
Having the Claude Code attribution is better than not. I think we should keep this PR over #5505. Assuming you're using AI for the other PRs, the same applies.

JacquesCarette:
Oops, sorry @balacij , I merged the other one before seeing this!

PR #5505, opened 2026-09-13
refactor(`drasil-makefile`): drop the extra `Make` namespace nesting

### LINKED ISSUE #5470: Move `Drasil.Makefile.Make.*.hs` files into `Drasil.Makefile.*` namespace

subodh-thallada:
Resolves #5470

Moved `Drasil.Makefile.Make.*` up to `Drasil.Makefile.*`. `Drasil.Makefile` is the only exposed module, so nothing outside the package referenced the old names and no other packages needed changes.

Also folded the now-empty `Make/` folder README into the parent one (it listed an `Import.hs` that no longer exists).

`stack build drasil-makefile` is clean.


JacquesCarette (approved):
Thanks!

PR #5506, opened 2026-09-13
SRS: Convert `[[Sentence]]` to `[Paragraph]` in `PurposeDescription`

### LINKED ISSUE #5468: Convert `[[Sentence]]`s to `[Paragraph]`s
Future issue: rather than a `[[Sentence]]`, shouldn't this be a `[Paragraph]` ?

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/5430#discussion_r3816545639_

subodh-thallada:
Closes #5468.

Adds a `Paragraph` alias in `Language.Drasil.Document.Core` and uses it for `CustomPurp`:

```haskell
type Paragraph = [Sentence]

data PurposeDescription = StdPurp Verbosity | CustomPurp [Paragraph]
```

`RawContent`'s `Paragraph` constructor is renamed to `Para` and now takes a `Paragraph`, which gets folded with `foldlSent_` when it's printed. `LayoutObj`'s `Paragraph` is a separate type, so it's unchanged.

No changes to `stable/`.


code/drasil-srs/lib/Drasil/SRS/DocumentLanguage/Core.hs
```diff
@@ -120,7 +120,13 @@ data IntroSub where
   -- | Organises the section.
   IOrgSec  :: Maybe Sentence -> IntroSub
 
-data PurposeDescription = StdPurp Verbosity | CustomPurp [[Sentence]]
+-- | A paragraph is a group of 'Sentence's that are folded into one when
+-- rendered.
+type Paragraph = [Sentence]
```
JacquesCarette: The correct place to introduce the concept of a paragraph is in `drasil-lang`, specifically `Language.Drasil.Document.Core`. Where it, kind of, already exists! But it is a constructor of `RawContent`, which is not what we want. So the right thing to do is to define this alias there, and then rename the constructor in `RawContent` (and adjust accordingly).

JacquesCarette (changes requested):

code/drasil-srs/lib/Drasil/SRS/DocumentLanguage/Core.hs
subodh-thallada: Done - moved the alias to `Language.Drasil.Document.Core` and renamed the `RawContent` constructor to `Para`. Left `LayoutObj`'s `Paragraph` as is since it's a different type. Kept the `"Par"` label in `prependLabel` so reference addresses don't change; all examples still match `stable/`.

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
```diff
@@ -66,7 +71,7 @@ data HasCaption = NoCaption | WithCaption
 -- | Types of layout objects we deal with explicitly.
 data RawContent =
     Table [Sentence] [[Sentence]] Title Bool -- ^ table has: header-row, data(rows), label/caption, and a bool that determines whether or not to show label.
-  | Paragraph Sentence                       -- ^ Paragraphs are just sentences.
+  | Para Sentence                            -- ^ Paragraphs are just sentences.
```
JacquesCarette, line 74: But `Para` should take a `Paragraph`. I know that's not what was implemented before, but what was there was a bit of a hack. Dealing with paragraphs properly is the better way to go.

JacquesCarette (changes requested):

code/drasil-lang/lib/Language/Drasil/Document/Core.hs
subodh-thallada: Updated, `Para` now takes a `Paragraph` and it gets folded with `foldlSent_` in the printer. `CustomPurp` passes its paragraphs straight through now instead of folding them first.

JacquesCarette (approved):

PR #5507, opened 2026-09-13
Give the portability NFR a default OS list

### LINKED ISSUE #5124: Abstract the hardcoded OS list out of each example's portability NFR
In the #5052 review ([BinaryStar/Requirements.hs](https://github.com/JacquesCarette/Drasil/pull/5052#discussion_r3360000103)), the portability NFR hardcodes 3 OS names:

  ```portable = mkPortableNFR "portable" ["Linux", "Mac OSX", "Windows"] "Portability"```

This is the same across every example (`dblpend, glassbr, pdcontroller, projectile, bss`). The OS names are also plain Strings rather than Drasil chunks.

Give default OS list (ideally proper OS concept chunks) so examples don't each repeat it.

subodh-thallada:
Resolves #5124.

Every example hardcoded the same three OS names in its portability NFR, as plain `String`s. Added `linux`, `macOSX` and `windows` as `IdeaDict`s in `Drasil.Metadata.Concepts.Computation` along with a `defaultOSs` list, and dropped the `[String]` argument from `mkPortableNFR`, so it now reads like the other `mk*NFR` helpers:

```haskell
portable = mkPortableNFR "portable" "Portability"
```

bss listed the three OSes in a different order than the rest, so its stable SRS is updated to match the default ordering. The other four examples generate the same text as before.

`stack build` is clean.

Also folded in two things from review: glassbr's `Portablity` label typo, and renaming "Mac OSX" to "macOS". Both are in their own commits.


samm82 (changes requested):
Other than this one nitpick (which might have unexpected side effects or be otherwise undesired) this looks good to me!

subodh-thallada:
Thanks for the look! The inline comment doesn't seem to have gone through. I only see the review body, no file comment. Could you re-post it by chance

code/drasil-metadata/lib/Drasil/Metadata/Concepts/Computation.hs
```diff
@@ -1,9 +1,22 @@
...
 
 algorithm :: ConceptChunk
 algorithm = cncpt''' (mkUid "algorithm") (cn' "algorithm")
   (S "a series of steps to be followed in calculations and problem-solving operations")
+
+-- * Operating Systems
+
+linux, macOSX, windows :: IdeaDict
+linux   = idea' (mkUid "linux")   (cn "Linux")
+macOSX  = idea' (mkUid "macOSX")  (cn "Mac OSX")
```
samm82: While we're here, might as well update this to "macOS" which is what Apple's operating system is now called (although it used to be "Mac OS X"; see [Wikipedia](https://en.wikipedia.org/wiki/MacOS) and the [Apple website](https://support.apple.com/en-ca/109033)). This _might_ have unforeseen consequences with capitalization, though.… 😅 
```suggestion
macOSX  = idea' (mkUid "macOS")   (cn "macOS")
```

samm82: Turns out actually submitting a comment is helpful! 😅 Since this PR changes stable already, it _might_ make sense to make this change (as well as the GlassBR typo; didn't even notice that!) here as well. @smiths might have some input as to whether or not we should change "Mac OSX" here or at all.

subodh-thallada: Done, both in separate commits so the macOS one is easy to drop if @smiths would rather leave it alone.

On the capitalization worry: `defaultOSs` is only used by `mkPortableNFR`, which folds the names in mid-sentence with `phrase`, so nothing capitalizes them. "macOS" comes out as written.

smiths: Using the right name makes sense to me.  macOS sounds good to me.  :smile:

subodh-thallada:
Took two follow-ups to get CI green. The OS chunks needed registering in `compcon`, since `phrase` resolves the term through the chunk database at render time and every example was dying with "Term: `windows` not found in TermMap". The longer `Portability` label also re-padded the glassbr traceability matrix, since it's a column header there.

All checks pass now.

samm82 (approved):

code/drasil-example/bss/lib/Drasil/BinaryStar/Requirements.hs
```diff
@@ -72,7 +72,7 @@ correct :: ConceptInstance
 correct = mkCorrectNFR "correct" "Correctness"
 
 portable :: ConceptInstance
-portable = mkPortableNFR "portable" ["Linux", "Mac OSX", "Windows"] "Portability"
+portable = mkPortableNFR "portable" "Portability"
```
JacquesCarette, line 75: Since this NFR is used in 5 different examples (at least), it should be moved to `drasil-data`.

code/drasil-srs/lib/Drasil/SRS/Sections/Requirements.hs
```diff
@@ -107,13 +107,11 @@ mkMaintainableNFR refAddress percent lbl = cic refAddress (foldlSent [
...
-mkPortableNFR :: String -> [String] -> String -> ConceptInstance
-mkPortableNFR _ [] _ = error "No operating systems specified; cannot create a requirement."
-mkPortableNFR refAddress [os] lbl = cic refAddress (S $ "The code shall be portable to " ++ os) lbl nonFuncReqDom
-mkPortableNFR refAddress osList lbl = cic refAddress (foldlSent [
+-- | Common Non-Functional Requirement for Portability, targeting 'defaultOSs'.
+mkPortableNFR :: String -> String -> ConceptInstance
+mkPortableNFR refAddress lbl = cic refAddress (foldlSent [
   S "The code shall be portable to multiple environments, particularly",
-  S $ stringList osList
+  foldlList Comma List $ map phrase defaultOSs
```
JacquesCarette, line 114: Hmm, isn't `stringList` exactly `foldlList Comma List` ?

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
```diff
@@ -2521,7 +2521,7 @@ <h2>Non-Functional Requirements</h2>
               </p>
               <p>
                 <div id="portable">
-                  Portability: The code shall be portable to multiple environments, particularly Windows, Mac OSX, and Linux.
+                  Portability: The code shall be portable to multiple environments, particularly Windows, macOS, and Linux.
```
JacquesCarette, line 2524: Why has the order in some of the examples (like BSS) changed, but not here? This is inconsistent. If we're going to tweak this stuff, we should always be increasing consistency. [The order in BSS changed to alphabetical, which I thought was on purpose.]

JacquesCarette (changes requested):

code/drasil-srs/lib/Drasil/SRS/Sections/Requirements.hs
balacij: I brushed up up `stringList` and `foldlList` in #5319:

https://github.com/JacquesCarette/Drasil/blob/6e99b7fd40ec5fdddae9ef14d7f94a072f9ee755/code/drasil-utils/lib/Utils/Drasil/English.hs#L13-L22

They have a slight difference; `stringList` filters empty strings. `foldlList` doesn't.

_Should_ there be a difference? No.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Requirements.hs
```diff
@@ -55,7 +55,7 @@ nonfuncReqs :: [ConceptInstance]
 nonfuncReqs = [portability, security, maintainability, verifiability]
 
 portability :: ConceptInstance
-portability = mkPortableNFR "portable" ["Windows", "Mac OSX", "Linux"] "Portability"
+portability = mkPortableNFR "portable" "Portability"
```
balacij: It seems like it would be better use of time to remove all `portability :: ConceptInstance`s, in favour of importing a single, shared one from `drasil-srs` or `drasil-data`, or `drasil-metadata` (listed in my preference order; I can see a reason to put them in any one of them, but they probably don't belong in any of them long-term).

code/drasil-metadata/lib/Drasil/Metadata/Concepts/Computation.hs
```diff
@@ -1,9 +1,22 @@
...
 algorithm :: ConceptChunk
 algorithm = cncpt''' (mkUid "algorithm") (cn' "algorithm")
   (S "a series of steps to be followed in calculations and problem-solving operations")
+
+-- * Operating Systems
+
+linux, macOS, windows :: IdeaDict
+linux   = idea' (mkUid "linux")   (cn "Linux")
+macOS   = idea' (mkUid "macOS")   (cn "macOS")
+windows = idea' (mkUid "windows") (cn "Windows")
```
balacij, line 18: @subodh-thallada Thank you for your work!

I'm going to dump some thoughts, but it's primarily intended for @JacquesCarette and @smiths:

Maybe we should capture these OSs with `ConceptChunk`s? How should they be captured? At the very least, I suspect `IdeaDict`s aren't enough.

My stream of thought:

First, why an `IdeaDict`? "Operational goals" is not a valid answer. Yes, it works, but we're not capturing anything further than that. What if we had Solaris or Redox? These are as well-known as Linux, macOS, Windows (alphabetical, ignoring case, being the order I prefer), we should be able to give more information in the SRS about these things. So, an `IdeaDict` is too shallow an encoding.

What about a `ConceptChunk`? A `ConceptChunk` carries a "definition" as well. That would be useful here. For example, we might write:

* "An operating system"? No, that's not enough. These operating systems are unique and our definitions should indicate so. If we left them as this, a `ConceptChunk` would be inappropriate because this would be an "alias" (a different kind of definition than the one we intend).
* "An operating system created by X/Y/Z?" Each of those people/companies have created other operating systems as well. So this would mean that any of the OSs created by Linus Torvalds, Apple, and Microsoft may be referred to as Linux, macOS, and Windows, respectively.
* Linux: "A monolithic operating system (kernel) created by Linus Torvalds in 1991, originally called Minix, used in $X$ devices as of 2026...". Not quite appropriate, either. This is more of a human-readable description than a definition (we need to clarify what we mean by description and definition). Linux is the _name of a project_. That (non-codified) project happens to be an operating system. Aha, perhaps we should create another chunk type (for non-codified things)? My intuition tells me I don't have enough information to make a better decision than this right now...

Going back to `IdeaDict`s: When is an `IdeaDict` an appropriate chunk type to use? Perhaps only when we don't have anything else to say about it (or it's not worth the effort to do so). What are situations where we don't have anything else to say about it? If there was nothing to say about it, why name it? Should `IdeaDict`s really exist or are they now a cost-saving measure we take when encoding the real knowledge is too costly?

code/drasil-example/bss/lib/Drasil/BinaryStar/Requirements.hs
subodh-thallada: Moved it to `Data.Drasil.Requirements` in `drasil-data`, and all five examples now import the one `portable` instead of building their own. `mkPortableNFR` had no callers left after that, so it's gone from `drasil-srs`.

code/drasil-srs/lib/Drasil/SRS/Sections/Requirements.hs
subodh-thallada: They were, but the argument isn't `[String]` any more. The OS names are chunks now, so the list going in is `[Sentence]` and `stringList :: [String] -> String` no longer typechecks. `foldlList Comma List` is the Sentence-level equivalent, and the rendered text comes out identical.

Good to know about the empty-string filtering. Nothing here can produce an empty entry, so it doesn't bite, but agreed the two should line up.

code/stable/dblpend/SRS/HTML/DblPend_SRS.html
subodh-thallada: Fair, that was backwards of me. BSS was the alphabetical one and I made it match the other four rather than the other way around.

Flipped it: `defaultOSs` is now alphabetical, so BSS goes back to its original order and the other five change to match. All six examples read "Linux, macOS, and Windows" now, from the one list.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Requirements.hs
subodh-thallada: Done, all five are gone in favour of one shared `portable`. I put it in `drasil-data` since that's where Jacques asked for it, which is second on your list rather than first, so say the word if you'd rather it sit in `drasil-srs`.

code/drasil-metadata/lib/Drasil/Metadata/Concepts/Computation.hs
subodh-thallada: Leaving these as `IdeaDict`s for now since that's a bigger call than this PR, and it sounds like one for Jacques and Spencer. Happy to file a ticket for it if that's useful.

Went alphabetical ignoring case, same as you and Jacques prefer.

code/drasil-srs/lib/Drasil/SRS/Sections/Requirements.hs
balacij: > Nothing here can produce an empty entry

`EmptyS` and `[]` are perfectly valid inputs! 😄

subodh-thallada: Fair, I overstated that. I meant this one call site, not the combinator: `defaultOSs` is a fixed list of three chunks with real terms, so nothing empty reaches it here. The combinator itself will happily take `EmptyS` and `[]`, and the mismatch with `stringList` is still worth sorting out on its own.

code/drasil-example/pdcontroller/lib/Drasil/PDController/Requirements.hs
JacquesCarette: `drasil-data` is where shared data for the examples go; `drasil-metadata` is where the data that is used internally to Drasil goes (and then some is re-exported via `drasil-data`).

code/drasil-metadata/lib/Drasil/Metadata/Concepts/Computation.hs
JacquesCarette: @balacij you're over-thinking this... you're not wrong, *but* we need to capture the information that we need _now_, and we can 'upgrade' these to capture more when we need it. If we've done our classy-lenses properly, that will be seamless.

JacquesCarette (approved):

PR #5508, opened 2026-09-13
SRS: Add smart constructors for the usually-empty `SRSDecl` arguments

### LINKED ISSUE #5459: `SRSDecl`: Why do we have so many empty list arguments?
> Hate to see these 'meaningless' empty lists -- smart constructor!

_Originally posted by @JacquesCarette in https://github.com/JacquesCarette/Drasil/pull/5423#discussion_r3816111164_
            
In our `SRSDecl` values (in `drasil-example/`), we have a bunch of empty lists. We should audit all of them asking ourselves:

1. Why do we allow empty lists in each place? Why are some arguments empty lists in some case studies but not in others?
2. Should we allow empty lists in those spots? Would using a `NonEmpty` list be better?
3. Should we create smart constructors that create those empty lists for us? i.e., to avoid empty lists.

subodh-thallada:
Closes #5459.

Went through the empty lists in all 12 `mkSRS` declarations. Most of them are trailing arguments that every (or nearly every) case study leaves empty, so this adds smart constructors that fill them in:

- `introProg` — `IntroProg` without the extra paragraph
- `iChar` — `IChar` with only the topics the reader is assumed to know
- `systCons` / `pdProg` — `SystCons` / `PDProg` with no related sections
- `phySysDesc` / `corrSolnPpties` — no extra contents
- `tms` / `gds` / `dds` / `ims` — no introduction (nothing passes one today)

and uses them across the case studies. The constructors are still exported, so the few spots that do need those arguments (GlassBR's terms section, SSP's system constraints and physical system contents, SWHS' properties derivation, PDController's reader characteristics) are left as they were.

I left the other empty lists alone, since there the empty list is real data rather than a placeholder: `AuxConsProg []`, `FReqsSub []`, and the empty `Fields`/`Goals`/terms in `hghc` and `template`.

Didn't go the `NonEmpty` route — none of these are "must have at least one", they're "usually have none".

Purely a refactor, the same values are passed as before, so generated artifacts are unchanged. `stack build` is clean for `drasil-srs` and all the examples.


samm82 (approved):
Code changes and rationale make sense to me!

Issue #5471, opened 2026-08-27
Majority of hand-written `HasChunkRefs` instances are wrong


balacij:
I was checking the chunk dependency lists with `make debug` and found that very very very few of our chunks have dependencies listed on other chunks. I double-checked all the code in `drasil-database` and the chunk dependency enumeration file generator, and all that code looks fine. The culprit however is our hand-rolled instances of `HasChunkRefs`. Of our 18 manually-wrtten instances, **12** are incorrect. The main issue is that expressions (`*Expr`), constraints, and inner content are omitted, resulting in missing direct dependencies.

### `QDefinition e`

https://github.com/JacquesCarette/Drasil/blob/1be5d71e310ac7639f8fd67d39a570cb195497be/code/drasil-lang/lib/Language/Drasil/Chunk/Eq.hs#L54-L59

Ignores `_expr :: e` (the right-hand side defining expression). Only collects references from `_qua` and `_inputs`.

### `RelationConcept`

https://github.com/JacquesCarette/Drasil/blob/1be5d71e310ac7639f8fd67d39a570cb195497be/code/drasil-theory/lib/Theory/Drasil/RelationConcept.hs#L23-L25

Ignores `_rel :: ModelExpr` (the relation expression). Only collects references from `_conc`.

### `ConstraintSet e`

https://github.com/JacquesCarette/Drasil/blob/1be5d71e310ac7639f8fd67d39a570cb195497be/code/drasil-theory/lib/Theory/Drasil/ConstraintSet.hs#L24-L26

Ignores `_invs :: NonEmpty e` (the invariant relations). Only collects references from `_con`.

### `DefiningExpr e`

https://github.com/JacquesCarette/Drasil/blob/1be5d71e310ac7639f8fd67d39a570cb195497be/code/drasil-theory/lib/Theory/Drasil/MultiDefn.hs#L39-L41

Ignores `_expr :: e` (the defining expression). Only collects references from `_rvDesc`. Because of this, `MultiDefn e` also misses references from its defining expressions.

### `DifferentialModel`

https://github.com/JacquesCarette/Drasil/blob/1be5d71e310ac7639f8fd67d39a570cb195497be/code/drasil-theory/lib/Theory/Drasil/DifferentialModel.hs#L105-L111

Ignores `_coefficients :: NonEmpty (NonEmpty Expr)` and `_dmConstants :: NonEmpty Expr`.

### `CodeDefinition`

[...]

Issue #2346, opened 2021-05-04
Separating cells of SRS


tingyuw:
How the document is printed currently is the printer calls printLO and maps the contents together. 
![image](https://user-images.githubusercontent.com/47548045/116967542-83164c00-ac80-11eb-8ced-d78e1a5aa936.png)

The HDiv maps the printLO layout objects to a huge section. 
![image](https://user-images.githubusercontent.com/47548045/116967735-00da5780-ac81-11eb-8ec5-868b57cd6868.png)
![image](https://user-images.githubusercontent.com/47548045/116975881-5406d700-ac8e-11eb-928c-f4ecc4bbe22c.png)

Thus, in the HTML, subsections are wrapped inside the ‘main’ sections, and each main section consisted of several subsections. If we separate the cells by the main sections, it would still be too large to maintain.
![image](https://user-images.githubusercontent.com/47548045/116974496-39336300-ac8c-11eb-9f95-0a5e5cb0053a.png)

So I was trying to separate the contents by each subsection, which means I need to ‘cut’ the printLO in the right place (maybe at the end of each subsection). That is, HDiv should not map the printLO, at least not at the time when we are still trying to print more information in the documents (the information for building a cell). The problem is how can I cut the printLO? I would like to print the document well without mapping the contents.

Another thing is, how do we know where is a section (or cell, if we would like to separate cells by sections)? This can depend on how we can cut the printLO (if applicable). For the SRS, if we can cut the printLO at any place we want (at the end of each data object for example), maybe we can use the headers as the separation standards at this point, or follow how the HTML handle subsections. And if we can cut the printLO properly, we might not need to expand the LayoutObj.
![image](https://user-images.githubusercontent.com/47548045/116974572-549e6e00-ac8c-11eb-8c90-adb3b44eabcd.png)

For the future application (such as the projectile example), to know what is a cell, my thought is to distinguish the headers and consider data objects. Some data objects could be a cell itself, such as tables, lists, and figures, and the other data objects can be separated by headers. Since it's hard to define what is a cell, we could figure out a pattern we want to separate the contents.

JacquesCarette:
You are explaining everything to us in terms of "here is what the code does now" and "here are the changes I would like to make to the code".  You have also only shown us how it is done in the HTML printer, but not the latex printer. There could be large differences.

[...]

PR #5514, opened 2026-09-14
refactor(`Section`): remove `SecCons`

balacij:
Contributes to #2346 

`SecCons` was a sum type, wrapping arbitrary document content with (sub)sections. All uses of `SecCons` were in the following form: contents ++ subsections. `Section` carried a list of `SecCons` as its "content."

Why remove it?

1. The structure of the "contents" was of the same consistent form. The sum type was unnecessary.
2. From a semantic POV, it doesn't quite make sense to intermix arbitrary content between sections. A reader would always perceive said content as belonging to the nearest top-level section.
3. For separating cells of the SRS (JacquesCarette/Drasil#2346), knowing which content is "preliminary content" versus "subsections" is useful. That's one way to split cells!
4. I want to fix the instance of `HasChunkRefs` for `Section` (JacquesCarette/Drasil#5471) and cleaning up this code before fixing that is a nice-to-have.


code/drasil-lang/lib/Language/Drasil/Document/Sections.hs
```diff
@@ -33,7 +33,8 @@ data SecCons = Sub Section
 -- and a shortname ('Reference').
 data Section = Section
              { tle  :: Title
-             , cons :: [SecCons]
+             , prelimCont :: [Contents]
```
JacquesCarette, line 36: This is a very good change -- that is the actual, proper structure of a section!

JacquesCarette (approved):
I was a little concerned that this "clean up" would be a little too hacky for my tastes, but it is the exact opposite: this introduces exactly the right structure that real sections have!

PR #5517, opened 2026-09-16
WIP: fix: `HasChunkRefs UnitDefn` instance

balacij:
Contributes to #5471 

This PR is going to need to be split up!! (and it's going to need a proper description)


balacij:
Some oddities:

1. We capture the following as `UnitDefn`s (they sound like 'just' variables... no?):
   1. "surface flaw parameter"
   2. "gravitational constant"
2. We have a bunch of missing chunks not placed in the `ChunkDB`!
3. `UnitDefn` carries a list of `UID`s that it says is its "list of chunks that make up the definition of this (new) unit." The problem: fundamental units always had their own unit in this list. This was a major hack that allowed us to automatically gather the units for the "Table of Units."

Questions:

1. What belongs in the "Table of Units" section of the SRS? Do "derived" units belong? according to the SRS blurb above the table, yes, but not all are displayed. For example, we have a "surface flaw parameter" unit. While we might not really want that to _be a unit_ (a `UnitDefn`), it currently is one, and it is used. Why shouldn't that be displayed as well?

@smiths Do you have any rules for what units should be displayed in the SRS's Table of Units? For example, should "squared metres" ($m^2$, made up of metres) be displayed? What about "A" ($\frac{XY}{Z}$, made up of X, Y, and Z)? The criteria embedded in the code is confusing: only direct (unit) dependencies of all units of all quantities displayed in the SRS. This happens to largely be the fundamental units, but misses the units of the quantities themselves (included in the `stable/` change shown here).

code/drasil-srs/lib/Drasil/SRS/ExtractDocDesc.hs
```diff
@@ -149,12 +149,15 @@ getDocDesc = fmGetDocDesc (sentencePlate id)
...
 
 -- | For a given list of 'Quantity's, collects the 'UnitDefn's dependencies of
 -- their units (i.e., what units their units are defined with).
-collectUnitDeps :: Quantity c => ChunkDB -> [c] -> [UnitDefn]
-collectUnitDeps db = map (`findOrErr` db) . concatMap getUnits . mapMaybe (getUnitLup db)
+collectUnits :: Quantity c => ChunkDB -> [c] -> [UnitDefn]
+collectUnits db qs = oneLevelDeps ++ uds
+  where
+    uds = mapMaybe (getUnitLup db) qs
+    oneLevelDeps = map (`findOrErr` db) (concatMap getUnits uds)
```
balacij, line 160: Depending on what we want to display in the "Table of Units," this code would need to be updated. Currently, I changed it to show all directly referenced units ++ all direct dependencies (units) of the quantities displayed in the SRS. I'm not sure if that's good policy yet (from the "what we want POV") but I know it has at least one issue: if we have a direct dependency that is also derived from $N$ other units, those units will not necessarily be caught unless they are also direct dependencies of the immediately referenced ones too.

code/stable/bss/SRS/HTML/BSS_SRS.html
```diff
@@ -138,11 +138,26 @@ <h2>Table of Units</h2>
...
+                <tr>
+                  <td><em>\(\frac{\text{m}}{\text{s}^{2}}\)</em></td>
+                  <td>acceleration</td>
+                  <td>acceleration</td>
+                </tr>
+                <tr>
+                  <td><em>\(\frac{\text{m}^{3}}{\text{kg}\text{s}^{2}}\)</em></td>
+                  <td>universal gravitational constant</td>
+                  <td>gravitational constant</td>
+                </tr>
```
balacij, line 160: https://github.com/JacquesCarette/Drasil/blob/e2e739c35680e5fec8326f747129ee19c83f6e6c/code/drasil-data/lib/Data/Drasil/Units/Physics.hs#L9-L21

Are these things really capturing "unit definitions"?

code/drasil-database/lib/Drasil/Database/ChunkDB.hs
```diff
@@ -217,10 +217,13 @@ insertAllOutOfOrder13 strtr as bs cs ds es fs gs hs is js ks ls ms =
...
     chDpdts = invert $ M.fromList $ map (\c -> (c ^. uid, S.toList $ chunkRefs c)) calt
 
+    fmtIDnTy c = show (c ^. uid) ++ " :: " ++ show (chunkType c)
+    dupeError c1 c2 = error $
+      "duplicate chunk found in mass insertion between `" ++ fmtIDnTy c1 ++ "` and `" ++ fmtIDnTy c2 ++ "`"
     -- Insert all incoming chunks with the existing chunk table, asserting that
     -- none of the inserted chunks were already inserted.
     chTab = M.unionWith
-      (\(x, _) _ -> error $ "duplicate chunk found in mass insertion: " ++ show (x ^. uid))
+      (\(c1, _) (c2, _) -> dupeError c1 c2)
```
balacij, line 226: Separate PR.

code/drasil-gen/lib/Drasil/Generator/CommonKnowledge.hs
```diff
@@ -39,16 +42,28 @@ withCommonKnowledge projN = insertAllOutOfOrder13 (insert projN basisCDB)
...
+basisIMs = []
+
+basisGDs :: [GenDefn]
+basisGDs = []
+
+basisTMs :: [TheoryModel]
+basisTMs = []
+
+basisUnits :: [UnitDefn]
+basisUnits = physicalPropertyUnits ++ physicsUnits ++ thermoUnits ++ siUnits
```
balacij, line 66: Separate PR.

smiths:
> @smiths Do you have any rules for what units should be displayed in the SRS's Table of Units? For example, should "squared metres" ( m 2 , made up of metres) be displayed? What about "A" ( X Y Z , made up of X, Y, and Z)? The criteria embedded in the code is confusing: only direct (unit) dependencies of all units of all quantities displayed in the SRS. This happens to largely be the fundamental units, but misses the units of the quantities themselves (included in the `stable/` change shown here).

I don't think rules have been recorded anywhere, but the idea of the table of units is to help the reader interpret the notation used in the document, not to be a comprehensive catalog.  We should include the basic SI units that come up and the derived units that are named, like pascal, joule, etc.  We don't want to include simple combinations or powers of units, like $m^2$.

The surface flaw parameter doesn't have a named unit.  We should show the units in terms of the basic SI units in the table of symbols, but it shouldn't be in the table of units.

balacij:
> not to be a comprehensive catalog.

Important thing to document.

> We should include the basic SI units that come up and the derived units that are named, like pascal, joule, etc....

Thank you for the clarification.

> We don't want to include simple combinations or powers of units, like m 2 .
>
> The surface flaw parameter doesn't have a named unit. We should show the units in terms of the basic SI units in the table of symbols, but it shouldn't be in the table of units.

It sounds like the design of "units" needs to be reevaluated.

I won't get into it _now_, but this is a good sized project, along with adding units to types. I'll write a ticket about it and update this comment when I've done that.

code/stable/bss/SRS/HTML/BSS_SRS.html
JacquesCarette: Good question. Certainly m/s^2 are the units of acceleration. And so on.

But whenever such questions arise, the [Wikipedia SI page](https://en.wikipedia.org/wiki/International_System_of_Units) is always worth a re-read! What we're doing is naming both the unit and the "quantity" traditionally associated to that unit (with a notion of quantity related but quite different than how it is used internally in Drasil, so potentially very confusing.)

smiths: I don't think  we want to introduce new units for acceleration, angular velocity, etc.  Acceleration has units of m/s^2, but we want to be careful about going the other way around.  Just because a quantity has units of m/s^2 doesn't mean it is acceleration.  Specific force (like how many g's a fighter pilot experiences) also has units of m/s^2.  It is obviously related to acceleration, but it isn't the same thing.  We've also talked many times about how the units of Nm can be Work or Torque.  I'm sure we could find many more examples.

If we have Work and Torque in the same document, are we going to define newUnit for "torque" and another newUnit for "work"?  We do want to know that both instances of Nm are different things.  I guess this could be a way to sneak that revelation into our knowledge, but it seems more implicit than explicit to do it that way.

What is the purpose of the definitions of the newUnits?  How do we use them?  If this is internally helpful to Drasil, then it could definitely have value.  If we are planning on putting these units in a table of units, it would be a strange thing for scientists to see.  I suppose we could have the option of a verbose table of units where we list everything, but by default, I think the table of units should be the base SI units used and the derived SI units used in the document.

smiths:
I added a note about what it typically in the table of units to the [capTemplate where the SRS for scientific software](https://smiths.github.io/capTemplate/SRS/SRS.pdf) is discussed.  I don't know if this is the best place to capture this information, but at least it is an improvement. :smile:

code/drasil-database/lib/Drasil/Database/ChunkDB.hs
balacij: Completed in #5524

code/stable/bss/SRS/HTML/BSS_SRS.html
JacquesCarette: It appears that our 'units' code works just fine, but is actually conceptually quite confused. Looks like it needs a re-design.

PR #5519, opened 2026-09-16
feat(`Sentence`): create `IsString` instance

balacij:
With this, in `OverloadedString` contexts, we can directly create `Sentence`s without the `S` constructor. It will also default to using `EmptyS` rather than `S ""` in cases of the empty string.

This is particularly helpful for reusing list-folding functions. This will be usable for simplifying:

https://github.com/JacquesCarette/Drasil/blob/e2e739c35680e5fec8326f747129ee19c83f6e6c/code/drasil-lang/lib/Language/Drasil/Sentence/Fold.hs#L18-L100


JacquesCarette (approved):
Nicely spotted!

PR #5521, opened 2026-09-17
Split NounPhrase.Core's generic types into drasil-natural-language

Purav-Kanda:
Moves the generic (G-suffixed) NounPhrase types, smart constructors, and raw constructors from drasil-lang into a new package, drasil-natural-language. The Symbol-filled aliases (NP, NPStruct, PluralForm, CapitalizationRule) stay in drasil-lang, defined locally in NounPhrase.hs, since Symbol is used throughout drasil-lang's math/expression core and isn't NL-specific.

NounPhrase.hs and Combinators.hs stay in drasil-lang for now  since they're written against the Symbol-filled aliases rather than the generic types; generalizing them is planned as a follow-up PR.

Repoints 13 files that imported NP/NPStruct directly from NounPhrase.Core to import from NounPhrase instead, since those aliases now live there.

Refs #5499


code/stack.yaml
```diff
@@ -31,6 +31,7 @@ packages:
 - drasil-makefile
 - drasil-meta-analysis
 - drasil-metadata
+- drasil-natural-language
```
balacij, line 34: There's just one more location in the Makefile we need to add the package name as well.

code/drasil-natural-language/package.yaml
```diff
@@ -0,0 +1,29 @@
...
+name:         drasil-natural-language
+version:      0.1.0.0
+author:       "Jacques Carette, Spencer Smith"
+maintainer:   "Jacques Carette"
+synopsis:     A framework for code and document generation for scientific software - Natural Language SubPackage
+github:       JacquesCarette/Drasil
+homepage:     https://jacquescarette.github.io/Drasil/
+description:  Please see the README on GitHub at <https://github.com/JacquesCarette/Drasil#readme>
+
+language: Haskell2010
```
balacij: We should use GHC2024

code/drasil-natural-language/package.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+ghc-options:
+- -Wall
+- -Wredundant-constraints
+- -Wmissing-export-lists
+
+library:
+  source-dirs: lib
+  exposed-modules:
+  - Language.Drasil.NaturalLanguage.English.NounPhrase.Core
```
balacij: Perhaps we should take this opportunity to rename the module as well? `Drasil.NaturalLanguage.English` as the single exposed module and `Drasil.NaturalLanguage.English.Core` as the 'main' file carrying the good stuff?

balacij (changes requested):

code/drasil-natural-language/package.yaml
JacquesCarette: Good idea.

Purav-Kanda: so like Drasil.NaturalLanguage.English (dropping the Language. prefix entirely) as the exposed module, with NounPhrase dropped from the path too? Or keeping Language.Drasil.NaturalLanguage.English and just flattening away the NounPhrase folder so Core.hs sits directly under English folder.

JacquesCarette: Hmm, now that I think about it, I don't want `NounPhrase` dropped, since we ought to add other parts of speech in due time. So just drop the leading `Language` prefix.

balacij (approved):
Looks good to me!

JacquesCarette (approved):

PR #5525, opened 2026-09-18
Removed all typeclasses with `VariableSym` as a constraint

B-rando1:
And probably some other ones too, while I was at it :)

Contributes to #5328 

I was second-guessing myself with how many of these constraints I'm removing (and how many I still have to remove), but I think it makes sense. As far as I can tell, the constraints only existed because we needed them to propagate the type families from one typeclass to another. Now that we don't have type families, the only "benefit" of the constraints is that they show some structure of how GOOL is intended to be used. From what I understand, that's not a good enough reason to have those constraints.


JacquesCarette (approved):
Wow, remove constraints from one measly little file, and tons need to be added all over! No worries, this increases the actual precision of everything.

PR #5528, opened 2026-09-20
Linter: Enable the `generalise` and `generalise-for-conciseness` HLint hint groups

### LINKED ISSUE #5279: Enable `generalise` and `generalise-for-conciseness` HLint hint groups
* `generalise`: https://github.com/jvoigtlaender/hlint/blob/ad9a1f8485eb34d40a59cc0bd1457d03fbffd50d/data/hlint.yaml#L1153-L1166

Warning: This one will have a lot of changes due to:

```yaml
    - warn: {lhs: map, rhs: fmap}
    - warn: {lhs: a ++ b, rhs: a <> b}
```

* `generalise-for-conciseness`: https://github.com/jvoigtlaender/hlint/blob/ad9a1f8485eb34d40a59cc0bd1457d03fbffd50d/data/hlint.yaml#L1168-L1204

Both look good to enable.

subodh-thallada:
Fixes #5279.

`generalise` accounts for nearly all of the diff (`map` to `fmap`, `++` to `<>`). `generalise-for-conciseness` only came up in 17 places.

Two notes for reviewers:

* `<>` binds tighter than `++` did (infixr 6 vs infixr 5), so a few expressions that mix `++` with `:` needed brackets to keep their original grouping.
* Modules that shadow Prelude's `(<>)` with the pretty printer's use a qualified `P.<>` for list concatenation, which a couple of them already did.

HLint is clean, and `build/` still matches `stable/` for every example and for the website.


code/drasil-code/lib/Language/Drasil/Code/Imperative/Build/Import.hs
```diff
@@ -70,15 +70,15 @@ renderExt (OtherExt e) _ = e
 
 -- | Helper that records the compiler input information.
 getCompilerInput :: BuildDependencies -> SoftwareDossierState -> ProgData -> [MakeString]
-getCompilerInput BcSource s _ = map makeS $ s ^. sources
+getCompilerInput BcSource s _ = fmap makeS $ s ^. sources
```
balacij:
```suggestion
getCompilerInput BcSource s _ = makeS <$> s ^. sources
```

Though, perhaps we should also remove the 3rd (always unused) parameter.

code/drasil-code/lib/Language/Drasil/Code/Imperative/Parameters.hs
```diff
@@ -79,7 +79,7 @@ getDerivedOuts :: GenState [CodeVarChunk]
 getDerivedOuts = do
   g <- get
   dvName <- genICName DerivedValuesFn
-  getParams dvName Out $ map codeChunk $ g ^. derivedInputs
+  getParams dvName Out $ fmap codeChunk $ g ^. derivedInputs
```
balacij:
```suggestion
  getParams dvName Out $ codeChunk <$> g ^. derivedInputs
```

balacij (commented):
Thanks! These changes are good. As a next step, I think can use `<$>` in a number of scenarios to tighten up the code a bit. `<$>` is more readable (to me) than `fmap`.

I've attached two suggestions where these are particularly helpful, but I think there are more applications (e.g., `dqdWr <$> [...]` instead of `fmap dqdWr [...]`).

code/drasil-code/lib/Language/Drasil/Code/Imperative/Build/Import.hs
subodh-thallada: Applied. The third parameter is used in the `BcSingle` equation just below (`renderBuildName s p nameOpts n`), so I left it alone.

subodh-thallada:
Both suggestions applied, and I went through the rest of the codebase for `<$>`.

I left `fmap` in two cases: where `<$>` would need brackets to keep the original grouping (e.g. `inputs' <> fmap quantvar derived`), and where `fmap` sits on its own line. Neither reads any tighter.

Also rebased on main to clear the conflicts.

JacquesCarette:
More conflicts...

subodh-thallada:
Rebased on main, conflicts cleared. The `future` group being enabled meant re-running the hints over the new base, so this is derived fresh rather than merged.

JacquesCarette:
Sorry, I merged in more stuff, so that there is again a conflict!

PR #5532, opened 2026-09-21
refactor(SRS): rebuild how Table of Units is constructed

balacij:
Split off from #5517 

The Table of Units will now be based the following rules applied to `UnitDefn`s of any quantity referenced in an SRS:

1. Fundamental and Derived SI units are always displayed.
2. "Defined" (Compound) units are never displayed. These are used for 'defining' $s^2$, $kW$, etc. <- I don't know if we should be considering these as "unit definitions." From an ontology POV, these aren't unit definitions -- if they were, their existence would imply that $s^2$ (from the chunk) is not necessarily the same as $s^2$ (from another chunk). We should leave "unit expressions" as they are. Names could probably be constructed automatically based on the expressions (?).

Note: We leave one exclusion hack in to avoid `stable/` being negatively changed, to avoid adding `mm` to the ToU.


smiths:
@balacij do you need a special rule to exclude mm?  There is only one fundamental unit for length in SI, the meter.  m should be the only symbol that shows up.  Modifications with powers of 10 (mm, cm, km, Mm etc, aren't fundamental units.

balacij:
The hack is unfortunately right now. `mm` is currently created as a "_derived_ SI" unit currently. I'm going to fix that in a subsequent PR.

code/drasil-lang/lib/Language/Drasil.hs
```diff
@@ -93,7 +93,7 @@ module Language.Drasil (
   , scale, shift
   , derUC, derUC', derUC''
   , fund, fund', compUnitDefn, derCUC, derCUC', derCUC''
-  , getCu, MayHaveUnit(getUnit)
+  , getCu, MayHaveUnit(getUnit), unitSymbol, UnitSymbol(..), UDefn(..)
```
JacquesCarette: Why are these exported now? I don't see (some of them) being actually used anywhere?

JacquesCarette (changes requested):

code/drasil-lang/lib/Language/Drasil.hs
balacij: It's normally easier to just export everything, I default to it unless I know there are invariants we need to uphold.

In [23199bb](https://github.com/JacquesCarette/Drasil/pull/5532/commits/23199bb3dc24f986db1185a0e77eada16a32e879), I restricted the `UDefn` export now to only exporting the type and its `UScale` constructor.

This PR uses the following:

1. `unitSymbol`: used in pulling the `UnitSymbol`s out of `UnitDefn` for examination.
2. `UnitSymbol` and constructors: used in examining specific `UnitSymbol`s; determining what kind of `UnitDefn` that it originated from really is.
3. `UDefn`'s `UScale` constructor: directly used in the hack to avoid `mm` from appearing in the ToU.

JacquesCarette: > It's normally easier to just export everything

and that's really horrible for modularity. We want the "export surface" of all modules to be as small as possible.

JacquesCarette (approved):

PR #5535, opened 2026-09-22
Removed final constraints in GOOL's typeclasses!

B-rando1:
Contributes to #5328


JacquesCarette (approved):
Hurray! And the last one was quite small, in comparison to many of the previous ones.

